PR #3, opened 2020-06-02
Add ngx-translate

davinciagf:
First test on the dropdown selector.
Warning - specific configuration is applied for this test (see sort-by.component.ts l26-l33) :

`
    translate.setDefaultLang('fr');
`

`
    translate.use('fr');
`


fgravin:
Thanks for investigating this library.
We are challenging default angular `i18n` librar https://angular.io/guide/i18n and this one.
In `i18n` there is a tool to automatically extract strings to translate from the ui, but it is less dynamic.

We are considering both options but still are not decided yet.

davinciagf:
For an older project, I had used  the extractor [ngx-translate-extract](https://github.com/biesbjerg/ngx-translate-extract) with the ngx-translsate. 
Sorry no experience with i18n

PR #15, opened 2020-10-02
I18n integration

fgravin:
See consquence of Angular i18n integration
- one build per locale
- heavy application configuration (angular.json)
- hard to maintain dynamic translation


fxprunayre:
> * one build per locale

This means that you can't load extra translation at run time ?
We are currently loading codelist from schema, translations from db eg. groups, category names

PR #17, opened 2020-10-13
Fix XSRF issues with angular serve and webcomponent demo

jahow:
This PR adjusts some stuff to make the search queries on GN work with the XSRF requirement:
* change the angular proxy config a bit
* add a new `lib/auth` module with an `AuthService` exposing an observable which emits when the auth has been queried; this has to be done _before_ sending search queries to make sure these succeed.

To do: make this work with web components as well.


libs/auth/src/lib/auth.service.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  let meApiMock
+  let me$
+
+  beforeEach(() => {
+    me$ = new Subject()
+    meApiMock = {
+      getMe: () => me$,
+    }
+
+    TestBed.configureTestingModule({
```
fgravin, line 17: you should not use `TestBed` for services, there are simple classes and should be instanciated with `new`, specially when all injection dependices are mocks.

It's way faster to execute tests this way.

libs/auth/src/lib/auth.service.ts
```diff
@@ -0,0 +1,17 @@
...
+import { map, shareReplay } from 'rxjs/operators'
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AuthService {
+  constructor(private meApi: MeApiService) {}
+
+  authReady() {
+    return this.meApi.getMe().pipe(
```
fgravin, line 12: Don't we want to store user info return by the `/me` service ?
Maybe it's out of scope now.

fgravin (changes requested):
Great step forward, thanks
A little change to do on service tests.

Does it work for web components ? :)

libs/auth/src/lib/auth.service.ts
jahow: exactly, for now we just want to be sure the browser received the correct cookies

fxprunayre:
If I got it right, for now to make GN-UI work with GN we need to turn off 
```xml
        <ref bean="csrfFilter" />
```
in `config-security-core.xml`. The proxy can not manage the cookie domain and path change. 


Maybe a multiple domain could work at some point? Something like
```xml
  <cookie-config>
      <domain>europa.eu</domain>
```    
in `web.xml` would make it work for all apps under sub domain of `europa.eu`? 


For the short term, would it make sense for version 4.0.0 to allows request without CSRF for the GN-UI endpoint we are using ie. search API.
See https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web/src/main/webapp/WEB-INF/config-security/config-security-core.xml#L366-L374 ? Like we are doing for CSW and xml.search which is used for GN harvester.

jahow:
> For the short term, would it make sense for version 4.0.0 to allows request without CSRF for the GN-UI endpoint we are using ie. search API.

I think this is a good workaround for now. Disabling CSRF token as a whole requires an analysis about whether the backend is secure enough without it. Or we can just leave it like that in GN4 and make sure to not include this system in GN micro-services.

PR #18, opened 2020-10-13
refactor record preview component

fgravin:
- Refactor `RecordSummaryComponent` -> `RecordPreviewList`
- Update Storybook to 6.0
- Create one story file for new component, the idea is to have one file per component


jahow:
Looks good, dropdown in storybook is broken because of a missing Translate import:
https://geonetwork.github.io/geonetwork-ui/record-preview/storybook/?path=/story/presentation-components--dropdownselectorcomponent

PR #19, opened 2020-10-13
Web component / handle API url and colors as input

jahow:
Api url and colors can be specified as attributes, eg:

```html
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title>Web Component Demo</title>
    <base href="/" />
  </head>
  <body>
    <script src="gn-search-input.js"></script>
    <gn-search-input
      api-url="https://apps.titellus.net/geonetwork/srv/api"
      primary-color="#e73f51"
      secondary-color="#c2e9dc"
      main-color="#212029"
      background-color="#fdfbff"
    ></gn-search-input>
  </body>
</html>
```


webcomponents/gn-search-input/src/app/gn-search-input.component.ts
```diff
@@ -13,7 +18,21 @@ import {
   encapsulation: ViewEncapsulation.ShadowDom,
 })
 export class GnSearchInputComponent implements OnInit {
+  @Input() apiUrl = '/'
+  @Input() primaryColor = '#9a9a9a'
```
fgravin, line 22: We might need an abstract class if we want to add this to all components.
Also, maybe add doc in web component README

webcomponents/common.module.ts
```diff
@@ -0,0 +1,32 @@
...
+  return new TranslateHttpLoader(http)
+}
+
+/**
+ * this module should be imported by all WebComponent modules
+ */
+@NgModule({
+  declarations: [],
+  imports: [
+    TranslateModule.forRoot({
```
fgravin, line 18: to use i18nModule when it's available in next PR

fgravin (commented):
Great adjusments thanks

webcomponents/gn-search-input/src/app/gn-search-input.component.ts
jahow: +1 for abstract class

PR #20, opened 2020-10-14
Clean libraries architecture

fgravin:
Try to keep libs scaffold as light and simple as possible.
This PR removes useless angular architect targets, configured in `angular.json` file.

It removes all libraries `lint` and `build` targets.

- The libs are not meant to be build nor published.
- The lint should be done in library through application lint

All libraries files for those tasks have been removed

- ng-package.json
- package.json
- tsconfig.lib.json 
- tsconfig.lib.prod.json


jahow (approved):
Awesome thanks!

PR #21, opened 2020-10-14
Add 3 new list items: card, item large and item small

MichelGabriel:
This PR adds 3 new list components: 
- card
- item large
- item small

and introduce new class: `.clamp-x`

gn-ui-list-components

### DIFF
--- a/libs/search/src/lib/model.ts
+++ b/libs/search/src/lib/model.ts
@@ -8,7 +8,7 @@ export interface RecordSimple {
   url: string
   thumbnailUrl: string
   category: string
-  read: string
+  more: string
   open: string
 }
 
--- a/libs/search/src/lib/model.ts
+++ b/libs/search/src/lib/model.ts
@@ -7,9 +7,6 @@ export interface RecordSimple {
   abstract: string
   url: string
   thumbnailUrl: string
-  category: string
-  more: string
-  open: string
 }
 
 export interface SearchState {
--- a/libs/search/src/lib/model.ts
+++ b/libs/search/src/lib/model.ts
@@ -7,6 +7,9 @@ export interface RecordSimple {
   abstract: string
   url: string
   thumbnailUrl: string
+  category: string
+  read: string
+  open: string
 }
 
 export interface SearchState {
--- a/libs/search/src/lib/state/effects.ts
+++ b/libs/search/src/lib/state/effects.ts
@@ -68,8 +68,8 @@ export class SearchEffects {
           category: hit._source.resourceCategoryObject
             ? hit._source.resourceCategoryObject.default
             : 'no category',
-          read: hit._source.resourceReadObject
-            ? hit._source.resourceReadObject.default
+          more: hit._source.resourceMoreObject
+            ? hit._source.resourceMoreObject.default
             : 'no read more',
           open: hit._source.resourceOpenObject
             ? hit._source.resourceOpenObject.default
--- a/libs/search/src/lib/state/effects.ts
+++ b/libs/search/src/lib/state/effects.ts
@@ -65,6 +65,15 @@ export class SearchEffects {
             : 'no abstract',
           thumbnailUrl: hit._source.overview ? hit._source.overview.url : '',
           url: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
+          category: hit._source.resourceCategoryObject
+            ? hit._source.resourceCategoryObject.default
+            : 'no category',
+          read: hit._source.resourceReadObject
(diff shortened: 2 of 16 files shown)


libs/ui/src/lib/record-preview-list/record-preview-list.component.html
```diff
@@ -1,17 +1,155 @@
...
+        {{ record.title }}
+      </h1>
+      <p class="leading-relaxed mb-3 clamp-3">{{ record.abstract }}</p>
+
+      <div class="hidden sm:flex-1 sm:flex sm:items-center sm:justify-between">
+        <a
+          class="inline-flex items-center md:mb-2 lg:mb-0 hover:underline"
+          [href]="record.url"
+        >
+          {{ record.more }}
```
fgravin, line 27: This should be a translation, it's not a field in metadata record
```html
<span translate>readMore</span>
```

libs/search/src/lib/state/effects.ts
```diff
@@ -65,6 +65,15 @@ export class SearchEffects {
             : 'no abstract',
           thumbnailUrl: hit._source.overview ? hit._source.overview.url : '',
           url: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
+          category: hit._source.resourceCategoryObject
```
fgravin, line 68: You misread the purpose of this mapping.
This is a mapper between ElasticSearch index and `RecordSimple` interface. `resourceCategoryObject`, `resourceMoreObject` and `resourceOpenObject` don't exist in ES search response.

libs/search/src/lib/model.ts
```diff
@@ -7,6 +7,9 @@ export interface RecordSimple {
   abstract: string
   url: string
   thumbnailUrl: string
+  category: string
+  more: string
```
fgravin, line 11: What is `more`, what is `open` ?
These are properties of the metadata itself, there is a miss understanding here.

fgravin (changes requested):
The HTML look fantastic but there is a misunderstanding about the RecordSimple interface and ES mappers.

libs/search/src/lib/state/effects.ts
MichelGabriel: Ok, I will remove them

libs/search/src/lib/model.ts
MichelGabriel: I removed those, they should be gone

MichelGabriel:
> don't forget to update the template accordingly
> I think you still want the metadata category don't you ?

Yes, category should be there.

fxprunayre:
> Yes, category should be there.

What do you mean by category? The db category ? Usually we prefer to use ISO topics or a thesaurus like INSPIRE themes that can be represented using icons.

MichelGabriel:
> > Yes, category should be there.
> 
> What do you mean by category? The db category ? Usually we prefer to use ISO topics or a thesaurus like INSPIRE themes that can be represented using icons.

I'm going to remove it altogether, it's not really needed in the list items

libs/ui/src/lib/record-preview-list/record-preview-list.stories.ts
```diff
@@ -20,10 +20,14 @@ RecordPreviewListComponentStory.args = {
...
     uuid: 'uiiudiiddeaafdjsqmlkfdq',
     title: 'A very nice record',
+    category: 'My Category',
     thumbnailUrl:
       'https://sextant.ifremer.fr/var/storage/images/_aliases/listitem_thumbnail/medias-ifremer/medias-sextant/accueil/cartes-thematiques/adcp/1595636-3-fre-FR/ADCP.png',
-    abstract: 'this is a great abstract',
+    abstract:
+      'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam ultrices turpis nec eleifend ornare. Etiam ante est, pulvinar ac diam nec, sollicitudin tristique leo. Vivamus a dui sed ex mattis tempus non in urna. Phasellus vel ante a risus rhoncus finibus ac sed nisl. Sed sapien lorem, gravida sit amet congue maximus, lobortis a sapien.',
     url: 'www.goto.com',
+    more: 'Read more',
```
fgravin, line 29: to be removed

MichelGabriel:
Yes, I will try to focus on just the dumb components. Problem was I didn't know where the dumb component stopped and the ES mapping began.

I'm going to try to split them into 3 components.

MichelGabriel:
Replaced by https://github.com/geonetwork/geonetwork-ui/pull/25

PR #22, opened 2020-10-14
Add record metrics component

jahow:
A new smart component does a search query to show record counts by a given field. Eg on the main search app, it shows the first 20 keywords:
image

Colors are generated based on the label. no click effect for now


jahow:
@MichelGabriel I went with the legacy design on this but it should be easy to make a new one, all the logic is there anyway

fgravin (approved):
Great work ! 
except the design that is too much GN legacy :)

PR #23, opened 2020-10-14
Results list

fgravin:
Rework results-list component

- split into smart/dumb components
- move models in common so we `lib-ui` does not depends on `lib-search`
- switch grid layout following `layout` input
- create stories for the dumb component

This work will be continued after https://github.com/geonetwork/geonetwork-ui/pull/21 is merged

- update html template to use new prensentation record-preview components
- add tests in that regard


jahow (approved):
Awesome thanks!

PR #24, opened 2020-10-14
Add new web component : aggregated records

jahow:
Shows records metrics by a given field and apply a filter on click

Also redone the record metric UI a bit

aggregated-records


apps/search/src/assets/i18n/en.json
```diff
@@ -4,5 +4,6 @@
 	"loadResults": "Load results",
 	"Sort by": "Sort by",
 	"last changed": "last changed",
-	"popularity": "popularity"
+	"popularity": "popularity",
+	"Back": ""
```
fgravin, line 8: translate `Back `at least in english

webcomponents/gn-aggregated-records/src/polyfills.ts
```diff
@@ -0,0 +1,64 @@
...
+/***************************************************************************************************
+ * Zone JS is required by default for Angular itself.
+ */
+import 'zone.js/dist/zone' // Included with Angular CLI.
+
+/***************************************************************************************************
+ * APPLICATION IMPORTS
+ */
+import 'document-register-element'
+import '@angular/localize/init'
```
fgravin, line 64: might be useless now

webcomponents/gn-aggregated-records/src/environments/environment.prod.ts
```diff
@@ -0,0 +1,3 @@
+export const environment = {
```
fgravin, line 1: environment could be clean up

fgravin (approved):
Great work and refactoring.
The target `serve:wc` might need to be adapted, cause sometimes wc won't accept default values.

PR #25, opened 2020-10-15
Add 3 new list items: card, item large and item small

MichelGabriel:
This PR replaces https://github.com/geonetwork/geonetwork-ui/pull/21

The PR changes one list record component: **List**. Add 2 new components: **Card** & **Text**

More changes:
- introduce new class: `.clamp-x`
- added new `i18n` strings
- change result list to display the 3 types 

**Card view**
gn-ui-card

**List view**
gn-ui-list

**Text view**
gn-ui-text

### DIFF
--- a/apps/search/src/assets/i18n/en.json
+++ b/apps/search/src/assets/i18n/en.json
@@ -1,8 +1,10 @@
 {
 	"My Geonetwork Catalog": "My Geonetwork Catalog",
 	"Find your metadata": "Find your metadata",
-	"loadResults": "Load results",
 	"Sort by": "Sort by",
+	"readMore": "",
+	"openRecord": "",
+	"loadResults": "Load results",
 	"last changed": "last changed",
 	"popularity": "popularity"
 }
--- a/apps/search/src/assets/i18n/da.json
+++ b/apps/search/src/assets/i18n/da.json
@@ -1,8 +1,10 @@
 {
 	"My Geonetwork Catalog": "",
 	"Find your metadata": "",
-	"loadResults": "",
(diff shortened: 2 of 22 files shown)


apps/search/src/assets/i18n/en.json
```diff
@@ -1,8 +1,10 @@
 {
 	"My Geonetwork Catalog": "My Geonetwork Catalog",
 	"Find your metadata": "Find your metadata",
-	"loadResults": "Load results",
 	"Sort by": "Sort by",
+	"readMore": "",
```
fxprunayre, line 5: @fgravin should we be mode specific on translation keys ? In master we have sometimes key that we are using in different part of the application and at some point, you want different translations because used in different context.
eg. `readMore` here is to have more info about the record and can be something slightly different somewhere else in the app.

Also translations are specific to a component or can be shared ?

Should we have a more fine grained conventions eg. `results.readMore`, `results.readMore.tooltip` for tooltips, `results.readMore.help` for helper text ... `sortRecordBy`, `sortRecordBy.resourceTitle`, ...

Some times we have camelCase key, sometime words eg. "last changed", sometimes capital letters, sometimes not...
Using english as key may not be a good idea, if there is a typo, then we have the key with a typo that may be painful to fix ?
Also related to that in master we sometimes have `:` in templates, sometimes in translations - what is best?

I'm raising this here, but maybe a dedicated discussion should be added to an issue?

MichelGabriel:
Replaced by https://github.com/geonetwork/geonetwork-ui/pull/26

PR #26, opened 2020-10-15
Add 3 new list components: Card, List and Text

MichelGabriel:
This PR replaces https://github.com/geonetwork/geonetwork-ui/pull/25

The PR changes one list record component: **List**. Add 2 new components: **Card** & **Text**

More changes:
- introduce new class: `.clamp-x`
- added new `i18n` strings
- change result list to display the 3 types 

**Card view**
gn-ui-card

**List view**
gn-ui-list

**Text view**
gn-ui-text

### DIFF
--- a/apps/search/src/assets/i18n/da.json
+++ b/apps/search/src/assets/i18n/da.json
@@ -1,8 +1,10 @@
 {
 	"My Geonetwork Catalog": "",
 	"Find your metadata": "",
-	"loadResults": "",
 	"Sort by": "",
+	"readMore": "",
+	"openRecord": "",
+	"loadResults": "",
 	"last changed": "",
 	"popularity": ""
 }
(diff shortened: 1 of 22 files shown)


fgravin:
@fxprunayre yes we need to define a policy for this. At the beginning I wanted that english texts are the keys, with no english translation file, but it can't work because of dynamic translations.

Translations are shared at the moment, the extract task extracts all strings from apps, libs and webcomponents and store them in application assets.
In webcomponent ATM we have a symbolic link to point to application assets to load translations.
We'll see how it works when we have a second application.

I will do a PR to update translations keys to something consistent.

fgravin:
Thanks @MichelGabriel I merge.
There might be an issue in the CI, ATM gh-pages branch is corrupted but it's out of the scope of the PR.
Thanks

fxprunayre:
> I will do a PR to update translations keys to something consistent.

Good.

Another related question, can we have the possibility to load translations from the webapp eg. schema translations, db translations ? It is something we are using a lot and we have to know how we will handle that and if we can't what is the alternative. It is also important for the multilingual support in order to help to decide to store for example codelist translations in the index or not.

PR #28, opened 2020-10-15
Fixes for the Card and List components

MichelGabriel:
This PR fixes height issues in the **Card** component: the links are aligned to the bottom of the Card. 

Is also fixes the missing thumbnail image, a fallback is shown for the **Card** and **List** component.

gn-card-fix

Changes:
- the links on the Card are aligned to the bottom
- a fallback image is added when there is no thumbnail


fgravin (approved):
Great changes, thanks
Please always add the translation in `en` file
https://github.com/geonetwork/geonetwork-ui/blob/master/apps/search/src/assets/i18n/en.json

PR #29, opened 2020-10-15
Catalog title component

fxprunayre:
image

It should probably go in another module.

I'm learning - be kind ;)

- [x] Also need a change in OpenAPI which is currently `get4` ! See https://github.com/geonetwork/core-geonetwork/pull/5094

Should we have some constant file for stuff like `'system/site/organization'` ?


libs/search/src/lib/catalog-title/catalog-title.component.html
```diff
@@ -0,0 +1,5 @@
+<h1 class="p-5 text-3xl font-bold"
```
fgravin, line 1: should be a dumb component in `lib-ui` lib, that takes `name `and `org` as inputs

libs/search/src/lib/catalog-title/catalog-title.component.ts
```diff
@@ -0,0 +1,20 @@
...
+  styleUrls: ['./catalog-title.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CatalogTitleComponent implements OnInit {
+  siteInfo$: Observable<SettingsListResponseApiModel>
+
+  constructor(private siteApiService: SiteApiService) {}
+
+  ngOnInit(): void {
+    this.siteInfo$ = this.siteApiService.get4()
```
fgravin, line 18: Should be shared through a `shareReplay observable` in a dedicated service.
Otherwise, all components that will need siteInfo will perform a request.

fgravin (changes requested):
Thanks for your first component 👍 
The approach should be rework though

- understand smart/dumb segragation
- observable sharing

libs/search/src/lib/catalog-title/catalog-title.component.html
fxprunayre: I understand why you have difficulties on finding component name - We are going to be short on naming with all these components ...

libs/search/src/lib/catalog-title/catalog-title.component.ts
fxprunayre: Ok. There is no cache mechanism like angularjs to not replay http calls on same URLs ?

fxprunayre:
> It should probably go in another module.

It is in search, should be move in another libs folder ? What will be the libs at the end
* auth
* common
* gn-api
* search
* map
* data
* edit
* admin

This one should go in common?

jahow:
> This one should go in catalog ?

I think this makes sense as it is a general concern, not really specific to searching. So then `catalog-title` would be a fitting name for the component. As for the dump component responsible for presentation, it may not be necessary to create one tied to the concept of catalog title; something like `ui-page-title` or `ui-header` might be enough?

libs/search/src/lib/catalog-title/catalog-title.component.ts
jahow: Ah my friend, welcome to the wonderful world of observables... In this case the HTTP request is _only sent when someone subscribes to_ `siteInfo$`. If there are three subscriptions made, the HTTP request is sent three times. There is no built-in caching mechanism that I'm aware of, but you can use an operator for the same result:
```js
 this.siteInfo$ = this.siteApiService.get4().pipe(
  shareReplay()
)
```

See https://blog.strongbrew.io/rxjs-best-practices-in-angular/#sharing-subscriptions for example

libs/search/src/lib/catalog-title/catalog-title.component.ts
```diff
@@ -0,0 +1,20 @@
+import {ChangeDetectionStrategy, Component, OnInit} from '@angular/core'
+
+import {SettingsListResponseApiModel, SiteApiService} from '@lib/gn-api'
+import {Observable} from 'rxjs'
+
+@Component({
+  selector: 'catalog-title',
```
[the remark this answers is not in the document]
fxprunayre: Move under the catalog lib, so name for now is `selector: 'catalog-site-title',`

libs/search/src/lib/catalog-title/catalog-title.component.html
fxprunayre: Dumb one created.

libs/search/src/lib/catalog-title/catalog-title.component.ts
fxprunayre: Done in https://github.com/geonetwork/geonetwork-ui/pull/29/files#diff-c59ec118eac6b193b7c3913e9d915874cda3023f17c08d8845d45c2bf1d11efcR17 but needs to move that to a service - any recommendation for this (location, naming, ...)?

apps/search/src/app/app.module.ts
```diff
@@ -11,6 +11,7 @@ import { storeFreeze } from 'ngrx-store-freeze'
 import { environment } from '../environments/environment'
 import { AppRoutingModule } from './app-routing.module'
 import { AppComponent } from './app.component'
+import { LibCatalogModule } from '../../../../libs/catalog/src/lib/catalog.module'
```
fgravin, line 14: you need to add an alias path in root tsconfig, links between libraries should be aliased, not relative.

libs/catalog/src/lib/site-title/site-title.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  templateUrl: './site-title.component.html',
+  styleUrls: ['./site-title.component.css'],
+})
+export class SiteTitleComponent implements OnInit {
+  siteInfo$: Observable<SettingsListResponseApiModel>
+
+  constructor(private siteApiService: SiteApiService) {}
+
+  ngOnInit(): void {
+    this.siteInfo$ = this.siteApiService.get4().pipe(shareReplay())
```
fgravin, line 17: I think this should be in a service, in your case, `siteInfo$` is a shared observable but only within the component.
If another component do the same thing, it will still perform the request again.

libs/catalog/src/lib/catalog.module.ts
```diff
@@ -0,0 +1,12 @@
+import { NgModule } from '@angular/core'
+import { SiteTitleComponent } from './site-title/site-title.component'
+import { UiModule } from '../../../ui/src'
```
fgravin, line 3: must be aliases like `@lib/gn-ui`

libs/catalog/src/lib/site-title/site-title.component.ts
```diff
@@ -0,0 +1,19 @@
+import { Component, OnInit } from '@angular/core'
+import { Observable } from 'rxjs'
+import { SettingsListResponseApiModel, SiteApiService } from '@lib/gn-api'
```
fgravin, line 3: Here an exemple of correct remote libray import via aliases.

libs/ui/src/lib/catalog-title/catalog-title.component.ts
```diff
@@ -0,0 +1,12 @@
+import { Component, Input } from '@angular/core'
+
+@Component({
+  selector: 'ui-catalog-title',
+  templateUrl: './catalog-title.component.html',
+  styleUrls: ['./catalog-title.component.css'],
```
fgravin, line 6: Add `changeDetection `strategy to `OnPush ` (search in code for example)

fgravin (changes requested):
Thanks for refactor, some comments and good practices about the import
Note that a smart component is not always mandatory, and can be made in applications containers ..
It's fine over all thanks

libs/search/src/lib/catalog-title/catalog-title.component.ts
fgravin: Well i don't know yet, we will need a loads of shared observables i guess like the settings or /me or info ...
You could have one or several services for that in common ..
we could create a new lib but i'm not sure it's usefull yet

apps/search/src/app/app.module.ts
fxprunayre: Can't this be manage by CLI? I probably did something wrong when creating `catalog` in lib? 

Added to `tsconfig.base.json`. No change.

Then got
```
Error: /data/dev/gn/ui/libs/catalog/src/index.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
```
Added `catalog/src/index.ts` was missing.

After a build it looks ok now but in Intellij
image

libs/catalog/src/lib/catalog.module.ts
fxprunayre: `import { UiModule } from '@lib/ui'`works. Would it make sense to prefix lib alias with `gn-` ?
There is no Intellij shortcut to properly write imports? or npm build task to clean it up?

fgravin: If everything is well configured
- alias in tsconfig
- classes exported in lib index.ts

Then Intellij do the correct import with the alias
If Intellij does not, it tells that there is mistake, fix it, remove the import and ask IntJ to redo the import, it should be correct then.

Note that Intellij sometimes uses alias to point to its own library, but path within the same library must be relatives.

fgravin:
@fxprunayre I have fixed some issues regarding the code and tests, check the commits changes for further information.
There is still an issue with Intellij trying to run the test, I am investigating

fgravin:
@fxprunayre see the last commit to set up your Intellij for tests.

This PR can be merged.

I think it will need some refactor at some point (`CommonService` is not a good name for that), let's wait we have more observables to bootstrap.

fxprunayre:
> This PR can be merged.

I let you check and merge then. Thanks for your help.

PR #30, opened 2020-10-16
Build / WC / Keep script simple

fxprunayre:


fgravin (approved):
Great thanks for simplification

PR #33, opened 2020-10-17
Update README.md

fxprunayre:
Those info are in CONTRIBUTING no? We can drop


fgravin:
Not exactly, all the instructions target `search` (what is wrong by the way should be `lib-search` now).
I think it can still have a place because it gives intructions to work with the library.

CONTRIBUTNIG.md gives global intructions, but does not tell how to build or test a specific library or application (could be added though)

Anyway, I think contributing doc should be improved shortly to help understanding npm scripts, storybook, webcomponent etc...

fxprunayre:
> I think it can still have a place because it gives intructions to work with the library.

But as we may end up with multiple libraries map/edit/admin/... maybe make sense to consolidate how to work with in a single place?

fgravin (approved):
Agreed, so we need to add a generic section in contributing to explain how to target a single library

PR #34, opened 2020-10-18
Rework the record preview in "list" layout

jahow:
I have reworked the layout on the following points:
* removed responsive classes: IMO small components like this one should not have a built in responsive layout (ie `sm:`, `md:` classes), this would be the responsibility of the container to decide which layout to use; besides such classes do not take into account the size of the actual container, which might be very different from the screen size
* removed the outer margin of the card: a presentation component should never care about its surrounding, this is also the responsibility of the container
* made the thumbnail of fixed size and the title/abstract/link layout more stable (ie not varying depending on content)
* added the last updated field as well as the source catalog logo
* added whether the record data is viewable and/or downloadable
* more subtle borders
* smaller font
* I've reverted a commit in https://github.com/geonetwork/geonetwork-ui/pull/22 that made CSS color variables be expressed in `r, g, b` form instead of `rgb(r, g, b)` form; this was made to be able to tweak the background opacity, but honestly I don't see it happen to often; also this would make it harder to override these variables as they are not in a standard CSS color format, as such there is no color picker available etc.

Still missing:
* temporal extent
* update frequency
* maybe factorize the icons into a `ui-icon` component instead of copy/pasting svg
* translations look like they're not kicking in... needs investigation

image

This is still draft, please let me know what you think and whether this is the right direction.

If this is accepted I can move on to do something similar with the card layout.


MichelGabriel:
I like it, but I think this should be the maximum of information on this item, otherwise it will become too cluttered.

fxprunayre:
> * added the last updated field as well as the org logo

Do we need the "changeDate"? it is related to the metadata and does not sound high priority information to display here.
It is always difficult to agree on what is best here but the content of
image
is probably quite relevant for most users.

ie.
* title
* abstract
* resource status
* download/view
* type of record with icon 
* resource point of contact - maybe the "logo" should be a component that can be configured. We know that we want sometimes to display the logo of the record owner group, the logo of the harvesting source, the logo of the contact in the record ...

Does the "readMore" really necessary? Clicking on the card it not enough to get more details? if we want to keep the UI light...

jahow:
True, it is hard to agree on what to show. I've looked at different catalogs and I feel like showing when was the data last updated makes sense (I though it was what the `changeDate` field was about BTW). 

I also considered adding the temporal extent, as seen in [data.gouv.fr](https://www.data.gouv.fr/fr/datasets/). To me, having a description of a dataset without at least some temporal indication feels weird, like the record could be sitting here for years or be fresh from a few days ago and there would be no way to tell.

To match the info shown in the GN4 UI we should also mention the update frequency, and whether the record is downloadable and viewable.

> Does the "readMore" really necessary? Clicking on the card it not enough to get more details? if we want to keep the UI light...

Agreed, this would be a way of making the component less cluttered.

fxprunayre:
> I though it was what the `changeDate` field was about BTW).

See https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/views/default/templates/recordView.html#L529-L575 for the index fields related to temporal extent and https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/views/default/templates/recordView.html#L379-L382 for update frequency

jahow:
I've updated the layout: the record title is now the link to GN4

> See https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/views/default/templates/recordView.html#L529-L575 for the index fields related to temporal extent and https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/views/default/templates/recordView.html#L379-L382 for update frequency

Thanks. About `changeDate`, should we still show it? in my understanding in GeoNetwork this field indicates the time the _metadata_ was updated, and not necessarily the data in itself.

fxprunayre:
> Thanks. About `changeDate`, should we still show it? in my understanding in GeoNetwork this field indicates the time the _metadata_ was updated, and not necessarily the data in itself.

For changeDate, I would say no - it is related to the metadata so no need in results list. Then which date to show ? eg.
http://metawal.wallonie.be/geonetwork/srv/fre/catalog.search#/metadata/4e00d23b-32a8-49c0-85ec-ab559fb088bc
Not that easy. Maybe:
* temporal range if exists because it will be the more precise (issue is that you could have multiple)
* then fallback to latest date in identification indicating the type - publication/revision/superseded/... (so we can discuss how we manage translation originating from schema codelist)

What do you think?

jahow:
I would have liked to be able to show the last revision date of the resource, but it looks like we just don't have this info (at least reliably). For temporal range, we can probably make a union of all ranges and show it in a very succinct way such as:
* "Covering 2018 to 3 months ago"
* "Covering 1841 to now"
* "Covering 1 month ago to 3 days ago"

This will be a good opportunity to dive into a robust, multi-lingual human-readable date rendering UI component :)

fxprunayre:
> This will be a good opportunity to dive into a robust, multi-lingual human-readable date rendering UI component :)

Some ideas in https://github.com/geonetwork/core-geonetwork/pull/5065#issuecomment-702776983

We also need a UI config option to set the timezone - See https://github.com/geonetwork/core-geonetwork/issues/1970

jahow:
> We also need a UI config option to set the timezone - See [geonetwork/core-geonetwork#1970](https://github.com/geonetwork/core-geonetwork/issues/1970)

I'm not sure I understand why we need such an option: the dates in the index should be in ISO format and contain either UTC dates or dates with a timezone shift specified (eg `2020-01-01T00:00:00+02:00`).

The only thing would be that dates without timezones in the index are in fact not UTC but expressed in a shifted timezone, and we'd need to read the backend configuration to know that. But this sounds like a situation we should avoid in the future.

fxprunayre:
> the dates in the index should be in ISO format and contain either UTC dates or dates with a timezone shift specified 

Currently we have issue because the server timezone may be different to client timezone and record may or may not have timezone. You may create a record, and then see created "2 hours ago" which is wrong with current setup. Dates in index and database will be all to the same server timezone config. It will be rendered on client side to default browser timezone or to a specific timezone as required in some projects (eg you don't want 01/01/2020T00:00 to be displayed as 31/12/2019T22:00 - some European wide catalog wants to be able to force CET on client side).

jahow:
Ok, I guess this is out of scope indeed. We'll have the opportunity to look into that later on, thanks for the explanation.

fgravin:
Thanks @jahow for this PR.
I agree, the layout must be in charge of the container.
Box size, at least height in this case must be fixed.

One thing: the organism image is broken link display a broken image if null.

IMO, we must keep it very simple, not too much info please. I wouldn't add the download and view links for the moment.
IMO it is still too big, I don't see a third party site integrate a list of 10 records, it will take 2 scroll page in their website .. do we have a way to customise the size, at least font size ?

So my advices: smaller (less lost spaces, maybe smaller font-size) and less info (not like gn4).
Maybe at some point we will create a special expert mode for the search application with tons of info, but for now keep it light for "grand public" scope.

MichelGabriel:
Just a remark, please don't make the font and whitespace too small, it should still be readable and accessible.

fxprunayre:
> Just a remark, please don't make the font and whitespace too small, it should still be readable and accessible.

Maybe you should explain if there is any minimum so we can learn about accessibility ?

MichelGabriel:
Concerning whitespace that will be difficult. Also for font size, as far as I know there is no official WCAG standard, but the common guideline is 16pt. 

And some other optimisations for typography:

- Alignment: left or right only (don’t justify)
- Line spacing: at least 1.5x the font size
- Paragraph spacing: at least 1.5x the line spacing
- Paragraph width: max. 80 characters; 40 for CJK characters

A handy tool (browser extension) I'm using for testing is this: `Siteimprove Accessibility Checker`

jahow:
Thanks for the indications @MichelGabriel. As for me I often use Lighthouse in Chrome, it also gives a lot of interesting insights. I agree that the layout should not be any more compact; if a host page wants to show a lot of records in a small space, I guess they'll have to use another layout (card probably). Having varying text size should be doable but IMO not that interesting, making it too small will break accessibility and website will never make it too large, that would not make sense.

jahow:
I've shown the update frequency instead of the update date in the record preview, please let me know if that makes sense and if this PR can be merged. Thanks

PR #36, opened 2020-10-23
Search / Records metric / Add aggregation config object.

fxprunayre:
To take advantage of the variety of bucket aggregation in Elasticsearch
(see
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket.html)
add support for a config object.

Move the query filter to a filter section of the query (which will
benefit of Elasticsearch cache for filter).

Add some input checks to explain what may be wrong with the component
config. Maybe there is a better approach ?

```html
  <search-records-metrics
    config='{"terms":{"field":"tag","size":20, "include": "IDP_DPSIR.*"}}'
  ></search-records-metrics>
```

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,8 @@ Run `ng test` to execute the unit tests via Jest.
 You can test a specific lib or app with
 ```shell script
 ng test (lib_name) (--prod)
+eg. 
+ng test lib-search --testNamePattern=RecordsMetric*
 ```
 
 To run the tests in Intellij, install the Jest plugin and run the test as usual.
--- a/apps/search/src/app/app.component.html
+++ b/apps/search/src/app/app.component.html
@@ -1,6 +1,8 @@
 <div>
   <catalog-site-title></catalog-site-title>
-  <search-records-metrics field="tag" count="20"></search-records-metrics>
+  <search-records-metrics
+    config='{"terms":{"field":"codelist_hierarchyLevel_text","size":20}}'
+  ></search-records-metrics>
   <div class="mb-3 mx-4 p-5 bg-gray-500 rounded flex flex-row">
     <div class="w-4/5 mr-3">
       <search-fuzzy-search></search-fuzzy-search>
--- a/libs/gn-api/src/lib/fixtures/search-responses.ts
+++ b/libs/gn-api/src/lib/fixtures/search-responses.ts
@@ -4,7 +4,7 @@ export const aggsOnly = {
   _shards: { total: 1, successful: 1, skipped: 0, failed: 0 },
   hits: { total: { value: 6073, relation: 'eq' }, max_score: null, hits: [] },
   aggregations: {
-    results: {
+    metric: {
       doc_count_error_upper_bound: 0,
       sum_other_doc_count: 910,
       buckets: [
--- a/libs/search/src/lib/records-metrics/records-metrics.component.spec.ts
+++ b/libs/search/src/lib/records-metrics/records-metrics.component.spec.ts
@@ -33,7 +33,7 @@ describe('RecordsMetricsComponent', () => {
   })
 
   describe('fetching record counts', () => {
-    it('parses the aggregation buckets', () => {
+    it('create a simple bucket aggregation of type terms.', () => {
       component.field = 'myfield'
       component.count = 4
       component.queryString = '+filter=abcd'
@@ -43,9 +43,13 @@ describe('RecordsMetricsComponent', () => {
       component.results$.subscribe((value) => (results = value))
       const req = httpMock.expectOne((req) => req.url.indexOf(`_search`) > -1)
       expect(JSON.parse(req.request.body)).toMatchObject({
-        query: { query_string: { query: '+filter=abcd' } },
+        query: {
+          bool: {
+            filter: { query: { query_string: { query: '+filter=abcd' } } },
+          },
+        },
         aggs: {
-          results: {
+          metric: {
             terms: {
               field: 'myfield',
               size: 4,
@@ -56,8 +60,36 @@ describe('RecordsMetricsComponent', () => {
       req.flush(aggsOnlyFixture)
 
       expect(results.length).toBe(
-        aggsOnlyFixture.aggregations.results.buckets.length
+        aggsOnlyFixture.aggregations.metric.buckets.length
       )
+    })
+
+    it('use the config for the aggregation.', () => {
+      component.config =
+        '{"terms":{"field":"tag","size":20, "include": "IDP_DPSIR.*"}}'
+      component.queryString = '+filter=abcd'
+      fixture.detectChanges()
+
+      let results = null
+      component.results$.subscribe((value) => (results = value))
+      const req = httpMock.expectOne((req) => req.url.indexOf(`_search`) > -1)
+      expect(JSON.parse(req.request.body)).toMatchObject({
+        query: {
+          bool: {
+            filter: { query: { query_string: { query: '+filter=abcd' } } },
+          },
+        },
+        aggs: {
+          metric: { terms: { field: 'tag', size: 20, include: 'IDP_DPSIR.*' } },
+        },
+      })
+    })
+
+    it('report warning in console in case of invalid config.', () => {
+      component.config =
+        '{"terms":{"field":"tag","size":2'
+      fixture.detectChanges()
+      // TODO: How to check config error is reported?
     })
 
     afterEach(() => {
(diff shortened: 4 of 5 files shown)


jahow:
Thanks for looking into this. As stated earlier I'm really not a fan of taking ES objects as inputs in web components, here's why:
- the API of the web component becomes immediately much more complex, and to document it we would basically have to redirect the user to the ElasticSearch doc (which is super hard for beginners); using a lucene query was IMO almost too complex already
- the given object has to be parsed and the user/developer needs to be able to understand if something is wrong and why; this is hard because problems can happen at different levels: JSON deserialization, using a malformed object, using the wrong nested object (i.e. how to we communicate what section of the whole ES query we expect?), using a valid object but getting no or unexpected response from ES, etc.
- the web component should _not_ write things in the console so we need another way of communicating all those potential issues to the user
- writing JSON is generally a pain, especially inline in an HTML attribute (a missing quote is pretty common and annoying)
- I don't see any major benefit here; applying a filter to the returned terms can be done using a separate input such as `include-only="IDP_DPSIR.*"` or something like that; maybe there are other uses/benefits that I don't see?

I hope that makes sense?

fxprunayre:
> I hope that makes sense?

Hum no. Maybe the target audience should be clarified. I don't think it is a question of being a fan of something or not. **Who is going to work with the GN-UI components?** Probably people using GeoNetwork so probably aware of the search API, so having basic knowledge of Elasticsearch API no? Why should we redefine a completely new model with a limited scope. **What is the purpose of GN-UI?** Be able to compose custom application on top of GeoNetwork no?  Same applies here https://github.com/geonetwork/geonetwork-ui/blob/master/libs/common/src/lib/model.ts#L5. Why not relying on the model we defined in the index ... or improve it if not good enough instead of creating something completely new.

If the metric component is supporting only terms aggregation, we are missing the benefits of using Elasticsearch and I don't really see the point of defining parameters for all combinations of ES bucket aggs (it will a large amount of work - would need to document each new or similar parameters - instead of relying on a good doc with full examples on ES side) and renaming `include` into `include-only` will not help much.

XML was a pain, now JSON is a pain, what are we supposed to use...? ;)

Also note that this PR was keeping a "dummy" mode for the configuration and propose to have and extended mode - so it could be a strategy to cover requirements of both basic and advanced users.

jahow:
I was just giving my opinion, which of course is not the absolute truth that we must follow.

I've stated the reasons why I'm not in favor of such a direction, and they come from a developer's point of view. Having to use loosely documented software exposing obscure and error-prone APIs is not fun. The ES doc is good, that I agree, but clearly not easy to use for newcomers.

Also, I believe strongly in YAGNI, or to say it differently: let's not open up possibilities for use cases that were not yet brought up. (But as stated above I may be missing some context on the whys of this PR). Adding one, two, or even five inputs later on is IMO worth it, instead of settling with one "catch-all" input to do everything.

> XML was a pain, now JSON is a pain, what are we supposed to use...? ;)

Plain text as much as possible. Writing JSON without assistance is error-prone, even if less annoying than XML.

But in the end that's my 2 cents! maybe someone else can drop in to give their opinion?

fgravin:
> Who is going to work with the GN-UI components

Also basic web site mainteners, like CMS, who want a list of metadatas with basic filter (and don't know about ES), expecting a config more like `sortBy:date, type:dataset`

> Why not relying on the model we defined in the index

There is no typescript export for the document model.

I think we must keep an easy way to configure the search, the filters, the aggs and we must think of what model and how to parse it. However, I think we can still propose the hardcore mode for people who really want something special and we can imagine passing an ES config if needed. But I also see limitation, in this component, the ES `config` (as it is called) only concerns a subpart of the ES object, the `aggs.metric` one. In other component, this part will be something else, maybe a new `bool`, `must` or `filter`, it's hard to put a scope to what to put in the config and what not ... here the final ES object is composed of 2 inputs (filters and config), it can be hard to understand for the users.
In some component the `config` will target the aggs, in other the filter, in other maybe the full ES object etc..

So my position is that i'm not against to have the hardcore mode overriding the whole config, but prefer to favor the easy mode, or at least allow and handle it.
It's true that explicit `Inputs` will lead the user to know what he can do or not, ES doc will not really.

jahow:
Allowing the user to bypass "simple" inputs altogether and just provide a full ES config, "use at your own risk" style, might be a good compromise. But then we'd still need to document that, in this case, the user should include a `terms` aggregation with a certain name for the component to render anything.

fxprunayre:
Note that this PR return a clear error message and example in case of errors https://github.com/geonetwork/geonetwork-ui/pull/36/files#diff-e2bc3405024780feb09e7bb466f86d721aca0d27a8e4219f8111014db11fee1dR25-R27 which might not be elegant enough but at least provide feedback that something is wrong and a working example.

fxprunayre:
So trying to summarize. 

#### Who is going to work with the GN-UI components?

* GeoNetwork developers
* Developers of SDI, portals, ...
* Web site and CMS mainteners


#### What are the use cases for this component?

> let's not open up possibilities for use cases that were not yet brought up. 

* GeoNetwork home page eg. https://sdi.eea.europa.eu/catalogue/srv/eng/catalog.search#/home - one goal is to improve GN default UI?
* Which can be customized by portal eg. https://sdi.eea.europa.eu/catalogue/idp/eng/catalog.search#/home
* Be in capacity to have one icon per values eg. something like country overview here https://inspire-geoportal.ec.europa.eu/overview.html?view=pdEuOverview&legislation=all maybe or the INSPIRE themes
* One number matching one query eg. "Downloadable data sets", "Viewable data sets" so a CMS can embed big numbers

To cover those, we need at least support for terms & filters. If we start adding parameters, we will end up with some parameters only applying to terms aggs, some for filters. The number of parameters to at least cover the cases above is not a matter of 5 (field, size, include, exclude, order, missing ... + for filters an array of query?). So we can keep a really dummy mode with term+size+order only and then consider that then it is the advanced mode, We can create one component per type of aggs, ... 


#### Component configurations

> hardcore mode 
> "use at your own risk" style 

Components related to the search can be configured using:
* a basic mode providing limited set of options in order to quickly plug the component for basic usage
* an advanced mode based on Elasticsearch API
* configuration are checked on init and errors are reported to developers using example and alternatives
* naming of configuration (eg. aggs, filters) MUST be consistent across components


#### Future requirements for the metric component

* Icon - How to plug an icon per value source? 
* Styling also probably need improvements


#### Model 

> There is no typescript export for the document model.

Why not adding one on the server API then? If the OpenAPI needs improvements, we can fix and improve it on server side - it will benefit to everyone - eg. https://github.com/geonetwork/geonetwork-ui/pull/35

jahow:
>    * One number matching one query eg. "Downloadable data sets", "Viewable data sets" so a CMS can embed big numbers

Won't that be part of the metadata record in ES? To be honest this sounds like such a common use case, we could maybe offer a separate web component to do just that.

> To cover those, we need at least support for terms & filters. If we start adding parameters, we will end up with some parameters only applying to terms aggs, some for filters. The number of parameters to at least cover the cases above is not a matter of 5 (field, size, include, exclude, order, missing ... + for filters an array of query?). So we can keep a really dummy mode with term+size+order only and then consider that then it is the advanced mode, We can create one component per type of aggs, ...

The way I view it, we can cover a lot of ground by using only terms aggregations with the following API:
* `field`
* `query` (lucene query)
* `size`
* `order-by` (eg `count` or `value`)
* `include` and `exclude` (either regex or array of values)
* `missing` (free text value)

Using a filters aggregation would be feasible but require going in full ES config mode.

> Components related to the search can be configured using:
> 
>    * a basic mode providing limited set of options in order to quickly plug the component for basic usage
> 
>    * an advanced mode based on Elasticsearch API
> 
>    * configuration are checked on init and errors are reported to developers using example and alternatives
> 
>    * naming of configuration (eg. aggs, filters) MUST be consistent across components

Agreed on this.

>    * Icon - How to plug an icon per value source?

Good question. I guess we need to :

1. Show a nice default icon

2. allow the user to provide a kind of value-to-icon mapping, where values would point at images (svg or png) either by url or value.
  Note that this would probably take place in Javascript like so:
```js
const metrics = document.getElementById('gn-metrics'); // this is the web component we defined in the html
metrics.iconMapping = {
  'DE': '../icons/germany.png',
  'FR': '../icons/france.png',
  'IT': '../icons/italy.png',
};
metrics.defaultIcon = '../icons/empty-flag.png';
```

fgravin:
> Why not adding one on the server API then? If the OpenAPI needs improvements, we can fix and improve it on server side - it will benefit to everyone - eg. #35

Yes would be great, with fields should be optionnal `?`.

fxprunayre:
Will be done in another branch.

PR #37, opened 2020-10-29
Bootstrap settings

fgravin:
Refactor `CommService` to `BootstrapService` that will host a battery of Observables that might be needed during application bootstrap.

Added a `uiConfReady(uiIdentifier: string): Observable<UiSettingApiModel>` method that will maintain a pool of Observable, one for each `uiIdentifier`. The Ui configuration will be parsed into a JSON object, error are handled with a console.warn.

I add a `LogService` to log information in the application for 2 reasons:
- centralized and can be enhanced (if for instance we want to display a notification as well as console messages)
- easier to test and mock


libs/common/src/lib/services/bootstrap.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        jest.clearAllMocks()
+      })
+    })
+
+    describe('When we require configuration already fetched', () => {
+      beforeEach(() => {})
+      it('calls api only once', () => {
+        service.uiConfReady('conf1').subscribe((conf) => (uiConf = conf))
+        service.uiConfReady('conf1').subscribe((conf) => (uiConf = conf))
+        expect(service['uiService'].getUiConfiguration).toHaveBeenCalledTimes(1)
```
fgravin, line 76: @jahow each `subscribe` actually trigger a call to the `service['uiService'].getUiConfiguration` method, that should not be the case the the `shareReplay` should it ?
It probably does not trigger the Http call though...

jahow: weird indeed, it should not work like that... needs some investigation

fgravin: Ok I understood ... 
The service is instanciated for each test, but the mock service is the same object, so a spy on the mock increases the call stack while `shareReplay` works as expected.
A `jest.clearAllMocks` might be usefull in `afterEach` for some cases.

jahow: yes! I've noticed that in other tests. Using mocks initalized once for the whole test suite is misleading for this reason. Maybe we should add `clearAllMocks` at the end of every spec file...

libs/common/src/lib/services/bootstrap.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    private siteApiService: SiteApiService,
+    private logService: LogService,
+    private uiService: UiApiService
+  ) {
+    this.uiConfigurations = {}
+  }
+
+  siteInfoReady(): Observable<SettingsListResponseApiModel> {
+    return this.siteApiService.getSiteOrPortalDescription().pipe(
+      map((r) => r),
```
jahow, line 28: what is that for?

libs/common/src/lib/services/bootstrap.service.ts
```diff
@@ -0,0 +1,59 @@
...
+      shareReplay()
+    )
+  }
+
+  settingsReady(): Observable<SettingsListResponseApiModel> {
+    return this.siteApiService.getSettingsSet().pipe(
+      map((r) => r),
```
jahow, line 35: same question

libs/common/src/lib/services/log.service.ts
```diff
@@ -0,0 +1,17 @@
...
+export class LogService {
+  constructor() {}
+  log(message: string): void {
+    console.log(message)
+  }
+  warn(message: string): void {
+    console.warn(message)
+  }
+  error(message: string): void {
+    console.error(message)
```
jahow, line 15: I don't think these shout output to the console eventually; like I said earlier, in the context of web components it would be super messy to log things to the console IMO; this service can be redirected to show notifications on the screen instead eventually I guess.

jahow (approved):
Thanks for tackling this. I just had a few questions on the code. Besides, I think we should avoid logging to the console in the future (although having a log service is smart as we can redirect it later on).

On the fact of using the catalog settings in geonetwork-ui, I'm a bit worried about simply taking the weird legacy config objects from GN and simply reading them in the app. Maybe having a mapper which exposes only what we need could be good in the long run? Also because we probably won't need most if it at first.

Last thing, have you considered putting the settings in some kind of state? Having a service exposing settings as observables sounds like a NgRx facade. I mean, this PR makes sense in that settings should not change during a session, but still it might be beneficial in some other way.

libs/common/src/lib/services/bootstrap.service.ts
fgravin: I saw it in the initial method so I kept it, thought you added that :) but yes it's useless.

libs/common/src/lib/services/log.service.ts
fgravin: Yes that's the purpose of the log service, we can change the implementation if required.
No opinion about console logging yet.

fgravin:
> On the fact of using the catalog settings in geonetwork-ui, I'm a bit worried about simply taking the weird legacy config objects from GN and simply reading them in the app. Maybe having a mapper which exposes only what we need could be good in the long run? Also because we probably won't need most if it at first.

You are completly right, we d'ont want the full ui config, here now i just need the facet, that may not changed (discussed with @fxprunayre )


> Last thing, have you considered putting the settings in some kind of state? Having a service exposing settings as observables sounds like a NgRx facade. I mean, this PR makes sense in that settings should not change during a session, but still it might be beneficial in some other way.

Yes we were just talking about that with LH, for instance, the facetConfig should go in SearchState so the service knows how to build the query. Note that the obersable provided by the boostrap serice is kind of a state as well, but yes some info will be put in ngrx store.

I will merge to move forward and help further development but I keep in mind the comments and will adress them, at least for facets in following PRs.

Thanks for the review !

PR #38, opened 2020-11-02
Facet keywords

LHBruneton-C2C:
* Create ui facets module
* Add block and item dumb components
* Create item and block models for inputs


libs/ui/src/lib/facet/checkbox-input/checkbox-input.component.ts
```diff
@@ -0,0 +1,33 @@
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core'
```
fgravin, line 1: I think the name is not explicit enough, there the checkbox, but also the invert. And `CheckboxInput` is confusing as a name.
I prefer something like `FacetItem`
We can agree on those terms:
- facet list (all facets on the left)
- facet block: a block for a particular definition (usually an index field)
- facet item (one line in a block)
What do you think ?

libs/ui/src/lib/facet/list/list.component.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+import { async, ComponentFixture, TestBed } from '@angular/core/testing'
+import { FormsModule } from '@angular/forms'
+
+import { ListComponent } from './list.component'
+
+describe('ListComponent', () => {
+  let component: ListComponent
+  let fixture: ComponentFixture<ListComponent>
+
+  const modelMock = {
```
fgravin, line 11: fixtures should be centralized, wait for my PR to be merged.

libs/ui/src/lib/facet/list/list.component.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    key: 'tag',
```
fgravin, line 12: We should declare a `Typescript` definition

libs/ui/src/lib/facet/checkbox-input/checkbox-input.stories.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Meta, moduleMetadata } from '@storybook/angular'
+import { CheckboxInputComponent } from './checkbox-input.component'
+
+const moduleMetadatas = {
+  imports: [I18nModule],
+}
+
+export default {
+  title: 'UI/Facet',
+  decorators: [moduleMetadata(moduleMetadatas), withKnobs, withA11y],
```
fgravin, line 14: No decorator needed for actions ??

fgravin (commented):
Good so far.
Some comment about naming convention.

I can't test the storybook unfortunatly cause it's a PR from a remote repository. I will see if I can fix that. Thanks

libs/ui/src/lib/facet/checkbox-input/checkbox-input.component.ts
LHBruneton-C2C: It will be more precise indeed. I'll also rename the facet module to facets to match with the search lib.

libs/ui/src/lib/facet/list/list.component.spec.ts
LHBruneton-C2C: I looked at your PR, but lib-ui depending on lib-search, only to be able to use fixtures in the unit tests, that feels strange to me.

libs/ui/src/lib/facet/list/list.component.spec.ts
LHBruneton-C2C: I agree, but doesn't it belong in lib-search ?

fgravin: If you put the model in `lib-search`, then `lib-ui` will depend on `lib-search` and we'll have a cycle dependency.

libs/ui/src/lib/facet/list/list.component.spec.ts
fgravin: hmm yes ...
We could create a new folder `test` somewhere with all the fixtures...

fgravin (changes requested):
Code looks good, missing some tests though.
Also use fixtures when you can for the tests and stories.

I have tested the storybook, I think we can improve a bit the rendering
- figures in lighter grey
- mouse hover with cursor and different color
- block title
- filter input position

Note that this component is missing some behavior (like isInSearch, color display in green or red..) but we can cover this later.
Thanks

.github/workflows/main.yml
```diff
@@ -71,3 +71,32 @@ jobs:
 
       - name: Build storybook
         run: npm run build:storybook
+
```
fgravin, line 74: should not have anything in this file

libs/ui/src/lib/facets/facets.model.ts
```diff
@@ -0,0 +1,18 @@
+export interface HasPath {
```
fgravin, line 1: I don't understand the name of the itnerface

tailwind.config.js
```diff
@@ -29,6 +29,9 @@ module.exports = {
       },
     },
   },
-  variants: {},
+  variants: {
```
fgravin, line 32: I didn't figure out what happens in storybook with the `group-hover`

fgravin (commented):
Very good overall.
Some minor comments.
Thanks

libs/ui/src/lib/facets/facets.model.ts
LHBruneton-C2C: I wanted to express interface composition, with both models extending over the capability to have a path. If it's not clear or not useful, I can revert to both model having a "path" property.

tailwind.config.js
LHBruneton-C2C: I don't understand your comment ? Is the group-hover functionality not working in your storybook ? Or is there something else ?

fgravin: Yes what should happen on group hover ?

libs/ui/src/lib/facets/facets.model.ts
fgravin: Oh ok got it, ok we can keep it.

tailwind.config.js
LHBruneton-C2C: In facet item, both label and count should darken independently of which one is hovered, and the minus or check at the end of the line should appear.
In facet block, both title and chevron should darken.

fgravin: Ok but why do you need to add TW variants for that ? It's just a question I didn't work on TW configuration.

PR #39, opened 2020-11-02
Facets module

fgravin:
Add a smart facet module in `lib-search`
- empty facets container will be responsible to read the request and the response and build the model required for the facet component to render itself and interact with the search after an user action
- service that only contain a mapping to the ui model

I also added ES fixtures to run the tests.


libs/search/src/lib/facets/facets.service.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  createFacetModel(
+    requestAggregations,
+    responseAggregations,
+    isNested: boolean,
+    path: string[] = []
+  ) {
+    if (requestAggregations == undefined) {
+      return
+    }
```
jahow, line 20: Is that really necessary?

jahow (commented):
Looking at the code, the first thing that I noticed was that there is no typing whatsoever for ES requests/responses. The plan is still to get the types from the OpenAPI documents in GN4?

Also, how do you envision plugging this to the search state? Will both the request and response objects be stored in the state, in order to render facets? Or just the facet model?

fgravin:
> Looking at the code, the first thing that I noticed was that there is no typing whatsoever for ES requests/responses. The plan is still to get the types from the OpenAPI documents in GN4?

Well i wanted to adresse it in another PR cause I just figured out that the @types/elasticsearch we are using are outdated and not compatible with version 7.9. The OpenAPI won't give any hints on Elasticsearch request/response interfaces, maybe, following @fxprunayre words, we will have a defition of the `sources` hits.

> Also, how do you envision plugging this to the search state? Will both the request and response objects be stored in the state, in order to render facets? Or just the facet model?

We are just discussing this this morning with Laure-Helene, at least I see that the config for facets (so requestion aggregations) will be stored in the state, and response aggregation as well. We need to clarify the structure of the state, maybe we can organize a smal meet on that with you @jahow and @fxprunayre.

More will come with facet-container logic implementation.

PR #40, opened 2020-11-03
Task GSGEODPT43-97

y-vynohradov:
Create a new application apps/datafeeder
Create a new called lib-editor in libs/editor


apps/datafeeder/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  })
+
+  it(`should have as title 'datafeeder'`, () => {
+    const fixture = TestBed.createComponent(AppComponent)
+    const app = fixture.componentInstance
+    expect(app.title).toEqual('datafeeder')
+  })
+
+  it('should render title', () => {
+    const fixture = TestBed.createComponent(AppComponent)
```
fgravin, line 26: wonder how it passes as i see nothing in the template

angular.json
```diff
@@ -508,6 +508,107 @@
...
+                {
+                  "type": "anyComponentStyle",
+                  "maximumWarning": "6kb",
+                  "maximumError": "10kb"
+                }
+              ]
+            }
+          }
+        },
+        "serve": {
```
fgravin, line 582: miss test target
check test target from search app

fgravin (changes requested):
Very good start !
Please add the test target for datafeeder application and fix your test.

Thanks

apps/datafeeder/src/app/app.component.spec.ts
y-vynohradov: It seems test was skipped because of miss test target in agular.json

fgravin (approved):
Thanks for the changes, excellent work

PR #41, opened 2020-11-03
Add a UI file drop input

y-vynohradov:
Add simple dnd file input conponent to the ui library.

### DIFF
--- a/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
+++ b/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
@@ -1,7 +1,3 @@
 ngx-dropzone {
   height: auto;
-  align-items: unset;
-  color: unset;
-  border: none;
-  border-radius: 0;
 }
--- a/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
+++ b/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
@@ -1,11 +1,7 @@
 ngx-dropzone {
   height: auto;
-  align-items: flex-start;
-  color: black;
+  align-items: unset;
+  color: unset;
   border: none;
   border-radius: 0;
-}
-
-ngx-dropzone-label {
-  margin: 10px 0 !important;
 }
--- a/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
+++ b/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
@@ -1,3 +1,4 @@
 ngx-dropzone {
   height: auto;
+  border:none;
 }
--- a/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
+++ b/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
@@ -1,0 +1,11 @@
+ngx-dropzone {
+  height: auto;
+  align-items: flex-start;
+  color: black;
+  border: none;
+  border-radius: 0;
+}
+
+ngx-dropzone-label {
+  margin: 10px 0 !important;
+}
--- a/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.html
+++ b/libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.html
@@ -1,9 +1,9 @@
-<div class="border-solid border border-black">
+<div>
   <ngx-dropzone multiple="false" (change)="selectFile($event)">
-    <div *ngIf="!fileName" class="text-black pl-4 py-4">
-      {{ placeholder | translate }}
+    <div *ngIf="!fileName" class="text-gray-900 pl-2 py-2" translate="">
+      {{ placeholder }}
     </div>
 
-    <div *ngIf="fileName" class="text-black pl-4 py-4">{{ fileName }}</div>
+    <div *ngIf="fileName" class="text-gray-900 pl-2 py-2">{{ fileName }}</div>
   </ngx-dropzone>
 </div>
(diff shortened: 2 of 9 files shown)


fgravin (changes requested):
Please add a description in the.
Also rebase it on master and push force.

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
```diff
@@ -0,0 +1,11 @@
+ngx-dropzone {
+  height: auto;
+  align-items: flex-start;
+  color: black;
+  border: none;
+  border-radius: 0;
+}
+
+ngx-dropzone-label {
```
fgravin, line 9: We use tailwindcss for the styles, it is made of many helper classes to set the `display`, `margin`, `padding` etc..
You almost never need to set css for this.

https://tailwindcss.com/docs/margin

It's the same for the color, border etc.. get inspired from https://github.com/geonetwork/geonetwork-ui/blob/master/libs/ui/src/lib/record-preview-list/record-preview-list.component.html for example.

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.ts
```diff
@@ -0,0 +1,28 @@
...
+import { BehaviorSubject, Observable } from 'rxjs'
+import { distinctUntilChanged } from 'rxjs/operators'
+
+@Component({
+  selector: 'ui-drag-and-drop-file-input',
+  templateUrl: './drag-and-drop-file-input.component.html',
+  styleUrls: ['./drag-and-drop-file-input.component.css'],
+})
+export class DragAndDropFileInputComponent implements OnInit {
+  @Input() placeholder = 'Nom de mon fichier'
```
fgravin, line 11: oh do you speak french ? :)

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.html
```diff
@@ -0,0 +1,5 @@
+<ngx-dropzone multiple="false" (change)="selectFile($event)">
+  <ngx-dropzone-label *ngIf="!fileName">{{ placeholder }}</ngx-dropzone-label>
```
fgravin, line 2: The placeholder should be internationalized. check out CONTRIBUTING doc i18n for more info, it should be a key, then translated in the template.

fgravin (changes requested):
Tested and looks very good.
Some comment about usage of Tailwindcss, and internationalistion.
I would also add a margin/padding left for the text inside of the input.
Thanks

fgravin (changes requested):
Thanks for the change.
You idea is  good to be able to pass css classes, but it is a bit too complex for the need that we have today. Just put the classes directly in the component template for now, keep it simple as we can !

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
```diff
@@ -1,11 +1,7 @@
 ngx-dropzone {
   height: auto;
-  align-items: flex-start;
-  color: black;
+  align-items: unset;
```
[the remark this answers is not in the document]
y-vynohradov: It will be impossible to override css styles with tailwindcss. For example, text will always be centered and colored in gray.

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.html
```diff
@@ -1,5 +1,7 @@
-<ngx-dropzone multiple="false" (change)="selectFile($event)">
-  <ngx-dropzone-label *ngIf="!fileName">{{ placeholder }}</ngx-dropzone-label>
+<div [class]="cssClassesForInput">
+  <ngx-dropzone multiple="false" (change)="selectFile($event)">
+    <div *ngIf="!fileName" [class]="cssClassesForLabel">{{ placeholder }}</div>
```
fgravin, line 3: add `translate` attribute for translation

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="border-solid border border-black">
+  <ngx-dropzone multiple="false" (change)="selectFile($event)">
+    <div *ngIf="!fileName" class="text-black pl-4 py-4">
+      {{ placeholder | translate }}
+    </div>
+
+    <div *ngIf="fileName" class="text-black pl-4 py-4">{{ fileName }}</div>
+  </ngx-dropzone>
+</div>
```
fgravin, line 9: I would go for no border, smaller margin, and grey color.
I also prefere the attribute `translate` when possible

```suggestion
<div class="">
  <ngx-dropzone multiple="false" (change)="selectFile($event)">
    <div *ngIf="!fileName" class="text-gray-900 pl-2 py-2" translate="">
      {{ placeholder }}
    </div>
    <div *ngIf="fileName" class="text-gray-900 pl-2 py-2">{{ fileName }}</div>
  </ngx-dropzone>
</div>
```

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.css
```diff
@@ -0,0 +1,7 @@
+ngx-dropzone {
+  height: auto;
+  align-items: unset;
+  color: unset;
+  border: none;
+  border-radius: 0;
```
fgravin, line 6: remove it

libs/ui/src/lib/drag-and-drop-file-input/drag-and-drop-file-input.component.ts
fgravin: On top of the file do a 
```js
export const placeholder = 'dropFile'
marker('dropFile')
```
This will add en entry for the label to translate in i18n translation files.
And here use `placeholder` variable for the default value.
Then run the script for label extraction for i18n, check CONTRIBUTING.md please.

fgravin (changes requested):
The i18n is not complete for your label.
Also some suggestion for a better rendering, closer to other components look.
Thanks

y-vynohradov:
> did you add it by hand or with the npm script ?

I used i18n:extract script

PR #44, opened 2020-11-05
Add a UI progress bar.

y-vynohradov:
Add progress bar to UI lib

image

### DIFF
--- a/libs/ui/src/lib/progress-bar/progress-bar.component.html
+++ b/libs/ui/src/lib/progress-bar/progress-bar.component.html
@@ -1,10 +1,12 @@
-<div class="flex bg-white rounded-t-lg rounded-b-lg">
-  <div class="flex w-full bg-blue-600 rounded-t-lg rounded-b-lg my-2 mx-2">
-    <div
-      [style.width.%]="validValue"
-      class="bg-blue-400 my-1 mx-1 transition-width duration-500 ease-in-out rounded-t-lg rounded-b-lg"
-    >
-      <span class="pl-1 py-1 font-bold text-4">{{ validValue }}%</span>
+<div class="flex h-full {{ color.outerBar }} rounded-t-lg rounded-b-lg">
+  <div
+    [style.width.%]="progress"
+    class="flex {{
+      color.innerBar
+    }} my-2 mx-2 transition-width duration-500 ease-in-out rounded-t-md rounded-b-md shadow-xl"
+  >
+    <div class="flex items-center pl-2 py-1 text-gray-900 font-bold text-4">
+      {{ progress }}%
     </div>
   </div>
 </div>
--- a/libs/ui/src/lib/progress-bar/progress-bar.component.html
+++ b/libs/ui/src/lib/progress-bar/progress-bar.component.html
@@ -1,0 +1,10 @@
+<div class="flex bg-white rounded-t-lg rounded-b-lg">
+  <div class="flex w-full bg-blue-600 rounded-t-lg rounded-b-lg my-2 mx-2">
+    <div
+      [style.width.%]="validValue"
+      class="bg-blue-400 my-1 mx-1 transition-width duration-500 ease-in-out rounded-t-lg rounded-b-lg"
+    >
+      <span class="pl-1 py-1 font-bold text-4">{{ validValue }}%</span>
+    </div>
+  </div>
+</div>
--- a/libs/ui/src/lib/progress-bar/progress-bar.component.ts
+++ b/libs/ui/src/lib/progress-bar/progress-bar.component.ts
@@ -1,5 +1,10 @@
 import { Component, Input, OnInit } from '@angular/core'
 
+interface ColorScheme {
+  outerBar: string
+  innerBar: string
+}
+
 @Component({
   selector: 'ui-progress-bar',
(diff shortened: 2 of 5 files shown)


libs/ui/src/lib/ui.module.ts
```diff
@@ -26,6 +27,7 @@ import { CatalogTitleComponent } from './catalog-title/catalog-title.component'
     RecordMetricComponent,
     ResultsListComponent,
     CatalogTitleComponent,
+    ProgressBarComponent,
   ],
   imports: [BrowserModule, TranslateModule.forChild()],
   exports: [
```
fgravin, line 33: should be exported as well to be used in other library/applications

libs/ui/src/lib/progress-bar/progress-bar.stories.ts
```diff
@@ -0,0 +1,23 @@
+import { I18nModule } from '@lib/common'
+import { withA11y } from '@storybook/addon-a11y'
+import { action } from '@storybook/addon-actions'
+import { number, object, text, withKnobs } from '@storybook/addon-knobs'
+import { Meta, moduleMetadata, Story } from '@storybook/angular'
+import { ProgressBarComponent } from './progress-bar.component'
+
+const moduleMetadatas = {
+  imports: [I18nModule],
```
fgravin, line 9: no text so not needed i think

fgravin (changes requested):
Great starts ! The code is good and clean thanks.
However the look is not exactly what we expect.

By the way, it would be great to have a color as input, either `primary, secondary`, or `blue, green` ...
Then in the component, you compute the brackground color and the progress color (lighten) from the input color. So we can declined the component in many colors.
You can have a look at the UI/button.components that do something like this.

libs/ui/src/lib/progress-bar/progress-bar.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="flex bg-white rounded-t-lg rounded-b-lg">
```
fgravin, line 1: I would put the white outcircle in the component, i think it's the responsability of the container.
The component should just display the progress bar.

libs/ui/src/lib/progress-bar/progress-bar.component.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Component({
+  selector: 'ui-progress-bar',
+  templateUrl: './progress-bar.component.html',
+  styleUrls: ['./progress-bar.component.css'],
+})
+export class ProgressBarComponent implements OnInit {
+  @Input() value = 0
+
+  get validValue() {
```
fgravin, line 11:
```suggestion
  get progress() {
```

libs/ui/src/lib/progress-bar/progress-bar.component.html
```diff
@@ -0,0 +1,10 @@
...
+  <div class="flex w-full bg-blue-600 rounded-t-lg rounded-b-lg my-2 mx-2">
+    <div
+      [style.width.%]="validValue"
+      class="bg-blue-400 my-1 mx-1 transition-width duration-500 ease-in-out rounded-t-lg rounded-b-lg"
+    >
+      <span class="pl-1 py-1 font-bold text-4">{{ validValue }}%</span>
```
fgravin, line 7: use `grey-900` black hurts a bit :)

libs/ui/src/lib/progress-bar/progress-bar.component.html
```diff
@@ -0,0 +1,10 @@
+<div class="flex bg-white rounded-t-lg rounded-b-lg">
+  <div class="flex w-full bg-blue-600 rounded-t-lg rounded-b-lg my-2 mx-2">
+    <div
+      [style.width.%]="validValue"
+      class="bg-blue-400 my-1 mx-1 transition-width duration-500 ease-in-out rounded-t-lg rounded-b-lg"
```
fgravin, line 5: I would round less the inner progress than the outter, visually it's better

libs/ui/src/lib/progress-bar/progress-bar.component.ts
```diff
@@ -1,17 +1,43 @@
...
+  get color(): ColorScheme {
+    switch (this.type) {
+      case 'default':
+        return {
+          outerBar: 'bg-blue-500',
+          innerBar: 'bg-blue-300',
+        }
+      case 'primary':
+        return {
+          outerBar: 'bg-green-500',
```
fgravin, line 30: no, if you pass `primary` as the input, you have to use primary colors and not forced blue or green ones.
primary is a css variable where you can put any value in, so the website keep the same theme.
Use something like `bg-primary` `bg-primary-lighter` (like in `ButtonComponent`)

Of course in your case maybe primary will not be blue, but in storybook you can in `Color scale` story change the primary color .. try to pick up a blue and return in progress bar story to see the result.

Otherwise you could pass `'blue'` as an input and build colors with ``bg-${inputColor}-500`` for instance.

fgravin (commented):
rebase on master and it should be mergeable !
```
git fetch upstream
git rebase upstream/master
git push -force origin progress-bar-for-data-analysis
```

fgravin (approved):
great work, thanks

PR #46, opened 2020-11-06
refactor: Change structure gn-api library. Change scripts for generat…

y-vynohradov:
refactor: Change structure of gn-api library. Change scripts for generating open openapi.


fgravin (commented):
Refactoring looks good but tests are failing, can you investigate ?
Did you move content to gn4 or did you generate it with the script ?

y-vynohradov:
> Refactoring looks good but tests are failing, can you investigate ?
I have fixed tests. 
> Did you move content to gn4 or did you generate it with the script ?
I generated content with sctript.

libs/gn-api/src/lib/gn4/openapi/model/regionResponse.api.model.ts
```diff
@@ -19,5 +19,4 @@ export interface RegionResponseApiModel {
   hasGeom?: boolean
   categoryId?: string
   id?: string
-  id?: string
```
fgravin: hmm, if it is auto generated, it will be put again at next generation ?
can you try to rerun the script for gn4 ? If the error persist, I will contact maintener of the openapi specification

y-vynohradov: So, I reruned the script and error persisted.

image

fgravin: @fxprunayre did you have to change something manually after api generation to fix some issues ?

Maybe we have an issue with this definition

```yml
        '@id':
          type: string
          xml:
            attribute: true
        id:
          type: string
```
where `@id` is put in the model as `id` which is already present.

fxprunayre: > @fxprunayre did you have to change something manually after api generation to fix some issues ?

I don't think so. Checking https://apps.titellus.net/geonetwork/doc/api/index.html
We can probably safely remove https://github.com/geonetwork/core-geonetwork/blob/4.0.x/services/src/main/java/org/fao/geonet/api/regions/ListRegionsResponse.java#L111-L112 and maybe remove the `@` for hasGeom/categoryId/id which does not make much sense?

fgravin: I don't want to edit the spec file manually, everything should be done in the Java so the generated file is Ok.
@fxprunayre could you provide a new file spec with those changes ?
I merge this PR cause it's not related, the refactoring is ok.

fgravin (approved):
Note that the task `generate-api:datafeeder` will fail but we can process that in a next PR when you'll start to have a spec file.

libs/gn-api/src/lib/gn4/openapi/model/regionResponse.api.model.ts
fgravin: it's not urgent at all, it can come with a next upgrade though

fxprunayre: > I will contact maintener of the openapi specification

@fgravin is not a maintainer of GeoNetwork? ;)
You can check https://github.com/geonetwork/core-geonetwork/pull/5179

PR #47, opened 2020-11-06
Aggregations in search state

LHBruneton-C2C:
Add aggregations to search state (request and response)
Add aggregations to query payload
Read aggregations from response
Display facets in app

### DIFF
--- a/libs/search/src/lib/facets/facets-container/facets-container.component.ts
+++ b/libs/search/src/lib/facets/facets-container/facets-container.component.ts
@@ -1,8 +1,12 @@
 import { Component, Input, OnInit } from '@angular/core'
-import { select, Store } from '@ngrx/store'
+import { Store } from '@ngrx/store'
+import { combineLatest } from 'rxjs'
 import { map } from 'rxjs/operators'
 import { SearchState } from '../../state/reducer'
-import { getSearchResultsAggregations } from '../../state/selectors'
+import {
+  getSearchConfigAggregations,
+  getSearchResultsAggregations,
+} from '../../state/selectors'
 import { FacetsService } from '../facets.service'
 
 @Component({
@@ -13,17 +17,14 @@ import { FacetsService } from '../facets.service'
 export class FacetsContainerComponent implements OnInit {
   @Input() uiConfig = 'srv'
 
-  // TODO: fixture
-  requestAggregations = {
-    tag: { terms: { field: 'tag', include: '.*', size: 10 } },
-  }
-
-  models$ = this.store.pipe(
-    select(getSearchResultsAggregations),
-    map((responseAggregations: any) => {
+  models$ = combineLatest([
+    this.store.select(getSearchConfigAggregations),
+    this.store.select(getSearchResultsAggregations),
+  ]).pipe(
+    map(([configAggregations, resultsAggregations]) => {
       return this.facets.createFacetModel(
-        this.requestAggregations,
-        responseAggregations,
+        configAggregations,
+        resultsAggregations,
         false
       )
     })
--- a/libs/search/src/lib/facets/facets-container/facets-container.component.ts
+++ b/libs/search/src/lib/facets/facets-container/facets-container.component.ts
@@ -1,4 +1,9 @@
 import { Component, Input, OnInit } from '@angular/core'
+import { select, Store } from '@ngrx/store'
+import { map } from 'rxjs/operators'
+import { SearchState } from '../../state/reducer'
+import { getSearchAggregations } from '../../state/selectors'
+import { FacetsService } from '../facets.service'
 
 @Component({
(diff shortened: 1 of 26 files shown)


libs/search/src/lib/facets/facets-container/facets-container.component.ts
```diff
@@ -1,4 +1,9 @@
 import { Component, Input, OnInit } from '@angular/core'
+import { select, Store } from '@ngrx/store'
+import { map } from 'rxjs/operators'
+import { SearchState } from '../../state/reducer'
+import { getSearchResultsAggregations } from '../../state/selectors'
+import { FacetsService } from '../facets.service'
```
fxprunayre, line 6: Would it make sense to use Aggregations or Facets everywhere for consistency ?

If yes, I would vote for Aggregation. 

Components in https://github.com/geonetwork/geonetwork-ui/pull/38 could be Aggs > Agg > AggBuckets > AggBucket

fgravin: `aggregations` is a way elastic aggregates the data, `facet` is a UI representation, other components like `metrics` can use aggregations: I would keep the distanciation.

libs/ui/src/lib/facets/facet-block/facet-block.component.ts
```diff
@@ -53,3 +53,13 @@ export class FacetBlockComponent implements OnInit {
     return this.model.more
   }
 }
+
+@Component({ selector: 'ui-facet-block', template: '' })
```
fgravin, line 57: I wouldn't put that here, why do you need this stub byt the way ?

libs/search/src/lib/state/reducer.ts
```diff
@@ -1,18 +1,33 @@
...
+import { RecordSummary, SearchFilters } from '@lib/common'
 import * as fromActions from './actions'
 
 export const SEARCH_FEATURE_KEY = 'searchState'
 
 export interface SearchState {
-  params: SearchParams
-  sortBy?: string
-  results: RecordSummary[]
+  requestParams: {
```
fgravin, line 7: I would change the structure of the state following our discussions.

fgravin (changes requested):
Good so far, the smart logic is well connected to the state.
Some refactoring of the state structure is expected though.
Thanks

fgravin:
@LHBruneton-C2C I don't see the translation of `facets.block.title.tag`

LHBruneton-C2C:
> @LHBruneton-C2C I don't see the translation of `facets.block.title.tag`

Neither do I, even though it was working OK before the rebase with the last commits on master. I will look into it.

libs/ui/src/lib/facets/facet-block/facet-block.component.ts
LHBruneton-C2C: I use it to test the facet-list component without additional schemas. Should we just use the NO_ERRORS_SCHEMA everywhere ?

fgravin: It depends what you do want to test. But usually, when testing a component, we just test the inner behavior of the component, and don't dive deeper in the components tree.

fgravin:
> Neither do I, even though it was working OK before the rebase with the last commits on master. I will look into it.

Ok, also I don't see the facet item text color change on hover, I just see the checkbox the slighlty get darker.

LHBruneton-C2C:
> I don't see the facet item text color change on hover, I just see the checkbox the slighlty get darker.
Don't know what's happening here. On my machine, no diff with the branch, npm run storybook, I see the color change on hover. I'm using chromium. What platform are you trying on ? Is this the result of the storybook compiled by the CI ?

LHBruneton-C2C:
> No, I'm trying on `app-search`

Well, we have a "It works on my machine" issue... ¯\_(ツ)_/¯
Is there a continuous deployment on this project ?

fgravin:
> Is there a continuous deployment on this project ?

storybook but disabled for now ..
I will try to find time to fix it and also deploy the applications as well

LHBruneton-C2C:
> > Is there a continuous deployment on this project ?
> 
> storybook but disabled for now ..
> I will try to find time to fix it and also deploy the applications as well

Thanks !

libs/ui/src/lib/facets/facet-block/facet-block.component.ts
LHBruneton-C2C: As explained here [https://angular.io/guide/testing-components-scenarios#nested-component-tests](url), the stub is only here to test the nested component interface, nothing deeper.

fgravin (approved):
Excellent.
Thanks for the refactoring and pluggin everything together.
Now we need to perform a search on facet item click :)

PR #49, opened 2020-11-09
Load more on aggregation

LHBruneton-C2C:
Add action and effect to load more on one aggregation
Output from dumb component
Dispatch action from container

### DIFF
--- a/libs/search/src/lib/state/reducer.ts
+++ b/libs/search/src/lib/state/reducer.ts
@@ -144,7 +144,7 @@ export function reducer(
     }
     case fromActions.REQUEST_MORE_ON_AGGREGATION: {
       const clone = JSON.parse(JSON.stringify(state.config.aggregations))
-      clone[action.key].terms.size += 20
+      clone[action.key].terms.size += action.increment
 
       return {
         ...state,
--- a/libs/search/src/lib/state/reducer.ts
+++ b/libs/search/src/lib/state/reducer.ts
@@ -154,6 +154,18 @@ export function reducer(
         },
       }
     }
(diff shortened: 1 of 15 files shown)


libs/search/src/lib/state/reducer.ts
```diff
@@ -127,6 +127,18 @@ export function reducer(
         },
       }
     }
+    case fromActions.REQUEST_MORE_ON_AGGREGATION: {
+      const clone = JSON.parse(JSON.stringify(state.config.aggregations))
+      clone[action.key].terms.size += 20
```
fgravin, line 132: no logic in reducers, `20` should be in the payload.

fgravin (changes requested):
Looks great except one thing to change in the new Action.
Can't test it cause the api has changed on titellus.

fgravin:
Great @LHBruneton-C2C, since the rebase, load more action set the `response.aggs` instead of patching it.
Also, see if you can just load the extra items instead of loading them all, using the `from` and size` `search options.
Thanks

LHBruneton-C2C:
On rebasing, I saw that there was a dependency from lib-ui to lib-search, in facet-block component, `import { AggregationsTypesEnum } from '../../../../../search/src/lib/facets/facets.model'` => is it OK ?

fgravin:
> On rebasing, I saw that there was a dependency from lib-ui to lib-search, in facet-block component, `import { AggregationsTypesEnum } from '../../../../../search/src/lib/facets/facets.model'` => is it OK ?

not it's not
model should be shared if needed

fgravin (approved):
Excellent thanks

PR #60, opened 2020-11-19
Results preview improvements & consistency

fxprunayre:
* Add a base record preview component 
* Extend base preview component for card/list/text
* Enum / ResultListLayout / Rename BLOCK > CARD to be consistent with component name
* Add a component for the thumbnail
* Consistent styling in each modes
* Add a target attribute to open a record in another browser tab or specific iframe with a name

### DIFF
--- a/libs/search/src/lib/elasticsearch/elasticsearch.mapper.ts
+++ b/libs/search/src/lib/elasticsearch/elasticsearch.mapper.ts
@@ -15,7 +15,7 @@ export class ElasticsearchMapper {
       title: hit._source.resourceTitleObject?.default || 'no title',
       abstract: hit._source.resourceAbstractObject?.default || 'no abstract',
       thumbnailUrl: this.getFirstValue(hit._source.overview)?.url || '',
-      url: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
+      metadataUrl: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
       downloadable: (hit as any).download,
       viewable: (hit as any).view,
       logoUrl: `/geonetwork${hit._source.logo}`,
--- a/libs/ui/src/lib/record-thumbnail/record-thumbnail.component.html
+++ b/libs/ui/src/lib/record-thumbnail/record-thumbnail.component.html
@@ -1,0 +1,23 @@
+<div class="h-full h-full relative flex-shrink-0 overflow-hidden">
+  <svg
+    class="text-gray-200 absolute h-12 w-12 transform -translate-x-1/2 -translate-y-1/2"
+    style="top: 50%; left: 50%;"
+    xmlns="http://www.w3.org/2000/svg"
+    fill="none"
+    viewBox="0 0 24 24"
+    stroke="currentColor"
+  >
+    <path
+      stroke-linecap="round"
+      stroke-linejoin="round"
+      stroke-width="2"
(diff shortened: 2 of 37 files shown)


libs/ui/src/lib/record-thumbnail/record-thumbnail.component.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(RecordThumbnailComponent)
+    component = fixture.componentInstance
+    fixture.detectChanges()
+  })
+
+  it('should create', () => {
+    expect(component).toBeTruthy()
+  })
```
fgravin, line 23: You should add some tests.
- img tag with correct url
- no img if no url

at least

libs/ui/src/lib/record-thumbnail/record-thumbnail.component.html
```diff
@@ -0,0 +1,23 @@
...
+      stroke-linejoin="round"
+      stroke-width="2"
+      d="M4 16l4.586-4.586a2 2 0 012.828 0L16 16m-2-2l1.586-1.586a2 2 0 012.828 0L20 14m-6-6h.01M6 20h12a2 2 0 002-2V6a2 2 0 00-2-2H6a2 2 0 00-2 2v12a2 2 0 002 2z"
+    />
+  </svg>
+  <img
+    *ngIf="thumbnailUrl"
+    class="relative h-full w-full object-cover object-left-top"
+    [src]="thumbnailUrl"
+    alt="thumbnail"
```
fgravin, line 21: Add a translation

libs/ui/src/lib/record-thumbnail/record-thumbnail.stories.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const Template: Story<RecordThumbnailComponent> = (args) => ({
+  component: RecordThumbnailComponent,
+  props: args,
+})
+
+export const RecordThumbnailComponentStory = Template.bind({})
+
+RecordThumbnailComponentStory.args = {
+  thumbnailUrl:
```
fgravin, line 20: use knobs so we can change the input from the story.

libs/search/src/lib/elasticsearch/elasticsearch.mapper.ts
```diff
@@ -15,7 +15,7 @@ export class ElasticsearchMapper {
       title: hit._source.resourceTitleObject?.default || 'no title',
       abstract: hit._source.resourceAbstractObject?.default || 'no abstract',
       thumbnailUrl: this.getFirstValue(hit._source.overview)?.url || '',
-      url: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
+      metadataUrl: `/geonetwork/srv/eng/catalog.search#/metadata/${hit._source.uuid}`,
```
fgravin, line 18: so you didn't manage to give the correct path url ?

fgravin (commented):
Thanks for the refactoring.
All look good, some minor comments though.
Take the good habit to test the components ;) You can get examples from other components (Css query selector).

libs/search/src/lib/elasticsearch/elasticsearch.mapper.ts
fxprunayre: not yet.

libs/ui/src/lib/record-thumbnail/record-thumbnail.component.html
fxprunayre: in fact thumbnail in the model should have a label and be 0..n - so I'll wait for refactoring this before adding label which may be useless after.

PR #66, opened 2020-12-07
Search results / Layout selector.

fxprunayre:
image

### DIFF
--- a/libs/search/src/lib/results-layout/results-layout.component.ts
+++ b/libs/search/src/lib/results-layout/results-layout.component.ts
@@ -1,0 +1,28 @@
+import { Component, OnInit } from '@angular/core'
+import { select, Store } from '@ngrx/store'
+import { UpdateResultsLayout } from '../state/actions'
+import { SearchState } from '../state/reducer'
+import { getSearchResultsLayout } from '../state/selectors'
+import { ResultsListLayout } from '@lib/common'
+
+@Component({
+  selector: 'search-results-layout',
+  templateUrl: './results-layout.component.html',
+})
+export class ResultsLayoutComponent implements OnInit {
+  choices = Object.values(ResultsListLayout).map((v) => {
+    return {
+      label: v,
+      value: v,
+    }
+  })
+  currentLayout$ = this.store.pipe(select(getSearchResultsLayout))
+
+  constructor(private store: Store<SearchState>) {}
+
+  ngOnInit(): void {}
+
+  change(layout: any) {
+    this.store.dispatch(new UpdateResultsLayout(layout))
+  }
+}
--- a/libs/search/src/lib/results-layout/results-layout.component.ts
+++ b/libs/search/src/lib/results-layout/results-layout.component.ts
@@ -1,6 +1,6 @@
 import { Component, OnInit } from '@angular/core'
 import { select, Store } from '@ngrx/store'
-import { UpdateResultsLayout } from '../state/actions'
+import { SetResultsLayout } from '../state/actions'
 import { SearchState } from '../state/reducer'
 import { getSearchResultsLayout } from '../state/selectors'
 import { ResultsListLayout } from '@lib/common'
@@ -23,6 +23,6 @@ export class ResultsLayoutComponent implements OnInit {
   ngOnInit(): void {}
 
   change(layout: any) {
-    this.store.dispatch(new UpdateResultsLayout(layout))
+    this.store.dispatch(new SetResultsLayout(layout))
   }
 }
--- a/libs/search/src/lib/results-list/results-list.container.component.ts
+++ b/libs/search/src/lib/results-list/results-list.container.component.ts
@@ -2,8 +2,16 @@ import { Component, Input, OnInit } from '@angular/core'
 import { BootstrapService, ResultsListLayout } from '@lib/common'
 import { select, Store } from '@ngrx/store'
 import { SearchState } from '../state/reducer'
-import { getSearchResults, getSearchResultsLoading } from '../state/selectors'
-import { RequestMoreResults, SetConfigAggregations } from '../state/actions'
+import {
+  getSearchResults,
+  getSearchResultsLayout,
+  getSearchResultsLoading,
+} from '../state/selectors'
+import {
+  RequestMoreResults,
+  SetConfigAggregations,
+  UpdateFilters,
+} from '../state/actions'
 import { map, pluck, take, tap } from 'rxjs/operators'
 
 @Component({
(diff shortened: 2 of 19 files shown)


libs/search/src/lib/state/reducer.ts
```diff
@@ -10,6 +10,7 @@ export interface SearchState {
   params: {
     filters: SearchFilters
     sortBy?: string
+    resultsLayout?: string
```
fgravin, line 13: It's not a search param, actually it has nothing related with the search process in itself.
I really don't know where to put this to be honest.. It should be in application state, but it's true that putting it somewhere in search state allow smart component to react from it.

At least put it at the root but not in params.

libs/search/src/lib/state/reducer.ts
```diff
@@ -54,6 +55,15 @@ export function reducer(
...
     }
+    case fromActions.UPDATE_RESULTS_LAYOUT: {
+      return {
+        ...state,
+        params: {
+          ...state.params,
+          resultsLayout: action.resultsLayout,
+        },
+      }
+    }
```
fgravin, line 66: tests ?

libs/search/src/lib/state/actions.ts
```diff
@@ -21,6 +22,12 @@ export class SortBy implements Action {
   constructor(public sortBy: string) {}
 }
 
+export class UpdateResultsLayout implements Action {
+  readonly type = UPDATE_RESULTS_LAYOUT
```
fgravin, line 26: `SET_RESULTS_LAYOUT` is more appropriate

libs/search/src/lib/results-list/results-list.container.component.ts
```diff
@@ -15,6 +23,7 @@ export class ResultsListContainerComponent implements OnInit {
   @Input() layout: ResultsListLayout = ResultsListLayout.CARD
```
fgravin, line 23: is it useless then ?

fgravin (changes requested):
Thanks for the work, it's globally good but details to fix from here to there.

libs/search/src/lib/state/reducer.ts
fgravin, line 13: add a value in default state

libs/search/src/lib/results-layout/results-layout.component.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@Component({
+  selector: 'search-results-layout',
+  templateUrl: './results-layout.component.html',
+})
+export class ResultsLayoutComponent implements OnInit {
+  choices = Object.values(ResultsListLayout).map((v) => {
+    return {
+      label: v,
+      value: v,
```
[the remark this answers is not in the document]
fxprunayre: Like in the sort by https://github.com/geonetwork/geonetwork-ui/blob/master/libs/search/src/lib/sort-by/sort-by.component.ts#L19 ;).

2 things about translations:
* We should clean it up with proper key scheme before it is getting too large https://github.com/geonetwork/geonetwork-ui/blob/master/apps/search/src/assets/i18n/en.json
* I'm still not 100% sure how this will be done at the end. Those type of values can't be really hardcoded in the app - search fields will change and may come from somewhere else (eg. https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/components/elasticsearch/directives/FacetDirective.js#L129 for facet in v4)

So I would dedicate some time to a proper i18n PR to make proposal for a key scheme, cleanup and how to handle translations (not only those for buttons). Something to put on the agenda for the sprint next week. Ok for you?

libs/search/src/lib/state/reducer.ts
fxprunayre: > It should be in application state,

No because it is related to a search. If you have 2 components making searches in the same page, one may have a layout and the other another one.

It could be considered related to the "_source" - we may link _source with only the fields required for a layout. But that is not really present in the config for now.

Ok moving it to root.

libs/search/src/lib/state/actions.ts
fxprunayre: Same for SET_SORT_BY instead of SORT_BY ?

fxprunayre: UPDATE_FILTERS ?

fgravin: Yes `SET_SORTBY`, we keep `UPDATE_FILTERS` cause the payload is a patch.

libs/search/src/lib/results-list/results-list.container.component.ts
fxprunayre: Changed to have this as default value.

libs/search/src/lib/results-layout/results-layout.component.ts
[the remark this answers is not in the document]
fxprunayre: Added to the agenda https://github.com/geonetwork/core-geonetwork/wiki/GeoNetwork-UI-and-microservices-codesprint-December-2020

libs/search/src/lib/state/actions.ts
```diff
@@ -17,13 +17,13 @@ export class UpdateFilters implements Action {
 }
 
 export class SortBy implements Action {
-  readonly type = SORT_BY
+  readonly type = SET_SORT_BY
```
fgravin, line 20: Should change action class export name as well to `SetSortBy`

libs/search/src/lib/results-list/results-list.container.component.ts
fgravin: I don't get it. `layout` input is not used anymore is it ?

fgravin (changes requested):
Thanks for the change.
2 last comments to address and you can merge

libs/search/src/lib/results-list/results-list.container.component.ts
fxprunayre: If you want to make an app using a custom layout without the layout switcher you can do
```
      <search-results-list-container layout="CARD"></search-results-list-container>
```
no?

fgravin: But in your code the input is not used anyway or i missed something.
The layout is not binded to the result-layout component but to the state. Your app could dispatch `SetResultLayout` action by its own.

fxprunayre: https://github.com/geonetwork/geonetwork-ui/pull/66/files#diff-040ddc6cd03fe81fc601cb3edff163da7c940c7a551b5e21c66f25d29fe0d11aR45

fgravin: ok sorry

fgravin (approved):

PR #67, opened 2020-12-07
Add filter and size parameter to results list

fxprunayre:
For now tentative. Filter is working in "normal" mode but not in webcomponent mode.

In normal mode:
```
      <search-results-list-container
        size="3"
        filter="GRID"
      ></search-results-list-container>
```

In webcomponent mode

```
 npm run serve:wc -- gn-results-list
```
the file built looks to be empty.
image

When building web component only colors stuffs are added and can't figure out how to add the useful properties. Story book starts locally but display grey pages only, with color knobs only.

Not sure to understand the difference between angular.stories.ts and elements.stories.ts

### DIFF
--- a/libs/common/src/lib/i18n/i18n.module.ts
+++ b/libs/common/src/lib/i18n/i18n.module.ts
@@ -11,6 +11,31 @@ export function HttpLoaderFactory(http: HttpClient) {
   return new TranslateHttpLoader(http)
 }
 
+export const LANG_3_TO_2_MAPPER = {
+  eng: 'en',
+  dut: 'nl',
+  fre: 'fr',
+  ger: 'de',
+  kor: 'ko',
+  spa: 'es',
+  cze: 'cs',
+  cat: 'ca',
+  fin: 'fi',
+  ice: 'is',
+  ita: 'it',
+  por: 'pt',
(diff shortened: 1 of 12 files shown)


fgravin:
> Not sure to understand the difference between angular.stories.ts and elements.stories.ts

GnResultList component is before all an Angular component, angular storybook tests it as an Angular component, as any other component from libs or apps. Once built, it is also exported as a webcomponent. WC storybook tests the web component mode and not the angular component mode. But for it to work, the web component must be built in dev mode (don't figure out why it is mandatory yet).

> the file built looks to be empty.

run ` npm run serve:wc -- gn-results-list --build` instead .. maybe this could be the default behavior and we could remove the `--build` ..

fgravin:
@fxprunayre I reworked the PR.
Run the serve command and check on http://localhost:8001/ it should work with correct styles.
The input parameters work well.
I have added a service to compute the metadata Url properly.

Let me know if it works for you.

fgravin:
@fxprunayre ? ok for a merge ?

fxprunayre:
Sure and https://github.com/geonetwork/geonetwork-ui/pull/66 also ?

PR #68, opened 2020-12-09
fix: search aggs property

fxprunayre:
See https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations.html


fgravin (approved):

fgravin:
Do you think we need to change the key in the state too ?

fxprunayre:
Don't know if it will change something or make some stuff more easy ?
Not sure why it is aggs in query and aggregations in response.

fxprunayre:
It may avoid to make mistakes if we keep the same as ES ? I remember some times blindly searching for typos for long times ... :)

PR #72, opened 2020-12-16
Optimize search API calls on page loading

fgravin:
Slight improvments and refactoring to avoid parasite search api calls:

- fix of `shareReplay` usage for `/me` api calls, that were called once per search request
- move `RequestMore` action dispatch out of `results-list-container` component. It is not the responsability of the component to perform a search request, it should be done at a higher level like during the application init. This was causing parasite search api calls while the initial state was not ready yet.

gn-results-list web component now just perform 2 request on load as expected
- `me`
- `_search`

### DIFF
--- a/apps/search/src/app/app.component.ts
+++ b/apps/search/src/app/app.component.ts
@@ -1,15 +1,41 @@
-import { Component } from '@angular/core'
-import { ColorService } from '@lib/common'
+import { Component, OnInit } from '@angular/core'
+import { BootstrapService, ColorService } from '@lib/common'
+import {
+  RequestMoreResults,
+  SearchState,
+  SetConfigAggregations,
+} from '@lib/search'
+import { Store } from '@ngrx/store'
+import { map, pluck, take, tap } from 'rxjs/operators'
 
 @Component({
   selector: 'app-root',
   templateUrl: './app.component.html',
   styleUrls: ['./app.component.css'],
 })
-export class AppComponent {
+export class AppComponent implements OnInit {
   title = 'search'
 
-  constructor() {
+  constructor(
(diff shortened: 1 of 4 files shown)


apps/search/src/app/app.component.ts
```diff
@@ -1,15 +1,41 @@
...
+  ngOnInit(): void {
+    this.bootstrap
+      .uiConfReady('srv')
+      .pipe(
+        take(1),
+        map((config) => config.mods.search.facetConfig),
+        // TODO: make the config work not just for tag
+        pluck('tag'),
+        tap((tagConfig) => {
+          this.store.dispatch(new SetConfigAggregations({ tag: tagConfig }))
```
LHBruneton-C2C, line 35: Careful here, `tag` has been changed to `tag.default` somehow.

LHBruneton-C2C (approved):
Thanks for the improvement ! And I understand better shareReplay now.
Be careful when rebasing, as the aggregations config has changed (see my comment).

apps/search/src/app/app.component.ts
fxprunayre: The field name change is related to https://github.com/geonetwork/core-geonetwork/pull/5193.
Why only adding the tag here ? it should be an array of facet as defined in the config no?

LHBruneton-C2C: Yes, see my TODO comment just above. For now, the aggregations configuration can't be used as is. There are some property to remove concerning roles and some other stuff.

fgravin: > Why only adding the tag here ? it should be an array of facet as defined in the config no?

I just moved code. The facet configuration is handled in other PRs.

PR #73, opened 2020-12-17
[datafeeder]: upload api

fgravin:
Plug datafeeder upload form to api
- api client genration
- refactoring
- api call


fgravin:
@y-vynohradov I have integrated the api, but the datafeeder backend is not yet working.
Check my refactoring to understand how to improve the workflow in upload-component. Also the translation in the component was way too complex.

Please add tests in upload-data.component
Thanks

apps/datafeeder/src/app/presentation/pages/dataset-validation-page/dataset-validation-page.ts
```diff
@@ -22,90 +30,47 @@ export class DatasetValidationPageComponent implements OnInit, OnDestroy {
...
-        [0.094757080078125, 47.18597932702905],
-        [0.070037841796875, 47.1813125359862],
-        [0.05630493164062499, 47.17757880776958],
-        [0.039825439453125, 47.17477833929903],
-        [0.01922607421875, 47.16637604771991],
-        [-0.05630493164062499, 47.127147777248645],
-      ],
-    },
-  }
+  geoJSONData: object
```
fgravin, line 33: what type is it ?

fgravin (commented):
Thanks @tonio it looks great.
One thing I'm not sure about: the progress bar page fetches the status while it's DONE, then move to validation page that fetch again the DONE status ? And both are handling the error ?
It does make sense because we can load any page with its route, but within the natural workflow, it should make sense maybe to save the status in a state or service.
Maybe we can consider that later, I think overall we will need to persist things, and to secure the page access as well with guards.

PR #74, opened 2020-12-17
suggestion for a mapping file from core strings

pvgenuchten:
in order to persist as many of current translations, i suggest to set up a mapping from core-app to ui-app, so we can develop a script to import existing translations into transifex. We can use this pr to discuss the approach, and if positive also draft a script to facilitate the actual import


fxprunayre:
For now translations for geonetwork-ui are defined in https://github.com/geonetwork/core-geonetwork/blob/main/web-ui/src/main/resources/catalog/locales/en-gnui.json and goes to transifex.

PR #75, opened 2020-12-17
web component: watch input changes

fgravin:
Show how to update the behavior of the web component from the source page, by changing wc inputs values.
Angular component should handle input changes within `onChanges` implementation. 

The PR implements the `onChanges` method in `gn-results-list component`, it must follow some rules
- don't call api request before the web component has initiliazed `API_BASE_PATH`
- be sure to trigger the change detection when it is expected, because the web component execution (even though it's in an angular custom element) is outside of an Angular zone, meaning the change detection is not triggered.

In my case, I had to watch for the `loading` property of the search state, informing me that the search is done. When it's done, I trigger the change detection otherwise the results are not shown.
I added a setTimeout to subscribe to loading selector after all other subscription to the same observable to be sure all angular components has been updated from loading observable before I run the change detection.

This is a bit complex overall, but no choices.

Here an exemple on how to update the input from source page:

```html
 <div>
    <button id="changeSizeBtn">Change size</button>
 </div>
 <gn-results-list api-url="https://apps.titellus.net/geonetwork/srv/api"></gn-results-list>

 <script>
    const wc = document.getElementsByTagName('gn-results-list')[0]
    const btn = document.getElementById('changeSizeBtn')
    btn.addEventListener('click', () => wc.size = 3)
 </script>
```

### DIFF
--- a/apps/search/src/app/app.component.spec.ts
+++ b/apps/search/src/app/app.component.spec.ts
@@ -1,14 +1,40 @@
-import { TestBed, async } from '@angular/core/testing'
-import { RouterTestingModule } from '@angular/router/testing'
-import { AppComponent } from './app.component'
 import { NO_ERRORS_SCHEMA } from '@angular/core'
+import { async, TestBed } from '@angular/core/testing'
+import { RouterTestingModule } from '@angular/router/testing'
+import { BootstrapService } from '@lib/common'
+import { EffectsModule } from '@ngrx/effects'
+import { StoreModule } from '@ngrx/store'
+import { of } from 'rxjs'
+import { AppComponent } from './app.component'
 
+const configFacetMock = {
+  mods: {
+    search: {
+      facetConfig: {
+        tag: {},
+      },
+    },
+  },
+}
+const boostrapServiceMock = {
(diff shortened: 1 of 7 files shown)


fgravin:
@fxprunayre for the review, just check 2 last commits

fxprunayre (approved):
Works well. So the only things you are not really convinced about is the `setTimeout(() => {` which relates to the init of API_BASE_PATH, right?

fgravin:
> So the only things you are not really convinced about is the `setTimeout(() => {` which relates to the init of API_BASE_PATH, right?

No, the init of the `BASE_PATH` is ok, I ensure it's done once and before any `ngInit` or `ngChanges`.

The timeout is here so the web component is the last one subscribing to the sate change, so it will be the last listeners that will be called. For instance, if you don't do that, it will be the first to register, then the change detection will be triggered before the result-list dumb component will be updated, so it will be useless.
The goal is that all angular subscriber to state change do their job as usual, once it's done, we ask the web component to update (what will update the nested angular component accordingly).

It's logic overall, it's like that anyway, it's just that it would have been easier not to have to manage the change detection life cycle manually.

PR #76, opened 2020-12-17
Make the styling more uniform

MichelGabriel:
Make the styling more uniform:
- introduce GN CSS variables
- same font size
- same font colors
- better aligning elements in search app
- hover action for Card
- improved focus border color

### DIFF
--- a/libs/ui/src/lib/record-preview-title/record-preview-title.component.html
+++ b/libs/ui/src/lib/record-preview-title/record-preview-title.component.html
@@ -1,6 +1,6 @@
 <!-- Record list item: List Small -->
 <div
-  class="h-10 border-b border-gray-200 transition duration-200 hover:bg-gray-100 border-gray-300"
+  class="h-10 w-full bg-white border border-gray-300 transition duration-200 hover:bg-gray-100 rounded-sm"
 >
   <a
     [href]="record.metadataUrl"
@@ -13,7 +13,7 @@
         [thumbnailUrl]="record.thumbnailUrl"
       ></ui-record-thumbnail>
 
-      <div class="flex-grow h-full px-1 py-1 flex flex-col overflow-hidden">
+      <div class="flex-grow px-3 py-1 flex flex-col overflow-hidden">
         <h1
           class="title-font text-sm font-bold text-gray-900 truncate md:overflow-clip"
         >
--- a/libs/ui/src/lib/record-preview-title/record-preview-title.component.html
+++ b/libs/ui/src/lib/record-preview-title/record-preview-title.component.html
@@ -1,6 +1,6 @@
 <!-- Record list item: List Small -->
 <div
-  class="h-10 w-full bg-white border border-gray-300 transition duration-200 hover:bg-gray-100 rounded-sm"
+  class="h-10 w-full border-b border-gray-200 transition duration-200 hover:bg-gray-100 rounded-sm"
 >
   <a
     [href]="record.metadataUrl"
@@ -14,9 +14,7 @@
       ></ui-record-thumbnail>
 
       <div class="flex-grow px-3 py-1 flex flex-col overflow-hidden">
-        <h1
-          class="title-font text-sm font-bold text-gray-900 truncate md:overflow-clip"
-        >
+        <h1 class="title-font text-sm font-bold truncate md:overflow-clip">
           {{ record.title }}
         </h1>
       </div>
--- a/libs/ui/src/styles.css
+++ b/libs/ui/src/styles.css
@@ -1,24 +1,14 @@
 @import './tailwind-generated.css';
 
-:root {
-  --gn-font-size-md: 1rem;
-  --gn-font-size-sm: 0.875rem;
-  --gn-title-color: #3B82F6;
-  --gn-link-color: #1D4ED8;
-}
-
 body {
   background-color: var(--color-background);
   color: var(--color-main);
 }
 a {
-  color: var(--gn-link-color);
+  color: var(--color-primary);
 }
 h1 {
-  color: var(--gn-title-color);
-}
-p {
-  font-size: var(--gn-font-size-sm);
+  color: var(--color-primary);
 }
 
 .clamp-1 {
--- a/libs/ui/src/styles.css
+++ b/libs/ui/src/styles.css
@@ -1,12 +1,26 @@
 @import './tailwind-generated.css';
 
+:root {
+  --gn-font-size-md: 1rem;
+  --gn-font-size-sm: 0.875rem;
+  --gn-title-color: #3B82F6;
+  --gn-link-color: #1D4ED8;
+}
+
 body {
   background-color: var(--color-background);
   color: var(--color-main);
 }
 a {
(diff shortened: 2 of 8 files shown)


libs/ui/src/styles.css
```diff
@@ -1,12 +1,26 @@
...
 body {
   background-color: var(--color-background);
   color: var(--color-main);
 }
 a {
-  color: var(--color-primary);
+  color: var(--gn-link-color);
+}
+h1 {
+  color: var(--gn-title-color);
```
fgravin, line 18: I don't think it's the way to go.
You should, in the h1 of the CARD, use class `color-primary`, this way people will be able to change the color theme just by changing the --color-primary css variable.
We should have elements in the ui with primary and secondray colors (like h1, button, link etcc..) so the themes is customizable

libs/ui/src/lib/record-preview-title/record-preview-title.component.html
```diff
@@ -1,6 +1,6 @@
 <!-- Record list item: List Small -->
 <div
-  class="h-10 border-b border-gray-200 transition duration-200 hover:bg-gray-100 border-gray-300"
+  class="h-10 w-full bg-white border border-gray-300 transition duration-200 hover:bg-gray-100 rounded-sm"
```
fgravin, line 3: I think just bottom border, very light is enough, full border is too much for this layout

fgravin (commented):
Thanks for the improvments.
I made some suggestion, also I think we are not using the css-variable potential correctly

libs/ui/src/styles.css
fxprunayre: Maybe check https://tailwindcss.com/docs/customizing-colors#naming-your-colors and https://levelup.gitconnected.com/tailwindcss-with-css-variables-513abe2e9a5

libs/ui/src/lib/record-preview-title/record-preview-title.component.html
fxprunayre: Ah les goûts et les couleurs - Consensus will be a challenge ;)

That's a difficulty/challenge I see with tailwind, you define a style somewhere eg. border-width and color in HTML. You expect to have some kind of consistencies all hover the places (even maybe all hover the apps) but that needs to be maintained. We used to define a css class gn-results and style the border in CSS/LESS file. Here we are defining the border in each layout - you can create a shortcut for a custom tailwind rule or create your themes in tailwind.config.js overriding the meaning of rounded-sm. Not sure what is better...

MichelGabriel: Yes, it gives you a lot of freedom 😄

libs/ui/src/styles.css
MichelGabriel: Ok, going to change it

libs/ui/src/lib/record-preview-title/record-preview-title.component.html
fgravin: > We used to define a css class gn-results and style the border in CSS/LESS file

True that defining our own classes that reference TW classes could have been great to homogeneize the theme.
How ever it is doable with less but not with css.

> you can create a shortcut for a custom tailwind rule or create your themes in tailwind.config.js overriding the meaning of rounded-sm

ok, let's see how it works. @MichelGabriel did you set up something like this in the PR or not ?

fgravin: > Ah les goûts et les couleurs - Consensus will be a challenge ;)

It's just that in EEA forest thing, the news list has no border, and in the TITLE card you have done, I only see a border at the bottom ... while I don't see any specific class for this (seems your HTML declare a full border).

Also @MichelGabriel I would rather go for a margin in list items themselves rather than a padding in item container, actually it gives the freedom to have different spacing depending on the CARD mode.

libs/ui/src/styles.css
```diff
@@ -1,24 +1,14 @@
...
-  --gn-link-color: #1D4ED8;
-}
-
 body {
   background-color: var(--color-background);
   color: var(--color-main);
 }
 a {
-  color: var(--gn-link-color);
+  color: var(--color-primary);
```
fgravin, line 8: I won't go for that either, what do you think @fxprunayre. Isn't it better to put the `text-primary` class explicitly where we want ?
I'm not sure we want all links to be primary, but maybe CARD metadata title yes, so we can easliy do the green metadata title as in https://forest.eea.europa.eu/datacatalogue

libs/ui/src/lib/record-preview-title/record-preview-title.component.html
fxprunayre: > the news list has no border, 

hum,
image


> and in the TITLE card you have done,

I know it's my fault.

> I only see a border at the bottom ... while I don't see any specific class for this (seems your HTML declare a full border).

See
https://github.com/geonetwork/geonetwork-ui/blob/master/libs/ui/src/lib/record-preview-title/record-preview-title.component.html#L3 `border-b` means border bottom.

libs/ui/src/styles.css
fxprunayre: > Isn't it better to put the `text-primary` class explicitly where we want ?

It looks like it is the way to go if we want named colors
https://tailwindcss.com/docs/customizing-colors#naming-your-colors

libs/ui/src/lib/record-preview-title/record-preview-title.component.html
MichelGabriel: > Also @MichelGabriel I would rather go for a margin in list items themselves rather than a padding in item container, actually it gives the freedom to have different spacing depending on the CARD mode.

This will probably give problems when you want to change the background color of a list item.

MichelGabriel: @fxprunayre What should the TITLE list look like? A border for the whole list?

fgravin: @MichelGabriel the image just above

MichelGabriel: But that has only the border at the bottom, so that's what I did, or not?

MichelGabriel: > you can create a shortcut for a custom tailwind rule or create your themes in tailwind.config.js overriding the meaning of rounded-sm

What does creating a shortcut mean exactly?

fgravin: > But that has only the border at the bottom, so that's what I did, or not?

When I tried the first time, the TITLE card had a full border. Indeed you removed `border-b` class.
My thought was just that for a very small list like TITLE, a full border is a little bit cumbersome

fgravin:
I will merge it anyway it improves the current state, and we'll wait for other PRs to continue the work.
Anyway, as the CSS framework might change, it's not worth to push on it for the moment.

MichelGabriel:
Thanks. To be continued.

PR #77, opened 2020-12-17
Small styling improvements to the facets and button

MichelGabriel:
Small improvements to the facets:
- add hover background colour
- added filter icon

Small styling changes to the button:
- border same as background
- added focus styles/colours


fgravin (approved):
Thanks

PR #80, opened 2020-12-23
Refactor web components architecture

fgravin:
The PR aims to add a web component for facets, and combine 2 different web components on the same page
- gn-facets
- gn-results-list

Unfortunatly, as web components are encapsulated in an Angular application, loading multiple Angular applications seems to randomly fails the web component loading. I decided to simplify the architecture, so all web components are served by the same angular application. It simplifies the workflow a lot, it fixes `storybook-wc`, the only drawback is that all wc are in the same bundle, so you could load useless code.

The documentation has been updated accordingly.

The results of this refactoring is an example of `gn-facets combined` with `gn-results-list`, you can see the result by running `npm run serve:wc` and go on http://localhost:8001/gn-facets.sample.html


fxprunayre:
Works fine. Minor changes proposed in https://github.com/geonetwork/geonetwork-ui/pull/81

Then we have other stuff to improve

http://localhost:8001/gn-results-list.sample.html
* If 2 results lists, changing one affect the other one. Maybe we can use the `bucket` as a searchState identifier in order to handle more than one search in a page. 

http://localhost:8001/gn-facets.sample.html
* changing the size attribute of result list, reset facets

fgravin:
> * If 2 results lists, changing one affect the other one. Maybe we can use the `bucket` as a searchState identifier in order to handle more than one search in a page.

I'm not sure it will be enough, cause the state will remain the same, won't it ?
I can work on that after the vacation: adapat the state structure to handle multiple search.

BTW i won't be able to follow github this week so I let you merge what you think opportune.
Cheers

fxprunayre:
> I'm not sure it will be enough, cause the state will remain the same, won't it ?

We could have a Map of states `default` => SearchState,  `homePageCountrySummary` => SearchStateForCountrySummaryBucket, `latestRecords` => ....

```
<search-facets-container class="w-1/4 h-full" 
                                        bucket="homePageCountrySummary" config=""></search-facets-container>
...

<search-results-list-container 
      bucket="latestRecords"
      [hidden]="record$ | async"
      class="w-3/4"
    ></search-results-list-container>

...

and later the main search using the default state:
<search-results-list-container 
      [hidden]="record$ | async"
      class="w-3/4"
    ></search-results-list-container>
```

Currently we have only one initialState. So if a component request a bucket that does not exist, it would create a search state for this bucket id (or search id if bucket is not meaningful).

> BTW i won't be able to follow github this week so I let you merge what you think opportune.

Ok.

fxprunayre (approved):

PR #81, opened 2020-12-28
wc: Add home page with a list. Fix filter and change  by .

fxprunayre:
image

and

image

### DIFF
--- a/webcomponents/src/app/components/gn-results-list/gn-results-list.sample.html
+++ b/webcomponents/src/app/components/gn-results-list/gn-results-list.sample.html
@@ -15,14 +15,13 @@
       filter="GRID"
     ></gn-results-list>
     <div>
-      <button id="changeSizeBtn"
-              onclick="moreRecords()">Load more ...</button>
+      <button id="changeSizeBtn" onclick="moreRecords()">Load more ...</button>
     </div>
   </body>
   <script>
     function moreRecords() {
       const wc = document.getElementsByTagName('gn-results-list')[0]
-      wc.size = parseInt(wc.size) + 10;
+      wc.size = parseInt(wc.size) + 10
     }
   </script>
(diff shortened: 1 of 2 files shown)


webcomponents/src/app/components/gn-results-list/gn-results-list.sample.html
```diff
@@ -8,19 +8,21 @@
...
     <script src="gn-wc.js"></script>
-    <div>
-      <button id="changeSizeBtn">Change size</button>
-    </div>
     <gn-results-list
       api-url="https://apps.titellus.net/geonetwork/srv/api"
       size="10"
       layout="TITLE"
-      filters="+tag.default:Soil"
+      filter="GRID"
```
fgravin, line 15: what is it ?

webcomponents/src/index.html
```diff
@@ -7,5 +7,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
   </head>
-  <body></body>
+  <body>
```
fgravin, line 10: Great, I had that in mind, if you look at my PR in the bash script I started to try to generate the list (in comment)... but was not very familiar with bash.. We can manage it manually.

Thanks

fgravin (approved):

webcomponents/src/app/components/gn-results-list/gn-results-list.sample.html
fxprunayre: https://github.com/geonetwork/geonetwork-ui/blob/master/webcomponents/gn-results-list/src/app/gn-results-list.component.ts#L24 is `filter` not `filters` but then the example can be whatever. But the attribute was wrong.

PR #82, opened 2020-12-29
All aggregations, Hits number, No config, Plurals

fxprunayre:
* Config / Load a default one if none set server side
* Config / Add adapter to convert GN config to whatever we need here. At the end we should converge to something if possible to avoid this step
* Aggs / Display all facets from config
* Aggs / Do not display a facet with no item

image

* Aggs / Display histogram and filters type
* Aggs / Filters / Do not display block if sum of count in buckets is 0
* Results / Add component to display number of results

image

* i18n / Test plural

image

* i18n / Test HTML in translation
* i18n / Apply key scheme for aggs and search
* StoryBook / addon-notes does not look to exist for the current version of storybook (see https://github.com/geonetwork/geonetwork-ui/pull/82/commits/3825be103e02c3bc1ec8a66bc17cce4e83b2b913). Removing it for now unless someone has another suggestion?


libs/ui/src/lib/facets/facet-block/facet-block.component.spec.ts
```diff
@@ -133,3 +134,31 @@ describe('FacetBlockComponent', () => {
     })
   })
 })
+
+describe('EmptyFacetBlockComponent', () => {
+  let component: FacetBlockComponent
+  let fixture: ComponentFixture<FacetBlockComponent>
+  let de: DebugElement
+
+  beforeEach(async(() => {
```
fgravin, line 143: should not be necessary just to test other component inputs

libs/ui/src/lib/facets/facet-block/facet-block.component.spec.ts
```diff
@@ -133,3 +134,31 @@ describe('FacetBlockComponent', () => {
...
+    component = fixture.componentInstance
+    de = fixture.debugElement
+    component.model = EMPTY_BLOCK_MODEL_FIXTURE
+    component.selectedPaths = []
+    fixture.detectChanges()
+  })
+
+  describe('Aggregation is hidden if no item', () => {
+    it('hides aggregation', () => {
+      expect(de.query(By.css('div > span'))).toBeFalsy()
```
fgravin, line 161: Add
```js
expect(component.hasItems).toBe(false)
```

libs/search/src/lib/results-hits-number/results-hits-number.component.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ResultsHitsNumberComponent)
+    component = fixture.componentInstance
+    fixture.detectChanges()
+  })
+
+  it('should create', () => {
+    expect(component).toBeTruthy()
+  })
```
fgravin, line 35: should have some tests

libs/ui/src/lib/color-scale/color-scale.component.html
```diff
@@ -1,7 +1,5 @@
 <div class="flex flex-row items-center">
-  <div class="text-xs text-gray-700 font-bold m-2 w-32 text-right">
-    primary
-  </div>
+  <div class="text-xs text-gray-700 font-bold m-2 w-32 text-right">primary</div>
```
fgravin, line 2: do you use prettier for html ?

fgravin (changes requested):
Thanks for the work.
The PR is too big and threat other topics that makes the review harder (hitscount component, translations etc...).
The commit history really helps in the beginning but is confusing in the end.
Overall, going slowly step by step with a clear history and dedicated PR helps the review, but also ensure to have a smoother and clearer progress when coding.

The new features seem great, the code is ok and quite tested, I haven't tested yet, it is red so might need to fix architecture issues.

libs/search/src/lib/state/reducer.ts
```diff
@@ -15,6 +15,10 @@ export interface SearchState {
   }
   params: SearchStateParams
   results: {
+    hits: {
+      value: number
+      eq: string
```
fgravin, line 20: what is it for ? `eq`

fxprunayre: "a value of "eq" indicates that "total.value" is the accurate count."  https://www.elastic.co/guide/en/elasticsearch/reference/master/search-your-data.html#track-total-hits

libs/ui/src/lib/color-scale/color-scale.component.html
fxprunayre: AFAIK I'm using `npm run format:fix` and prettier in intellij. Not sure what introduced that change.

fgravin (approved):

PR #83, opened 2020-12-29
i18n: Add loader for core translation + codelist

fxprunayre:
* i18n / Load the main translations + codelist translations (mainly used for
translating aggregation bucket values)
* Aggs / Translate values (for all cl_*.key fields)

image

* OpenAPI / Update to GN 4.0.2 and fix doc examples.


fgravin (approved):
Thanks for the enhancement, it's way better to load translations for external dedicated files.
Try to have a separated commit for generate api changes.

libs/common/src/lib/i18n/i18n.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    private standardApiService: StandardsApiService
+  ) {}
+
+  getTranslation(lang: string): Observable<any> {
+    const observables: Observable<any>[] = []
+    const translationFiles = [`assets/i18n/${lang}.json`]
+
+    const core = this.http.get(translationFiles[0])
+    observables.push(core)
+    observables.push(
```
fgravin, line 23: How does it work exactly ? does it fetch another translation file ?
Why do we still have `facets.block.title.cl_hierarchyLevel.key` in the assets ?

fxprunayre: See https://apps.titellus.net/geonetwork/srv/api/standards/iso19139/codelists?codelist=gmd%3AMD_TopicCategoryCode&codelist=gmd%3AMD_ScopeCode&codelist=gmd%3AMD_MaintenanceFrequencyCode&codelist=gmd%3AMD_ProgressCode&codelist=gmd%3ADS_InitiativeTypeCode&codelist=gmd%3AMD_SpatialRepresentationTypeCode

The API call loads the codelist values translations. Not its label. If we want to load codelist labels in one call, we may need to extend the API...

PR #84, opened 2021-01-04
Sorting usually define asc or desc order.

fxprunayre:


libs/search/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -17,7 +17,15 @@ export class ElasticsearchService {
 
   buildPayload(state: SearchState): SearchParams {
     const { size, sortBy, filters } = state.params
-    const sort = sortBy ? [sortBy] : undefined
+    const sort: SortParams = sortBy
```
fgravin, line 20: Should be externalized cause it might be used to build another payload ?
Also, not tested yet

libs/search/src/lib/elasticsearch/elasticsearch.model.ts
```diff
@@ -0,0 +1,52 @@
...
+} from 'elasticsearch'
+
+export type SortOrder = 'asc' | 'desc'
+interface SortParam {
+  [key: string]: SortOrder
+}
+export type SortParams = string | SortParam | (string | SortParam)[]
+
+export interface SearchParams extends GenericParams {
+  analyzer?: string
```
fgravin, line 16: missing `aggs` ?
I think the `SearchParams` interface need a complete refactoring, where this change could be part of.
Old interface might still be used somewhere, some object might not be correctly typed (otherwise `aggs` would trigger an error), es-types dep still in the project etc..

fgravin (changes requested):
Good !
Missing tests though.
Also try to restrict the PR to a clear scope, it's easier to maintain.
Thanks

fxprunayre:
> Also try to restrict the PR to a clear scope, it's easier to maintain.

The scope is not clear here ?

libs/search/src/lib/elasticsearch/elasticsearch.model.ts
fxprunayre: > missing `aggs` ?

the type comes from the `@types/elasticsearch` https://github.com/geonetwork/geonetwork-ui/commit/5bfae11199fe1f3f49515db2cdadf779f210cb91

fgravin:
> The scope is not clear here ?

Well you work on sorting, but I see stuff about partial refactoring of ES types. To me this has no place here, and the **full** refactoring should be done in a dedicated PR.

libs/search/src/lib/elasticsearch/elasticsearch.model.ts
fgravin: Why closing ? Just remove the `SearchParams` interface from the PR and it's great !

fxprunayre: > Why closing ?

Un peu fatigué.



> Just remove the `SearchParams` interface from the PR and it's great !

If I remove it then it does not build. If we want buildPayload to return a typed response we have to create a SearchParams which is not an Elasticsearch 5 version type.

PR #85, opened 2021-01-06
Add featuring demo pages on gh-pages

fgravin:
This PR proposes a way to publish and share static demo pages to showcase the usage of webcomponents better than storybook only.

The command `npm run serve:wc` has been moved to `npm run demo` and do:
- build webcomponents + sample pages in `dist/demo`
- copy `/demo` files in `dist/demo`
- serve `dist/demo` on `localhost:8001`

The `storybook` gh-action has been moved to `gh-pages` action and will both publish storybook and demo folder.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Attempt production build
         run: npm run build
 
-  storybook:
+  gh-pages:
     runs-on: ubuntu-latest
     steps:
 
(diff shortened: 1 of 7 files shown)


fgravin:
@fxprunayre https://geonetwork.github.io/geonetwork-ui/demo-pages/demo/pages/eea/results-list.html
I think of adding a page to illustrate facets and results-list with web component like in https://forest.eea.europa.eu/datacatalogue

PR #87, opened 2021-01-07
Facets filter

fgravin:
Enable the facet block filter input to filter the facet items in the list.
If the facet configuration contains an `include` field (ie `include: '.*'`) then the model will have `includeFilter: true` and the block is considered as filterable.

The filter input is debounced, and dispatches a `SetIncludeOnAggregation` what trigger a new ES search with only the new aggregation definition for the current field. Only the facet items list is refreshed. The whole block is not refreshed anymore due to `trackby` setup.

Some refactoring has been made to handle request aggregation term patch overall (both `size` on load more action and `include` value).

Screenshot from 2021-01-07 20-50-51


libs/search/src/lib/facets/facets.service.spec.ts
```diff
@@ -7,11 +7,16 @@ import { FacetsService } from './facets.service'
...
+  warn: jest.fn(),
+  log: jest.fn(),
+  error: jest.fn(),
+}
 describe('FacetsService', () => {
   let service: FacetsService
 
   beforeEach(() => {
-    service = new FacetsService()
+    service = new FacetsService(logServiceMock)
```
LHBruneton-C2C, line 19: Is there a reason for not using test bed here ?

libs/search/src/lib/state/effects.ts
```diff
@@ -74,28 +78,52 @@ export class SearchEffects {
...
           'bucket',
           JSON.stringify(
-            this.esService.buildMoreOnAggregationPayload(state, action.key)
+            this.esService.buildMoreOnAggregationPayload(state, aggregationKey)
           )
         )
       ),
       map((response: SearchResponse<any>) => {
         const aggregations = response.aggregations
-        const key = Object.keys(aggregations)[0]
```
LHBruneton-C2C, line 97: Thanks for cleaning this part !

libs/search/src/lib/facets/facets.service.spec.ts
fgravin: It's slower and useless.

libs/ui/src/lib/facets/facet-block/facet-block.component.ts
```diff
@@ -40,7 +70,7 @@ export class FacetBlockComponent implements OnInit {
   }
 
   onFilterChange(value: string) {
-    this.filterChange.emit(value)
+    this.filterChange.emit(`.*${value}*`)
```
LHBruneton-C2C, line 73: Shouldn't this be done at the smart container level ? The UI components are not aware of business logic, just user interactions.

LHBruneton-C2C (approved):
Thanks for the new features and cleaning up ! LGTM, even though I could not test. The GH page link gives me a 404, and running the branch locally doesn't seem to be configured to use apps.titellus.net anymore.

libs/ui/src/lib/facets/facet-block/facet-block.component.ts
fgravin: You are perfectly right ! What a mistake ;)

fgravin:
Thanks for the review @LHBruneton-C2C I appreciate.

> The GH page link gives me a 404, 

Yes with rebasing https://geonetwork.github.io/geonetwork-ui/facets-filter/demo/eea/datacatalogue.html works.

> and running the branch locally doesn't seem to be configured to use apps.titellus.net anymore.

Yes it is, weird

PR #91, opened 2021-01-25
[datafeeder]: Sum-up page

y-vynohradov:
chore: Add wizard summarize page


fgravin:
Still red :/

libs/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+        {
+          display: 'tagName2',
+          value: 'tagName2',
+        },
+      ])
+    )
+    window.localStorage.setItem('datepicker', String(new Date().getTime()))
+    window.localStorage.setItem('dropdown', JSON.stringify('10000'))
+    window.localStorage.setItem('description', 'description')
+  })
```
tonio, line 66: while you're at dealing with localstorage, you should refactor to be able to handle multiples import => store with the id as key ?

tonio: storage should be done in a service, implementation being private. If you change implementation then you won't be forced to rewrite all tests.

apps/datafeeder/src/app/presentation/pages/summarize-page/summarize-page.component.css
```diff
@@ -0,0 +1,43 @@
+:host {
+  display: flex;
+  flex: 1;
```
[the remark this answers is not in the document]
y-vynohradov: Do you mean <app-root class="flex flex-col flex-1"></app-root>? 

But component will be rendered inside root and this required for proper displaying 

Look, it is simular to this

`<app-root class="flex flex-col flex-1">
<app-summirize-page class="flex-1 flex"></app-summirize-page>
</app-root>`

PR #93, opened 2021-01-26
Allow multiple searches in the state

fgravin:
This PR aims to change the search state structure to handle multiple searches in the same page.
Before, the state structure allowed to store a unique search, matching the `SearchSate` interface. This structure has been enhanced so the state stores a dictionnary of searches, the `SearchSate` has become a dictionnary of `SearchSateSearch`.
The dictionnary key is the search id, used to bind any component and action to a specific search, the default key is `default`.

I tried to keep the code simple so the change is quite transparent:
- I kept the original reducer that is now called to reduce one element of the state pool
- If the state pool does not contain the search id provided by the action, it initiates a new search structure for this new id
- If an action does not provide any id, then it passes the `default` value
- All effects propagate the search id along the actions flow

By default, the `initialState` contains a search state for the `default` id, and all actions work on this id in a transparent way.
At the moment, the application looks to run exactly the same.

- [x] Refactor the search state
- [x] Add 2 searches in the same page
- [x] Update the effects and the facade to dispatch the correct search id
- [x] Demonstrate the work in a web component

**State Facade**

- The facade has been reworked so it is transparent for any component to be binded with the correct search.
- Facade is no more a singleton, there should be one facade per search
- Facade is not injected at root level
- Facade is instanciated when you declare the `searchSearchStateContainer="mainSearch"` directive
- If you don't provide the directive (I made it almost mandatory to force the developper to know what is is doing), you are responsible to inject and initiate the Facade service (as done in the web component)
- Any component that is below the directive in the DOM, and that inject the facade, will retrieve the directive instance of the facade, binded to a specific `searchId`
- When you don't provide a `searchId`, the default value is `default`.

I had to refactor many components and tests that were using the store directly.

I don't guarantee everything is working perfectly but it looks ok with the tests I've made.
Demo: https://geonetwork.github.io/geonetwork-ui/multiple_search/demo/webcomponents/gn-results-list-multiple.sample.html

The commit https://github.com/geonetwork/geonetwork-ui/pull/93/commits/1057df235ddaf2c56030977700b95ba4aa02529e should be adressed (usage of `flatMap` operator).

### DIFF
--- a/libs/search/src/lib/state/effects.ts
+++ b/libs/search/src/lib/state/effects.ts
@@ -32,7 +32,7 @@ import {
   SET_PAGINATION,
 } from './actions'
 import { SearchState } from './reducer'
-import { getSearchState } from './selectors'
+import { getSearchState, getSearchStateSearch } from './selectors'
 
 @Injectable()
 export class SearchEffects {
@@ -70,7 +70,7 @@ export class SearchEffects {
     this.actions$.pipe(
       ofType(REQUEST_MORE_RESULTS),
       switchMap(() => this.authService.authReady()), // wait for auth to be known
-      withLatestFrom(this.store$.pipe(select(getSearchState))),
+      withLatestFrom(this.store$.pipe(select(getSearchStateSearch))),
       switchMap(([_, state]) =>
         this.searchService.search(
           'bucket',
@@ -129,7 +129,7 @@ export class SearchEffects {
     return updateTermAction$.pipe(
       tap((action) => (aggregationKey = action.key)),
       switchMap(() => this.authService.authReady()), // wait for auth to be known
-      withLatestFrom(this.store$.pipe(select(getSearchState))),
+      withLatestFrom(this.store$.pipe(select(getSearchStateSearch))),
       switchMap(([_, state]) =>
         this.searchService.search(
           'bucket',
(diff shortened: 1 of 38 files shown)


libs/search/src/lib/state/effects.ts
```diff
@@ -55,53 +56,67 @@ export class SearchEffects {
...
-            this.esService.search(state, ElasticsearchMetadataModels.SUMMARY)
-          )
-        )
-      ),
-      switchMap((response: SearchResponse<any>) => {
-        const records = this.esMapper.toRecordSummary(
-          response,
-          this.searchService.configuration.basePath
+      switchMap((action: SearchActions) =>
+        this.authService.authReady().pipe(
```
LHBruneton-C2C, line 78: Is it intentional to pipe all the work inside the authReady stream ? Why not go back to the actions$ stream like before ?

LHBruneton-C2C:
I'm a little confused with the name 'getSearchStateSearch'. Will there be anything else than searches in the state ? And if it's to recognize that we only get one search in the dictionary, why not getSearchById or getSearchByKey ?

fgravin:
> I'm a little confused with the name 'getSearchStateSearch'. Will there be anything else than searches in the state ? And if it's to recognize that we only get one search in the dictionary, why not getSearchById or getSearchByKey ?

Yes good idea, I used the common pattern but it's not appropriate here, thanks.

Beside that, what do you think of the global approach ?

LHBruneton-C2C:
It's really nice that you tried to stay as close as possible to the previous state. Now that I think about it, managing the state as ngrx entities would have been a step too far. 
Great idea keeping a default search, it makes the PR much more readable.
I never worked with the web components, so I can't say much there.
And a last question, won't there ever be interactions between searches ? Otherwise, one facade per search is great.

PR #94, opened 2021-01-27
Fix for the height of the search result list items

MichelGabriel:
This PR has a small fix for the height of the items, it now hasn't a full height anymore, but still no fixed height. It now depends on title, 3 lines of abstract and the link. So if the font-size is increased, the height of the list item 'grows' with it

Small fixes for the height of the search result list item:
- removed the class `h-full`, gave the item the full height
- added `clamp-3` to display only the first 3 lines and end with `...`


fgravin:
Hi @MichelGabriel thanks, it's better if you push your branch upstream when doing a PR, this allows the CI to push changes on gh-pages, so we can test immediatly the result. For the next time ;)

PR #96, opened 2021-02-01
[datafeeder]: Publish page 

y-vynohradov:
Add page of publishing data.
Add page of success published data.


fgravin (approved):
Ok nice, I tested it and the look is great.
We will have a polish task to fix the remaining design issues
- bg color too blue
- button not rounded
Also I would like that you check the :host css of all your component, I d'ont undesrtand why you alway put css rules about flex, you should fix all of them please.

PR #97, opened 2021-02-02
[datafeeder] allow to switch encoding & preview 

tonio:


fgravin (changes requested):
The code is good but I think the behavior is not the one that is expected:
- if the bbox code returned by the server is null, then bbox dropdown should contains the config projections list (without the unknown)
- if the bbox code is not null, then the combo should be disabled, and display the correct code
- encoding list should not contain unknown

fgravin (approved):
Nice thank you, I couldn't test if the ref system is null though: is unknown selected then ?

There are conflicts, could you rebase please and merge both PR ? Thanks

tonio:
> I couldn't test if the ref system is null though: is unknown selected then ?

yes

PR #98, opened 2021-02-03
[datafeeder]: Local storage

y-vynohradov:
refactor: Change local storage structure

### DIFF
--- a/libs/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
+++ b/libs/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
@@ -8,14 +8,38 @@ import { HttpClientTestingModule } from '@angular/common/http/testing'
 import { NO_ERRORS_SCHEMA } from '@angular/core'
 import { WizardService } from '../../services/wizard.service'
 
-const localStorageMock = () => {
-  let storage = {}
-  return {
-    getItem: (key) => (key in storage ? storage[key] : null),
-    setItem: (key, value) => (storage[key] = value || ''),
-    removeItem: (key) => delete storage[key],
-    clear: () => (storage = {}),
-  }
+const wizardServiceMock = {
+  getWizardFieldData: jest.fn((id) => {
+    switch (id) {
+      case 'title': {
+        return 'title'
+      }
+      case 'abstract': {
+        return 'abstract'
+      }
+      case 'tags': {
+        return JSON.stringify([
+          {
+            display: 'tagName1',
+            value: 'tagName1',
+          },
+          {
+            display: 'tagName2',
+            value: 'tagName2',
+          },
+        ])
+      }
+      case 'datepicker': {
+        return String(new Date().getTime())
+      }
+      case 'dropdown': {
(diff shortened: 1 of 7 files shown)


libs/editor/src/lib/services/wizard.service.ts
```diff
@@ -13,8 +13,25 @@ export class WizardService {
...
-    this.initialize()
+  private id: string
+
+  constructor(private translateService: TranslateService) {}
+
+  initialize(id: string): void {
+    this.id = id
+    this.wizardData.clear()
+
+    const savedWizardData = localStorage.getItem(id)
```
fgravin, line 24: almost, the id of the batch should not be ids of the localStorage, but ids of the dictionnary stored in the localStorage
```js
localStorage
 - datafeeder-state: {
     d7c14181-2e66-49ff-868e-8c55671f955f: [{},{}]...,     
     trtrez543534543543534: [{},{}]...,
}
```

libs/editor/src/lib/services/wizard.service.ts
```diff
@@ -39,21 +56,35 @@ export class WizardService {
...
+    const config = []
+
+    this.wizardData.forEach((value, key) => {
+      config.push({
+        id: key,
+        value,
+      })
+    })
+
+    const data = {
```
fgravin, line 76: why do you save the step exactly ? how is using it ?

fgravin (changes requested):
Quite good but not exactly the good structure still.
Please add some test as well.

Also I don't really understand the purpose to store the step in the localStorage as it is already in the url, explain me you approach please.

Thanks

fgravin:
Also please rebase on master, thanks

libs/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
```diff
@@ -8,14 +8,38 @@ import { HttpClientTestingModule } from '@angular/common/http/testing'
...
 
-const localStorageMock = () => {
-  let storage = {}
-  return {
-    getItem: (key) => (key in storage ? storage[key] : null),
-    setItem: (key, value) => (storage[key] = value || ''),
-    removeItem: (key) => delete storage[key],
-    clear: () => (storage = {}),
-  }
+const wizardServiceMock = {
```
fgravin, line 11: good !
but you should test that it is called with the correct params.

libs/editor/src/lib/services/wizard.service.ts
```diff
@@ -13,8 +13,25 @@ export class WizardService {
   private wizardStep = 1
   private wizardData: Map<string, any> = new Map()
 
-  constructor(private translateService: TranslateService) {
-    this.initialize()
+  private id: string
+
+  constructor(private translateService: TranslateService) {}
```
fgravin, line 18: please add tests in this service, where you mock the localStorage

fgravin (changes requested):
Thanks for the refactoring.
You should add some tests please.

tonio:
Integrated into #103

PR #99, opened 2021-02-03
[datafeeder] Uniformise pages design

tonio:
And allow to go until publication from interface.


fgravin (approved):
Ok perfect thanks.
I think step things are a bit too hard coded (everything should be computed from wizard configuration object size), but we can adress that later.

I note also something for later: seems the router does not really work with the steps (when you play with the previous/next of the browser history).

But thank you it's very nice and it works well for the demo, great job

tonio:
> I think step things are a bit too hard coded

Totally agree, just miss some time before the demo. I'll rework this afterwards.

PR #100, opened 2021-02-11
i18n: Apply translation scheme.

fxprunayre:
* Apply translation scheme to english translations
* Remove non english ones.

Next steps will be:
* Transifex / Register gn-ui english translation file in transifex
* GN / Add a new translation pack (see https://github.com/geonetwork/core-geonetwork/pull/5387) for GN-UI composed of GN-UI JSON + Schema codelist required for facet translations
* GN-UI / Load the translation pack (https://github.com/geonetwork/geonetwork-ui/blob/master/libs/common/src/lib/i18n/i18n.service.ts#L21)


apps/search/src/assets/i18n/en.json
```diff
@@ -1,29 +1,30 @@
...
+  "results.layout.selectOne": "Results layout",
+  "results.records.hits.found": "{hits, plural, =0{No documents match the specified search.} one{} other{{hits} records found.}}",
+  "results.records.hits.found=0.help": "Suggestions: <ul class='list-disc list-inside'><li>Try other words</li><li>Specify fewer words</li></ul>",
+  "results.sortBy.createDate": "New records",
+  "results.sortBy.dateStamp": "Last updates",
+  "results.sortBy.popularity": "Popularity",
+  "results.sortBy.relevancy": "Relevancy",
+  "search.field.any.placeholder": "Search datasets, services and maps ...",
+  "search.field.sortBy": "Sort by",
+  "search.loading": "Loading ..."
```
fgravin, line 29: how did you change the order ? Did you use the npm script to extract the translations ?

libs/search/src/lib/sort-by/sort-by.component.ts
```diff
@@ -5,25 +5,22 @@ import { SearchState } from '../state/reducer'
...
-
 @Component({
   selector: 'search-sort-by',
   templateUrl: './sort-by.component.html',
 })
 export class SortByComponent implements OnInit {
   choices = [
     {
-      label: 'relevancy',
+      label: 'results.sortBy.relevancy',
```
fgravin, line 15: how those strings will be reconized to be extracte d?

fgravin (changes requested):
> * Apply translation scheme to english translations

What does it mean ?

Thanks to refactor the keys, some questions though.

apps/search/src/assets/i18n/en.json
fxprunayre: JSON sort plugin.

> > * Apply translation scheme to english translations
> What does it mean ?

We discussed to apply a common pattern/scheme to translation keys - like in
https://github.com/geonetwork/geonetwork-ui/blob/master/apps/datafeeder/src/assets/i18n/en.json or https://github.com/geonetwork/geonetwork-microservices/blob/main/modules/services/ogc-api-records/service/src/main/resources/messages/api.properties
no?

libs/search/src/lib/sort-by/sort-by.component.ts
fxprunayre: In this particular case, I don't think we can rely on the extractor script - at some point this will not be hardcoded and we may have a list or config for this ... but I can restore the marker and use the npm script if it makes more sense ...

apps/search/src/assets/i18n/en.json
fgravin: ok, I was confused by the term "schema"

libs/search/src/lib/sort-by/sort-by.component.ts
fgravin: Well, I think it's better to keep the translation keys being extracted. It's a good practice to let the script update the file.


> at some point this will not be hardcoded and we may have a list or config for this

What do you mean exactly ?
If we update `en.json` manually to add foreign keys, it's the same effort to add them as markers in a dedicated file.

fxprunayre: Also if we use ` npm run i18n:extract`

```json
{
	"search.field.any.placeholder": "Search datasets, services and maps ...",
	"results.layout.selectOne": "Results layout",
	"search.field.sortBy": "Sort by",
	"records": "records",
	"record.action.view": "viewable",
	"record.action.download": "downloadable",
	"record.more.details": "Read more",
	"results.records.hits.found=0.help": "Suggestions: <ul class='list-disc list-inside'><li>Try other words</li><li>Specify fewer words</li></ul>",
	"results.records.hits.found": "{hits, plural, =0{No documents match the specified search.} one{} other{{hits} records found.}}",
	"search.loading": "Loading ...",
	"results.sortBy.relevancy": "Relevancy",
	"results.sortBy.dateStamp": "Last updates",
	"results.sortBy.popularity": "Popularity",
	"dropFile": ""
}
```

which looks to be incomplete. Should we then add markers for all facets stuffs ? Web component translation like `nav.back` are also missing.

fxprunayre: Does it work well for the datafeeder app ?

fgravin: > Does it work well for the datafeeder app ?

Not sure they are using, nobody read the manual ! :)

fgravin: > Should we then add markers for all facets stuffs ? 

IMO yes

> Web component translation like `nav.back` are also missing.

Maybe it's just the path config in the package.json script

It helps to see some translations are missing, or some key are not well formatted, I keep thinking it's a good practice but I won't force it you find it painfull to maintain.

fgravin:
Note that for the moment, it is always english and there is no mecanism to detect or change the langage.

libs/search/src/lib/sort-by/sort-by.component.ts
fxprunayre: Let's see if it works. Markers added. Format consistent with prettier. Web component path fixed.

fgravin (approved):
Thanks for the changes, LG !

PR #101, opened 2021-02-12
Paging fix

fxprunayre:
* Fix from/size when setting a filter
* Fix text search
* Stop scrolling at the end
* Cosmetic / Avoid blank space in search results

image

### DIFF
--- a/libs/search/src/lib/results-list/results-list.container.component.ts
+++ b/libs/search/src/lib/results-list/results-list.container.component.ts
@@ -1,22 +1,24 @@
-import { Component, Input, OnInit } from '@angular/core'
+import { Component, Input, OnDestroy, OnInit } from '@angular/core'
 import {
   InfiniteScrollModel,
   InfiniteScrollOptionsDefault,
   ResultsListLayout,
 } from '@lib/common'
+import { Subscription } from 'rxjs'
+import { distinctUntilChanged } from 'rxjs/operators'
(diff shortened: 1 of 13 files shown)


libs/search/src/lib/results-list/results-list.container.component.ts
```diff
@@ -25,6 +26,9 @@ export class ResultsListContainerComponent implements OnInit {
       ...this.scrollableOptions,
     }
     this.facade.setResultsLayout(this.layout)
+    this.facade.isEndOfResults$.subscribe((isTheEnd) => {
```
fgravin, line 29: all subscription must be unsubscribed on component destroy

fgravin, line 29: `distinctUntilChanged` may avoid useless change detection cycles

libs/search/src/lib/state/reducer.ts
```diff
@@ -125,6 +125,15 @@ export function reducerSearch(
...
       }
     }
+    case fromActions.CLEAR_PAGINATION:
+      return {
+        ...state,
+        params: {
+          ...state.params,
+          from: 0,
+          size: state.params.size,
+        },
```
fgravin, line 135:
```suggestion
        params: {
          ...state.params,
          from: 0,
        },
```
size should be already good with spread operator

fgravin (approved):
Very good and readable, thanks

PR #102, opened 2021-02-17
[Datafeeder] Design polish

y-vynohradov:
- [x] All inputs should be aligned in the wizard (text, textarea, date ...)
- [ ] Scale input should not be a combo, but should allow text and propose values
- [x] on each wizard page, the first input must be focused
- [x] inputs focus should have the same color for all inputs
- [ ] the keyword tagsinput should allow focus whereever you click in the box

### DIFF
--- a/apps/datafeeder/src/styles.css
+++ b/apps/datafeeder/src/styles.css
@@ -50,6 +50,28 @@ ui-button button[type=button].secondary:hover {
   color: white;
 }
 
+.focus\:border-primary:focus {
+  border-color: var(--color-primary-darker) !important;
+}
+
+my-date-picker .mydp {
+  border-width: 2px;
+  border-color: var(--color-primary);
+}
+
+my-date-picker .mydp:focus-within {
+  border-color: var(--color-primary-darker) !important;
+}
+
+my-date-picker::ng-deep input::ng-deep:focus {
+  border-color: var(--color-primary-darker) !important;
+}
+
+tag-input:focus-within, my-date-picker:focus-within {
+  border-color: var(--color-primary-darker) !important;
+}
+
+
 .grey1 {
   fill: #f2f3f4;
 }
--- a/apps/datafeeder/src/styles.css
+++ b/apps/datafeeder/src/styles.css
@@ -50,26 +50,17 @@ ui-button button[type=button].secondary:hover {
   color: white;
 }
 
-.focus\:border-primary:focus {
-  border-color: var(--color-primary-darker) !important;
-}
-
-my-date-picker .mydp {
-  border-width: 2px;
-  border-color: var(--color-primary);
-}
-
-my-date-picker .mydp:focus-within {
-  border-color: var(--color-primary-darker) !important;
-}
-
-my-date-picker::ng-deep input::ng-deep:focus {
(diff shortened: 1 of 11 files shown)


libs/editor/src/lib/components/wizard/wizard.component.ts
```diff
@@ -37,11 +50,23 @@ export class WizardComponent implements OnInit {
...
   }
+
+  private setFocus() {
+    setTimeout(() => {
+      const inputEl =
+        this.wizardFieldsEl.nativeElement.getElementsByTagName('input')[0] ||
+        this.wizardFieldsEl.nativeElement.getElementsByTagName('textarea')[0]
+
+      inputEl.focus()
+    }, 100)
```
tonio, line 70: Doesn't look like a robust solution. I guess `afterViewInit` would be more error-proof.

apps/datafeeder/src/styles.css
```diff
@@ -50,6 +50,28 @@ ui-button button[type=button].secondary:hover {
...
+  border-color: var(--color-primary-darker) !important;
+}
+
+my-date-picker::ng-deep input::ng-deep:focus {
+  border-color: var(--color-primary-darker) !important;
+}
+
+tag-input:focus-within, my-date-picker:focus-within {
+  border-color: var(--color-primary-darker) !important;
+}
```
tonio, line 72: I would try to avoid `!important` if possible. Moreover, try to avoid duplicating declaration  when shared accross selector:
```css
foo,
foo:focus,
foo:focus-within {
  border-color: …;
}
```

libs/editor/src/lib/components/wizard-field/wizard-field.component.ts
```diff
@@ -28,6 +29,7 @@ import { Subscription } from 'rxjs'
 })
 export class WizardFieldComponent implements OnInit, AfterViewInit, OnDestroy {
   @Input() wizardFieldConfig: WizardFieldModel
+  @Input() focus: boolean
```
tonio, line 32: is that used somewhere?

libs/ui/src/lib/chips-input/chips-input.component.css
```diff
@@ -28,3 +28,19 @@ tag-input::ng-deep
...
+  width: 100% ;
+}
+
+tag-input::ng-deep form {
+  height: 100%;
+}
+
+tag-input::ng-deep input {
+  width: 100%;
+}
```
tonio, line 46: same here, declaration is shared, shouldn't be redundant.

tonio (changes requested):
Thanks, see comments inline.

fgravin (approved):
LG thanks

fgravin:
@tonio I am testing, can you check the boxes of what is done in the PR description please, thanks.

fgravin:
Some new feedback after testing:

- [ ] padding in tex-input and text-area are not the same
- [ ] text color in tex-input and text-area are not the same
- [ ] tags-input: focus position when the list is not empty is weird
- [ ] date: go on step 3 then back step 2, when back step 3 date is: `01 undefined 1970` (in english)

PR #106, opened 2021-02-24
I18n interceptor

fgravin:
Add an interceptor to set 'Accept-Language` header to the current application language.

Also do some refactoring in `i18n` folder.

### DIFF
--- a/libs/common/src/lib/i18n/i18n.constants.ts
+++ b/libs/common/src/lib/i18n/i18n.constants.ts
@@ -1,0 +1,59 @@
+import { HttpClient } from '@angular/common/http'
+import { ToolsApiService } from '@lib/gn-api'
+import { TranslateCompiler, TranslateLoader } from '@ngx-translate/core'
+import { TranslateHttpLoader } from '@ngx-translate/http-loader'
+import { TranslateMessageFormatCompiler } from 'ngx-translate-messageformat-compiler'
+import { Gn4TranslateLoader } from './gn4.translate.loader'
+
+export const DEFAULT_LANG = 'en'
+
+export const LANG_3_TO_2_MAPPER = {
+  eng: 'en',
+  dut: 'nl',
+  fre: 'fr',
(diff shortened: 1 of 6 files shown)


libs/common/src/lib/i18n/i18n.constants.ts
```diff
@@ -0,0 +1,59 @@
+import { HttpClient } from '@angular/common/http'
+import { ToolsApiService } from '@lib/gn-api'
+import { TranslateCompiler, TranslateLoader } from '@ngx-translate/core'
+import { TranslateHttpLoader } from '@ngx-translate/http-loader'
+import { TranslateMessageFormatCompiler } from 'ngx-translate-messageformat-compiler'
+import { Gn4TranslateLoader } from './gn4.translate.loader'
+
+export const DEFAULT_LANG = 'en'
+
+export const LANG_3_TO_2_MAPPER = {
```
fxprunayre, line 10: We could rely on https://github.com/geonetwork/geonetwork-ui/blob/eddf2bee37f8ca1f03dd64c646417e33949813d7/libs/common/src/lib/services/constant.ts#L30-L45 ?

fgravin: I've just moved them from the module to avoid circular dependencies.

To answer the question, I would say yes and no, to me, the mapping between 2 and 3 letters languages should be somewhere (either hardcoded or in the config or db) but the list of available languages for the UI is something different, ideally, an admin that would manage the available languages should not have to care about the languages codes (neither 2 or 3 letter ones), but would pick languages up from a list. I think we have a table for that in the DB with codes and language labels.

But I agree that we should rely on a remote config to avoid duplicating codes.

fxprunayre: Yep we have all in db, we should let the admin select the language they used in the catalog in the admin and then just provide the list (with mapping) from the db.... but needs some work in backend and API.

PR #107, opened 2021-02-25
Aggregation minor improvements

fxprunayre:
Restore some GN4 logic.

* fix(aggs): Preserve aggregation order as defined in the configuration. See https://github.com/geonetwork/core-geonetwork/blob/4.0.x/web-ui/src/main/resources/catalog/components/elasticsearch/EsFacet.js#L287.
* fix(aggs): Aggregation search field is not always the key.
    
User may need to label differently an aggregation. The label is based on the key. The search field can then be different. Same as in GN4.
    
Example of configuration of a more restricted facet on resourceType labeled differently:
    
    ```json
    facetConfig: {
            DataServiceOrMap: {
              terms: {
                field: 'cl_hierarchyLevel.key',
                include: 'dataset|service|map'
              },
              meta: {
                field: 'cl_hierarchyLevel.key'
              }
            },
    ```

### DIFF
--- a/demo/eea/datacatalogue.html
+++ b/demo/eea/datacatalogue.html
@@ -611,7 +611,24 @@
                     >
                       <gn-facets
                         api-url="https://apps.titellus.net/geonetwork/srv/api"
-                        facet-config='{"tag.default":{"terms":{"field":"tag.default","include":".*","size": 10}}}'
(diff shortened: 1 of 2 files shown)


fgravin (approved):
LG so far

PR #109, opened 2021-02-26
Add filter params in ES query

fgravin:
Allows to set permanent filters to all elasticsearch search requests.
These filters are stored in the `filter` property of es request body (in `query.bool` path).
```
request
  - query
    - bool
      - must (user filters)
      - filter (permanent filters)
```
**State**
The permanent filters can be configured 2 ways:

- custom way: a friendly user format (based on internal state filters structure)
- elastic: a pure filter object passed as it is

Example:
```js
this.searchFacade.setConfigFilters({
  custom: {
    'cl_hierarchyLevel.key': {
      service: true,
    },
  },
  elastic: [
     { term: { 'cl_hierarchyLevel.key': 'service' } },
     { query_string: { "query": "-resourceType:service" }
  ],
})
```

custom config will generate:
```js
{ query_string: { query: '(cl_hierarchyLevel.key:"service")' } }
```

elastic config will generate:
```js
[
   { term: { 'cl_hierarchyLevel.key': 'service' } },
   { query_string: { "query": "-resourceType:service" }
]
```

Both configurations can't coexist, the elastic one has the priority

**Web components**
`gn-result-list` can now take 2 inputs to define the search state: `filter` and `query`

Example

```html
<gn-results-list
      query="+tag.default:forest"
      query='{"tag.default": { "land use": true }}'
      filter='{"custom": { "cl_hierarchyLevel.key": { "service": true } }}'
      filter='{"elastic": [{ "term": { "cl_hierarchyLevel.key": "service" } }]}'
></gn-results-list>

```

### DIFF
--- a/demo/eea/results-list.html
+++ b/demo/eea/results-list.html
@@ -842,7 +842,7 @@
                       api-url="https://apps.titellus.net/geonetwork/srv/api"
                       layout="TITLE"
                       size="4"
-                      filter="forest"
+                      query="forest"
                       primary-color="#251"
                       secondary-color="#c2e9dc"
                       main-color="#212029"
--- a/libs/common/src/lib/i18n/gn4.translate.loader.spec.ts
+++ b/libs/common/src/lib/i18n/gn4.translate.loader.spec.ts
@@ -5,7 +5,7 @@ import { Gn4TranslateLoader } from './gn4.translate.loader'
 
 const toolsApiServiceMock = {
   getTranslationsPackage1: jest.fn(() =>
-    of({ farming: 'Farming', legacy: '{{ id }} id' })
+    of({ farming: 'Farming', legacy: ' {{ id }} id' })
   ),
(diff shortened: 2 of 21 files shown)


fgravin:
@fxprunayre the body with
```js
filter: { query_string: { query: '(cl_hierarchyLevel.key:"service")' } }
```
does not impact the search, any idea ?

fxprunayre:
> does not impact the search, any idea ?

```json
GET /gn-records/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "match_all": {}
        }
      ],
      "filter": [
        {
          "query_string": {
            "query": "+cl_hierarchyLevel.key:service"
          }
        }
      ]
    }
  }
}
```
works fine. See https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html 

Any error reported by ES ?

PR #112, opened 2021-03-31
[datafeeder] fix i18n path when built with --base-href

tonio:


tonio:
`basehref` is `/` by default: should be safe to have this by default. PR updated.

fgravin (approved):
Sorry for the delay

PR #114, opened 2021-04-12
datafeeder: support ENV for settings

fgravin:
Update the settings mecanism to use angular `environnement` object instead of a custom Class `SETTINGS` for specific ENV settings.
It also avoids an asynchronous load to the `settings.json` file as it's now done through a javascript file, imported directly in the index.html file.

The UI has been tested alone, the backend was not working, so not sure about the `config/frontend` stuff, that's why I put it in draft while I have not tested the backend.

### DIFF
--- a/apps/datafeeder/README.md
+++ b/apps/datafeeder/README.md
@@ -30,4 +30,4 @@ envsubst < assets/env.template.js > assets/env.js
 ```
 
 ### Other settings
-Other settings are fetched as a second step from the API `/datafeeder/config/setting`
+Other settings are fetched as a second step from the API `/datafeeder/config/frontend`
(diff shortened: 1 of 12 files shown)


apps/datafeeder/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+# Set environment variable
+export DATAFEEDER_API_URL="/my-datafeeder";
+
+# Replace variables in env.js
+envsubst < assets/env.template.js > assets/env.js
+```
+
+### Other settings
+Other settings are fetched as a second step from the API `/datafeeder/config/setting`
```
pmauduit, line 33: seems that it is `/datafeeder/config/frontend` instead: https://github.com/georchestra/georchestra/blob/datafeeder/integration/datafeeder/api.yaml#L18-L32

tonio (approved):
Untested, code looks good. I think PR #112 is not necessary anymore with relative base-href  ?

PR #115, opened 2021-04-16
[datafeeder] add ability to mark wizard field required

tonio:


libs/ui/src/lib/text-input/text-input.component.ts
```diff
@@ -7,15 +14,36 @@ import { Observable, Subject } from 'rxjs'
...
+
   rawChange = new Subject<string>()
 
   constructor() {}
 
   ngOnInit() {
     this.valueChange = this.rawChange.pipe(distinctUntilChanged())
   }
+
+  ngAfterViewInit() {
```
fgravin, line 33: this could have been factorized a bit by using a super class

libs/editor/src/lib/components/wizard/wizard.component.ts
```diff
@@ -51,7 +55,17 @@ export class WizardComponent implements OnInit, AfterViewInit {
     this.stepChanges.emit(this.currentStep)
   }
 
-  handleNextBtnClick() {
+  async handleNextBtnClick() {
```
fgravin, line 58: why using a promise ?

fgravin (commented):
Tested and works well.
I'm not sure about the user experience though. It's a bit weird because input as marked as error (orange border) while we haven't entered anything yet. Also, the focused input is not orange then.
Also all fields are mandatory anyway.

libs/editor/src/lib/components/wizard/wizard.component.ts
tonio: The use of `await` inside the function require it to be async.

fgravin: Yes but why to you use `await` inside the function ?
```js
this.translate.get('datafeeder.wizard.emptyRequiredValuesMessage').pipe(take(1))
   .subscribe( msg => alert(msg} )
```

fgravin (approved):

PR #117, opened 2021-04-30
WC facet sample fixes

fgravin:
This PR adresses severals fixes en enhancements

- [x] i18n file is load from each WC depending on the url passed as an input
- [x] facet more fixes when no intial size is given in the configuration
- [x] hide facet block when no item
- [x] display thumbnails as data url when available


fgravin:
@fxprunayre I finally merge it without review

PR #122, opened 2021-06-09
[datafeeder] fix scale & crs not sent.

tonio:


fgravin (approved):
Looks good but i don't see what happen concretely when you select a crs in the list ?
Do we request something ? Do we use the crs code somewhere ?

tonio:
It's just stored in the LS to be sent at the publication step. A format mismatch prevent it to be correctly done before.

PR #123, opened 2021-06-11
[datafeeder] Avoid multiple requests to thesaurus

tonio:
& then decrease debounce time.

### DIFF
--- a/libs/ui/src/lib/chips-input/chips-input.component.html
+++ b/libs/ui/src/lib/chips-input/chips-input.component.html
@@ -5,6 +5,7 @@
   [placeholder]="placeholder"
   [secondaryPlaceholder]="placeholder"
   [animationDuration]="{ enter: '0ms', leave: '0ms' }"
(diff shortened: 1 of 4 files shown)


fgravin (approved):
It looks a bit hacky but anyway i plan to remove this shity component and use angualr-material one in a near future so we can go forward, thanks.

libs/ui/src/lib/chips-input/chips-input.component.html
```diff
@@ -5,6 +5,7 @@
   [placeholder]="placeholder"
   [secondaryPlaceholder]="placeholder"
   [animationDuration]="{ enter: '0ms', leave: '0ms' }"
+  [onTextChangeDebounce]="100"
```
fgravin, line 8:
```suggestion
  onTextChangeDebounce="100"
```
try that maybe

fgravin: Also the name presumes it's an event, not a input ?

PR #124, opened 2021-06-15
Add table component to display result (simple component)

davinciagf:
To be completed with angular material aiming to use mat-table.

The main goal is to be able to create a webcomponent which allows to embed a simple table with a list of records. eg.

A simple table can display simple fields eg. titles:

```html
<gn-result-list 
  layout="TABLE"
  fields="resourceTitle.default,resourceAltTitle.default"/>
```

Here the question is : currently we have RecordSummary which is a limited object and the Elasticsearch index document - How can we have whatever field from the index document ?

or more advanced fields eg. a link to the map ...

Here the question is : How to render advanced field eg. overview, links?

### DIFF
--- a/libs/common/src/lib/models/search.model.ts
+++ b/libs/common/src/lib/models/search.model.ts
@@ -36,6 +36,12 @@ export enum ResultsListLayout {
   LIST = 'LIST',
   TEXT = 'TEXT',
   TITLE = 'TITLE',
+  TABLE = 'TABLE',
+}
+
+export enum FieldsList {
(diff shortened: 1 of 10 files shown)


fgravin:
@davinciagf you should find anything you want to implement your features:

- `TableComponent`
- `_source` in search state with `facade.setIndexFields` action
- generic mappers if you need specific ones

davinciagf:
@fgravin 
Great, I will test it in the next few weeks. 
Thanks

PR #126, opened 2021-06-18
geo-table dataviz component

jahow:
This PR introduces a new smart component leveraging the existing table and map:
image

For now the component uses a hardcoded geojson dataset taken from Sextant.

To do:
- [ ] allow specifying data as a link to a file or WFS service
- [ ] allow filtering on columns
- [x] synchronize map view with table
- [ ] continuous loading of table view (to avoid slowing down the DOM)


libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.ts
```diff
@@ -1,8 +1,40 @@
...
+  geojson = FEATURE_COLLECTION_POINT_FIXTURE_4326
+
+  get tableData() {
+    return this.geojson.features.map((f) => ({
+      id: f.id,
+      ...f.properties,
+    }))
+  }
+
+  get mapData() {
```
maelleAubert, line 24: this is executed a lot of time, rather init the config in the `ngInit `once

jahow: maybe if I switch to OnPush it will only happen on input change? using OnInit means we assume the input will not change later on

fgravin: `OnPush` is really expected here right, but I don't think it's enough, put a console log in your getter and you'll understand :)

fgravin (commented):
Looks good so far

fgravin:
@jahow I have worked on your PR to implement the bindings between the map and the table, with the last architecture it was pretty straightforward :)

So this implement just create interaction between the map and the table as it's done in GranLyon
I think we can address the other bullets of your description in other PR
We will also wait for the UX/UI to improve the global behavior.

Screenshot from 2021-08-09 16-36-32

libs/ui/layout/src/lib/table/table.component.ts
```diff
@@ -1,22 +1,35 @@
...
 
+const rowIdPrefix = 'table-item-'
+
 @Component({
   selector: 'gn-ui-table',
   templateUrl: './table.component.html',
   styleUrls: ['./table.component.css'],
 })
 export class TableComponent implements OnInit, AfterViewInit {
   @Input() data: any
```
jahow, line 24: The component should explicitly ask for an `id` field on the input data now

libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.ts
```diff
@@ -1,40 +1,168 @@
...
-  get tableData() {
-    return this.geojson.features.map((f) => ({
+  constructor(
+    private manager: MapManagerService,
+    private featureInfo: FeatureInfoService,
+    private changeRef: ChangeDetectorRef
+  ) {}
+
+  ngOnInit(): void {
+    this.tableData = this.geojsonToTableData(this.data)
```
jahow, line 61: The component should advertise that it will not update if the input data changes IMO (or a `ngChanges` handler should be added, but that might be non trivial)

libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.ts
```diff
@@ -1,40 +1,168 @@
...
+
+  private map: Map
+  private view: View
+  private vectorLayer: VectorLayer<VectorSource<Geometry>>
+  private vectorSource: VectorSource<Geometry>
+  private features: Feature<Geometry>[]
+
+  tableData: any
+  mapContext: MapContextModel
+  selectionId: any
```
jahow, line 49: Ideally we should refrain from using `any`; can't all ids be `string` here?

libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.ts
```diff
@@ -1,40 +1,168 @@
...
+  @Input() data: FeatureCollection = FEATURE_COLLECTION_POINT_FIXTURE_4326
+  @ViewChild(TableComponent) uiTable: TableComponent
+
+  private map: Map
+  private view: View
+  private vectorLayer: VectorLayer<VectorSource<Geometry>>
+  private vectorSource: VectorSource<Geometry>
+  private features: Feature<Geometry>[]
+
+  tableData: any
```
jahow, line 47: same, avoid `any` if possible

jahow (commented):
Tested, the map <> table interaction is really nice now.

I have a few comments on the code, I'll make another review/test when the stories are ready for the dataviz component.

jahow:
@fgravin this PR is ready as far as I'm concerned! the story for the geotable component is working fine.

Thanks for your help

PR #127, opened 2021-06-21
Migrate nx + upgrades

fgravin:
Core refactoring of the architecture

- use Nx workspaces
- upgrade storybook
- upgrade TailwindCSS
- add angular-material
- add search suggest autocomplete component
- add first components of dataviz (grid, map)


libs/ui/map/jest.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   displayName: 'ui-map',
   preset: '../../../jest.preset.js',
+  transformIgnorePatterns: ['node_modules/(?!(ol))'],
```
fgravin, line 4: can't we do that for all projects ? at root level ?

jahow: probably, that would be more future-proof

PR #132, opened 2021-08-03
Migrate ol 6.6

fgravin:
Mostly to have a better typescript support.

`definitlytyped` are now offically exported in the package what eases the development under IDE such as IntelliJ
- auto import
- completion
- types

This needed some fixes in the actual type definition of GN code related to openlayers.


jahow (approved):
Nice! good that you could get rid of the TS path alias as they were potentially misleading. Thanks!

PR #135, opened 2021-08-06
Restore Storybook for UI components

jahow:
This PR begins with a migration to Nx 12.6, followed by an installation of the `@nrwl/storybook` plugin.

All the stories for the UI components in `ui-search`, `ui-inputs`, `ui-widgets`, `ui-layout` and `ui-catalog` have been migrated to storybook 6 and new ones have been written.

Also, by default Nx creates one storybook per lib which is not practical to look at all UI components at once. To work around that I have created a new app called `demo` which is empty but has a storybook as well, only its storybook will aggregate the stories of all apps and libs in the repo. As such, launching `nx run demo:storybook` will give an overview of _all_ UI stories in the repo:
image

Note that the storybook currently does not showcase the webcomponents but I think we should be able to do that in a second PR as well as restore the automatic build and deploy of the storybook using the CI.

Finally, the translations in the demo app are just a copy of the ones in the search app; this is IMO a very temporary way of making things work, a more future-proof system should be put in place.


fgravin:
> Also, by default Nx creates one storybook per lib which is not practical to look at all UI components at once

I wonder then if it's usefull to keep `.storybook` configuration files in all libraries ?

jahow:
> > Also, by default Nx creates one storybook per lib which is not practical to look at all UI components at once
> 
> I wonder then if it's usefull to keep `.storybook` configuration files in all libraries ?

You're right, I thought about that afterwards. Running the `@nrwl/storybook` generator against each library was useful because it scaffolded stories for each component automatically, but I think we can now remove the `.storybook` folders and storybook-related targets for these.

libs/ui/search/src/lib/results-hits-number/results-hits-number.component.ts
```diff
@@ -1,10 +1,13 @@
...
+interface HitsNumberInput {
+  value: number
+}
 @Component({
   selector: 'gn-ui-results-hits-number',
   templateUrl: './results-hits-number.component.html',
 })
 export class ResultsHitsNumberComponent {
-  @Input() hits: number
+  @Input() hits: HitsNumberInput
```
jahow, line 11: why was that required?

fgravin: Because the type was wrong indeed, `hits` is not a number, but an object with a `value` property.
So in the test, giving `component.hits = 10` was wrong I guess.
In the template, it clearly uses `hits.value` so the typing was incorrect somewhere.

jahow: Weird that this component does not take a number as input. The reasoning behind the wrapping object is unclear. IMO this has to be revisited in the future.

jahow:
Ok, I've removed the storybook-related stuff in each individual ui libs, now only the `demo` app has storybook enabled.

There is still an issue with Tailwind not purging the CSS (this makes the build fail for size reasons), looking into it

libs/ui/search/src/lib/results-hits-number/results-hits-number.component.ts
fxprunayre: We may on some setup have a non exact hit number with stuff like ""there are at least 10000 hits"" See  https://www.elastic.co/guide/en/elasticsearch/reference/current/search-your-data.html#track-total-hits

fgravin: Yes but it should be the parent responsibility to handle that, and not the UI component who's responsibility is just to display a number, not a big deal thought for now :)

PR #137, opened 2021-08-13
Restore auto deploy of storybook through gh-pages on branches

jahow:
Now that Storybook is functional we can start deploying it on PRs again


jahow:
There's work to do left for the demo app and the webcomponents but this is straightforward enough, merging

PR #138, opened 2021-08-19
Proposal for generic ES mappers

fgravin:
The PR proposes a way to deal with ES mapped objects to store a metadata.
ES response format is not optimized so we need to transform it into kind of a metadata object.
At the moment, we wanted to have several type definitions for different level of completion of the metadata object
- RecordSummary
- RecordBrief
- RecordFull

Even though we want to keep these concepts with predefined set of metadata properties, this format is a bit static.
Actually, the content of the metadata object should directly depend on the `_sources` from the ES request.

This draft shows a way to have a generic mapper that maps all the properties from the response `_source` object with a dedicated named function from the `ElasticFieldMapperService` (eg `mapResourceTitleObject`).
It contains generic mappings, and specific for particular fields.

I haven't touched the test and it still pass, meaning this is the minimum implementation to keep the actual behavior.

If you have a custom index with custom fields which are not generic, you can inherit `ElasticFieldMapperService` and implement your own mapping field functions, and inject it in the application module.
If you want more flexibility in the mapping (like in sextant where the content of the abstract depends on another field), you can inherit and inject a new `ElasticMapperService`.

Another approach would have been to have a class `Metadata` (or `Record`) that do all the mapping in the constructor. So `Record` is not a type but a class, and we wouldn't have a mapping service. The problem with that is that we can't use Angular injection system to override the behavior of the Record mapping.

I'm also interested by your opinion @fxprunayre 
I worked on this mostly to prepare @davinciagf work to have custom index fields in a metadata result table (the other part of this work is to set the `_source` in search state).


libs/feature/search/src/lib/elasticsearch/mapper/elasticsearch.mapper.ts
```diff
@@ -0,0 +1,39 @@
...
+        this.getMappingFn(fieldName) || this.fieldMapper.mapGenericField
+      ).bind(this.fieldMapper)
+      mappingFn(record, _source, fieldName)
+    })
+
+    return record as RecordSummary
+  }
+
+  private getMappingFn(fieldName: string) {
+    return this.fieldMapper[`map${this.captitalizeFirstLetter(fieldName)}`]
```
jahow, line 34: I'm a bit uncomfortable with using indirection for a method name. This often leads to code that is harder to search. For example with this system, if I look for `mapResourceTemporalExtentDateRange` globally in the project I will never find explicit usages even though it is actually used.

How about using a dict for the field mappers in `ElasticsearchFieldMapper`? The key would be the full, untransformed field name and the value is the mapper.

jahow (commented):
Using injectable services for mapping is a great solution, flexible and clean. I just think the field mapper service should not define explicit methods for each file but more like a dict or something like that (see my comment in the code).

Thanks!

libs/feature/search/src/lib/elasticsearch/mapper/elasticsearch.mapper.ts
fgravin: > How about using a dict for the field mappers in `ElasticsearchFieldMapper`? The key would be the full, untransformed field name and the value is the mapper.

Yes I agree it's more readable, thanks !

jahow (approved):
looking good now, thanks!

PR #139, opened 2021-08-19
Add ES _source in search state

fgravin:
- Add the request `_source` param in the state so it is configurable
- build the ES request payload from the `_source` in the state.

By default, the set `ES_SOURCE_SUMMARY` is used.

We could imagine getting this list from the DB config if needed.

The PR also get rid of the `@types/elasticsearch` types as they are not releavant anymore.
It also clean up the `search-ui` from being used as a test UI fro the `geo-table` component.


jahow (approved):
Good step forward, thanks. Feels like we're in the right direction to have a truly modular search system!

I was thinking that "Source" is not a very explicit term in that context. Things like:
* `SourceParam` type
* `SetConfigSource` action
* and most importantly `setConfigSource()` on the facade
might become confusing with time, and actually refer to a technical term that is specific to Elasticsearch configuration.

How about (simple suggestion):
* `ReturnedFields` type
* `SetReturnedFields` action
* `setReturnedFields()` on the facade

I'm approving and letting you judge on this :)

fgravin:
Thanks @jahow for the suggestions.
I went for `IndexFields` to talk about the `_source` param.

It's more readable in the long run.

PR #141, opened 2021-08-23
Restore Storybook for Web Components

jahow:
This PR transforms the `webcomponents` directory into an actual app handled by Nx. The content mostly stays the same and angular/elements stories have been rewritten to work with Storybook 6.

The approach for the storybook was the same as in #135: a new app was added to "host" the storybook. This makes things easy because: 1. an app is always buildable, whereas a lib is not (and making a lib buildable adds other issues) and 2. an app has e2e tests, which might come in handy when we want to ensure all components render correctly on the CI.

Note that for Web Components stories, the user can change the whole HTML template of the story.

Remaining to do:
- [x] ~Fix issue with web components only rendering once:  ~
      ~this is due to custom elements being registered only once for all the storybook, but when switching stories SB actually tears down all the Angular runtime and recreates it, thus the valid injector changes and the existing custom elements are using an deprecated injector~
      ~Apparently [there is currently no way to redefine a custom element](https://github.com/WICG/webcomponents/issues/754)~
      Went with the previous method which is building the webcomponents app prior to the storybook
- [x] Adjust documentation and NPM scripts
- [x] deploy storybook on the CI


fgravin (approved):
Thanks for the work, works well on the gh-pages except for the facet.
Config looks ok.

PR #143, opened 2021-08-24
Add Mdview feature state

fgravin:
Add a new feature module and feature state `MdView` to manage the metadata detail page informations.

`loadFull` action trigger an ES `getById` request and dispatch `loadFullSuccess/loadFullFailure` accordingly (through effects).
`loadFull` can be dispatched manually or is trigger when a `setPreview` is sent.

The preview is in the state cause already loaded when you come from the results list, so you can have first basic infromations while the full metadata is loading.
We avoid to load to many _source fields for the search page, and do an additional full loading for the specific metadata.

The module and actions dispatching is done in the application, so you can use `resultsListContainer` in a different scope (eg from web components).

The metadata UI preview item click is propagated to the `result-list.container` and to the application which is responsible to set the `preview` in the store.


fgravin:
@jahow ready for review, thanks

libs/ui/search/src/lib/results-list/results-list.component.html
```diff
@@ -5,13 +5,17 @@
   ></ng-container>
   <div *ngSwitchCase="layoutEnum.LIST" class="gap-4 p-4">
     <div class="pb-4" *ngFor="let record of records">
-      <gn-ui-record-preview-list [record]="record"></gn-ui-record-preview-list>
+      <gn-ui-record-preview-list
+        [record]="record"
+        (mdSelect)="mdSelect.emit($event)"
```
jahow, line 10: To be honest I'm not sure if the `mdSelect` event brings any value; why not just do `(click)="mdSelect.emit(record.uuid)"` here, and get rid of all the emitter logic in `gn-ui-record-preview-list`?

libs/feature/search/src/lib/metadata-view/state/mdview.actions.ts
```diff
@@ -1,6 +1,17 @@
...
 )
+
+export const setPreview = createAction(
+  '[Metadata view] Set preview',
+  props<{ preview: RecordSummary }>()
+)
+
+export const setFull = createAction(
+  '[Metadata view] Set full metadata',
+  props<{ full: RecordSummary }>()
```
jahow, line 16: Should be `RecordFull` maybe?

jahow (approved):
Made some suggestions, especially to simply the logic in the md results preview components. Also it's weird to see the GN url in the `environment.ts` files, we should probably put it somewhere else eventually but this has to be discussed.

Other than that nothing to add, solid foundation for what's coming, thanks!

jahow:
@fgravin actually the GN url is currently in the proxy config of the project, so it's strange that you had to add it to the environments files. Why was that necessary?

libs/feature/search/src/lib/metadata-view/state/mdview.actions.ts
fgravin: I don't adress the typing in this PR.
We need to refine how we use `RecordSummary` `RecordFull` etc...

libs/ui/search/src/lib/results-list/results-list.component.html
fgravin: Yes should be way better than listening to the event in the `Preview` abstract component.
Thanks

libs/feature/search/src/lib/metadata-view/state/mdview.actions.ts
jahow: No problem, I though it was a mistake

fgravin:
> @fgravin actually the GN url is currently in the proxy config of the project, so it's strange that you had to add it to the environments files. Why was that necessary?

`BASE_PATH` must be injected cause I use it in `MetadataUrlService`, also I think we have to set it up and it must be configurable.
In dev we use a proxy, not in prod so the url are differents (even though we could use titellus direct url from the dev too). 

Anyway, I think we might update this value in production app by passing ENV variable in the docker composition, it's usually the way to go, see https://github.com/geonetwork/geonetwork-ui/blob/master/apps/datafeeder/src/assets/env.js and the corresponding `environment.ts` files.

I think we can address that later anyway, the point here is that i wanted different values for dev and prod (gh-pages).

libs/ui/search/src/lib/results-list/results-list.component.html
fgravin: Actually it does not work because the `PreviewComponent` is a HREF that by default opens the `metadataUrl` link in an other tab.
My event was breaking that anyway. I'm wondering what is the best approach:
1. remove all the a-href  link stuff and propagate an event from `gn-ui-results-list`. In the application (or the webcomponent), get the event and put it in the state or manually open the link in another tab in javascript.
2. pass down a property to tells if we register to the event or not, if not the normal HTML href processes the click.

jahow: having `<a>` elements with href pointing to a full metadata view is pretty nice I think, it makes sense in some cases and lets the browser handle things natively.

I think it might work with 2: we stop event propagation when the record is clicked if we want to intercept it

PR #145, opened 2021-08-26
Dataviz - Figure component

fgravin:
Add dataviz components to display figures from a dataset.

- icon
- figure + unit
- text

Screenshot from 2021-08-25 17-24-00

The figure is computed from an expression over a dataset.
Tha dataset is an array of objects.
The expression is something like operation|column (eg sum|pop)
Only average and sum are available for now, on a single column.

Web component stuff will be part of another PR, as the extraction of the dataset from the metadata is not done.

Contains:
- [x] Dumb component
- [x] Storybook
- [x] Smart component
- [x] figure computation from expression
- [ ] ~~Extract dataset from metadata~~
- [ ] ~~Web component~~

### DIFF
--- a/libs/ui/layout/src/lib/figure/figure.component.ts
+++ b/libs/ui/layout/src/lib/figure/figure.component.ts
@@ -14,6 +14,6 @@ import {
 export class FigureComponent {
   @Input() icon: string
   @Input() title: string
-  @Input() figure: number
+  @Input() figure: string
   @Input() unit: string
 }
--- a/libs/ui/layout/src/lib/figure/figure.component.ts
+++ b/libs/ui/layout/src/lib/figure/figure.component.ts
@@ -1,0 +1,19 @@
+import {
+  Component,
+  OnInit,
+  ChangeDetectionStrategy,
+  Input,
+} from '@angular/core'
+
+@Component({
+  selector: 'gn-ui-figure',
+  templateUrl: './figure.component.html',
+  styleUrls: ['./figure.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FigureComponent {
+  @Input() icon: string
+  @Input() title: string
+  @Input() figure: number
+  @Input() unit: string
+}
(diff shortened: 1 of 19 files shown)


libs/ui/layout/src/lib/figure/figure.component.ts
```diff
@@ -0,0 +1,19 @@
...
+} from '@angular/core'
+
+@Component({
+  selector: 'gn-ui-figure',
+  templateUrl: './figure.component.html',
+  styleUrls: ['./figure.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class FigureComponent {
+  @Input() icon: string
```
jahow, line 15: I think it would be nice to be able to specify the style of the icon as well, since [Material provides many](https://fonts.google.com/icons) (outlined, filled, rounded...).

Maybe this could be done by specifying `[rounded]people` as the icon name for example? (adding another parameter would maybe tie us too much to the Material Icons lib)

Thus anything specified in brackets would be considered a "variant" and simply ignored if not valid/found

libs/ui/layout/src/lib/figure/figure.component.ts
```diff
@@ -0,0 +1,19 @@
...
+  @Input() title: string
+  @Input() figure: string
```
jahow, line 17: I wonder why you switched back to string for the actual figure? Having a number means it can be formatted correctly for the user locale, and also shown in different ways depending on available space, e.g. : the same data could be shown as `12k`, `12,127`, `12,127.84` depending on the width of the component

Having figures that are strings makes sense as well, for qualitative data for example, eg. : "Most common name in 2021: **John**", but this is a less common use case I think.

jahow (approved):
This is looking really good, thanks!

I'm a bit concerned about the expression system introduced here. Will it stay that simple? We should try to think of a system that provides good developer experience (e.g. meaningful errors in case of invalid expression) as well as flexibility. Also introducing a new syntax is maybe unnecessary; we could rely on SQL-like queries for example:
```
SELECT AVG(pop) WHERE country_code = 'FR'
```

At least this is a very well known DSL and there's probably libraries out there to parse it.

Anyway, this can be merged like that for now and revisited later based on actual use cases.

jahow:
@fgravin I've rebased the PR to merge it and will move on to implementing a web component using this, thanks for your work

PR #146, opened 2021-08-27
Improve Elasticsearch mappers by using a functional approach

jahow:
This PR refactors the Elasticsearch mapper service to have more of a functional style, i.e. not mutating objects while mapping.

The field mappers now rely on atomic operations that can be combined in different ways, and which provide a safe framework for parsing ES responses, handling missing and/or unexpected properties without crashing (hopefully).

Consumers can still extend the field mapper service and add or replace specific mappers according to their needs, eg:
```ts
class CustomMapper extends ElasticsearchFieldMapper {
  protected fields = {
    ...super.fields,
    toto: (source) =>
      ({
        ...source,
        newProp: selectField(source, 'toto')
      })
  }
}
```

Using pure functions for mapping instead of relying on mutating objects is a safeguard against bugs where objects end up being mutated incorrectly (which is very hard to track).

This PR also reduces the setup time for tests by approx. 50% by setting `isolatedModules: true` in ts-jest config.

### DIFF
--- a/libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
+++ b/libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
@@ -17,7 +17,7 @@ export const selectFallbackFields = <T>(
     null
   )
 
-export const selectFallback = (field, fallback) =>
+export const selectFallback = <T, U>(field: T, fallback: U): T | U =>
   field === null ? fallback : field
 
 export const selectTranslatedField = <T>(
--- a/libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
+++ b/libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
@@ -1,0 +1,47 @@
+import { MetadataLink } from '@geonetwork-ui/util/shared'
+
+export type SourceWithUnknownProps = { [key: string]: unknown }
+
+export const selectField = <T>(
+  source: SourceWithUnknownProps,
+  fieldName: string
+): T | null =>
+  source !== null && fieldName in source ? (source[fieldName] as T) : null
+
+export const selectFallbackFields = <T>(
+  source: SourceWithUnknownProps,
+  ...fieldNames: string[]
+): T | null =>
+  fieldNames.reduce<T>(
+    (prev, curr) => (prev === null ? selectField(source, curr) : prev),
+    null
+  )
+
+export const selectFallback = (field, fallback) =>
+  field === null ? fallback : field
+
+export const selectTranslatedField = <T>(
+  source: SourceWithUnknownProps,
+  fieldName: string
+): T | null => selectField(selectField(source, fieldName), 'default')
+
+export const toDate = (field) => new Date(field)
+
+export const getFirstValue = (field) =>
+  Array.isArray(field) ? field[0] : field
+
+export const getAsArray = (field) =>
+  Array.isArray(field) ? field : field !== null ? [field] : []
+
+export const mapLinks = (
+  sourceLinks: SourceWithUnknownProps[],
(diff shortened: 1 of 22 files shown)


fgravin:
> This PR also reduces the setup time for tests by approx. 50% by setting `isolatedModules: true` in ts-jest config.

Sick, thanks god

libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
```diff
@@ -0,0 +1,47 @@
...
+export const selectFallbackFields = <T>(
+  source: SourceWithUnknownProps,
+  ...fieldNames: string[]
+): T | null =>
+  fieldNames.reduce<T>(
+    (prev, curr) => (prev === null ? selectField(source, curr) : prev),
+    null
+  )
+
+export const selectFallback = (field, fallback) =>
```
fgravin, line 20: missing types

libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
```diff
@@ -0,0 +1,47 @@
...
+  field === null ? fallback : field
+
+export const selectTranslatedField = <T>(
```
fgravin, line 23: What value does `selectTranslatedField(source, 'title')` bring more than  `source?.title` ? The `null` set ?
Brings complexity IMO but why not

libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
```diff
@@ -0,0 +1,47 @@
...
+  source: SourceWithUnknownProps,
+  ...fieldNames: string[]
+): T | null =>
+  fieldNames.reduce<T>(
+    (prev, curr) => (prev === null ? selectField(source, curr) : prev),
+    null
+  )
+
+export const selectFallback = (field, fallback) =>
+  field === null ? fallback : field
```
fgravin, line 21: And if it's a string `undefined` or `''` ? Default value is not applied.

fgravin (approved):
Thanks
To me it brings a little bit of complexity but overall it's cleaner and safer.
I hesitated in mutating or not the `output`, I didn't really saw the benefit of forbidding mutation in our case but you might be right.
You can go ahead.

libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
jahow: in these operations `null` is used when a key is missing; basically the idea is that you can chain operations like `getAsDate(selectFallback(selectField()))` and not risk a runtime error; `null` will just propagate through

libs/feature/search/src/lib/elasticsearch/mapper/atomic-operations.ts
jahow: For now, it only reads the `default` value, but I guess at some point we'll read the value from the current UI language

I agree that the added value isn't huge right now but I felt it was important to give some kind of structure to the mapping code, esp. since we explicitly support extending the mapping logic with custom fields

jahow:
> To me it brings a little bit of complexity but overall it's cleaner and safer.
I hesitated in mutating or not the output, I didn't really saw the benefit of forbidding mutation in our case but you might be right.

FYI when mutating the object (before this PR), mapping a large record took about 1-2ms. With this PR it takes about 8-10ms. There is a significant performance hit, so I hesitated a bit, but overall I'd rather have a mapping logic that is safe and will not crash at runtime. Also I anticipate that the mapping code will steadily grow in complexity over time, and that people might be interested in customizing it.

jahow:
Thanks for the review!

PR #148, opened 2021-08-31
Implement basic dataset visualization

jahow:
Currently rebased on #146 

This PR aims at providing foundations for smart components to be able, when given a metadata UUID, to fetch and render part of a dataset in different forms. The following steps are made to achieve this result:

* Introduces a `DatasetFinderService` used for determining the usage of each link in a metadata record (render on a map, download); this can be overridden with custom logic using Angular injection
* ~(WIP) Adds a `data-fetcher` lib in `libs/util` for downloading and parsing data in various formats, and apply expressions to filter/aggregate data~
  * ~This lib is **not** an angular library but in typescript only, also it is marked as buildable and publishable in Nx, meaning that it can eventually be published as a NPM package under the name `@geonetwork-ui/data-fetcher`~
  * ~Currently only reads a CSV file using a dedicated parser~

### DIFF
--- a/libs/feature/dataviz/src/lib/dataset-finder.service.spec.ts
+++ b/libs/feature/dataviz/src/lib/dataset-finder.service.spec.ts
@@ -1,0 +1,16 @@
+import { TestBed } from '@angular/core/testing'
+
+import { DatasetFinderService } from './dataset-finder.service'
+
+describe('DatasetFinderService', () => {
+  let service: DatasetFinderService
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({})
+    service = TestBed.inject(DatasetFinderService)
+  })
+
+  it('should be created', () => {
+    expect(service).toBeTruthy()
+  })
+})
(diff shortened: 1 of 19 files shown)


libs/feature/dataviz/src/lib/dataset-finder.service.spec.ts
```diff
@@ -1,16 +1,137 @@
...
+    }
+    describe('for download', () => {
+      it('returns the downloadable resources in priority order', () => {
+        expect(
+          service.getDistinctLinksByUsage(record, LinkUsage.DOWNLOAD)
+        ).toEqual([dataCsv, geodataWfs, geodataWfs2])
+      })
+      it('returns the resources renderable on a map in priority order', () => {
+        expect(service.getDistinctLinksByUsage(record, LinkUsage.MAP)).toEqual([
+          dataCsv,
```
fgravin, line 101: why csv in MAP, shouldn't we have WFS instead ?

fgravin (commented):
Great contribution, well tested !
The `data-fetcher` is great until we have the Data API I guess.
Thanks, maybe change the title of the PR cause it's not related ATM. All what you've done are needed for the full view in the datahub.

libs/feature/dataviz/src/lib/dataset-finder.service.spec.ts
jahow: CSV files could be used on a map if they have lon/lat columns, which we could check at some point using for example the `data-fetcher` util. But yeah this can be confusing (also the test label is wrong here, there is no sort in priority order)

PR #149, opened 2021-09-02
Datahub: use ENV to set API_BASE_PATH

fgravin:
Set up API base url from ENV var, to be customizable for production.

Follow the followings:
```
# Set environment variable
export GN4_API_URL="https://gn4.georchestra/...";
# Replace variables in env.js
envsubst < assets/env.template.js > assets/env.js
```

### DIFF
--- a/apps/datafeeder/README.md
+++ b/apps/datafeeder/README.md
@@ -1,0 +1,32 @@
+# Datafeeder
+UI of the web application `datafeeder`.
+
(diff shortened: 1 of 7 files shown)


jahow (approved):
looking good, thanks!
any idea how this could go in a lib eventually? so that all applications do not need to do the same trick everytime

fgravin:
> any idea how this could go in a lib eventually? so that all applications do not need to do the same trick everytime

hmm I can think about it, even though all applications might not have the same ENVs.

PR #150, opened 2021-09-03
Add a routing module in feature/search

jahow:
This PR adds an optional routing module in the `feature/search` lib. When imported in an app, it will handle synchronizing the search and metadata view state with the URL by dispatching actions.

Two routes are defined in this module:
* `/metadata/:uuid`: load and open a metadata record
* `/search` : currently does nothing except closing the opened metadata

The mdView module has been adapted to work with this. Also the Datahub app was modified to use this module as well. Note that the impact on the apps should be minimal, the idea is that the routing is completely optional and a full app can be built without it (e.g. in a web component).

~The routing module needs some documentation, will add it before merging.~ Added a bot of documentation for the `feature-search` lib and also routing in particular.


jahow:
@fgravin let me know what you think on this one, hopefully it makes sense for the changes to come

apps/datahub/src/app/app.component.ts
```diff
@@ -2,32 +2,34 @@ import { Component } from '@angular/core'
...
-        : of('Recherche')
-    )
+  breadcrumb$ = this.isPresentOrLoading$.pipe(
+    switchMap(([present, loading]) => {
+      if (present)
+        return this.mdViewFacade.metadata$.pipe(
+          map((md) => `Jeu de donnée > ${md.title}`)
+        )
+      if (loading) return of('Chargement...')
+      return of('Recherche')
```
fgravin, line 28: translation

libs/feature/search/src/lib/router/state/router.effects.ts
```diff
@@ -0,0 +1,70 @@
...
+        tap(({ path, query: queryParams }) => {
+          this._router.navigate([path], { queryParams })
+        })
+      ),
+    { dispatch: false }
+  )
+
+  loadMetadata$ = createEffect(() =>
+    this._actions$.pipe(
+      navigation(MetadataRouteComponent, {
```
fgravin, line 33: this comes from Nrwl, actually, the router state management took from carto-ui was hand made, maybe nrwl had more recent/clean implementation of it somewhere.

libs/feature/search/src/lib/router/constants.ts
```diff
@@ -0,0 +1,6 @@
+import { Component } from '@angular/core'
+
+export const ROUTER_STATE_KEY = 'router'
+
+export class SearchRouteComponent extends Component {}
+export class MetadataRouteComponent extends Component {}
```
fgravin, line 6: not very comfortable with this but it seems mandatory.

fgravin (approved):
I think it's cleaner this way. Thanks for the code, the tests and the doc
Great work

libs/feature/search/src/lib/router/state/router.effects.ts
jahow: yeah I figured I might as well not write a custom serializer and rely on nrwl helpers.

libs/feature/search/src/lib/router/constants.ts
jahow: I know, I feel like we're misusing the router somehow, but in our case I think using `router-outlet` isn't the best approach. And these components are harmless and not exported outside of the module.

apps/datahub/src/app/app.component.ts
jahow: I haven't registered any translation here for now since this is all considered placeholder, hope that makes sense...

jahow:
Thanks for the review!

PR #151, opened 2021-09-16
Export dataset

tkohr:
This PR adds components to download datasets in the different formats available.

TODOs:
- [x] Create UI components
- [ ] Adapt UI components to UI recommendations
- [ ] Handle events/logic of components
- [ ] Adapt mapping from elasticsearch
- [ ] Integrate components in datahub app

### DIFF
--- a/libs/util/shared/src/lib/models/search.model.ts
+++ b/libs/util/shared/src/lib/models/search.model.ts
@@ -33,6 +33,17 @@ export interface MetadataLink {
   url: string
 }
 
+// we will need to change model to contain format:
+// export type MetadataLinkFormats = 'OGC:WMS'|'OGC:WFS'|'OGC:WPS'|'ESRIFeature'|'csv'|'geosjon'|'json'|'shp'|'xls'|'xlsx'|'kml'|'gpkg'|'outbound'|'unknown'
+
+// export interface MetadataLink {
+//   format: MetadataLinkFormats
+//   resourceName: string // layerName WMS, featureType WFS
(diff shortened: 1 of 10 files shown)


libs/util/shared/src/lib/models/search.model.ts
```diff
@@ -33,6 +33,17 @@ export interface MetadataLink {
   url: string
 }
 
+// we will need to change model to contain format:
+// export type MetadataLinkFormats = 'OGC:WMS'|'OGC:WFS'|'OGC:WPS'|'ESRIFeature'|'csv'|'geosjon'|'json'|'shp'|'xls'|'xlsx'|'kml'|'gpkg'|'outbound'|'unknown'
+
+// export interface MetadataLink {
```
fgravin, line 39: why ?

jahow: this is something that we discussed, this model would be more appropriate for all kind of links, be it OGC services or simple files; the concept of "protocol" does not make sense for files, and the model currently does not contain a `title` (only a `name` which is more of a technical identifier)

fgravin:
@tkohr 
tips when you are contributing:
run the following commands to avoid useless CI runs:
`npm run format`
`npm run affected:lint`

you can run `npm run affected:test` sometimes but it might be slow

Be sure to configure prettier so it automatically formats your code on the fly (plugins exist for IDEs)

tkohr:
moved to #153

PR #153, opened 2021-09-17
Export dataset

tkohr:
This PR adds components to download datasets in the different formats available.

TODOs:
- [x] Create UI components
- [ ] Adapt UI components to UI recommendations
- [x] Handle events/logic of components
- [x] Adapt mapping from elasticsearch
- [x] Integrate components in datahub app
- [x] Focus on opendata catalogs
- [ ] Retrieve available formats from WFS getcapabilites (and ESRI REST API?) => will be done in a different PR
- [ ] Tests


libs/ui/layout/src/lib/download-entry/download-entry.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  templateUrl: './download-entry.component.html',
+  styleUrls: ['./download-entry.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DownloadEntryComponent implements OnInit {
+  @Input() format: string
+  @Input() resourceName: string
+  @Input() title: string
+  @Input() description: string
+  @Input() url: string
```
fgravin, line 20: missing the ouput on the click + action in the story (actions illustrates the outuput)

fgravin (commented):
Good strat.
I'm wondering about the name `Download` that is a bit confusing. I would talk more about `exports` `distributions` `resources`.
Note that the download is not the only action available in there, we could just copy paste a link to the resource as well for instance.

libs/ui/layout/src/lib/download-entry/download-entry.component.stories.ts
```diff
@@ -0,0 +1,38 @@
...
+  Story,
+  Meta,
+  componentWrapperDecorator,
+} from '@storybook/angular'
+import { DownloadEntryComponent } from './download-entry.component'
+import { UiLayoutModule } from '../ui-layout.module'
+import { BrowserAnimationsModule } from '@angular/platform-browser/animations'
+
+export default {
+  title: 'Layout/DownloadEntryComponent',
```
fgravin, line 12: I would start a section `Dataset`, to me putting everything in the `layout` one will be a mess in the end.

tkohr: Ok, thanks, do I understand correctly that you would create a `ui-dataset.module`?

fgravin: I would at least create a new section in the storybook tree.
```js
title: 'Dataset/DownloadEntryComponent
```
but actually i think it would be good to refactor a bit the libraries yes, the `layout` is a bit too general, also we need to think about what we do put in `feature-search`, is it a library for the search in itself (state, services, eventually smart component), or the logic for the search UI ?
I would have at least an Angular module for `mdView` and `dataset` which would contain the smart components.
@jahow any opinion ?

jahow: Yes, maybe we should clarify and agree on how smart and presentation modules are categorized.

I'll create a documentation PR with a proposition and we can keep talking about it there.

For now though, I think the `ui-layout` module is fine for these components.

jahow: doc PR here https://github.com/geonetwork/geonetwork-ui/pull/154

tkohr: The discussed refactoring has been in merged in #157

tkohr:
@fgravin @jahow a part from the retrieval of available WFS formats (which is currently mocked) and some more tests the PR is ready for review.

However, I'm still wondering if we should also query the ESRI REST API for formats ?

Regarding the format mapping in dcat2 plugin (https://github.com/camptocamp/geonetwork-dcat2/pull/2) I think it could be interesting to return "shp" instead of "zip" as "format" for shapefiles, if possible. This would allow to be more specific in the UI. And perhaps something shorther for "arcgis geoservices rest api".

PR can be tested by searching for g2f datasets from https://gn4.georchestra.org/geonetwork/srv/eng/catalog.search#/search?isTemplate=n&resourceTemporalDateRange=%7B%22range%22:%7B%22resourceTemporalDateRange%22:%7B%22gte%22:null,%22lte%22:null,%22relation%22:%22intersects%22%7D%7D%7D&sortBy=relevance&from=1&to=30&query_string=%7B%22sourceCatalogue%22:%7B%22972b12aa-08dc-4d29-8a10-ff44ecef1dda%22:true%7D,%22cl_hierarchyLevel.key%22:%7B%22dataset%22:true,%22series%22:true%7D%7D

jahow (approved):
Thanks! Two small comments overall. Note that this PR has probably too many commits for its own good (maybe the naming discussions are partially to blame though...), I think we can squash and merge it if you're OK.

libs/feature/search/src/lib/metadata-view/data-exports/data-exports.component.ts
```diff
@@ -7,6 +15,48 @@ import { MdViewFacade } from '../state'
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class DataExportsComponent {
+export class DataExportsComponent implements OnInit {
+  links$: Observable<Array<MetadataLink>>
+
   constructor(public facade: MdViewFacade) {}
+
+  ngOnInit() {
+    this.links$ = this.facade.downloadLinks$.pipe(
```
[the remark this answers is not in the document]
tkohr: ah yes, was wondering about this one

PR #154, opened 2021-09-20
Improve README and CONTRIBUTING, describe how libraries are organized

jahow:
This PR follows [this comment](https://github.com/geonetwork/geonetwork-ui/pull/153#discussion_r711940174) and attempts to clarify how libraries are organized in the project.

Please note that the current organization of libs in the project does not match this description, and things will have to be moved around once agreed on.

The gist of it is:

1. Presentation libraries are in the `ui` folder and are categorized by their forms:
   * `ui-inputs` for reusable components made to collect input from the user (e.g.: form fields, buttons...)
   * **`ui-elements` for components focused on rendering specific types of information in an elaborate way, which may or may not be related to business usages; (THIS DOES NOT EXIST YET)**
     examples include download links, facet or selection tree, etc.
   * `ui-layout` for components which occupy a large part of the screen and might contain variable content or other components
   * `ui-map` for map-specific components (map container, controls, etc.)
   * `ui-widgets` for reusable, small, self-contained components which show information in a visual way, similar to icons but more elaborate (e.g.: icon with tooltip, status indicator, progress bar...)

**Note: `ui-catalog` and `ui-search` WILL BE ABSORBED BY OTHER LIBS IN THIS SCENARIO**

2. Libraries providing business or data logic and state management are in the `feature` folder and are categorized by their intended use: 
   * `feature-auth` for logic and components related to authentication
   * `feature-catalog` for logic and components related to general catalog topics (title, logo, etc.)
   * `feature-dataviz` for logic and components related to data visualization
   * `feature-editor` for logic and components related to editing metadata
   * `feature-map` for logic and components related to interactive maps
   * `feature-search` for logic and components related to searching through the catalog

3. Libraries used for interacting with backend services are in the `data-access` folder

4. Libraries providing common services or shared models are in the `util` folder

@fgravin @tkohr does that make sense to you?


fgravin:
Thanks @jahow for the proposal.
I think `ui-component` could be a good idea, but almost all components will land into that. I like to have a business entry but i can't tell what is the best (`ui-dataviz, ui-search, ui-dataset`), it might help to quickly find what you look for.

For the feature, I wonder if `mdView` and `dataset` wouldn't have a better place than `feature-search` lib. To me, it's not really a search.

My concerns is that we don't easily find a component cause we don't know where it is, it was the case in previous project and it was very boring. 
In Nexsis, there are 3 level of granularity for the components, in our case, we are not talking about a granularity but almost just about context.

Anyway, I think generic component like inputs or widgets should be common, but the `downloadLink` could be in a business directory. We could at least add folders in the `ui-components` lib to separate concerns, as well as angular modules if needed.

But i don't really know what is the best i have to admit.

jahow:
> For the feature, I wonder if mdView and dataset wouldn't have a better place than feature-search lib. To me, it's not really a search.

Maybe `feature-metadata`?

> My concerns is that we don't easily find a component cause we don't know where it is, it was the case in previous project and it was very boring.

I agree, that's why it would be useful to have a clear logic for how libraries are organized. Then, if I'm looking for a component, I can ask myself a series of questions to find it:
* does it handle data? (yes => smart, no => presentation)
* does it collect user input? (yes: `ui-inputs`)
* is it used to contain varying content ? (yes: `ui-layout`)

And then if everything else fails the component is probably in `ui-elements`.

> We could at least add folders in the ui-components lib to separate concerns, as well as angular modules if needed.

Yes, I think this is the right approach. Many components might end up in `ui-elements` (not `ui-components` btw, not sure if you noticed), but they could end up in directories such as `metadata-record` (records preview & list).

jahow:
I'm going with `feature-record` for the smart components used in the record view (metadata, links, data and map preview, exports, APIs...).

PR #155, opened 2021-09-20
metadataview: add tabs to navigate to sections

fgravin:
Basic tab integration in metadataview panel
- info
- data
- map
- exports
- api

Screenshot from 2021-09-20 17-40-35


jahow (approved):
Looking good! Material provides a powerful and easy to use tab component so I guess we don't need to redefine one in the `ui-layout` module.

PR #156, opened 2021-09-21
Datahub: add a map to preview data on dataset view

jahow:
This PR introduces a new smart component calls `DataPreviewMapComponent`, which is bound to the `mdView` feature state. When a metadata is opened, the component will look for the first available link that can be rendered on a map and show it, alongside a dropdown to switch to other links.

Remaining to do:
- [ ] handle translation for placeholder value
- [x] hide map if no valid layer can be rendered in it

image


apps/datahub/src/app/dataset-view/dataset-view.component.html
```diff
@@ -1 +1,4 @@
-<p>dataset-view works!</p>
+<div class="float-right w-96 h-96 m-3 border rounded overflow-hidden">
+  <gn-ui-data-preview-map></gn-ui-data-preview-map>
+</div>
+<gn-ui-metadata-full-view></gn-ui-metadata-full-view>
```
fgravin, line 4: `metadata-full-view` was the larger container, and NOT the information tab or whereever you put the abstract and the preview map.
We must define a clear treeview, let's discuss about that.
- metadata(dataset)-view (the whole view)
  - summary
    - metadata-info
    - dataset-preview
  - data table
  - map
  - exports
  - apis

or something like this.
In my PR, i put the tabs in `ui-metadata-full-view`

apps/datahub/src/app/app.component.html
```diff
@@ -33,8 +33,6 @@ <h2 class="text-white text-4xl uppercase">Data Hub</h2>
     <gn-ui-main-search
       *ngIf="(isRecordOpened$ | async) === false"
     ></gn-ui-main-search>
-    <gn-ui-metadata-full-view
-      *ngIf="isRecordOpened$ | async"
-    ></gn-ui-metadata-full-view>
+    <gn-ui-dataset-view *ngIf="isRecordOpened$ | async"> </gn-ui-dataset-view>
```
fgravin, line 36: I think this needs some refactoring.
`dataset-view` is the larger container for the metadata view ?

libs/feature/search/src/lib/metadata-view/data-preview-map/data-preview-map.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    ),
+    map((links) =>
+      links.filter(
+        (link) =>
+          this.datasetFinder.getLinkUsages(link).indexOf(LinkUsage.MAP) > -1
+      )
+    ),
+    startWith([])
+  )
+  dropdownChoices$ = this.availableLinks$.pipe(
```
fgravin, line 34: this list is not customizable.
`availableLinks` should be in a service/class which could be extended ?
Not mandatory for now.

fgravin (changes requested):
Very good and excellent testing.
I think it can be improved though with some helpers, and maybe by centralizing the links computation, and make it configurable.
Also the component tree structure should be discussed.
Thanks

fgravin:
Screenshot from 2021-09-21 11-25-40

- does not display WFS data.
- overlap with the link boxes (does not really matter cause anyway we will rework the whole styling after UX, hopefully).

jahow:
> does not display WFS data.

I know, I think implementing data fetching should be part of another PR (besides this was not requested in the user story).

> overlap with the link boxes (does not really matter cause anyway we will rework the whole styling after UX, hopefully).

Exactly. The float thing was just eye candy but really the whole layout is completely provisional.

fgravin:
> I know, I think implementing data fetching should be part of another PR (besides this was not requested in the user story).
> 
> >

yes I agree, I think the logic in this PR is good, could be slighlty improved though. But the whole data fetching stuff are a bit out of scope.

jahow:
I've rebased this PR but hasn't changed much, waiting on #157 (no pressure @tkohr ;D) to go forward

apps/datahub/src/app/dataset-view/record-view.component.ts
```diff
@@ -8,9 +9,14 @@ import { MdViewFacade } from '@geonetwork-ui/feature/search'
...
-  constructor(public facade: MdViewFacade) {}
+  constructor(
+    public facade: MdViewFacade,
+    private mapManager: MapManagerService
+  ) {}
 
   onTabIndexChange(index: number): void {
-    console.log(index)
+    if (index === 2) {
+      setTimeout(() => this.mapManager.map.updateSize(), 0)
```
jahow, line 24: this brings back memories... Ideally the map presentation component should be able to do that itself by using e.g. [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

apps/datahub/src/app/dataset-view/record-view.component.ts
```diff
@@ -9,6 +10,10 @@ import { MdViewFacade } from '@geonetwork-ui/feature/search'
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RecordViewComponent {
+  displayMap$ = this.facade.mapApiLinks$.pipe(
+    map((links) => !!links && links.length > 0)
```
jahow, line 14: this observable should not emit anything other than arrays... this may be why also there are sometimes errors when opening a record

jahow: but this can be addressed in another PR I guess

jahow (commented):
Looking good, thanks for picking this up! I think we can merge this and move forward with WFS and geojson later on

PR #157, opened 2021-09-22
Record View refactoring

fgravin:
Starting with implementing the `data-table` smart component of the `metadata-view` module, the PR is more about the module refactoring.

What has been decided
- MetadataView module should provide smart component for section in the full record views
  - metadatas (or information)
  - data-preview
  - data-table
  - data-exports
  - data-api
- The record-view container (containing tabs) is part of the application
- `metadata-full-view` component is useless
- `metadata-page` name is not meaningfull and should be part of ui-components future library

The work is in progress
To do in this PR or in futures:

- [x] create `ui-elements` library
- [x] move `metadata-page` to `metadata-info` in `ui-elements`
- [ ] rename Metadata to Record in the whole project (may not be part of the PR)
- [x] refactor ui components in meaningfull libraries
- [x] better link classification
- [ ] fetch the dataset and pass it to the table


jahow:
> MetadataView module should provide smart component for section in the full record views
>   * data-preview
>   * data-table

What is the difference between `data-preview` and `data-table`? Why not `data-preview-map` and `data-preview-table` for example?

> metadata-page name is not meaningfull and should be part of ui-components future library

Agreed, a component like `gn-ui-record-metadata` would make sense IMO (just like we have record-preview etc.)

> create ui-component library

@fgravin I was actually suggesting `ui-elements` for the new ui lib, not `ui-component`, since that sounds too generic to me :)

Other than that, I agree with the approach. I think this is going in the right direction, thanks!

libs/feature/search/src/lib/metadata-view/links/link-helper.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  isMapLink(link: MetadataLink): boolean {
+    return this.datasetFinder.getLinkUsages(link).indexOf(LinkUsage.MAP) > -1
+  }
+  isDownloadLink(link: MetadataLink): boolean {
+    return (
+      this.datasetFinder.getLinkUsages(link).indexOf(LinkUsage.DOWNLOAD) > -1
+    )
+  }
+}
```
jahow, line 31: To me this could almost be simple function exports without an injectable service. Would make usage less verbose! Just a suggestion.

fgravin:
> What is the difference between `data-preview` and `data-table`? Why not `data-preview-map` and `data-preview-table` for example?

ok sure, I thought `preview` was about the map from your PR. `data-preview-map` and `data-preview-table` seem good, but what is preview actually, compared to a view ? :)

> @fgravin I was actually suggesting `ui-elements` for the new ui lib, not `ui-component`, since that sounds too generic to me :)

Ok sorry, go for ui-elements ;)

fgravin:
@jahow note that with the Classifier, the `DatasetFinderService` would only be used to attach a Usage to a link, which is the role of a classifier... BTW is it finding Dataset ATM ? I don't think so
I think what you have done in this service is good though, but a method like `getDistinctLinksByUsage` may be useless as all kind of links and usages will be served as observables from the service ... the point is that the observables are always related to the state metadata$, we there is not way to just retrieve all links by usage from a record, which might be usefull at some point !

So maybe, we should keep such a method, call it in the `metadata$.pipe`, use a `shareReplay` and then pass the filtered links to each observable

libs/feature/search/src/lib/metadata-view/links/link-helper.service.ts
fgravin: I wanted to to this but as I need the `datasetFinder` which is a service, so i need to create a service to inject the `datasetFinder`

jahow: Ah, true, too bad then :)

libs/feature/search/src/lib/metadata-view/links/link-classifier.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    map((links) => links.filter((link) => this.helper.isDataLink(link)))
+  )
+  otherLinks$ = this.allLinks$.pipe(
+    map((links) => links.filter((link) => this.helper.isOtherLink(link)))
+  )
+  validMapLinks$ = this.allLinks$.pipe(
+    map((links) => links.filter((link) => this.helper.isValidLink(link))),
+    map((links) => links.filter((link) => this.helper.isMapLink(link))),
+    startWith([])
+  )
```
jahow, line 23: I realized that we now have two more services: LinkClassifier and LinkHelper. How about simply putting these observables in the facade? that would make sense IMO

jahow:
> but what is preview actually, compared to a view ? :)

I guess the idea is that these components will show you only part of the data (either a rendered image in a map or 20 results out of 1000 in a table for example). But map-view and table-view would make sense as well, maybe even more actually.

> @jahow note that with the Classifier, the DatasetFinderService would only be used to attach a Usage to a link, which is the role of a classifier... BTW is it finding Dataset ATM ? I don't think so
I think what you have done in this service is good though, but a method like getDistinctLinksByUsage may be useless as all kind of links and usages will be served as observables from the service ... the point is that the observables are always related to the state metadata$, so there is no way to just retrieve all links by usage from a record, which might be usefull at some point !
> So maybe, we should keep such a method, call it in the metadata$.pipe, use a shareReplay and then pass the filtered links to each observable

So I agree that the DatasetFinderService has to evolve because its name and structure are not relevant. To me the LinkClassifier should be enough and could replace it. Maybe the following logic could work out?
1. LinkClassifier (maybe `RecordLinkClassifier`?) contains logic for deciding which links are suitable for which usages; so basically `getUsagesForLink` and not much else; this replaces the dataset finder service altogether
2. MdView Facade uses LinkClassifier (using helpers for example, but I'm not even sure they're necessary) to provide observables with links by usage: `mapApiLinks$`, `downloadLinks$`, `otherLinks$` (no usage) etc.

Does that make sense @fgravin ?

fgravin:
> 2\. MdView Facade uses LinkClassifier (using helpers for example, but I'm not even sure they're necessary) to provide observables with links by usage: `mapApiLinks$`, `downloadLinks$`, `otherLinks$` (no usage) etc.

I agree.
So we should now decide what `LinkUsage` we'll have, and what observables we want for the links in the facade, and see if they are the same ...
pnig @tkohr to follow the discussion and give your idea if needed.

fgravin:
I wouldn't do to much in this PR as well.
We should move forward by pushing this one (after suggested changes) and pushing @jahow's one.
And do another one to follow the refactoring afterwards

tkohr:
> I guess the idea is that these components will show you only part of the data (either a rendered image in a map or 20 results out of 1000 in a table for example). But map-view and table-view would make sense as well, maybe even more actually.

@jahow just to stay coherent with the naming of the other views I would call them `data-map` and `data-table` if you are ok with this. In the end all components in `metadata-view` are views.

Like, MetadataView:
- metadatas (or information)
- data-map
- data-table
- data-exports
- data-api

jahow:
> > I guess the idea is that these components will show you only part of the data (either a rendered image in a map or 20 results out of 1000 in a table for example). But map-view and table-view would make sense as well, maybe even more actually.
> 
> @jahow just to stay coherent with the naming of the other views I would call them `data-map` and `data-table` if you are ok with this. In the end all components in `metadata-view` are views.

Hmm, I think `data-map` is a bit too misleading... These components will provide a view on the dataset contents so IMO `data-view-map` and `data-view-table` are more description.

> metadatas

"Metadata" is already plural :) you can probably go for `record-metadata` and anticipate on when we'll have to rename "metadata" to "record".

> data-exports
> data-api

Same as previously, these are not descriptive enough IMO. I'd rather have `record-exports` and `record-api-endpoints` for example.

What do you think @tkohr? Thanks

PS: we agree that all of these are smart components, right?

tkohr:
> Hmm, I think `data-map` is a bit too misleading... These components will provide a view on the dataset contents so IMO `data-view-map` and `data-view-table` are more description.

ok, I agree

> "Metadata" is already plural :) you can probably go for `record-metadata` and anticipate on when we'll have to rename "metadata" to "record".

that's true

> > data-exports
> > data-api
> 
> Same as previously, these are not descriptive enough IMO. I'd rather have `record-exports` and `record-api-endpoints` for example.

I think they are rather exports and APIs to the data than to the record, but I don't have a strong preference for the naming. I would perhaps facilitate/shorten `record-api-endpoints` to `record-apis`. What do you think @jahow ?
 
> PS: we agree that all of these are smart components, right?

yes

jahow:
> I think they are rather exports and APIs to the data than to the record, but I don't have a strong preference for the naming. I would perhaps facilitate/shorten record-api-endpoints to record-apis. What do you think @jahow ?

Agreed! sounds nice and much shorter

tkohr:
@jahow @fgravin I hope I covered most of the refactoring we discussed and that it goes into the direction you had in mind. The structure of the refactoring should now be laid out.

The logic of the link classification is kept very simple for now and might evolve further, depending on the `protocol`s and `format`s we target.

The DatasetFinderService and its prioritization are not used anymore. We might delete it before merging.

I haven't been able to adapt all the tests, like 
- checking the link observables in the facade.
- checking the inputs in record-view component (which will still change component anyway)

Just noticed that the LinkClassifierService and helper could be simplified into one, maybe just using simple functions. Not sure what's the most convenient separation here.

Don't hesitate to leave your comments.

libs/feature/search/src/lib/metadata-view/metadata-view.module.ts
```diff
@@ -6,17 +6,33 @@ import { StoreModule } from '@ngrx/store'
...
+import { RecordExportsComponent } from './record-exports/record-exports.component'
+import { RecordApisComponent } from './record-apis/record-apis.component'
 
 @NgModule({
-  declarations: [DataTableComponent],
+  declarations: [
+    RecordMetadataComponent,
+    DataViewMapComponent,
+    DataViewTableComponent,
+    RecordExportsComponent,
```
fgravin, line 20: I think exports and apis are related to a dataset, not a record

fgravin (commented):
Thanks @tkohr for the follow up on the refactoring regarding what we've discussed.
I think it is in a good shape right now and we should merge it so we can implement the smart components independantly.
@jahow do you agree at this stage ?
Thanks

libs/feature/search/src/lib/metadata-view/metadata-view.module.ts
tkohr: I agree, would you be ok with `data-exports` and `data-apis` @fgravin, @jahow ?

jahow: well I think you went with these names [because of me](https://github.com/geonetwork/geonetwork-ui/pull/157#issuecomment-926733874) so... I guess whichever is fine :)

jahow (approved):
looking good, thanks for the heavy work! if you can remove the DatasetFinderService as well that would be great, thanks

tkohr:
thanks for reviewing @fgravin @jahow

PR #162, opened 2021-10-20
Expand map preview to work with WFS and GeoJSON files

jahow:
This PR expands on the data-view-map component to make it work with WFS layers as well as geojson files provided in the metadata record.

* `data-fetcher` is used to download all data (either from files or WFS) as geojson
* a loading mask component was added to handle the time where the dataset is being downloaded
* use ogc-client 0.2.2 to resolve an issue with output formats not parsed in the WFS

Remaining:
- [x] show errors/warnings when something goes wrong (CORS, parse error etc.)

image
popup-alert

### DIFF
--- a/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
+++ b/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
@@ -45,6 +45,7 @@ export class DataViewMapComponent {
   selectedLinkIndex$ = new BehaviorSubject(0)
 
   loading = false
+  error = null
 
   currentLayers$ = combineLatest([
     this.compatibleMapLinks$,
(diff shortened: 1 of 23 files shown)


libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
```diff
@@ -59,10 +84,56 @@ export class DataViewMapComponent {
...
+      return fromPromise(
+        new WfsEndpoint(link.url).isReady().then((endpoint) => {
+          if (
+            !endpoint
+              .getServiceInfo()
+              .outputFormats.some(
+                (format) => format.toLowerCase().indexOf('json') > -1
+              )
+          ) {
+            throw new Error('map.wfs.geojson.not.supported')
```
tkohr, line 108: :+1: we could generalize this to i18n error messages, if I see right

PR #168, opened 2021-10-29
Datahub / navigation and user experience improvements

jahow:
* Search view:
  * Adds a bit more logic in the search router: the router will close the current metadata if new search filters are sent (e.g. when pressing <kbd>enter</kbd> on the search input)
  * The datahub app will trigger a search with no filters at initialization
* Record view :
  * disable download and api tabs if no link available
  * fix a bug where links with a protocol such as `OGC:WMS-http-getmap-etc` would not render on the map
  * improve layout of the tabs content
  * added keywords, lineage, usage constraints, contact info, other links
Misc:
* use ogc-client v0.3.0 which have better cache! (not limited in size anymore)
* better style for the record count under the table

image

### DIFF
--- a/apps/datahub/src/app/app.component.html
+++ b/apps/datahub/src/app/app.component.html
@@ -31,7 +31,7 @@
   </div>
   <div class="container-lg mx-auto py-4">
     <gn-ui-main-search
-      *ngIf="(isRecordOpened$ | async) === false"
+      [ngClass]="{ hidden: (isRecordOpened$ | async) }"
     ></gn-ui-main-search>
     <gn-ui-record-metadata *ngIf="isRecordOpened$ | async">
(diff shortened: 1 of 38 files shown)


tkohr (approved):
Nice work @jahow. Thanks for the great effort! I did not get all the details of the es mapping, but otherwise the code looks good to me. Tested and works well. Maybe the logo could be replaced with one of a better resolution or a placeholder (for now).

jahow:
> Maybe the logo could be replaced with one of a better resolution or a placeholder (for now).

Which logo? the one displayed under "contact" is taken from the metadata record. We're supposed to show the producer logo but I think that the records only contain the source catalogue logo, so it doesn't really match. Not sure we can get a better result for now.

Thanks for the review :)

PR #169, opened 2021-10-29
Zoom to layer

tkohr:
This PR zooms the map preview to the selected displayed data.

TODOs:

- [x] implement logic to determine extent of data (WFS, WMS, files)
- [x] fix problem with `WmsEndpoint`
- [x] make existing tests pass


libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -104,4 +109,34 @@ export class MapUtilsService {
...
+      return from(
+        new WmsEndpoint(layer.url).isReady().then((endpoint) => {
+          const wmsLayer = endpoint.getLayerByName(layer.name)
+          return this.extentFromLonLat(wmsLayer.boundingBoxes['EPSG:4326'])
+        })
+      ) as Observable<Extent>
+    } else {
+      return of(undefined)
+    }
+    return of(extent)
```
jahow, line 133: This is not needed (will never be executed)

libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -104,4 +109,34 @@ export class MapUtilsService {
       return []
     }
   }
+
+  getLayerExtent(layer: MapContextLayerModel): Observable<Extent> {
+    let extent = []
+    if (layer.type === 'geojson' || layer.type === 'wfs') {
```
jahow, line 115: WFS layers do not have the data inline, they rely on a url like WMS; for the datahub we don't use WFS layers, only geosjon and WMS

libs/feature/map/src/lib/map-context/map-context.model.ts
```diff
@@ -23,3 +24,4 @@ export interface MapContextViewModel {
   center: [number, number]
   zoom: number
 }
+export type MapContextExtentModel = [number, number, number, number]
```
jahow, line 27: I think we should document that this is expressed in the map view projection, so web mercator for now

jahow (commented):
Some comments!

libs/feature/map/src/lib/utils/map-utils.service.ts
tkohr: ok, shouldn't we keep a fall back for the case once we might have other types than `wms` and `geojson`?

libs/feature/map/src/lib/utils/map-utils.service.ts
tkohr: ok, got it

libs/feature/map/src/lib/map-context/map-context.model.ts
tkohr: yes I agree, where should I do so, here in a comment or in a readme?

jahow: a comment is enough for now I think

libs/feature/map/src/lib/utils/map-utils.service.ts
jahow: we definitely need a fallback :) it's good now, thanks

jahow (approved):
Looking good, thanks a lot!

libs/feature/map/src/lib/map-context/map-context.model.ts
tkohr: ok, I added that

PR #170, opened 2021-11-03
Add translations for the datahub app, rework i18n workflow

jahow:
The i18n workflow should be simpler now:
* all translation files are in the `translations` folder at the root of the project
* all apps can use these files as assets instead of having translations files inside the app-specific assets folder (see datahub config in `angular.json`)
* added a cli tool to merge translation files (it was necessary to merge the datafeeder translations with the search ones and probably also the transifex ones)
* improved the `i18n:extract` task to read all apps and libs

The datahub app now relies on the translation files only (not on GN4 api anymore). I also added a custom translation loader to make sure that keys translated by a blank string (which is the default with `i18n:extract`) will fallback to the default language, which is supposedly completely translated.

Support for custom translations in an override file will be done in another PR I guess, since we'll be a bit short on time for this week...

Rely on #168


libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -5,7 +5,7 @@
   <mat-tab>
     <ng-template mat-tab-label>
       <mat-icon class="mr-3">info</mat-icon>
-      <span translate="">mdview.tab.info</span>
+      <span translate="">record.tab.info</span>
```
tkohr, line 8: I guess we can omit the empty quotes.

tkohr (commented):
Thanks @jahow. I might have not gotten all the details, but overall makes sense to me.

Worked fine in firefox in French, just noticed a strange behavior in chromium in French: App was displayed in English and once I had executed the extract script, I got 404 errors for the French translations.

PR #171, opened 2021-11-04
Datahub / Inform user when download links could not be fetched from a WFS service

jahow:
image


tkohr (approved):
Thanks @jahow, good catch! Looks good and works fine :+1:

PR #173, opened 2021-11-05
Fix sort for component init when inside ngIf

tkohr:
PR fixes that mat-sort currently does not work directly after the table component is rendered on an `ngIf` condition.


jahow (approved):
Thanks!

PR #177, opened 2021-11-09
Results list layout: dynamic component instanciation

fgravin:
Refactor the way record previews are rendered in the search result list. It was static, with a big switch in the template.
The PR introduces a dynamic component instanciation, based on a configuration which is injectable (customizable).
```js
{
  LIST: new ResultsLayoutConfigItem(
    RecordPreviewListComponent,
    'h-24',
    'grid md:grid-cols-2 lg:grid-cols-4 gap-4 p-4'
  ),
  TEXT: new ResultsLayoutConfigItem(
    RecordPreviewTextComponent
  )
}
```

Inject your config in your application module to override the default one
```js
  providers: [
    { provide: RESULTS_LAYOUT_CONFIG, useValue: DEFAULT_RESULTS_LAYOUT_CONFIG },
  ],
```

The layout switcher now uses this config to list the available layouts.
You can create your own layout by adding a new component and injecting a new configuration.

The system is used to customize the datahub `RecordPreview`
Screenshot from 2021-11-09 18-30-58


apps/datahub/src/assets/i18n/en.json
```diff
@@ -0,0 +1,47 @@
...
+  "results.layout.selectOne": "",
+  "results.records.hits.found": "",
+  "results.records.hits.found=0.help": "",
+  "results.sortBy.dateStamp": "",
+  "results.sortBy.popularity": "",
+  "results.sortBy.relevancy": "",
+  "search.field.any.placeholder": "",
+  "search.field.sortBy": "",
+  "search.loading": ""
+}
```
jahow, line 47: These files are not needed any more; to extract keys, run `npm run i18n:extract`, the keys will be added to the files in the `translations` directory

apps/datahub/src/assets/i18n/fr.json
```diff
@@ -0,0 +1,47 @@
...
+  "results.layout.selectOne": "",
+  "results.records.hits.found": "",
+  "results.records.hits.found=0.help": "",
+  "results.sortBy.dateStamp": "",
+  "results.sortBy.popularity": "",
+  "results.sortBy.relevancy": "",
+  "search.field.any.placeholder": "",
+  "search.field.sortBy": "",
+  "search.loading": ""
+}
```
jahow, line 47: Same as `en.json`

apps/datahub/src/app/record-preview-datahub/record-preview-datahub.component.css
```diff
@@ -0,0 +1,10 @@
+.record-item {
+  height: 10.5em;
+  cursor: pointer;
+}
+.publisher {
+  width: 15rem;
```
jahow, line 6: Not tested but on the screenshot the width seems to vary from one record to the next

libs/util/shared/src/lib/models/search.model.ts
```diff
@@ -54,11 +54,13 @@ export interface RecordMetric {
...
 
-export enum ResultsListLayout {
+export enum ResultsListLayoutEnum {
   CARD = 'CARD',
   LIST = 'LIST',
   TEXT = 'TEXT',
   TITLE = 'TITLE',
 }
 
+export type ResultsListLayout = ResultsListLayoutEnum | string
```
jahow, line 64: This feels weird. Why keep the enum there since the layout configuration is now in `ui/search` and can be overridden in apps?

I think in the state the result layout should just be a string, and the enum should be removed altogether. Then we can handle an incorrect layout value to avoid any unintentional misconfiguration.

libs/util/shared/src/lib/utils/strip-html.ts
```diff
@@ -0,0 +1,6 @@
+export const stripHtml = function (html: string): string {
+  if (!html) {
+    return ''
+  }
+  return html.replace(/<[^>]*>/g, '')
```
jahow, line 5: I think we should rely on the `DOMParser` API instead, see: https://stackoverflow.com/questions/822452/strip-html-from-text-javascript/47140708#47140708
This should be more robust than a regex

libs/util/shared/src/lib/utils/strip-html.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+      const html = '<b>hello</b>'
+      expect(stripHtml(html)).toBe('hello')
+    })
+  })
+  describe('when no HTML tags', () => {
+    it('return same string', () => {
+      const html = 'hello'
+      expect(stripHtml(html)).toBe('hello')
+    })
+  })
```
jahow, line 15: It'd be interesting to add a test to make sure that any js in the html is _not_ executed, something like a console log (but this is just nice to have)

libs/ui/search/src/lib/results-list/results-layout.config.ts
```diff
@@ -0,0 +1,39 @@
...
+import { RecordPreviewTitleComponent } from '../record-preview-title/record-preview-title.component'
+import { RecordPreviewComponent } from '../record-preview/record-preview.component'
+
+const DEFAULT_ITEM_CLS = 'pb-4'
+const DEFAULT_CONTAINER_CLS = 'gap-4 p-4'
+
+export class ResultsLayoutConfigItem {
+  constructor(
+    public component: Type<RecordPreviewComponent>,
+    public itemClass: string = DEFAULT_ITEM_CLS,
```
jahow, line 15: Having the possibility to specify a css style here would be nice as well, e.g. for having a precise height (this would remove the need for defining custom css classes)

jahow (changes requested):
Thanks for the solid refactoring; the result layout is much more flexible this way. This is actually a great way to allow custom result layouts! I've made a few comments for changes, especially the `ResultsListLayoutEnum` which now feels really out of place.

apps/datahub/src/app/record-preview-datahub/record-preview-datahub.component.css
fgravin: yes it is because of the image size and shape... never really understood what is the best way to render thumnails so they always have the same size.. Using `width` and `height` attributes of the `img` tag ?

jahow: There is a `gn-ui-record-thumbnail` component, see: https://github.com/geonetwork/geonetwork-ui/blob/master/libs/ui/search/src/lib/record-preview-card/record-preview-card.component.html

Maybe you'll have more luck with it :)

fgravin:
Thanks for the review @jahow, check the last commits i have addressed all your comments.
I improved slightly the datahub preview list as well.

Screenshot from 2021-11-10 10-53-15

jahow (approved):
All good, thanks!

PR #178, opened 2021-11-09
Display ESRI:REST WFS links in download component

tkohr:
This PR displays ESRI:REST WFS links in the download component. Unfortunately, the metadata link names of the tested ESRI WFS servers did not match the layer/featuretype names, which does not allow inferring the available formats. An according error message is displayed.

### DIFF
--- a/libs/feature/search/src/lib/utils/links/link-utils.spec.ts
+++ b/libs/feature/search/src/lib/utils/links/link-utils.spec.ts
@@ -3,8 +3,28 @@ import {
   DownloadFormatType,
   checkFileFormat,
   getLinksWithEsriRestFormats,
+  getLinksWithWfsFormats,
 } from './link-utils'
 import { LINK_FIXTURES } from './link.fixtures'
+
+jest.mock('@camptocamp/ogc-client', () => ({
+  WfsEndpoint: class {
+    constructor(private url) {}
+    isReady() {
+      return Promise.resolve({
+        getFeatureUrl: () => this.url + '?GetFeature',
+        getFeatureTypeSummary: (name) => {
+          return name !== 'NOT_A_LAYER_NAME'
+            ? {
+                name,
+                outputFormats: ['geojson', 'csv'],
+              }
+            : null
+        },
+      })
+    }
+  },
+}))
 
 describe('link utils', () => {
   describe('#getDownloadFormat', () => {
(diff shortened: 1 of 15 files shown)


libs/feature/record/src/lib/links/link-utils.ts
```diff
@@ -47,7 +47,12 @@ export function getLinksWithWfsFormats(
   link: MetadataLinkValid
 ): Promise<MetadataLinkValid[]> {
   return new WfsEndpoint(link.url).isReady().then((endpoint) => {
-    const featureType = endpoint.getFeatureTypeSummary(link.name)
+    let featureType = endpoint.getFeatureTypeSummary(link.name)
+    if (!featureType) {
+      //fallback: if link name does not match layer name, use first layer
+      featureType = endpoint.getFeatureTypes()[0]
+      featureType.outputFormats = endpoint.getServiceInfo().outputFormats
```
jahow, line 54: Why is this needed? this should be the responsibility of ogc-client I think

jahow (commented):
Nice, thanks! The changes make sense overall. I just think the fallback logic is a bit weird though. Why exactly is there no matching layer name in the WFS endpoint? Could you please give a concrete example? Thanks!

libs/feature/record/src/lib/links/link-utils.ts
tkohr: You are right that this should perhaps rather be handled by ogc-client. One example is https://dservices8.arcgis.com/rxZzohbySMKHTNcy/arcgis/services/ind_hdf/WFSServer?SERVICE=WFS&REQUEST=GetCapabilities (which you find by searching for "ind hdf" on datahub). Here the `link.name` in the metadata is `ArcGIS GeoService` whereas the `FeatureType` name is `ind_hdf:ind_hdf_2021`. Hum, just noticing, that this PR does not work on the gh-pages deployment :-(

jahow: Ok so the feature type seems to be the hard to guess based on the record metadata so this probably cannot be handled consistently in the JSON harvester. On the other hand I'm not a big fan of a catch-all fallback logic like this because it produces inconsistent behaviour. For example if a record points to a dataset in a WFS endpoint but for whatever reason this dataset disappeared and there's only one other completely unrelated dataset left, then we'll show this unrelated dataset and that would not make any sense.

I think what we should do is if the feature type name does not match then too bad, throw an error and move on. We can improve the JSON harvester to "guess" the feature type name which apparently would work in many cases, or we could also assume that if a `WFSServer` is available then a `FeatureServer` will be as well (to be discussed with the platform maintainers maybe?).

What do you think?

jahow: > Hum, just noticing, that this PR does not work on the gh-pages deployment :-(

No worries, that's because the deployed app points to gn4.georchestra.org which does not have up to date harvesters for ESRI.

tkohr: Yes, I agree, that this fallback might lead to unexpected behavior as you described. Let's go for a clear error handling, then!

Looking at `https://data-atmo-hdf.opendata.arcgis.com/data.json` we cannot assume that data with a `WFSServer` is often also available via a `FeatureServer`, but I don't think this would help anyways, as `FeatureServer` metadata does not give better information and we do not consume these services for data.

fgravin:
What is the state of this PR, is it ready for a merge or asking to changes ?
Anyway it needs a rebase on master because of the refactoring, sorry for that but it's easier in the way.
Thanks

libs/feature/search/src/lib/utils/links/link-utils.ts
```diff
@@ -47,17 +50,16 @@ export function getLinksWithWfsFormats(
...
-      featureType.outputFormats = endpoint.getServiceInfo().outputFormats
+    const featureType = endpoint.getFeatureTypeSummary(link.name)
+    if (featureType) {
+      return featureType.outputFormats.map((format) => ({
+        ...link,
+        url: endpoint.getFeatureUrl(featureType.name, { outputFormat: format }),
+        format: format,
+      }))
+    } else {
+      return Promise.reject(new Error('downloads.wfs.featuretype.not.found'))
```
jahow, line 61: you can probably just throw the error here, less confusing I think

libs/feature/search/src/lib/utils/links/link-utils.spec.ts
```diff
@@ -3,8 +3,28 @@ import {
...
+        getFeatureTypeSummary: (name) => {
+          return name !== 'NOT_A_LAYER_NAME'
+            ? {
+                name,
+                outputFormats: ['geojson', 'csv'],
+              }
+            : null
+        },
+      })
+    }
```
jahow, line 24: You can probably do:
```
isReady() { return Promise.resolve(this); }
getFeatureUrl() { ... }
getFeatureTypeSummary() { ... }
```

jahow (approved):
added a comment to help with the ogc-client mock, other than that for me it's good to go, thanks!

libs/feature/search/src/lib/utils/links/link-utils.spec.ts
tkohr: Thanks!

PR #180, opened 2021-11-10
Recognize JPG type to display in download list

tkohr:
PR recognizes JPG file type in links with protocol `WWW:DOWNLOAD` and displays it in the download list, which is currently not the case ("unknown"). To test this behavior, search for entry "Schémas de cohérence territoriale dans l'Oise" (471011ac-b927-4c8d-9589-ce3c530ea08d) for example.

### DIFF
--- a/libs/feature/search/src/lib/utils/links/link-classifier.service.spec.ts
+++ b/libs/feature/search/src/lib/utils/links/link-classifier.service.spec.ts
@@ -1,77 +1,7 @@
 import { LinkClassifierService, LinkUsage } from './link-classifier.service'
+import { LINK_FIXTURES } from './link.fixtures'
 
 describe('LinkClassifierService', () => {
-  const readmeLink = {
-    protocol: 'WWW:LINK',
-    description: 'Readme page',
-    url: 'http://envlit.ifremer.fr/resultats/quadrige',
-  }
-  const doiLink = {
-    protocol: 'WWW:DOI',
-    description: 'DOI for the resource',
-    url: 'http://doi.org/123-456-678',
-  }
-  const dataCsv = {
-    protocol: 'WWW:DOWNLOAD',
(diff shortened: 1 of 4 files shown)


fgravin:
Thanks @tkohr, please add a description, with an exemple of a failing link (you can take the one i wrote in Jira).
Also, you must add a test which would break without your fix and which now passes.

Thanks

tkohr:
Yes right @fgravin, I added the missing tests for the file types.

libs/feature/record/src/lib/links/link-utils.spec.ts
```diff
@@ -4,24 +4,86 @@ import {
   checkFileFormat,
   getLinksWithEsriRestFormats,
 } from './link-utils'
+import { LINK_FIXTURES } from './link.fixtures'
 
 describe('link utils', () => {
   describe('#getDownloadFormat', () => {
-    describe('for a FILE link', () => {
+    describe('for a csv FILE link', () => {
       it('returns download format', () => {
```
fgravin, line 12:
```suggestion
      it('returns csv format', () => {
```

libs/feature/record/src/lib/links/link.fixtures.ts
```diff
@@ -0,0 +1,108 @@
+export const LINK_FIXTURES = {
```
fgravin, line 1: good !

fgravin (changes requested):
minor comment but it looks good to me thanks.
You'll have to wait for the big refactoring PR to be merged and then update your PR accordingly, sorry for that.

tkohr:
You are right @fgravin, the description wasn't quite clear. Updated that, rebased and integrated your suggestion. If CI ends up green and you are ok with it, ready to merge I think.

fgravin (approved):

PR #181, opened 2021-11-15
feat(datahub): breadcrumb datahub link reset the search

fgravin:
The `Data hub` link in the breadcrumb no longer set the route to `/search` but also reset the search, as a link to _home_.
Quick win implementation.
I don't want to create a RESET search action, neither to manage default search params in the config etc... Keeping it simple for now.

After implementing this, i figured out that we could refactor a bit to move all what is related to the search in the `main-search.component` of the `datahub` application, this would avoid to call a children method from the parent (this is mandatory cause the `app.component` can't inject the `searchFacade` as it's injected underneath from the `gnUiSearchStateContainer` directive ).

### DIFF
--- a/apps/datahub/src/app/app.component.html
+++ b/apps/datahub/src/app/app.component.html
@@ -24,9 +24,10 @@
           truncate
           no-underline
           hover:text-primary-lighter hover:underline
+          cursor-pointer
         "
-        routerLink="search"
         translate
+        (click)="resetSearch()"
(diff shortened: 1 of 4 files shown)


jahow (approved):
Not tested but looking good. I think ideally the breadcrumb should be a smart component on its own, maybe even in a lib somewhere, but for now this will do just fine.

fgravin:
Thanks for the review @jahow

PR #182, opened 2021-11-16
Datahub / add Proxy service an associated configuration

jahow:
This PR introduces a very simple Proxy service with a `PROXY_PATH` injection token so that each app can configure the proxy path as they see fit. If no path is set, the proxy service does not alter URLs.

The proxy path for the datahub is present in the `env.js` file alongside the GN4 API. It can be customized using an env variable, including in the docker image. By default no proxy is set.

I also added a development proxy using webpack on the `/dev-proxy?` path, but this is not used by default and more of a convenience for investigating CORS-related issues. I don't want to enable this by default, otherwise things that work on dev will stop working in production without a doubt.

Added documentation, hopefully that will be clear enough.


fgravin (approved):
Good thanks.
Shall we set the proxy ENV somewhere in our fork ?

tkohr:
As discussed, I noticed that the proxy is not used in the data-fetcher, when trying to display CSV in the table for example.

jahow:
> Shall we set the proxy ENV somewhere in our fork ?

I think this can be set in the deployment environment

PR #183, opened 2021-11-16
datahub: open metadata on any autocomplete suggestion selection

fgravin:
Change the behavior of the `fuzzy-search` input suggestion selection, instead of copying the suggestion in the input and triggering a search, it opens directly the record.

This required several changes:

- refactor the `AutocompleteComponent` to handle objects instead of string
- refactor the `FuzzySearchComponent` to distinguish the enter key press and the suggestion selection event, allowing to override the behavior by subscribing to the output. We now pass the suggestion `MetadataRecord` as item for the suggestion list.


jahow (approved):
Looking good, except that I don't understand what is the use of the `displayWithFn` input on the autocomplete component. If I'm not mistaken it is the same function as `itemToStringFn`, also its name is confusing and I feel like it's the angular material API leaking in. But I'll let you judge on that.

Thanks!

fgravin:
Thanks @jahow, I first added a custom `itemToStringFn` to allow to pass objects instead of strings.
`itemToStringFn` is used to display the text from the object in the dropdown list.
I discoverd afterwards the `displayWithFn` option, which is almost the same but different, this one is used to know what to display in the input after an item selection. 
I agree it is almost always the same, but could be different.

In my case, after picking an item in the list, i clear the input so `displayWithFn = () => ''`, but it could be managed in another way, cause actually the item is still selected internally (this should be done in the ItemSelected intern event).

I think this component needs some refactoring anyway, I just wonder if we shouldn't have 2 components, one which is a regular autocomplete, and one which is an autocomplete with a custom action on selection.

I also think that the `hostCpt -> fuzz-search -> autocomplete -> materialAutocomplete` chains would be a good candidate to owns its serviceManager beside the line, so we could interact with any nested component method from the host (eg using the clearFn).

Let's discuss that in a future coding campaign.

From my explanation, do you expect me to change something with the `itemToStringFn` and `displayWithFn` or not ?
Thanks for the review.

jahow:
> From my explanation, do you expect me to change something with the `itemToStringFn` and `displayWithFn` or not ?
> Thanks for the review.

No it's good, let's move on. Maybe add a `// FIXME` to remind us to reevaulate all this in the future. Thanks for the explanation!

PR #184, opened 2021-11-17
datahub: fix state conflicts between search and mdview

fgravin:
The router effect which navigates to `/search` on `RequestMoreResults` action was too eager and has been removed.
It was causing several issues
- scrolling on record view page triggers a search (infinite scrolling bug) so it goes back to search page
- race condition on record view relaoding page, the full metadata comes after the search results list so it goes back to /search and then display the record view while on `/search` route

So we manually redirects to` /search` when desired in the datahub application.

- on fuzzy search input submission
- on datahub click from the breadcrumb

### DIFF
--- a/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
+++ b/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
@@ -27,6 +27,7 @@ import { ElasticsearchMapper } from '../utils/mapper'
 export class FuzzySearchComponent implements AfterViewInit {
   @ViewChild(AutocompleteComponent) autocomplete: AutocompleteComponent
   @Output() itemSelected = new EventEmitter<MetadataRecord>()
+  @Output() inputSubmited = new EventEmitter<string>()
   @Input() displayWithFn: (MetadataRecord) => string = (record) => record?.title
   itemToStringFn = (record: MetadataRecord) => record?.title
 
@@ -70,5 +71,6 @@ export class FuzzySearchComponent implements AfterViewInit {
 
   handleInputSubmission(any: string) {
     this.searchFacade.setFilters({ any })
+    this.inputSubmited.emit(any)
   }
 }
--- a/apps/datahub/src/app/app.component.html
+++ b/apps/datahub/src/app/app.component.html
@@ -15,6 +15,7 @@
         <gn-ui-fuzzy-search
           [displayWithFn]="autocompleteDisplayWithFn"
           (itemSelected)="onFuzzySearchSelection($event)"
+          (inputSubmited)="onFuzzySearchSubmission()"
         ></gn-ui-fuzzy-search>
       </div>
     </div>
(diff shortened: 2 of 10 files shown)


libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -27,6 +27,7 @@ import { ElasticsearchMapper } from '../utils/mapper'
 export class FuzzySearchComponent implements AfterViewInit {
   @ViewChild(AutocompleteComponent) autocomplete: AutocompleteComponent
   @Output() itemSelected = new EventEmitter<MetadataRecord>()
+  @Output() inputSubmited = new EventEmitter<string>()
```
jahow, line 30:
```suggestion
  @Output() inputSubmitted = new EventEmitter<string>()
```

jahow (commented):
Still not perfect apparently :crying_cat_face: 

* On first load I am not taken automatically to the `/search` route, for example I stay on http://localhost:4200
* When selecting a result in the autocompletion list with Enter:
  * I am transported to the `/metadata/<uuid>` route but immediately taken to the `/search` route
  * UI is showing me the record metadata
  * This does not happen on mouse click (behaviour should not be differentiated!)
  * Afterwards the UI is broken, I cannot reach the search results again

I think it makes sense that the router brings the user back to search automatically when search results are updated. It feels like this change would give too much responsibility to the app. Why not make sure that the search is really disabled in the datahub when not on the `/search` route? This should fix the issue as well.

fgravin:
> * On first load I am not taken automatically to the `/search` route, for example I stay on http://localhost:4200

Not related, we can do a default redirect in the route definition if needed.

> * When selecting a result in the autocompletion list with Enter:
>   
>   * I am transported to the `/metadata/<uuid>` route but immediately taken to the `/search` route
>   * UI is showing me the record metadata
>   * This does not happen on mouse click (behaviour should not be differentiated!)
>   * Afterwards the UI is broken, I cannot reach the search results again

The UI is broken because of the mdview state, not because of the bug. The bug here is that it triggers a search on the `enter key` selection.
I implemented a workaround, which is not purely satisfying because it changes the `Autocomplete` generic behavior to our needs, anyway, it still work in a default way as well (tested on search app and works as before).

> I think it makes sense that the router brings the user back to search automatically when search results are updated. It feels like this change would give too much responsibility to the app. Why not make sure that the search is really disabled in the datahub when not on the `/search` route? This should fix the issue as well.

Yes, but i think it's more work and refactoring, maybe we can handle that when we manage a search router state if we do one day. Also, we wouldn't be able to have background search task, even though it might not be an issue.

I personally would go for the merge event though it's not perfect and try to improve this when we have more time and when we will need to handle different behaviors for different usecases.

jahow (approved):
That's a workaround by the books :slightly_smiling_face: and it works well; let's go with this and rework the autocomplete component later on. Thanks!

PR #185, opened 2021-11-17
datahub: improve download list layout

fgravin:
I wanted to work on ODS format but it seems more backend related, i started to update the design anyway so doing a PR just for this.

Screenshot from 2021-11-17 14-21-01


tkohr (approved):
LGTM, thanks!

PR #186, opened 2021-11-17
Display data from ESRI:REST services in table and map

tkohr:
PR displays data from ESRI:REST services in table and on map. This includes `FeatureServers` and `WFSServers`. The tested `WFSServers` responded with errors however, as their `FeatureType` names did not match the link names indicated in the metadata.
Can be tested on the datahub with any service found in https://data-atmo-hdf.opendata.arcgis.com/data.json (eg. "emi hdf region 2012").
I had to enable the map for data-links to display the data on the map (which should have already been the case for a WFS without WMS before), hoping this is not a big drawback if there are only non-geographic data-links and no map-api-link.

### DIFF
--- a/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
+++ b/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
@@ -26,6 +26,7 @@ import { MetadataLinkValid } from '@geonetwork-ui/util/shared'
 import { readDataset } from '@geonetwork-ui/data-fetcher'
 import { fromPromise } from 'rxjs/internal-compatibility'
 import { WfsEndpoint } from '@camptocamp/ogc-client'
+import { getEsriRestDataUrl } from '@geonetwork-ui/feature/search'
 
 @Component({
   selector: 'gn-ui-data-view-map',
@@ -144,6 +145,20 @@ export class DataViewMapComponent {
     } else if (link.protocol.startsWith('WWW:DOWNLOAD')) {
       return fromPromise(
         readDataset(link.url, 'geojson').then((features) => ({
+          type: MapContextLayerTypeEnum.GEOJSON,
+          data: {
+            type: 'FeatureCollection',
+            features,
+          },
+        }))
+      )
+    } else if (
+      /^ESRI:REST/.test(link.protocol) &&
+      /FeatureServer/.test(link.url)
+    ) {
+      const url = getEsriRestDataUrl(link, 'geojson')
+      return fromPromise(
+        readDataset(url, 'geojson').then((features) => ({
           type: MapContextLayerTypeEnum.GEOJSON,
           data: {
             type: 'FeatureCollection',
--- a/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
+++ b/libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
@@ -121,7 +121,10 @@ export class DataViewMapComponent {
         type: MapContextLayerTypeEnum.WMS,
         name: link.name,
       })
-    } else if (link.protocol.startsWith('OGC:WFS')) {
+    } else if (
+      link.protocol.startsWith('OGC:WFS') ||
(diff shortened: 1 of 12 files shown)


libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -36,12 +36,20 @@
     </div>
   </mat-tab>
 
-  <mat-tab [disabled]="(displayMap$ | async) === false">
+  <mat-tab
+    [disabled]="
+      (displayMap$ | async) === false && (displayData$ | async) === false
```
fgravin, line 41: What is `displayMap$` if it's not about displaying the map ?

libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
```diff
@@ -121,7 +121,10 @@ export class DataViewMapComponent {
         type: MapContextLayerTypeEnum.WMS,
         name: link.name,
       })
-    } else if (link.protocol.startsWith('OGC:WFS')) {
+    } else if (
+      link.protocol.startsWith('OGC:WFS') ||
```
fgravin, line 125: I would put all this kind of checks in helpers

libs/feature/record/src/lib/data-view-table/data-view-table.component.ts
```diff
@@ -70,7 +70,10 @@ export class DataViewTableComponent {
   constructor(private mdViewFacade: MdViewFacade) {}
 
   fetchData(link: MetadataLinkValid): Observable<{ id: string | number }[]> {
-    if (link.protocol.startsWith('OGC:WFS')) {
+    if (
```
fgravin, line 73: Check use for the second time here => mutualise in a helper if possible.
Notice that if it was already the case, you would have to modify the code once and don't care about where it's used, now you need to change the test everywhere you use it.

fgravin (changes requested):
Not tested cause i don't really understand what to test.
Some coding comment to improve the readability and the efficiency, I let @jahow checks if it implements the expected behavior.
Thanks

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
tkohr: True, not very meaningful what I wrote

libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
tkohr: Good point, I'll adapt that

tkohr: So I finally put them in the `link.helper`, but indeed we might do some more refacto/merging of the three link services

tkohr:
Thanks for the review @fgravin. I updated the PR accordingly

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
```diff
@@ -10,8 +11,15 @@ import { map } from 'rxjs/operators'
...
-    map((links) => !!links && links.length > 0)
+  displayMap$ = combineLatest([
+    this.facade.mapApiLinks$,
+    this.facade.dataLinks$,
+  ]).pipe(
+    map(
+      ([mapLinks, dataLinks]) =>
+        (!!mapLinks && mapLinks.length > 0) ||
+        (!!dataLinks && dataLinks.length > 0)
+    )
```
jahow, line 22: This is not ideal, eventually we'll need another type of usage like `GEODATA` or something, but for now this will do I guess

jahow (approved):
Thanks! I agree, the link utils and link helper really feel weird now, we really have to refactor that part to make it simpler, but this will have to be done in another PR

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
tkohr: I agree, didn't feel that well with this. I think we'll have to get back to the classification at some point. Thanks for the review!

PR #187, opened 2021-11-18
Fix: Rename empty string property ids to 'unknown'

tkohr:
PR adds a small fix, renaming column names that consist of an empty string to "unknown" which currently creates a `properties` object with a missing key which breaks the table.

Can be tested with "Accroches vélos MEL" once its data formats will be harvested and available in the datahub.

### DIFF
--- a/libs/util/data-fetcher/src/parsers/json.spec.ts
+++ b/libs/util/data-fetcher/src/parsers/json.spec.ts
@@ -1,4 +1,4 @@
-import { parseJson } from './json'
+import { jsonToGeojsonFeature, parseJson } from './json'
 
 describe('parseJson', () => {
   describe('valid JSON with id', () => {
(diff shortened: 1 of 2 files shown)


jahow:
@tkohr I don't understand how this problem can happen, do you have a sample file that we can check?

For reference here is the record: https://metropole-europeenne-de-lille.opendatasoft.com/explore/dataset/accroche_velos

jahow:
> Note: Data-fetcher messages are not yet translated. Is this possible with the current approach as it is not an Angular but ES module ?

It might be a problem but hopefully not. The idea is that the data-fetcher throws errors with flags on them to let the app know what happened, and also specify a message with a technical explanation (in english). See here for the errors that can be thrown:
https://github.com/geonetwork/geonetwork-ui/blob/master/libs/util/data-fetcher/src/lib/data-fetcher.ts#L12

tkohr:
> @tkohr I don't understand how this problem can happen, do you have a sample file that we can check?
> 
> For reference here is the record: https://metropole-europeenne-de-lille.opendatasoft.com/explore/dataset/accroche_velos

@jahow it happens with the aforementioned dataset (having to adapt the `link-classifier` for now to test it to let it be classified as a `DATA` link). Actually, I thought it was related to interpreting the "," in one column as separator, but I just realize that there really is a last "unnamed" column. So we should probably not throw an error, but handle this and display the data as the portal of Lille does.

jahow:
It depends what exchange format we use: with CSV this dataset is parsed incorrectly because of commas in URLs, but it should work fine using GeoJSON.

Without this fix, does the UI crash or is the error handled somewhere else?

tkohr:
> It depends what exchange format we use: with CSV this dataset is parsed incorrectly because of commas in URLs, but it should work fine using GeoJSON.


No, this is actually not caused by the commas in the URL, which are handled well by the parser, but by the last column that has an empty string as name. I've changed the commit to rename an empty string to 'unknown' to not break the table and display the data. `GeoJSON` has the problem here to be interpreted as `JSON`.

> Without this fix, does the UI crash or is the error handled somewhere else?

Yes, without this fix the app crashes.

jahow (approved):
Ok, that looks better now! thanks!

jahow:
Actually, there is still something unresolved here: an error during parsing should not have been able to make the whole UI crash. So we'll have to come up with a following PR to address that.

PR #188, opened 2021-11-18
Datahub / add file-based configuration and use it for the main header

fgravin:
This PR introduces a way to configure the Datahub application using a [TOML](https://toml.io) file. TOML is nice because it is easy to author, supports comments and [has a JS parser](https://github.com/LongTengDao/j-toml).

The current approach for configuring the Datahub (and other applications later on) is this:
* The global app configuration can be used to:
  * specify GN4 api url and proxy path (both can also still be done through env vars)
  * specify theme parameters: colors, font families etc.
  * override default translations with custom ones
* The app configuration file is an asset of the app, and as such is loaded by the frontend dynamically (instead of being baked in the app's static files)
* As such, the configuration can be either directly modifies in the app assets when deploying without docker, or overridden with a custom one when using docker

Note that in another PR we'll have to introduce CSS variables specific to UI components (e.g. autocomplete) which can then be overridden in apps to modify them accordingly; this will _not_ be possible in the global configuration file, but hardcoded in each application stylesheet.

To do:
- [x] write default config file
- [x] write config file loading logic to apply custom colors
- [x] modify translation loaded to use custom translations from config (possibly changed at runtime?)
- [x] add CSS variables for font families & logic for setting them based on config

image


jahow:
Some thoughts:
* I think the font family should be handled like other CSS variables and added to the tailwind config, actually two families: `--font-family-main` and `--font-family-title`; then either font can be used using CSS classes anywhere in the components
* for the main title I would actually rely on translations; I think using HTML with inline classes styles (for color and size) is possible in translations, and since it's readable text it has to be translatable; so no settings, but we need a way to customize translations using override files
* I would replace the `--catalog-title-size` var with inline style in the title HTML; might as well use just one parameter for everything!
* As such, the catalog-title component doesn't really bring any value, we might as well use a `<div [innerHtml]="title">` right away
* +1 for using `datahub` as a prefix, great idea

Having a user-friendly way to customize translations is important IMO, especially if we use that for titles & introductions. The requirements would be:
* the app should be able to get translations both from default files (`translations/en.json` etc.) and override files, without having dozens of 404 errors if no override files exist in some languages
* when using docker, ability to specify override files for any language at run time (and not compile time), maybe using a volume for example
* when not using docker, ability to add and/or edit override files for any language
* it should be possible to access the app in a mode that disables translation altogether, so that people can actually look at the keys that are used and know which one to customize, eg by using https://my.host/datahub?disable_i18n

As for the colors and fonts, maybe we should introduce a proper settings system with a file that can be modified more easily than `env.js` (which might be a bit cryptical and error-prone). Then:
* when using docker, individual settings can be modified using env vars at run time, eg:
  ```shell
  $ docker run -e GN4_API_URL="../geonetwork/srv/api" -e PROXY_URL="../proxy?url=" -e PRIMARY_COLOR="blue" -e SECONDARY_COLOR="green" geonetwork-ui/datahub
  ```
* when not using docker, the settings file can be easily edited for each instance manually

jahow:
I have rebased this PR and will rework the title & body to match advancement

jahow:
This is ready for a first review I think. I still need to document how app configuration works.

Docker workflow is broken as well since we don't have `env.js` any more.

apps/datahub/src/index.html
```diff
@@ -15,9 +15,9 @@
       href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined"
       rel="stylesheet"
     />
-    <script src="assets/env.js"></script>
+    <link rel="preload" href="assets/configuration/default.toml" as="fetch" />
```
fgravin, line 18:
```suggestion
    <link rel="preload" href="assets/configuration/default.toml" as="fetch" importance="high"/>
```
may be better

fgravin (commented):
Very good work seems clean. I will test that on monday before approving.
Nevertheless, i don't see where you actually download the fonts.

fgravin (commented):
I can't approve cause it's my PR but it's ok for me, great work.
I have tested it a bit and looks to work well.

We will now need to rely more on the correct css var and classes, like it seems we are not using `main_color`, or what is `background` ? We need some rigor about this, thanks.

tkohr (approved):
Nice, thanks @jahow! As far as I can see, looks very good and works well. Adds some complexity, but I think this is unavoidable to allow customization.

Didn't quite see where the fonts are downloaded, neither. Does tailwind take care of this?

I think we can still remove the buttons in the header, but this might as well be part of further UI adaptions.

fgravin:
Thanks @jahow 
Merging to have this refactoring while starting new components
TODO in a next PR:

- load font
- load background image
- be sure `color-main`, `font-main` are used by defaut

PR #189, opened 2021-11-19
Datahub / Use proxy for showing data in table and map, rely on mime types in protocol if any

jahow:


libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
```diff
@@ -151,24 +153,28 @@ export class DataViewMapComponent {
...
       const url = getEsriRestDataUrl(link, 'geojson')
       return fromPromise(
-        readDataset(url, 'geojson').then((features) => ({
-          type: MapContextLayerTypeEnum.GEOJSON,
-          data: {
-            type: 'FeatureCollection',
-            features,
-          },
-        }))
+        readDataset(this.proxy.getProxiedUrl(url), 'geojson').then(
```
fgravin, line 169: duplicated code, should the if just about the `url` var set ?

libs/feature/record/src/lib/data-view-table/data-view-table.component.ts
```diff
@@ -111,7 +113,7 @@ export class DataViewTableComponent {
     } else if (this.linkHelper.isEsriRestFeatureServer(link)) {
       const url = getEsriRestDataUrl(link, 'geojson')
       return fromPromise(
-        readDataset(url, 'geojson').then((features) =>
+        readDataset(this.proxy.getProxiedUrl(url), 'geojson').then((features) =>
```
fgravin, line 124: again, i would have done a shared `readDatasetFn` somewhere, it's like for fetching the dataset we have to change the code at 3 or 4 different places.

fgravin (approved):
Thanks.
Some comments but not related with your work.

libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
jahow: yeah I agree it's bad... sorry, don't have much time now, we'll revisit this later on I guess

libs/feature/record/src/lib/data-view-table/data-view-table.component.ts
jahow: agreed!

jahow:
@fgravin thanks for the review, I don't have time to address the comments now, we will have to revisit this code and also the format detection because there is way too much duplicated logic

PR #191, opened 2021-11-24
Fix initial map zoom on data layer

tkohr:
PR intends to fix the initial zoom on the data layer when accessing the map. Currently, the app does not wait for the map to be rendered when it needs to calculate the map view from a layer extent. Since the map size is needed for this calculation, a fitting view can not be determined before rendering.

This PR uses a `ResizeObserver` to wait for the DOM, before calculating a map view from a layer extent.

TODOs

- [ ] finish `map-context.component` (tests)


libs/feature/map/src/lib/map-context/component/map-context.component.ts
```diff
@@ -25,19 +26,35 @@ export class MapContextComponent implements OnChanges {
   @Input() context: MapContextModel
   @Output() featureClicked = new EventEmitter<Feature<Geometry>[]>()
 
+  resizeObserver = new ResizeObserver(() => {
+    if (this.context.extent) {
+      this.map.updateSize()
+      this.context.view = this.utils.getViewFromExtent(
```
fgravin, line 32: I may missed something but why not using the `map.fit()` method ?
Has the map already a `View `or not BTW ?

libs/feature/map/src/lib/map-context/component/map-context.component.ts
```diff
@@ -25,19 +26,35 @@ export class MapContextComponent implements OnChanges {
...
   ) {
     this.map = manager.map
   }
 
   ngOnChanges() {
     if (this.context) {
-      this.service.resetMapFromContext(this.map, this.context)
+      this.mapEl = this.map.getTargetElement()
+      if (!this.mapEl) return
+      this.resizeObserver.observe(this.mapEl)
```
fgravin, line 57: Could we observe the map element in the low level map component instead or not ?

fgravin (commented):
Thanks LG, I need to test that

libs/feature/map/src/lib/map-context/component/map-context.component.ts
tkohr: No, when accessing the map for the first time it has no `View` yet, so this prevents creating a default view and fitting on the extent afterwards. Apart from that, `fit()` should also work now with the `ResizeObserver`.

libs/feature/map/src/lib/map-context/component/map-context.component.ts
tkohr: Which component do you mean here?

tkohr: ok, so we've moved it to the `map.component`. Thanks @fgravin !

fgravin (approved):
Looks good thanks

PR #192, opened 2021-11-25
Proxy service now encodes urls

jahow:
This required an adaptation in ogc-client, see: https://github.com/camptocamp/ogc-client/commit/7813e712b230699d3f53f2a3e35813f64fccf89f


fgravin (approved):
LG gn-ui side

PR #193, opened 2021-11-26
Add GEODATA classification

tkohr:
PR adds GEODATA classification to prevent proposing to display DATA without spatial reference on the map.


jahow (approved):
Thanks!

fgravin (commented):
Is geojons both `DATA` and `GEODATA` or just `GEODATA` ?

tkohr:
> Is geojons both `DATA` and `GEODATA` or just `GEODATA` ?

I chose to only classify it as GEODATA and to let the table display DATA and GEODATA.

PR #195, opened 2021-11-30
Add card components for APIs, otherlinks, related records

tkohr:
PR implements card components following the UI spec to display:

- APIs (modifies current UI component)
- Otherlinks (adds UI component)
- Related records (adds UI component)

Note: Cards currently appear a little wider, than on the screens. This might still be adapted, once the cards will be used.
Also, the color "calculation" for badges in API and Download link representations are not intended in the UI screens, if I see right. Shall we remove them?
api
link
relatedrecord


fgravin (commented):
Can you put back the geonetwork api path to dev.geo2france plz so the application works in gh-pages.
Also, when working on UI components, try to upload screenshot when possible.
I have looked in the published story book and it looks nice, great.
I will review the code afterwards.
Thanks

tkohr:
Hum, I didn't touch the api path, doesn't seem to work on master neither anymore. Did we break this with the customization? Currently, `default.toml` points at "/geonetwork/srv/api" and `proxy.config.js` at "https://dev.geo2france.fr".

fgravin:
> Hum, I didn't touch the api path, doesn't seem to work on master neither anymore. Did we break this with the customization? Currently, `default.toml` points at "/geonetwork/srv/api" and `proxy.config.js` at "https://dev.geo2france.fr".

yes @jahow  removed it from the ENV but I think it's better for the dev to have it at the moment.
Put geo2france in the toml plz

libs/ui/elements/src/lib/apis-list/apis-list.component.spec.ts
```diff
@@ -1,8 +1,9 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing'
 
 import { ApisListComponent } from './apis-list.component'
-import { ApisListItemComponent } from '../apis-list-item/apis-list-item.component'
```
fgravin, line 4: remove the api-list-component

fgravin (approved):
Good !
Now we need to add this new components in the record view page ;)

conf/default.toml
```diff
@@ -3,11 +3,11 @@
 
 [global]
 # This URL (relative or absolute) must point to the API endpoint of a GeoNetwork4 instance
-geonetwork4_api_url = "/geonetwork/srv/api"
+geonetwork4_api_url = "https://dev.geo2france.fr/geonetwork/srv/api"
```
fgravin, line 6: remove those 2 changes cause it actually break the dev mode

fgravin (changes requested):

libs/ui/elements/src/lib/apis-list/apis-list.component.spec.ts
tkohr: Ok, done. Then would you loop over the cards in `DataApisComponent` or even get rid of this component and use them directly in `RecordMetadataComponent`?

fgravin: I think `DataApisComponent` was there because of the tabs. I'm not sure but maybe we could do it directly in `RecordMetadataComponent`.
`DataApisComponent` or `ApiListComponent` don't bring much value

tkohr: For sure, they don't add much logic. For now, I'm keeping `DataApisComponent` in the UI PR (#196), so that `RecordMetadataComponent` does not become too massive. It's rather a presentational component than a smart component though. Don't hesitate to get back to this topic in the other PR to discuss what makes most sense.

PR #198, opened 2021-11-30
Handle landing page link in metadata record

fgravin:
Ad a new `LinkUsage.LANDINGPAGE` if the link has the protocol equal to `WWW:LINK:LANDING_PAGE`.
The landing page link is added as an observable in the record state and passed to the `metadata-info-component`.


jahow (approved):
Looking good but the landing page is not shown yet, right?

fgravin:
> Looking good but the landing page is not shown yet, right?

Yes, I wait for @tkohr to finish the refactoring of the metadata-info component.
Also waiting for the expandable panel.
Thanks for the review @jahow

PR #199, opened 2021-12-01
Datahub header layout

fgravin:
Improve header layout from mockup

- cleaner style definition and overrides for fuzzy search input
- remove unused elements (favorites, organisation, etc...)
- dataset tab click reset the search
- add header background image
- adapts style from mockups (font-size, margins, colors, etc..

Screenshot from 2021-12-01 14-30-36
)

PR #200, opened 2021-12-02
Restore docker build with the new config system

jahow:
This PR reworks the generic dockerfile for apps and associated docker-entrypoint.sh to work with the new config file. It is still possible to specify a GN4 url and proxy path using env var, or a full config file can be given to the docker container at startup.

Also:
* apply "main" color to all text by default (in the datahub only for now)
* allow specifying custom translations without quotes in the TOML file; this is a workaround for a quirk of the TOML syntax
* allow specifying a stylesheet link for fonts (e.g. google fonts) in config

Rebased on #199

### DIFF
--- a/apps/datahub/README.md
+++ b/apps/datahub/README.md
@@ -54,3 +54,17 @@ Notice how the `GN4_API_URL` and `PROXY_PATH` variables are used to override any
 **This override will happen everytime the docker container is started.**
 
 The application will be available on http://localhost:8080/datahub/.
+
+### Specifying a custom configuration file
+
+If the `GN4_API_URL` and `PROXY_PATH` environment variables are not enough and you want to specify a full configuration file,
+you can do so like this:
+
+```bash
+# this assumes a custom `default.toml` file is located in the /home/user/custom-conf directory:
+$ docker run -p 8080:80 \
+             -v /home/user/custom-conf:/conf \
+             geonetwork-ui/datahub
+```
+
+If a `default.toml` file is found in the `/conf` folder of the app container at startup, it will be used instead of the default one.
--- a/apps/datahub/README.md
+++ b/apps/datahub/README.md
@@ -61,10 +61,10 @@ If the `GN4_API_URL` and `PROXY_PATH` environment variables are not enough and y
 you can do so like this:
 
 ```bash
-# this assumes a custom `default.toml` file is located in the /home/user/custom-conf directory:
+# this assumes a file named `default.toml` is located in the /home/user/custom-conf directory:
 $ docker run -p 8080:80 \
              -v /home/user/custom-conf:/conf \
              geonetwork-ui/datahub
 ```
 
-If a `default.toml` file is found in the `/conf` folder of the app container at startup, it will be used instead of the default one.
+If a file named `default.toml` is found in the `/conf` folder _of the app container_ at startup, it will be used by the application.
--- a/apps/datahub/README.md
+++ b/apps/datahub/README.md
@@ -16,11 +16,7 @@ will run `datahub` as an Angular application available on localhost:4200.
 
 ## Configuration
 
-### API url
-
-By default it uses GN4 API `/geonetwork/srv/api`.
-
-You can override this URL with the ENV variable `${GN4_API_URL}`.
+See the [main README section for more info](../../README.md#application-configuration).
 
 ### Proxy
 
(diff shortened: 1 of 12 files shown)


conf/default.toml
```diff
@@ -18,24 +21,31 @@ primary_color = "#0f4395"
...
 # [translations.fr]
 # "results.sortBy.dateStamp" = "Dernière fois que quelqu'un a modifié quelque chose"
 #
-# Note that translation keys ending with ".html" _can_ contain HTML markup, including inline CSS:
-# "catalog.welcome.html" = """
-# Welcome to <span class="text-primary">Organization</span>'s<br>
-# wonderful <span style="font-size: 1.2em;">data catalogue</span>
-# """
+# Note 1:
+#   translation keys ending with ".html" _can_ contain HTML markup, including inline CSS:
```
fgravin, line 40: nice !

libs/util/shared/src/lib/services/theme.service.ts
```diff
@@ -81,6 +82,14 @@ export class ThemeService {
         titleFont
       )
     }
+
+    if (fontsStylesheetUrl) {
```
fgravin, line 86: what if we need several fonts ?

apps/datahub/src/index.html
```diff
@@ -6,11 +6,8 @@
     <base href="/" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap"
-      rel="stylesheet"
-    />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
```
fgravin, line 9: can you explain what it is for ?

tools/docker/docker-entrypoint.sh
```diff
@@ -1,17 +1,26 @@
...
+    echo "[INFO] Replaced GN4 url in conf with: ${GN4_API_URL}"
+  fi
+  if [ ! -z "${PROXY_PATH}" ]
+  then
+    sed -i "s%proxy_path = \".*\"%proxy_path = \"${PROXY_PATH}\"%" /usr/share/nginx/html/${APP_NAME}/${CONFIG_FILE_PATH}${CONFIG_FILE_NAME}
+    echo "[INFO] Replaced proxy path in conf with: ${PROXY_PATH}"
+  fi
 fi
 
+
```
fgravin, line 25: I'm not sure to understand how we can override the config file, shall we edit the one from `assets/configuration/default.toml`, or from `conf/default.toml`, can we override the `CONFIG_FILE_NAME` ?

fgravin (approved):
Looks good thanks.
I just asked some question to be sure to understand the workflow, thank great job.

libs/util/shared/src/lib/services/theme.service.ts
jahow: so if it's several google fonts you can specify them all in the same url; I don't know of any other apis like that but if the need arises then we can support an array of values; on the other hand, I'd be more keen in allowing custom fonts as files in the assets instead

apps/datahub/src/index.html
jahow: It's what google recommends to use for google fonts:
image

Notice the two `preconnect` links, which work a bit like `preload` except they [only open a connection](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preconnect).

tools/docker/docker-entrypoint.sh
jahow: to override the config file you need to essentially provide your own, custom file. For example, you can copy the `conf/default.toml` file in another directory, modify it, and provide this directory as a volume to the docker container.

Or you can just modify the `conf/default.toml` file in place and do:
```shell
$ docker run -v `pwd`/conf:/conf -p 8080:80 geonetwork-ui:datahub
```

apps/datahub/README.md
```diff
@@ -54,3 +54,17 @@ Notice how the `GN4_API_URL` and `PROXY_PATH` variables are used to override any
...
+you can do so like this:
+
+```bash
+# this assumes a custom `default.toml` file is located in the /home/user/custom-conf directory:
+$ docker run -p 8080:80 \
+             -v /home/user/custom-conf:/conf \
+             geonetwork-ui/datahub
+```
+
+If a `default.toml` file is found in the `/conf` folder of the app container at startup, it will be used instead of the default one.
```
tkohr, line 70: Did you mean ?
```suggestion
If a `custom.toml`...
```

conf/default.toml
```diff
@@ -36,17 +36,12 @@ background_color = "#fdfbff"
 # [translations.fr]
 # "results.sortBy.dateStamp" = "Dernière fois que quelqu'un a modifié quelque chose"
```
tkohr, line 37: Shouldn't this example be without quotes? Did I understand right, that quotes are only used for custom translations that display quotes in the app text?

tkohr (commented):
Looks nice, as far as I can judge. Thanks @jahow !

apps/datahub/README.md
jahow: No, actually only a file named `default.toml` can be used currently. Maybe this is misleading since we're providing a custom configuration, but I could not find a better name :)

conf/default.toml
jahow: > Did I understand right, that quotes are only used for custom translations that display quotes in the app text?

Not sure what you mean:
* quotes around the key (left part) are optional (they weren't before this PR)
* quotes around the translated text (right) are mandatory since TOML requires strings to be quoted

tkohr: Ok, got it

apps/datahub/README.md
tkohr: Ok, maybe it would make sense to write "If a custom `default.toml` file is found..." then, just to be more clear ?

apps/datahub/src/index.html
jahow: Also, to explain a bit more, I chose to "hardcode" this in the datahub template in anticipation of people using google fonts; ideally this should be added dynamically and not rely on modifications in each app's template

jahow:
Thanks for the reviews!

PR #201, opened 2021-12-04
datahub: add related record section

fgravin:
Handle related records fetching through state effect on `LoadRecordFullSuccess` action.
Add the related record section in the record-metadata component.

This PR does not address layout issues with the `related-card` component, neither the action on the click, it takes the component as it is except:

- add gray box when no image
- fix sanitize for unsafe urls

Screenshot from 2021-12-04 17-46-45

### DIFF
--- a/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
+++ b/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
@@ -1,9 +1,11 @@
-<div class="w-64 overflow-hidden rounded-lg shadow-lg">
-  <img
-    class="object-cover w-full h-52"
-    [src]="record.thumbnailUrl"
-    [alt]="record.title"
-  />
+<div class="w-64 overflow-hidden rounded-lg shadow-lg bg-white">
+  <div class="object-cover w-full h-52 bg-gray-100">
+    <img
+      *ngIf="record.thumbnailUrl"
+      [src]="record.thumbnailUrl | safe: 'url'"
+      [alt]="record.title"
+    />
+  </div>
   <div class="px-5 pt-4 pb-6">
     <h4 class="max-h-20 mb-5 text-xl overflow-ellipsis overflow-hidden">
       {{ record.title }}
--- a/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
+++ b/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
@@ -1,6 +1,7 @@
 <div class="w-64 overflow-hidden rounded-lg shadow-lg bg-white">
-  <div class="object-cover w-full h-52 bg-gray-100">
+  <div class="h-52 bg-gray-100">
     <img
+      class="h-52 w-full object-cover"
       *ngIf="record.thumbnailUrl"
       [src]="record.thumbnailUrl | safe: 'url'"
       [alt]="record.title"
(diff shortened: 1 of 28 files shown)


libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -48,6 +49,45 @@ export class ElasticsearchService {
...
+    title: string,
+    size: number = 6,
+    _source = ES_SOURCE_SUMMARY
+  ): EsSearchParams {
+    return {
+      query: {
+        bool: {
+          must: [
+            {
+              more_like_this: {
```
jahow, line 62: I think this is not in line with customer requirements: the idea for related records was to show records which have an association with the current one: sibling, parent, child, service etc.

libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
```diff
@@ -1,9 +1,11 @@
...
-    [alt]="record.title"
-  />
+<div class="max-w-xs overflow-hidden rounded-lg shadow-lg bg-white">
+  <div class="object-cover w-full h-52 bg-gray-100">
+    <img
+      *ngIf="record.thumbnailUrl"
+      [src]="record.thumbnailUrl | safe: 'url'"
+      [alt]="record.title"
+    />
+  </div>
```
jahow, line 9: I feel like this was already addressed in `gn-ui-record-thumbnail` (having a placeholder in case no thumbnail was given)

jahow (changes requested):
Thanks! Not tested, just looked at the code. If I'm not mistaken the method used for fetching related records is not right, see comments. Also I think the `gn-ui-record-thumbnail` component can be reused to avoid duplicating logic for placeholder, sanitizing etc. Other than that, all good for me!

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
fgravin: .... the mockup doesn't tell anything about what kind of relation it is.
doesn't tell either what happens if more than 3
we can keep the code cause it's going to be usefull anyway, if they want the `gn-related` directive, then I won't do it personally, no more functional dev for this contract. And I don't like the semi-dozens of requests sent to fetch the `related` in gn3

libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
fgravin: Yes, so maybe `gn-ui-record-thumbnail` should be refactored to only display the image, there are way to many divs and css classes to be used as it is in other components.

jahow: it does display the image only: https://github.com/geonetwork/geonetwork-ui/blob/master/libs/ui/search/src/lib/record-thumbnail/record-thumbnail.component.html

Honestly I think you can probably just use it straight away without modification

tkohr:
I've rebased and used `<gn-ui-record-thumbnail>` (which is in `UiSearchModule`). It would've looked more logical to me to move it to `UiElementsModule`now, but I ended up with a circular dependency this way. Would like to have your confirmation for my commit and an adaption of https://github.com/geonetwork/geonetwork-ui/pull/201/files#diff-94088355691a4fe01c7531d9ec1597b658b6ff6cdb2f7c26fa01f03349cfe6a5R109 (which does not pass) if possible, before merging as discussed @fgravin .

fgravin:
Thanks @tkohr , i merge it cause it goes in your branch, not in master.

PR #202, opened 2021-12-08
Datahub / disable worker in ogc-client if a proxy path is set

jahow:
ogc-client now offers the possibility to disable using a worker altogether.

Also fixes an issue where the data url would be going through the proxy twice for the map data view.


fgravin (approved):
Great you added a way to keep the worker.
You can merge thanks

PR #205, opened 2021-12-10
Rework error handling for data preview in the datahub

jahow:
This PR contains a small scale refactoring on how data is queried for the components in the record view (table, map, download list).

Most of the logic is now handled in a new `DataService` in `feature-record`, which guarantees a uniform handling of errors, proxy, messages, edge cases etc. This was previously handled by the `link-utils` module (part of `feature-search`) which was not a service and, as such, could not use the proxy. This also takes a bit of responsibility off the table & map preview components.

Also:
- errors in map & table were made less intrusive (yellow, no technical error shown)
- more robust CSV parsing & extent computation from GeoJSON; should fix a runtime error that was happening [on this record](https://dev.geo2france.fr/datahub/metadata/n_tri_lill_inondable_s_059)
- no more error shown when queyring downloads from a WFS fails (see https://github.com/geonetwork/geonetwork-ui/pull/205/commits/da9ae2e6884ce9c470f05274a2be1f3e5c6ec645)

image


jahow:
This is ready for review @tkohr @fgravin

libs/feature/record/src/lib/service/data.service.ts
```diff
@@ -0,0 +1,150 @@
...
+      })
+    )
+  }
+
+  getGeoJsonDownloadUrlFromWfs(
+    wfsUrl: string,
+    featureType: string
+  ): Observable<string> {
+    return this.getDownloadUrlsFromWfs(wfsUrl, featureType).pipe(
+      pluck('geojson'),
```
tkohr, line 79: Didn't know this one, but looking in the docs, it looks deprecated. Maybe use `map` here?

libs/feature/record/src/lib/service/data.service.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private getDownloadUrlFromEsriRest(apiUrl: string, format: string): string {
+    return this.proxy.getProxiedUrl(
+      `${apiUrl}/query?f=${format}&where=1=1&outFields=*`
+    )
+  }
+
+  getGeoJsonDownloadUrlFromEsriRest(apiUrl: string): string {
+    return this.getDownloadUrlFromEsriRest(apiUrl, 'json')
```
tkohr, line 95: I think this should be 'geojson'.

tkohr (changes requested):
Nice, big refactoring that takes off data querying logic from the components. Thanks @jahow !

When testing the "test datasets" with the dev proxy, most data querying issues seem fixed (like CSV parsing and GeoJSON extent issue).
I just notice that the ESRI:REST services don't work anymore. I think for the map and table this could be fixed with the change I proposed in a comment, but I think there is another issue to list the downloads.

Examples of servers that should work are:
FeatureServers: fec89c2ec71f4a9ca5f5dd193fd0515d_0, 547c2ffffcb04e8dad2baaafda597989_2
WFSServer: fe1fb1ec84c7493592cab0b88650a2ed_0

libs/feature/record/src/lib/service/data.service.ts
jahow: good catch, thanks!

libs/feature/record/src/lib/service/data.service.ts
jahow: true, thanks

jahow:
> Examples of servers that should work are:
> FeatureServers: fec89c2ec71f4a9ca5f5dd193fd0515d_0, 547c2ffffcb04e8dad2baaafda597989_2
> WFSServer: fe1fb1ec84c7493592cab0b88650a2ed_0

I've updated the PR to fix the issue you identified, now all those records seem to work correctly. Thanks!

PR #206, opened 2021-12-14
datahub: Add default map style based on PRIMARY_COLOR

fgravin:
In the datahub record/preview section, the PR aims to improve the look of the features rendered on the map.
It adds a new service `MapStyleService` which is responsible to manage the styling in Openlayers.
It comes with first method `createDefaultStyle` which replaces the one by default in OL code, and is applied on each vector layer from the map context.
It takes options, as argument, by default it uses the `PRIMARY_COLOR` for the feature color.
It's a bit more beautiful in real than on the screenshot :) (click on the images to see the correct zoom)

Screenshot from 2021-12-14 11-46-51
Screenshot from 2021-12-14 11-46-35

### DIFF
--- a/libs/feature/map/src/lib/map-context/map-context.service.spec.ts
+++ b/libs/feature/map/src/lib/map-context/map-context.service.spec.ts
@@ -6,7 +6,9 @@ import Map from 'ol/Map'
 import TileWMS from 'ol/source/TileWMS'
 import VectorSource from 'ol/source/Vector'
 import XYZ from 'ol/source/XYZ'
+import { Style } from 'ol/style'
 import View from 'ol/View'
+import { MapStyleService } from '../style/map-style.service'
 import {
   MAP_CTX_FIXTURE,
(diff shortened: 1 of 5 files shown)


jahow (approved):
Untested but code looks good, thanks!

Only one remark: the app config might not always be loaded (it is not a mandatory mechanism) so it has to be checked first, like here: https://github.com/geonetwork/geonetwork-ui/blob/master/libs/util/i18n/src/lib/file.translate.loader.ts#L23

Nothing else to add!

jahow:
Also I think for lines the style might look weird (only white lines, no outline), but this can be improved later on

fgravin:
> Only one remark: the app config might not always be loaded (it is not a mandatory mechanism) so it has to be checked first, like here: https://github.com/geonetwork/geonetwork-ui/blob/master/libs/util/i18n/src/lib/file.translate.loader.ts#L23

thanks, fixing this in a next PR

PR #207, opened 2021-12-14
datahub: add a feature info panel in the map view

fgravin:
Bind the map preview with the `FeatureInfoService` and `FeatureInfoComponent`.
Works for WMS and vector data.
Update the feature style with `SECONDARY_COLOR` for the selection.

Screenshot from 2021-12-14 13-54-16


libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
```diff
@@ -111,9 +128,45 @@ export class DataViewMapComponent {
...
+    private changeRef: ChangeDetectorRef,
+    private styleService: MapStyleService
   ) {}
 
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe()
+  }
+
+  ngOnInit(): void {
+    this.selectionStyle = this.styleService.createDefaultStyle({
```
jahow, line 142: why not something like `this.styleService.createHighlightStyle` instead? So that we don't need to read the app config again

jahow (approved):
Looking good! Just one comment, maybe providing a highlight style in the style service would make sense?

libs/feature/record/src/lib/data-view-map/data-view-map.component.ts
fgravin: What do you mean ? a function that takes the default style and change something in it (like the color) ?
I usually do this but i'm not sure it is a use case for the `MapStyleService`.
Also, I see more _highlight_ term for the hover thing, than for the selection.

What would your function do ?

jahow: just generate a style similar to the default one but with larger circles and the secondary color, for example

jahow (approved):
Thanks, let's merge and deploy it live

PR #208, opened 2021-12-15
Do not display current md-record in related records

tkohr:
PR requests four related records from API, filters the currently displayed record from the related records not to be displayed, and displays a maximum of three related records.


jahow (approved):
looking good, thanks!

PR #209, opened 2021-12-15
datahub: Fix infinite scroll

fgravin:
The infinite-scroll was plugged on the HTML top element (document), so scrolling in the record view was triggering the infinite-scroll which was requesting search requests. It ended with tons of records loaded when moving back to the search page, and was causing useless xhrs.

The PR does a refactoring of the Datahub application to separate the 2 pages `search` and `record`.
It adds the container option in the infinite-scroll plugin, and the Datahub passes the container of the search page to it, to prevent useless requests.
Note that the HTML overflow is hidden, and the scroll is done in each page container.


apps/datahub/src/app/search/search-page/search-page.component.html
```diff
@@ -0,0 +1,13 @@
+<header>
+  <datahub-search-header></datahub-search-header>
+</header>
+<main>
```
tkohr, line 4: Might we get rid of these tags like in the `record-page` ?

tkohr (approved):
Nice, this makes a big difference, preventing many useless requests on the record pages. Separating components into search and record was still open to refactoring, thanks @fgravin.

PR #210, opened 2021-12-15
Datahub: Last tuning

fgravin:
- Use CARTO basemaps
- Change record route to` /dataset`


tkohr (commented):
Works fine. Not sure if attribution is not more urgent for this baselayer.

PR #211, opened 2021-12-16
Various fixes

tkohr:
PR fixes

- calling `getDownloadLinksFromEsriRest`
- default height for table when loading

TODOs:

- [ ] fix scroll to top
- [ ] handle 404 error on WFS to prevent displaying empty "Donwloads" on http://localhost:4200/dataset/urn:isogeo:metadata:uuid:1355be63-7f12-41f7-bcc8-724146679eb3
- [ ] close feature-info when clicking related record
- [ ] navigating from record with WMS to search and then other record displays WMS from other record until new context in loaded

### DIFF
--- a/libs/feature/record/src/lib/data-downloads/data-downloads.component.ts
+++ b/libs/feature/record/src/lib/data-downloads/data-downloads.component.ts
@@ -28,8 +28,8 @@ export class DataDownloadsComponent {
     switchMap((links) => {
       const wfsLinks = links.filter((link) => this.linkHelper.isWfsLink(link))
       const esriRestLinks = links
-        .filter(this.linkHelper.isEsriRestFeatureServer)
-        .flatMap(this.dataService.getDownloadLinksFromEsriRest)
+        .filter((link) => this.linkHelper.isEsriRestFeatureServer(link))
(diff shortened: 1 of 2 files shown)


libs/feature/record/src/lib/data-downloads/data-downloads.component.ts
```diff
@@ -28,8 +28,8 @@ export class DataDownloadsComponent {
     switchMap((links) => {
       const wfsLinks = links.filter((link) => this.linkHelper.isWfsLink(link))
       const esriRestLinks = links
-        .filter(this.linkHelper.isEsriRestFeatureServer)
-        .flatMap(this.dataService.getDownloadLinksFromEsriRest)
+        .filter((link) => this.linkHelper.isEsriRestFeatureServer(link))
+        .flatMap((link) => this.dataService.getDownloadLinksFromEsriRest(link))
```
jahow, line 32: Actually the real problem here is that this is not covered by the tests... do you think you can add a test for this? thanks :)

tkohr: I think we have some tests, but they return false positifs. I will try to improve this.

jahow (approved):

PR #212, opened 2021-12-16
datahub: Fix router scroll

fgravin:
The `Router` was configured with the [scrollPositionRestoration](https://angular.io/api/router/ExtraOptions#scrollPositionRestoration) option:
```js
      scrollPositionRestoration: 'enabled',
```
which manages the scroll position on the global viewport, using the [ViewportScroll](https://angular.io/api/common/ViewportScroller) class.

As the scroll is not managed in the viewport anymore, to avoid infinite scroll issues, we must manually intercept the router scrolling event and scroll to top of the record page.
We do what the router option does, but for our own div instead of for the viewport.

### DIFF
--- a/apps/datahub/src/app/app.component.html
+++ b/apps/datahub/src/app/app.component.html
@@ -6,7 +6,11 @@
   >
     <datahub-search-page></datahub-search-page>
   </div>
-  <div class="h-screen overflow-auto" *ngIf="isRecordOpened$ | async">
+  <div
+    id="record-page"
+    class="h-screen overflow-auto"
+    *ngIf="isRecordOpened$ | async"
+  >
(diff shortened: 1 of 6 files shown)


tkohr:
Works well for the navigation to other `related records`, but does not scroll to the top when navigating back to `search`.

fgravin:
> Works well for the navigation to other `related records`, but does not scroll to the top when navigating back to `search`.

yes it's expected.
To me, going back to search, or going backward with the browser should be the same and get back where you were in the search

tkohr:
Ok I agree, it might just be a little strange if you navigate to another (or more) related record in between and then come back to the scroll position of the first record you visited.

PR #214, opened 2022-01-12
Datahub: Document building without docker

tkohr:
PR adds some doc for building datahub app without docker.

I'm not sure what the possibilities are to overload `GN4_API_URL`, `PROXY_PATH` and the entire config file without docker and in how far we should document this as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ GeoNetwork UI provides a standard way of configuring applications using the [con
 This file can be used to:
 
 - customize the URL used to reach the GeoNetwork 4 API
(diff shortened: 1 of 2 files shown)


fgravin (approved):
To me it's enough, maybe @fvanderbiest could approve as well it fits non docker users.

jahow:
> I'm not sure what the possibilities are to overload GN4_API_URL, PROXY_PATH and the entire config file without docker and in how far we should document this as well.

This only works with the Docker build. I think this documentation is fine, thanks @tkohr :)

PR #216, opened 2022-01-26
Datafeeder: Fix translations

tkohr:
PR adapts datafeeder build config to use reworked translations. Sets language the same way the datahub does, using `en` as default.


apps/datafeeder/src/app/app.module.ts
```diff
@@ -86,7 +86,6 @@ export function apiConfigurationFactory() {
 })
 export class AppModule {
   constructor(translate: TranslateService) {
-    document.documentElement.lang = getLangFromBrowser()
```
fgravin, line 89: wonder if this line is usefull though
could you check please ?

angular.json
```diff
@@ -64,7 +64,12 @@
             "aot": true,
             "assets": [
               "apps/datafeeder/src/favicon.ico",
-              "apps/datafeeder/src/assets"
+              "apps/datafeeder/src/assets",
+              {
+                "glob": "*",
+                "input": "translations",
+                "output": "assets/i18n/"
+              }
```
jahow, line 72: Sounds great! But the translations are still part of the assets right? https://github.com/geonetwork/geonetwork-ui/tree/6e0ebc5f8e75c008995e921688a7f14c11881d08/apps/datafeeder/src/assets/i18n

Shouldn't they be removed? See https://github.com/geonetwork/geonetwork-ui/pull/170 for inspiration :)

PR #217, opened 2022-02-01
Exclude templates and subtemplates from search

fgravin:
YAGNI implementation to remove the templates and subtemplates from the search.
It will always be the case for the searches we have at the moment, so i don't put it in the state yet.
I think the `esService` will need some refactoring but it's a bit too soon, I think what could be good is too have functions which merge all of the pieces of the payload together to build the query.

I think that having something like `isTemplate` as JSON `bool` ES param is better than to mix them in the `query_string`.
Open to discussion.

Thanks

### DIFF
--- a/libs/util/shared/src/lib/models/elasticsearch.model.ts
+++ b/libs/util/shared/src/lib/models/elasticsearch.model.ts
@@ -32,3 +32,6 @@ export interface EsSearchParams {
 }
 
 export type EsSearchResponse = any
+
+export type EsTemplateValues = 'y' | 'n' | 's'
+export type EsTemplateType = EsTemplateValues | EsTemplateValues[]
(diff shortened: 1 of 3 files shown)


libs/util/shared/src/lib/models/elasticsearch.model.ts
```diff
@@ -32,3 +32,6 @@ export interface EsSearchParams {
 }
 
 export type EsSearchResponse = any
+
+export type EsTemplateValues = 'y' | 'n' | 's'
```
jahow, line 36: you're missing the `'t'` :sob:

jahow (approved):
Looking good, but indeed we'll need to refactor the ES service to have something more manageable. Also there are libraries out there to help build ES payloads, we should look into it IMO.

fgravin:
Thanks for the review @jahow, I missed the template of subtemplate damn !
Yes clearly if can discuss about this service soon.

PR #218, opened 2022-02-01
Multilingual search

fgravin:
Add multilingual full text search in geonetwork-ui, following up the PR in core https://github.com/geonetwork/core-geonetwork/pull/6145

Search and suggestion are built uppon a pattern which contains a placeholder for the lang. Either we search in a single language eg. `any.langfre` or in all language eg. `any.\\*`

The search is also boosted on the title eg. `resourceTitleObject.${searchLang}:(${any})^2`

I don't really know if we should rely on the GN4 ui config or not for the search configuration (the search pattern in this case).


jahow (approved):
Looking good, thanks (untested)

jahow:
I've rebased this on master and made some adaptations to make it work with the changes made by @tkohr. I think we might try to rethink this a bit more before merging though.

PR #220, opened 2022-02-08
Search Router - Add full text search param in URL

fgravin:
This PR setup the router architecture and configuration to be able to add search params in the URL.
- eg. `http://datahub.fr/search?q=scot`

It makes the router works along with  `<router-outlet></router-outlet>`.
It means that the router is responsible to instantiate the components binded to each route (while it was done manually before).

### Configuration
To keep the router generic, the routes are generated dynamically from a configuration injected during `RouterModule` creation

```
DefaultRouterModule.forRoot({
  searchStateId: 'mainSearch',
  searchRouteComponent: SearchPageComponent,
  recordRouteComponent: RecordPageComponent,
})
```
The routes are generated in the `RouterInitService`, from the `DefaultRouterModule` constructor.

Unfortunately, it appears that the router does not work well along ReduxDevTools, Nrwl operators, and dev mode options (freeze), so I had to remove some dev checks.

### Reuse Strategy
A Router ReuseStrategy is added to alway keep the searchPage component alive (usually the router destroy and create again the component on route change), so the search is not launched again on search route navigation.
Therefore, the infinite-scroll container option is kept cause the searchPage component body still exists in background.

### search route dispatch
On page reloading, the `q` url params is taken from the router state and passed as `initialValue` to the autocomplete.
The autocomplete now emits the initial value as first emission of the submitted event, if you don't want to catch it, you can ignore it with the rxJs `skip()` operator.
This way, the datahub does not trigger a search manually on page loading, the autocomplete is responsible for it and it does it with the `q` parameter of the router state, if any.

When navigating to the record page and coming back to the search page, the `q` parameter is restored.

### Autocomplete improvments
When selecting a suggestion in the autocomplete list, if `resetOnSelection` is set, the normal behavior of Material Autocomplete is overriden, we set the last text input value in the full text input
eg: if you type `first` then the autocomplete show off `first season` and you click on `first season`, then the input will contain `first` instead of `first season`. The suggestion selection redirects to the record page, this way when you come back to the search page you have the same value in the input that you typed just before.

- [x] Enable `router-outlet`
- [x] Make routes configurable
- [x] Keep search page present in the page even though it is not the active route
- [x] Set full text input in search route state
- [x] Init autocomplete input from router state
- [x] Restore search route when back from record
- [x] Refactor and improve autocomplete behavior  

---------------
**Notes**
I think we can go further and create an effect to bind route and search states. 
`RouterEffect`
```ts
  searchFromRoute$ = createEffect(() =>
    this._facade.anySearch$.pipe(
      map(
        (any: string) =>
          new SetFilters({ any }, this.routerConfig.searchStateId)
      )
    )
  )
```

We would 
- get rid of the router state listening in the datahub `SearchPageComponent`
-  listen the search state `any` from the `FuzzySearchComponent`.
- pass it as `Autocomplete.initialValue` (even as value and the autocomplete input would be binded with the `searchState.filters.any` field)

`FuzzySearchComponent`
```ts
  textInput$ = this.searchFacade.searchFilters$.pipe(
    filter((filters) => !!filters),
    pluck('any'),
    take(1),
    map((any) => ({ title: any }))
  )
```

The problem here is that we should not trigger a search on route change, but also checks that the search filter state has not changed, otherwise it triggers a search when coming back from the record page.
Also, we will need to handle other params in the router state url (eg. keywords), so I postpone the thinking for later.


fgravin:
Thanks for the review @jahow

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -8,7 +7,10 @@ import {
   ViewChild,
 } from '@angular/core'
 import { SearchApiService } from '@geonetwork-ui/data-access/gn4'
-import { AutocompleteComponent } from '@geonetwork-ui/ui/inputs'
+import {
+  AutcompleteItem,
```
tkohr, line 11: Just had a quick look over the big PR, seeing one typo here.

fgravin: Good catch, it was introduced in a previous PR, I will fix it thanks

PR #222, opened 2022-02-11
Enable to configure map from md record

tkohr:
The goal of this PR is to enable customizing the map with the following possibilities:

- [x] limit map's max zoom
- [x] limit map's navigation extent

Adding the following custom layers will be implemented in a different PR:
- configure a base layer
- add an additional layer (WMS / WMTS / TMS)
- add semi transarent vector layer

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -12,9 +12,12 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 proxy_path = ""
 
 ### MAP SETTINGS
-[map]
-max_zoom = 10
-max_extent = [-418263.418776, 5251529.591305, 961272.067714, 6706890.609855]
+# The map section allows to override default settings of the map view.
+# Example:
+#
+# [map]
+# max_zoom = 10
+# max_extent = [-418263.418776, 5251529.591305, 961272.067714, 6706890.609855]
 
 ### VISUAL THEME
 


libs/feature/map/src/lib/map-context/component/map-context.component.ts
```diff
@@ -41,25 +41,17 @@ export class MapContextComponent implements OnChanges {
...
     if (this.context) {
-      if (this.context.extent) {
-        if (!this.map.getSize()) {
-          this.map.once('change:size', () => {
-            this.updateContextByExtent()
-          })
-        } else {
-          this.updateContextByExtent()
-        }
+      if (this.context.extent && !this.map.getSize()) {
```
fgravin, line 44: `extent` should go in the view

libs/feature/map/src/lib/map-context/map-context.service.ts
```diff
@@ -37,7 +36,7 @@ export class MapContextService {
       mapContext = this.mergeMapConfigWithContext(mapContext, mapConfig)
     }
     if (mapContext.view) {
-      map.setView(this.createView(mapContext.view))
+      map.setView(this.createView(mapContext, map))
```
fgravin, line 39: this change should have told you that the extent was not in the correct model :)

fgravin (commented):
Good start, I think the approach is good.
Note @jahow that I told @tkohr that I was ok if the config is passed to anywhere in the application (all modules will have a dep on the config). I think it is the typical kind of object that should be injected everywhere, if there is not config there must be a fallback. Otherwise, if you can only load the config from the application, you will have to pass it deep in the DOM tree which is open to discussion.

libs/feature/map/src/lib/map-context/component/map-context.component.ts
tkohr: I agree that the `view` appears fitting to include the `extent`. I was hesitating doing this, as it will mean to make `zoom` and `center` optional in our view model, which are mandatory to create an `ol.view`. I know, we will calculate them from the `extent` in time, but it seemed somehow misleading at first glance. Will see to adapt this.

fgravin:
@tkohr I would like not to have a huge PR, maybe we can merge the first setting to have already the logic implemented

libs/feature/map/src/lib/map-context/component/map-context.component.ts
```diff
@@ -38,26 +40,18 @@ export class MapContextComponent implements OnChanges {
...
-    if (this.context) {
-      if (this.context.extent) {
-        if (!this.map.getSize()) {
-          this.map.once('change:size', () => {
-            this.updateContextByExtent()
-          })
-        } else {
-          this.updateContextByExtent()
-        }
+    if (this.context && this.context.view) {
```
fgravin, line 43:
```suggestion
    if (this.context?.view) {
```

fgravin (approved):
Very good thank you.
One minor suggestion then I merge it.

libs/feature/map/src/lib/map-context/component/map-context.component.ts
tkohr: Yep, I adapted that

libs/util/app-config/src/lib/app-config.ts
```diff
@@ -70,14 +82,14 @@ export function loadAppConfig() {
     .then((conf) => {
       let parsed
       try {
-        parsed = TOML.parse(conf, { joiner: '\n' }) as any
+        parsed = TOML.parse(conf, { joiner: '\n', bigint: false }) as any
```
fgravin, line 85: why did you need this ?

tkohr: Otherwise, I got `max_zoom = 10` in config file parsed as `max_zoom: 10n`

PR #227, opened 2022-02-15
Enable configuring custom layers

tkohr:
The goal of this PR is to enable adding layers to the map view to allow :

- configuring a base layer
- adding another layer (WMS / WMTS / TMS)
- adding a semi transparent vector layer

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -15,9 +15,27 @@ proxy_path = ""
 # The map section allows to override default settings of the map view.
 # Example:
 #
-# [map]
+[map]
 # max_zoom = 10
 # max_extent = [-418263.418776, 5251529.591305, 961272.067714, 6706890.609855]
+# use_basemap_from_layers = false
+# 
+# Layers (basemaps, overlays, masks) can be added to the map view with three properties:
+# - type (mandatory): Indicates the layer type. Possible values are 'xyz', 'wms', 'wfs'.
+# - url (mandatory): Layer endpoint URL.
+# - name: Mandatory for 'wms', 'wfs' types where it indicates the layer name.
+# Layer order in the config is the same as in the map, the first one being the bottom one.
+# If you want to override the default basemap with the first layer of your config set use_basemap_from_layers = true (default: false)
+# Each layer is defined in its own [[layers]] section.
+# Example:
+# [[layers]]
+# type = "xyz"
+# url = "https://{a-c}.basemaps.cartocdn.com/light_all/{z}/{x}/{y}.png"
+# [[layers]]
+# type = "wfs"
+# url = "https://www.geo2france.fr/geoserver/cr_hdf/ows"
+# name = "masque_hdf_ign_carto_latin1"
+#
 
 ### VISUAL THEME
 
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -12,30 +12,34 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 proxy_path = ""
 
 ### MAP SETTINGS
-# The map section allows to override default settings of the map view.
-# Example:
-#
+
+# The map section allows to customize how maps are configured.
 [map]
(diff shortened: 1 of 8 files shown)


fgravin (changes requested):
Thanks for this draft, I think the approach should be reworked, at a first sight we see the complexity of `DataViewMap` component growing for no reason.
Try not to stick too much to the code you find before you start your work, eg you keep the `getBackgroundLayer()` behavior and the way the array of layers is built and the way the map is built while it appears clearly that the needs were different in that timen don't be afraid to refactor for the current needs.

Your code does not guarantee that a config with a background layer or custom layer passed to the `map-context` will be applied, cause you transform the config at 2 different places.

I think the `data-view-map-component` could not change much, pass the `mapContext` and the `config` to the `MapContextComponent` as usual, and it should be in the `mapContextService`, that you merge the background and custom layers in the resulting context. It means you must not add a default background layer in the context, if you need one, you can add a default one in a config object if no config or no config-background is given.

You should mutualise a lot of code between background and custom layers.

We could just add a `layers` section in the config with z-index for the order.

Did you think how you will handle the mask overlay option ? 

Let me know if it makes sense for you.
Thanks !

tkohr:
Thanks for the detailed feedback @fgravin ! I've reworked the approach as discussed, hoping it looks better now:)

The arrays don't feel ideal in TOML, but it provides a large flexibility for the customlayers.

Not sure if I should still add a transparency param, as vector layers are displayed with transparency by default.

conf/default.toml
```diff
@@ -18,6 +18,26 @@ proxy_path = ""
 # [map]
 # max_zoom = 10
 # max_extent = [-418263.418776, 5251529.591305, 961272.067714, 6706890.609855]
+#
+# Baselayer: The default baselayer can be overridden with three properties:
+#   type: Indicates the baselayer type. Possible values are 'xyz', 'wms', 'wfs'. Mandatory if you configure a baselayer.
+#   url: Baselayer endpoint URL. Mandatory if you configure a baselayer.
+#   name: Baselayer name. Mandatory for 'wms', 'wfs' types where it indicates the layer name.
+# baselayer.type = "xyz"
```
jahow, line 26: I think `baselayer_type` would make more sense, and would be consistent with other keys

conf/default.toml
```diff
@@ -18,6 +18,26 @@ proxy_path = ""
...
+# baselayer.url = "https://{a-c}.basemaps.cartocdn.com/rastertiles/voyager/{z}/{x}/{y}.png"
+# baselayer.name = "some-layername"
+#
+# Customlayer: Besides the baselayer, addtional customlayers can be configured with the same three properties.
```
jahow, line 30: What's not clear here is the difference between custom layers and the base layer. To me it feels like the same thing. How about simply allowing users to specify one or several default map layers, and the default config pointing to the basemap we're currently using? This is probably something you discussed cc @fgravin

jahow: Just to be clear, the config schema is extremely important because we won't be able to change it later on except by providing a migration system, so getting the schema right is really critical IMO.

conf/default.toml
tkohr: Ok, I had this reflex, as it will be in the same JS object later, but indeed can adapt this to stay consistent with other configs.

conf/default.toml
tkohr: Yes, I was thinking about simply defining the baselayer as a customlayer. It would have the drawback that we cannot check in the app anymore if a baselayer is defined in order to set a default one. What we could do in this case is to set a default layer in the app if no customlayer is defined, but this didn't feel ideal to me when the admin only wants to define a data or mask layer.
Then again, we could also always define the baselayer as the first customlayer in the config, but this would make this section of the config mandatory for the app.
Also, the separation has the advantage that the admin does not have to worry about the layer order for the baselayer, which will always be placed at the bottom.

jahow (approved):
Thanks!

PR #229, opened 2022-02-18
Improve search by setting priorities on search fields

tkohr:
PR weighs the searched fields in the following order:

- title
- keywords
- abstract
- lineage
- (any)

### DIFF
--- a/libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
+++ b/libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core'
 import { Observable } from 'rxjs'
 import { map, take } from 'rxjs/operators'
+import { ES_QUERY_STRING_FIELDS } from './constant'
 import {
   EsSearchParams,
   EsTemplateType,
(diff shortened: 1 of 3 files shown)


libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core'
 import { Observable } from 'rxjs'
 import { map, take } from 'rxjs/operators'
+import { ES_QUERY_STRING_FIELDS } from '.'
```
jahow, line 4:
```suggestion
import { ES_QUERY_STRING_FIELDS } from './constants'
```

fgravin:
This should work in multilingual ?
Should be based on https://github.com/geonetwork/geonetwork-ui/pull/218

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -114,7 +115,15 @@ export class ElasticsearchService {
 
     return {
       bool: {
-        must: [{ query_string: { query } }, this.addTemplateClause('n')],
+        must: [
```
fgravin, line 118: what about the suggestions ?

jahow: the autocomplete query is currently taken from the GN4 ui config; since there may not be a query string, I don't know how we can change the fields weight there

PR #230, opened 2022-02-21
Datahub / Implement sticky headers for search and record view

jahow:
This PR introduces a `gn-ui-sticky-header` presentation component which allows making a sticky header with varying height depending on the current scroll.

Both the search and record headers have been reworked to rely on this.

headers

Note that currently the performance isn't great and I feel like this will give a mixed user experience. I have spent quite a while trying to optimize this and have a few more ideas, but I think this can be done in another PR.

Also this PR introduces a new `header_background` setting which allows customizing the header background.

Remaining to do:
- [x] restore the logic for disabling buttons in the record nav bar (since the nav bar is now part of the header component)
- [ ] rework app config tests


tailwind.base.config.js
```diff
@@ -66,12 +66,6 @@ module.exports = {
           },
         ],
         28: '28px',
-        48: [
-          '48px',
-          {
-            lineHeight: '58px',
-          },
-        ],
```
tkohr, line 74: Wondering if we shouldn't prevent mixing the use of pre-generated and JIT classes for font sizes. Maybe we should set `text-13` and `text-21` via JIT as well if this is the preferred way!?

libs/ui/layout/src/lib/ui-layout.module.ts
```diff
@@ -26,12 +28,16 @@ import { StickyHeaderComponent } from './sticky-header/sticky-header.component'
     FigureComponent,
     ExpandablePanelComponent,
     StickyHeaderComponent,
+    // AnchorLinkComponent,
```
tkohr, line 31: Might be removed, I guess.

tkohr (approved):
Thanks, @jahow ! I was a little surprised by the complexity of the PR for visual interactions, but as far as I can see the code LGTM and reusable. The result looks really nice, especially on chromium. Seems a little less fluid on firefox when testing.

tailwind.base.config.js
jahow: yes, absolutely agree, these preset classes do not make sense now that we have JIT mode for tailwind. I didn't do it here since I feared the PR would get too big. I guess we should do this progressively with other PRs.

jahow:
thanks for the review @tkohr :)

fgravin:
Thanks it's really pleasant to see the smooth animation.
However, it seems the home header is way higher than before when expanded, too high IMO.

jahow:
I'm going to check again the height from the mockup :)

PR #231, opened 2022-02-23
Refactor app-config utils

jahow:
This is an attempt to simplify the code used for parsing the app configuration file and make it more readable. Hopefully this will improve the app-config code maintanability.

### DIFF
--- a/libs/util/app-config/src/lib/app-config.ts
+++ b/libs/util/app-config/src/lib/app-config.ts
@@ -1,4 +1,9 @@
 import * as TOML from '@ltd/j-toml'
+import {
+  parseConfigSection,
+  parseMultiConfigSection,
+  parseTranslationsConfigSection,
+} from './parse-utils'
 
 const MISSING_CONFIG_ERROR = `Application configuration was not initialized correctly.
 This error might show up in case of an invalid/malformed configuration file. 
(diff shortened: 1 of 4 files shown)


libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+): Record<string, unknown>[] | null {
+  const sections = fullConfigObj[sectionName]
+  if (!sections) {
+    return []
+  }
+  if (!Array.isArray(sections)) {
+    outErrors.push(`Expected [[${sectionName}]] to be an array of sections.`)
+    return null
+  }
+  const result = sections.map((section) =>
```
tkohr, line 109: Nice, thanks for adding this and checking the layer properties as well!

libs/util/app-config/src/lib/app-config.ts
```diff
@@ -95,150 +84,99 @@ export function loadAppConfig() {
...
-        BACKGROUND_COLOR: theme.background_color,
-        HEADER_BACKGROUND: theme.header_background,
-        TITLE_FONT: theme.title_font,
-        MAIN_FONT: theme.main_font,
-        FONTS_STYLESHEET_URL: theme.fonts_stylesheet_url,
       }
-      customTranslations = translations || {}
 
       appConfigLoaded = true
     })
```
tkohr, line 182: Do we actually still need the `_reset()` function below ?

tkohr (approved):
LGTM, thanks @jahow !

libs/util/app-config/src/lib/app-config.ts
jahow: I think so, for tests in `app-config.spec`

PR #232, opened 2022-02-24
Datahub: Add a button to open layers in external viewer

tkohr:
PR adds a button to the map that allows to open the currently selected layer in an external viewer. Currently, supports `WMS` and `WFS`. Tested with mapstore where both work fine and mapfishapp, which did not open the `WFS`s I've tried. Not sure if this is supported by mapfishapp.

TODO
- [x] fix extract i18n

externalviewer


jahow (approved):
Thanks! looking good!

PR #233, opened 2022-02-24
Fix i18n extract

tkohr:
PR temporarily installs another fork of `ngx-translate-export` to solve Angular 13 compatibility.

Tested running `npm run i18n:extract` with translations in HTML templates and marker translations.

see issue/PR:
- https://github.com/biesbjerg/ngx-translate-extract/issues/246
- https://github.com/biesbjerg/ngx-translate-extract/pull/247
- (https://github.com/angular/angular/issues/43833#issuecomment-980568711)

Note that prettier HTML formatting has changed with version [2.5.0](https://prettier.io/blog/2021/11/25/2.5.0.html#collapse-html-class-attributes-onto-one-line-11827httpsgithubcomprettierprettierpull11827-by-jlongsterhttpsgithubcomjlongster), which explains the HTML modifications in this PR.

### DIFF
--- a/apps/datahub/src/app/record/header-badge-button/header-badge-button.component.html
+++ b/apps/datahub/src/app/record/header-badge-button/header-badge-button.component.html
@@ -1,15 +1,5 @@
 <button
-  class="
-    items-center
-    justify-center
-    px-2
-    py-1
-    bg-primary
-    text-white
-    rounded
-    opacity-25
-    capitalize
-  "
+  class="items-center justify-center px-2 py-1 bg-primary text-white rounded opacity-25 capitalize"
   (click)="onClick()"
 >
   <mat-icon *ngIf="icon" class="align-middle">{{ icon }}</mat-icon>
--- a/libs/feature/record/src/lib/data-view-map/data-view-map.component.html
+++ b/libs/feature/record/src/lib/data-view-map/data-view-map.component.html
@@ -1,28 +1,12 @@
 <div
-  class="
-    relative
-    w-full
-    h-full
-    bg-white
-    border border-gray-500
-    rounded-lg
-    overflow-hidden
-  "
+  class="relative w-full h-full bg-white border border-gray-500 rounded-lg overflow-hidden"
 >
   <gn-ui-map-context
     [context]="mapContext$ | async"
     [mapConfig]="mapConfig"
   ></gn-ui-map-context>
   <div
-    class="
-      feature-info
-      p-3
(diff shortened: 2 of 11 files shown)


jahow (approved):
good enough, hopefully we can go back to the original package soon

apps/datahub/src/app/record/header-badge-button/header-badge-button.component.html
```diff
@@ -1,15 +1,5 @@
...
-    justify-center
-    px-2
-    py-1
-    bg-primary
-    text-white
-    rounded
-    opacity-25
-    capitalize
-  "
+  class="items-center justify-center px-2 py-1 bg-primary text-white rounded opacity-25 capitalize"
```
fgravin, line 2: I don't understand why we have such changes...
Prettier or `npm run format` should guarantee to always have the correct indentation.

tkohr: Prettier has changed its strategy for html formatting with version 2.5 (https://prettier.io/blog/2021/11/25/2.5.0.html#collapse-html-class-attributes-onto-one-line-11827httpsgithubcomprettierprettierpull11827-by-jlongsterhttpsgithubcomjlongster), and does not wrap lines > `printWidth` anymore. I didn't see another param to conserve the old behavior, which might not make sense anyway, since this evolution.

fgravin: Yes but i wonder why all those changes were not automatic and mandatory for the CI when angular and prettier were updated.

jahow: That's because prettier is only run on `affected` libraries and apps, so the new formatting strategy has not been applied everywhere but will be gradually. I guess it would be preferable to have all HTML reformatted correctly to avoid confusing diffs in the future.

fgravin:
yes, at some point we should run `test`, `lint` and `format` on all sources to avoid having broken stuff while didn't change anything about them.
We should add npm scripts for that IMO.

apps/datahub/src/app/record/header-badge-button/header-badge-button.component.html
fgravin: yes, at some point we should run test, lint and format on all sources to avoid having broken stuff while didn't change anything about them.
We should add npm scripts for that IMO.

PR #234, opened 2022-02-25
Improve map and table dropdown labels

tkohr:
The goal of this PR is to unify map and table dropdown labels and to make them more precise. There might be more refactoring necessary in the long term (like moving the `link.utils` somewhere else). I tried to simplify them for now, looking for a compromise between refactoring and implementing the demanded functionality.

TODOs
- [ ] tests
- [ ] translate 'unkown' fallback


libs/feature/search/src/lib/utils/links/link-utils.ts
```diff
@@ -1,67 +1,74 @@
...
-        case 'text/csv':
-        case 'application/csv':
-          return 'csv'
-        case 'x-gis/x-shapefile':
-          return 'shp'
-        case 'application/vnd.ms-excel':
-        case 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet':
-          return 'excel'
-      }
+      const mimeType = checkMimeType(matches[1])
```
fgravin, line 37:
```suggestion
      return checkMimeType(matches[1])
```

libs/feature/search/src/lib/utils/links/link-helper.service.ts
```diff
@@ -78,4 +79,13 @@ export class LinkHelperService {
   hasProtocolDownload(link: MetadataLinkValid): boolean {
     return /^WWW:DOWNLOAD/.test(link.protocol)
   }
+
+  getLinkLabel(link: MetadataLinkValid): string {
+    let source = getFileFormat(link)
+    if (this.isWmsLink(link)) source = 'WMS'
```
fgravin, line 85: `else`
It's not clear what have the priority to what.

libs/util/shared/src/lib/models/search.model.ts
```diff
@@ -46,6 +46,7 @@ export interface MetadataLinkValid {
   protocol?: string
   format?: string
   description?: string
+  label?: string
```
fgravin, line 49: why not

libs/feature/search/src/lib/utils/links/link-utils.ts
```diff
@@ -1,67 +1,74 @@
...
-  type: DownloadFormatType
-): string {
+export function getWfsFormat(link: MetadataLinkValid): string {
+  for (const format in FORMATS) {
+    for (const alias of FORMATS[format]) {
+      if ('format' in link && new RegExp(`${alias}`, 'i').test(link.format))
+        return `WFS:${format}`
+    }
+  }
+  return undefined
```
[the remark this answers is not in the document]
tkohr: Ok, I see, but I might be missing something here. If I remove the `undefined` return values, I end up with `TS7030: Not all code paths return a value.` errors.

libs/feature/search/src/lib/utils/links/link-utils.ts
tkohr: Ah thanks, not seeing the forest because of all the trees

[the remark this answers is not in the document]
tkohr: Good point, I'll adapt

libs/util/shared/src/lib/models/search.model.ts
tkohr: Should not occur in practice I guess, but I put it optional as `name` and `description` are optional, as well

libs/feature/search/src/lib/utils/links/link-helper.service.ts
tkohr: ok, I'll adapt

libs/feature/search/src/lib/utils/links/link-utils.ts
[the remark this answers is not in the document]
fgravin: Ok so maybe you need to keep it.
Would it change something to change the returned type of the function to `string | void` or `string | undefined` ?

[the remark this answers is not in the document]
tkohr: Ah sure, `string | void` makes sense and works well :+1:

tkohr:
Thanks for the review @fgravin. Hope now it looks better to merge.

fgravin (approved):
Thanks for last changes

PR #235, opened 2022-02-25
Layout / improve performance of sticky header animation

jahow:
* Change size of default header background  
  Using `cover` means the background image gets resized according to the header size, which makes the browser spend much more time in the layout/painting phase.

* Removed the height transition which does not really change anything at all.

* Avoid changing `height` of header, change `transform` instead; this will avoid reflows of the layout

Performance is probably in the range of acceptable now.

### DIFF
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -5,7 +5,13 @@
     <div
       *ngIf="metadata"
       class="font-title text-center px-2 mx-48 text-[28px] mb-[41px]"
-      [style.transform]="'scale(' + (0.85 + 0.15 * expandRatio) + ')'"
+      [style.transform]="
+        'scale(' +
+        (0.85 + 0.15 * expandRatio) +
+        ') translate(0, ' +
+        (1 - expandRatio) * 70 +
+        'px)'
+      "
     >
       {{ metadata.title }}
     </div>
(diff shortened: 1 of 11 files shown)


fgravin:
I noticed a bug with the dataset sticky header.
When you click on an item (eg: preview): it scrolls to the top of the section which is below the header, we need an offset.

jahow:
> I noticed a bug with the dataset sticky header. When you click on an item (eg: preview): it scrolls to the top of the section which is below the header, we need an offset.

Yes true, and it's going to be tricky because the header height can change so the offset is not constant...

fgravin:
Here the text overflow is not well managed.
Overall, Lucile mockup just mentionned that the dataset header should be sticky, but not be resized, I prefer this option.
The scroll should just be in the content, above the header. It will fix the scroll to section issue as well i guess.

fgravin:
Hello,
Sorry i have again another UI issue with the header of the dataset.
When opening a dataset from the search page, the background of the dataset header slightly flickers on opening.
It appears like a small glitch from the user perspective IMO.

jahow:
Updated, the header of the record view is not sticky any more, only the navbar:
record-view

Also now the title correctly scrolls into view (thanks `scroll-padding` CSS property).

jahow:
> Did you fix the header background flickering issue ?

See my previous comment, I can't fix it from the application code. This has to be done on the server.

fgravin:
> > Did you fix the header background flickering issue ?
> 
> See my previous comment, I can't fix it from the application code. This has to be done on the server.

What is the reason about that ? Is it about the sticky things or about the head background configuration setting ?
There was no issue before those changes.
Thanks

jahow:
> > > Did you fix the header background flickering issue ?
> > 
> > 
> > See my previous comment, I can't fix it from the application code. This has to be done on the server.
> 
> What is the reason about that ? Is it about the sticky things or about the head background configuration setting ? There was no issue before those changes. Thanks

The issue is definitely present on master. A fix in the application code might be to make sure that headers both for the search and record view are not destroyed when the route changes, but it feels more like a workaround.

PR #236, opened 2022-03-01
Fix External Viewer: Uri encode service url

tkohr:
PR fixes opening layers in external viewer, that contain query params. The layer's service is uri encoded, before it is passed it to the external viewer. Currently, service urls containing `request=getCapabilities` do not allow mapstore to add the layer for example. Can be tested with: https://dev.geo2france.fr/datahub/dataset/35d867a1-9420-449c-b0c4-442e7662eca6


fgravin (approved):
LG thanks

PR #237, opened 2022-03-01
Fix scolling for navigation

tkohr:
PR recovers the div id that allows to scroll to top when navigating to related record.

TODO:

- [ ] do not scroll to top when returning to search


fgravin:
> * [ ]  do not scroll to top when returning to search

I saw this behavior while refactoring the router but I considered it as an acceptable side effect.
We can discuss with them or lucile to check exactly what is the expected behavior and see if we can do something or not. Cause I am not exactly sure how the router keeps the rendered component while not on the matching route.

fgravin (approved):

PR #238, opened 2022-03-03
Datahub: Display catalog in search results and md record

tkohr:
The PR displays the catalog of the MD origin in the search results and in the metadata record. The PR is WIP, as it currently takes the catalog's UUID for demonstration/implementation. This will need to be adapted in the field mapper, once the catalog's name will be part of GN's index.

TODOs:

- [x] replace catalog's UUID by its name in field mapper or fetch catalog names from API
- [x] rework tests
- [x] add tests for source-label.component and service

search_cat
record_cat


fgravin:
2 options to get the translations:

- source API https://dev.geo2france.fr/geonetwork/srv/api/sources
- i18n api https://dev.geo2france.fr/geonetwork/srv/api/i18n/packages/search in `source-*`

libs/feature/record/src/lib/state/mdview.facade.ts
```diff
@@ -69,7 +74,11 @@ export class MdViewFacade {
     map((links) => links.filter((link) => this.helper.isOtherLink(link)))
   )
 
-  constructor(private store: Store, private helper: LinkHelperService) {}
+  constructor(
+    private store: Store,
+    private helper: LinkHelperService,
+    private sources: SourcesService
```
fgravin, line 80: `sourcesService`

libs/ui/search/src/lib/record-preview/record-preview.component.ts
```diff
@@ -27,9 +29,13 @@ export class RecordPreviewComponent implements OnInit, OnDestroy {
     return this.record.hasDownloads
   }
 
-  constructor(protected elementRef: ElementRef) {}
+  constructor(
+    protected elementRef: ElementRef,
+    protected sources: SourcesService
```
fgravin, line 34: `sourcesService`

libs/ui/elements/src/lib/ui-elements.module.ts
```diff
@@ -26,6 +27,7 @@ import { UiSearchModule } from '@geonetwork-ui/ui/search'
     UtilSharedModule,
     UiSearchModule,
     RouterModule,
+    FeatureCatalogModule,
```
fgravin, line 30: no `ui` should not depend on `feature`
put the model elsewhere if it's what you need.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -97,12 +97,12 @@
...
       </p>
     </div>
-    <div *ngIf="metadata.catalog">
+    <div *ngIf="source">
       <p class="text-gray-600 text-xs mb-3 uppercase" translate>
         record.metadata.catalog
       </p>
       <p class="text-primary font-title text-21 mb-1">
-        {{ metadata.catalog }}
+        {{ source.name }}
```
fgravin, line 105: what is `source.name`, it is not about the `label` ? with the translation ?

libs/ui/search/src/lib/ui-search.module.ts
```diff
@@ -47,6 +48,7 @@ import { UiWidgetsModule } from '@geonetwork-ui/ui/widgets'
     TagInputModule,
     UtilSharedModule,
     UiWidgetsModule,
+    FeatureCatalogModule,
```
fgravin, line 51: i d'ont think `search` depends on `catalog`, but `records` does.

fgravin (changes requested):
Quite good but the dependencies directions are not ok.
Maybe the simplest would be to have a smart component that just translates the uuid to the label, which you can use in the record preview and the record page.

libs/ui/elements/src/lib/ui-elements.module.ts
tkohr: Yes, I see, removed this

libs/ui/search/src/lib/ui-search.module.ts
tkohr: Yep, also removed this

libs/ui/search/src/lib/record-preview/record-preview.component.ts
```diff
@@ -19,6 +20,7 @@ export class RecordPreviewComponent implements OnInit, OnDestroy {
   @Input() linkTarget = '_blank'
   @Output() mdSelect = new EventEmitter<MetadataRecord>()
   subscription = new Subscription()
+  source$: Observable<CatalogSource>
```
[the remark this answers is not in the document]
tkohr: Yes, thanks for the review, this is what I did now:)

tkohr:
> Quite good but the dependencies directions are not ok. Maybe the simplest would be to have a smart component that just translates the uuid to the label, which you can use in the record preview and the record page.

Thanks for the hints @fgravin. I've reworked the approach as you suggested. Still need to rework the tests. Don't hesitate if you have further comments.

One thing I've noticed is that the share of the observable works fine for multiple instances of the same component (`RecordPreviewComponent` and `RecordMetadataComponent` respectively) that use the `SourceLabelComponent`, but not among the two, when we navigate from the search to one entry.

libs/feature/catalog/src/lib/source-label/source-label.component.ts
```diff
@@ -0,0 +1,32 @@
...
+})
+export class SourceLabelComponent {
+  sourceLabel$: Observable<CatalogSource>
+  @Input() set catalogUuid(uuid: string) {
+    this.sourceLabel$ = this.sourcesService
+      .getSource(uuid)
+      .pipe(
+        map(
+          (source) =>
+            source.label[LANG_2_TO_3_MAPPER[this.translateService.currentLang]]
```
fgravin, line 23: Can you cherry-pick https://github.com/geonetwork/geonetwork-ui/pull/218/commits/b731b0aebb3f29fe5dd717bb281fc5845f040860 from multilingual PR and use this service instead plz, thanks

libs/feature/catalog/src/lib/sources/sources.model.ts
```diff
@@ -0,0 +1,26 @@
+export interface CatalogSource {
+  label: CatalogSourceLabel
+  name: string
+  type: string
+  uuid: string
+}
+
+export interface CatalogSourceLabel {
```
fgravin, line 8: Bof
You should have somewhere a type like `export type Iso3Langs = 'fre' | 'eng' |` ...
Then `label` is a `Record<Iso3Langs, string>`

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -1,22 +1,41 @@
...
+      <gn-ui-metadata-info
+        class="block"
+        *ngIf="(facade.isPresent$ | async) === true"
+        [metadata]="facade.metadata$ | async"
+        [incomplete]="facade.isIncomplete$ | async"
+        [landingPages]="facade.landingPageLinks$ | async"
+      >
+      </gn-ui-metadata-info>
+    </div>
+    <div>
```
fgravin, line 22: `(facade.isPresent$ | async) === true` can be put here, it avoids to multiply this check.
BTW, is it the same as `(facade.metadata$ | async)` ? If yes, you could use it to create an alias and avoid complexity with too many `async` pipes in nested markup.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -1,22 +1,41 @@
...
+      <gn-ui-metadata-contact
+        *ngIf="(facade.isPresent$ | async) === true"
+        [metadata]="facade.metadata$ | async"
+      ></gn-ui-metadata-contact>
+      <div *ngIf="(facade.isPresent$ | async) === true">
```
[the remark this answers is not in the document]
tkohr: Yes, I thought about this, but how do I prevent making `feature-catalog` a dependency of `ui-elements` again in this case ?

libs/feature/catalog/src/lib/sources/sources.service.ts
```diff
@@ -0,0 +1,22 @@
...
+import { map, share } from 'rxjs/operators'
+import { CatalogSource } from './sources.model'
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SourcesService {
+  $sources: Observable<Array<CatalogSource>> = this.sourcesApiService
+    .getSources1()
+    .pipe(share())
```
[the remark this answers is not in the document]
tkohr: Yes, thanks `shareReplay` indeed fits better here and does the job

libs/feature/catalog/src/lib/source-label/source-label.component.ts
```diff
@@ -0,0 +1,32 @@
...
+@Component({
+  selector: 'gn-ui-source-label',
+  templateUrl: './source-label.component.html',
+  styleUrls: ['./source-label.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SourceLabelComponent {
+  sourceLabel$: Observable<CatalogSource>
+  @Input() set catalogUuid(uuid: string) {
+    this.sourceLabel$ = this.sourcesService
```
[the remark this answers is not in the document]
tkohr: Reworked this again, adding a `setSourceUuid()` method to the service

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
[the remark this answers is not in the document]
fgravin: no you can pass the source label, as you pass the metadata for the contact.

libs/feature/catalog/src/lib/sources/sources.service.ts
```diff
@@ -1,22 +1,34 @@
...
+import { BehaviorSubject, combineLatest, Observable } from 'rxjs'
+import { map, shareReplay } from 'rxjs/operators'
 import { CatalogSource } from './sources.model'
 
 @Injectable({
   providedIn: 'root',
 })
 export class SourcesService {
-  $sources: Observable<Array<CatalogSource>> = this.sourcesApiService
+  sources$: Observable<Array<CatalogSource>> = this.sourcesApiService
```
fgravin, line 12: Prefer `CatalogSource[]` notation

libs/feature/catalog/src/lib/sources/sources.service.ts
```diff
@@ -1,22 +1,34 @@
...
+    .pipe(shareReplay())
+  sourceUuid$ = new BehaviorSubject('')
+  sourceLabel$: Observable<CatalogSource> = combineLatest(
+    this.sources$,
+    this.sourceUuid$
+  ).pipe(
+    map(
+      ([sources, uuid]) => sources.filter((source) => source.uuid === uuid)[0]
+    ),
+    map((source) => (source ? source.label[this.langService.iso3] : undefined))
```
fgravin, line 23: add a `filter` in pipe if source is falsy

libs/feature/catalog/src/lib/source-label/source-label.component.ts
```diff
@@ -12,15 +8,9 @@ import { SourcesService } from '../sources/sources.service'
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SourceLabelComponent {
-  sourceLabel$: Observable<CatalogSource>
   @Input() set catalogUuid(uuid: string) {
-    this.sourceLabel$ = this.sourcesService
-      .getSource(uuid)
-      .pipe(map((source) => source.label[this.langService.iso3]))
+    this.sourcesService.setSourceUuid(uuid)
```
fgravin, line 12: I don't like this sorry, cause if the component is not in the page, then the source service won't update its source.
2 options:
- we keep to store the current source uuid in the `sourceService`, wich might not be conceptually ideal, sorry i was not very aware and clear in my last review. If we do this, we should update the source uuid on record-state metadata change, not within this component.
- we don't store any uuid in the service, but the `getSourceLabel(uui: string) `method is in the source Service, not in the component. I think it's better for now.

libs/feature/catalog/src/lib/source-label/source-label.component.html
```diff
@@ -1,3 +1,3 @@
 <div>
-  {{ sourceLabel$ | async }}
+  {{ sourcesService.sourceLabel$ | async }}
```
fgravin, line 2: should be tested in the test

fgravin (commented):
I've added some small comment but it looks good otherwise thanks

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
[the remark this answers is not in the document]
tkohr: I guess I'm missing something here, how to pass the source label from `source-label.component` (feature) to `metadata-catalog.component` (ui). Using an `output` that emits from `source-label.component` to `record-metadata.component` (parent) and then pass it as input to `metadata.catalog.comonent`?
What I've done now is just use `ng-content` in the ui component. What do you think about that?

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -1,40 +1,43 @@
 <div id="about" class="container-lg mx-auto mb-16" style="margin-top: 60px">
-  <div class="grid grid-cols-3 gap-6">
+  <gn-ui-metadata-info
+    class="block"
+    *ngIf="(facade.isPresent$ | async) === false"
+    [metadata]="{
+      title: 'Loading...'
```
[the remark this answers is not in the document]
tkohr: The text is actually not used anymore, but has been replaced by a `content-ghost.component` in the past. The content ghost just needs an empty metadata object, so I adapted this.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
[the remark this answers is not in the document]
fgravin: You have the `uuid` of the source, you use the `SourceService` to translate it, and you pass the result to the ui component ? I don't have the whole thing in mind but does it make sense ?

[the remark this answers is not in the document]
tkohr: Ah ok thanks, I see, I'll keep as is like discussed

PR #239, opened 2022-03-07
Datahub: Add a thumbnail placeholder to config

tkohr:
The PR introduces an optional `thumbnail_placeholder` param to the config, that allows to indicate a URL to an image that is used if a metadata record has no thumbnail image, or it is broken.

A new `image-fallback.directive` replaces broken thumbnail URLs by the placeholder URL.

The thumbnail component uses the placeholder URL if no thumbnail URL is present. (I would have liked to place this logic in the directive as well, but could not find a way to manipulate values within the directive, before the error is triggered by the `undefined` URL.)

Potential improvements:
- Indicate placeholder URL as `SafeUrl` (not sure if this is necessary?)
- Allow setting `object-fit: cover/fill` to image via config, or keep `fill` for all. => I decided to keep `fill` for all to not add complexity to the config. Also played around with the CSS background possibilities used for the header, but this does not fit the use case of images that are not used as a background.


conf/default.toml
```diff
@@ -71,6 +71,11 @@ background_color = "#fdfbff"
 # This mandatory parameter indicates which background should be used for the main header
 header_background = "center / cover url('assets/img/default_header_bg.webp')"
 
+# This mandatory parameter indicates which fallback image should be used for thumbnails,
```
jahow, line 74: I don't think this parameter should be mandatory actually, and I also realized that the header background above shouldn't be mandatory as well; this is justified by the fact that we then won't be able to update the default placeholder (or the default header) in subsequent versions without asking users to update their configs manually.

Maybe we made too many config parameters mandatory in the first place?

libs/util/shared/src/lib/image.directive.ts
```diff
@@ -0,0 +1,13 @@
+import { Directive, ElementRef, HostListener, Input } from '@angular/core'
+
+@Directive({
+  selector: 'img[gnUiImage]',
+})
+export class ImageDirective {
+  @Input() placeholder: string
```
jahow, line 7: this is actually more of a fallback than a placeholder

libs/ui/search/src/lib/record-thumbnail/record-thumbnail.component.html
```diff
@@ -15,9 +15,10 @@
     />
   </svg>
```
jahow, line 16: This SVG element will never show up with the logic in this PR; this SVG is IMO the placeholder that's supposed to be made customizable

apps/datahub/src/assets/img/geo2france.svg
```diff
@@ -0,0 +1,73 @@
...
+    </g>
+    <defs>
+        <radialGradient id="_Radial1" cx="0" cy="0" r="1" gradientUnits="userSpaceOnUse" gradientTransform="matrix(72.8075,0,0,72.8075,401.11,128.943)"><stop offset="0" style="stop-color:rgb(230,231,232);stop-opacity:1"/><stop offset="0.3" style="stop-color:rgb(207,208,210);stop-opacity:1"/><stop offset="0.75" style="stop-color:rgb(167,169,172);stop-opacity:1"/><stop offset="1" style="stop-color:rgb(167,169,172);stop-opacity:1"/></radialGradient>
+        <linearGradient id="_Linear2" x1="0" y1="0" x2="1" y2="0" gradientUnits="userSpaceOnUse" gradientTransform="matrix(25.3501,-20.5671,20.5671,25.3501,400.223,87.2502)"><stop offset="0" style="stop-color:rgb(230,231,232);stop-opacity:0.37"/><stop offset="0.11" style="stop-color:rgb(207,208,209);stop-opacity:0.37"/><stop offset="0.34" style="stop-color:rgb(149,148,149);stop-opacity:0.37"/><stop offset="0.67" style="stop-color:rgb(57,54,55);stop-opacity:0.37"/><stop offset="0.75" style="stop-color:rgb(35,31,32);stop-opacity:0.37"/><stop offset="1" style="stop-color:rgb(35,31,32);stop-opacity:0.37"/></linearGradient>
+    </defs>
+</svg>
```
jahow, line 73: This cannot be the default placeholder, it has to be something more generic (e.g. the current one)

conf/default.toml
tkohr: Ok, I agree, I'll rework this

libs/ui/search/src/lib/record-thumbnail/record-thumbnail.component.ts
```diff
@@ -1,10 +1,15 @@
...
 
 @Component({
   selector: 'gn-ui-record-thumbnail',
   templateUrl: './record-thumbnail.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class RecordThumbnailComponent {
-  @Input() thumbnailUrl: string
+  @Input() set thumbnailUrl(url: string) {
+    this.imgUrl = url || this.placeholderUrl
```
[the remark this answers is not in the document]
tkohr: Hum, it's not exactly the same. The svg fallback embedded in the HTML still shows during loading for thumbnails that don't need a fallback. If we do not judge this "loading placeholder" necessary, indeed I can remove it.  
The setter here takes care to use the placeholder from the config for thumbnail urls that are `undefined`.
Making the `thumbnail_placeholder` config optional will also have an impact here, however, as the embedded svg will have to be used as the `img` `src`, I think, to ensure coherent default and custom behavior.

libs/ui/search/src/lib/record-thumbnail/record-thumbnail.component.html
tkohr: See comment above, it's still displaying during image fetching.

apps/datahub/src/assets/img/geo2france.svg
tkohr: Yep, I agree

conf/default.toml
tkohr: I adapted the param to be optional in the config, providing a default placeholder in the `app-config` if necessary and removing the embedded SVG "loading placeholder" to be consistent.

libs/ui/search/src/lib/record-thumbnail/record-thumbnail.component.ts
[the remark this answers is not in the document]
tkohr: I've removed the embedded SVG, to always display the same placeholder/fallback

jahow (approved):
Thanks! I think we can go with that. Any reason why the original image was converted to WEBP? Why not keep it as SVG?

tkohr:
> Thanks! I think we can go with that. Any reason why the original image was converted to WEBP? Why not keep it as SVG?

Thanks! The originally embedded SVG included inline styling that would not be applied, using it as an `img` `src`.

PR #242, opened 2022-03-29
Datahub - UI improvments

fgravin:
Based on https://github.com/geonetwork/geonetwork-ui/pull/241 to avoid future conflicts.
Check commit list for content.

### DIFF
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -46,7 +46,7 @@
     class="uppercase font-medium tracking-wider"
     translate
   >
-    record.metadata.access
+    record.metadata.download
   </button>
   <button
     gnUiAnchorLink="links"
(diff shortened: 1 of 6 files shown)


apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -46,7 +46,7 @@
     class="uppercase font-medium tracking-wider"
     translate
   >
-    record.metadata.access
+    record.metadata.download
```
jahow, line 49: The old key should be removed from translations I think

jahow (approved):
Not tested, changes look good, thanks!

apps/datahub/src/app/record/header-record/header-record.component.html
fgravin: Indeed i forgot to run the npm command !
Thanks

PR #247, opened 2022-04-04
Handle GeoJSON layers as remote files in map-context

jahow:
This PR includes:
* handling of GeoJSON files given as URL instead of inline data for the map-context component
* support for GeoJSON layers in app-config, either inline of with a URL
* stricter types for the map context layers

A valid map layer in the config can now be:
```toml
[[map_layer]]
type = "geojson"
url = "https://france-geojson.gregoiredavid.fr/repo/regions.geojson"
```

or

```toml
[[map_layer]]
type = "geojson"
data = """
{
"type": "FeatureCollection",
"features": [{"type": "Feature", "geometry": {"type": "Point", "coordinates": [125.6, 10.1]}}]
}
"""
```

### DIFF
--- a/libs/feature/map/src/lib/map-context/map-context.service.ts
+++ b/libs/feature/map/src/lib/map-context/map-context.service.ts
@@ -14,9 +14,9 @@ import TileWMS from 'ol/source/TileWMS'
 import TileLayer from 'ol/layer/Tile'
 import XYZ from 'ol/source/XYZ'
 import VectorSource from 'ol/source/Vector'
+import GeoJSON from 'ol/format/GeoJSON'
 import { MapUtilsService } from '../utils/map-utils.service'
 import { bbox as bboxStrategy } from 'ol/loadingstrategy'
-import GeoJSON from 'ol/format/GeoJSON'
 import { LayerConfig, MapConfig } from '@geonetwork-ui/util/app-config'
 
 export const DEFAULT_BASELAYER_CONTEXT: MapContextLayerModel = {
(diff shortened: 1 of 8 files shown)


libs/feature/map/src/lib/map-context/map-context.service.ts
```diff
@@ -54,46 +55,70 @@ export class MapContextService {
...
         })
       case MapContextLayerTypeEnum.WFS:
         return new VectorLayer({
           source: new VectorSource({
             format: new GeoJSON(),
             url: function (extent) {
-              return `${url}?service=WFS&version=1.1.0&request=GetFeature&outputFormat=application/json&typename=${name}&srsname=EPSG:3857&bbox=${extent.join(
-                ','
-              )},EPSG:3857`
+              return `${
```
fgravin, line 80: could have kept the dedicated extractions of var in the case statement, more readable IMO, not a big deal.

fgravin (approved):
Not tested but clean, good job thanks

libs/feature/map/src/lib/map-context/map-context.service.ts
jahow: this was actually not working with the stricter types, since some properties might not be present (e.g. for geojson it's either url or data, not both)

fgravin: you can do the residual extraction in the `case` statement, casting with `as` to the correct type.

PR #249, opened 2022-04-05
Datahub: Responsive design

tkohr:
This PR arranges the content of the datahub app to be responsive to the screen size it is displayed in. This includes the search with its results as well as the metadata record view. The PR focuses on the implementation of the static mockups and leaves some adaptions concerning user interaction open to other PRs.

TODOs concerning the mockups :
- [x] check/rework description line height
- [x] check/rework spacing

TODOs to be done in separate PRs :

Search:
- reduce size of sticky header

MD record:
- block interactive map on mobile
- make nav menu responsive
- rework download panels (for desktop and mobile)


apps/datahub/src/app/search/record-preview-datahub/record-preview-datahub.component.html
```diff
@@ -1,42 +1,43 @@
...
-          class="material-icons-outlined text-gray-200 mr-4"
-          >cloud_download</mat-icon
-        >
-        <mat-icon
-          *ngIf="isViewable"
-          class="material-icons-outlined text-gray-200 mr-4"
-          >map</mat-icon
-        >
-      </div>
+    <div class="order-2 text-gray-400 sm:order-3 sm:row-end-4 uppercase">
```
fgravin, line 27: I think the organisation should be displayed on one line, at least in desktop mode, and could take more space in the row.
- before 
Screenshot from 2022-04-13 18-33-14
- after
Screenshot from 2022-04-13 18-33-09

fgravin (approved):
Nice work.
I haven't checked all the classes to see if it's optimized but the result is quite good.
I noticed that some scroll bar do not work though in desktop mode
- data table
- api

apps/datahub/src/app/search/record-preview-datahub/record-preview-datahub.component.html
tkohr: Ah, thanks for noticing, this. Should be fixed with the commit below.

fgravin: Ok, take care, if it's very long it should be an elipsis, confirmed by Lucile.
Thanks

tkohr: Yep, I added a `truncate` for the "desktop" layout and let "mobile" layout wrap on multiple lines, as one line is much shorter here and multiple lines don't "break" the design.

tkohr:
> I noticed that some scroll bar do not work though in desktop mode
> 
>     * data table
>     * api

As discussed, they seem to work fine. I've added a commit for the other comment. Thanks for the review @fgravin !

PR #251, opened 2022-04-13
Datahub: Make navigation bar responsive

tkohr:
Based on #249 the PR implements a responsive navigation bar, specific to the datahub app. There might still be room for improvement in the approach.

https://user-images.githubusercontent.com/6329425/163205398-d1f43d40-fd9d-4b27-b913-5fc59a3fbc14.mp4


fgravin (approved):
Very nice.
It works pretty well.
The mobile management could have been wraped in a UI component but we can do that afterwards if we need it for another toolbar.
Thanks

fgravin:
I think that the mode mobile should be activate from a largest width, I wonder how it looks like on a tablet for instance.

PR #252, opened 2022-04-14
Datahub: set favicon

fgravin:
- Use the GeoNetwork favicon as default for the datahub.
- Move it to assets cause it can be overridden in the docker composition.
  ```
  
  -v /home/user/dev/geonetwork-ui/myconfig/assets:/assets \
  ```
- For non docker build, just replace the default one by your own in the `assets` folder.


tkohr (approved):
LGTM, thanks!

PR #253, opened 2022-04-14
Datahub: add a default theme

fgravin:
Update configuration for the default theme

- header background
- fonts
- colors

To address in another PR

- choose `black` or `white` record head color
- choose `primary` or `secondary` associated resources background

Screenshot from 2022-04-14 16-54-26

Screenshot from 2022-04-14 16-54-34


conf/default.toml
```diff
@@ -72,22 +72,22 @@ proxy_path = ""
...
 # This optional parameter indicates which background should be used for the main header
-# header_background = "center url('assets/img/default_header_bg.webp')"
+header_background = "center /cover url('assets/img/header_bg.webp')"
 
 # This optional parameter allows to override the fallback image that should be used for thumbnails, 
 # if the metadata record has no thumbnail image url or it is broken.
 # thumbnail_placeholder = 'assets/img/my_custom_placeholder.png'
 
 # These optional parameters allow changing fonts used in the app
-# main_font = "'My Custom Font', fallback-font"
```
tkohr, line 88: Maybe it is worth keeping in a comment, that the second is a `fallback-font`.

tkohr (approved):
Thanks @fgravin LGTM. Just noticing that the optional background and font parameters might seem less optional now, as they are set by default. Anyway, the comments tell the admin that they are optional.

>     * choose `black` or `white` record head color
Indeed seems necessary.
>     * choose `primary` or `secondary` associated resources background
What do you mean here?

conf/default.toml
```diff
@@ -72,22 +72,22 @@ proxy_path = ""
...
-# title_font = "'My Custom Title Font', fallback-font-title"
-# fonts_stylesheet_url = "https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&family=Permanent+Marker&display=swap"
+main_font = "'Rubik', sans-serif"
```
jahow, line 88: I don't think these parameters should be set in the default config; once someone deploys the datahub, we will never be able to change the default values for them since they will be written in their config file;

This is why most settings are not set in the default config,

jahow (changes requested):

conf/default.toml
```diff
@@ -106,6 +106,3 @@ background_color = "#fdfbff"
 #     Welcome to <span class="text-primary">Organization</span>'s<br>
 #     wonderful <span style="font-size: 1.2em;">data catalogue</span>
 #     """
-
-[translations.en]
-datahub.header.title.html = '<div class="text-white">Toutes les <span class="text-primary">données<br>publiques</span> de mon organisation</div>'
```
jahow, line 111: +1 this has nothing to do in the default config, can't remember why we put it there!

fgravin:
> choose primary or secondary associated resources background

@tkohr this background color
would be better in darkblue for this theme, which is the secondary color, while for geo2france, the blue is the primary

Screenshot from 2022-05-03 11-36-04

jahow (approved):
Looking good, thanks! There is a failing test on the app config I think.

PR #254, opened 2022-04-22
Proxy: don't use the proxy if the url is on the same host

fgravin:
https://www.geo2france.fr/proxy/?url=https%3A%2F%2Fwww.geo2france.fr%2Fgeoserver%2Fcr_hdf%2Fows%3FSERVICE%3DWMS%26REQUEST%3DGetCapabilities should not happen.

### DIFF
--- a/libs/util/shared/src/lib/services/proxy.service.ts
+++ b/libs/util/shared/src/lib/services/proxy.service.ts
@@ -20,7 +20,8 @@ export class ProxyService {
     const proxyUrl = new URL(this.proxyPath, current.toString()).toString()
     if (
       current.hostname === urlObj.hostname &&
-      current.protocol === urlObj.protocol
+      current.protocol === urlObj.protocol &&
+      current.port === urlObj.port
     ) {
       return url
     }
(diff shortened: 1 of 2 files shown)


libs/util/shared/src/lib/services/proxy.service.ts
```diff
@@ -15,10 +15,15 @@ export class ProxyService {
...
-    const proxyUrl = new URL(
-      this.proxyPath,
-      window.location.toString()
-    ).toString()
+    const urlObj = new URL(url)
+    const current = window.location
+    const proxyUrl = new URL(this.proxyPath, current.toString()).toString()
+    if (
+      current.hostname === urlObj.hostname &&
+      current.protocol === urlObj.protocol
```
jahow, line 23: We should also check that the port is the same, otherwise CORS will apply as well. E.g. going from http://localhost:8080 to http://localhost:8081 is considered a cross-origin request.

jahow (approved):
Left a comment about checking the host port as well, but other than that it's all good, thanks!

fgravin:
Thanks @jahow i updated the function and the tests accordingly.

PR #255, opened 2022-04-22
datahub: Update Organisation active filter

fgravin:
On the search, if an organization is active in the filter (publisher param in the route), then the layout of the active filter is différent, and we can now remove the active filter.
Screenshot from 2022-04-22 15-46-16


apps/datahub/src/app/search/search-summary/search-summary.component.html
```diff
@@ -2,6 +2,19 @@
...
+  >
+    <div class="flex">
+      <div
+        class="close bg-white bg-opacity-50 text-primary w-4 cursor-pointer h-4 text-xl leading-3 hover:bg-opacity-100 rounded-full text-center transition-all"
+        style="line-height: 14px"
+        (click)="removeOrg()"
+      >
+        ×
+      </div>
+      <div class="publisher ml-3 mt-px flex-grow">{{ publisher }}</div>
```
tkohr, line 17: Do the `publisher` and `close` classes only serve in the tests as a target? If so, I'm wondering if there is a way to prevent adding the classes for the tests and still targeting the divs in a clean way.

tkohr (approved):
Nice, thanks @fgravin!

Not directly related to this PR, but I'm noticing that entering a search term, navigating to its entry and clicking on the organization, keeps the search term in the search box (with the filter active and also once you delete the filter)

apps/datahub/src/app/search/search-summary/search-summary.component.html
fgravin: yes they are...
haven't hear about a better solution so far, what do you suggest ?

fgravin:
> Not directly related to this PR, but I'm noticing that entering a search term, navigating to its entry and clicking on the organization, keeps the search term in the search box (with the filter active and also once you delete the filter)

Not sure to understand the steps to reproduce.
Can you create a ticket for that please @tkohr  ? THanks

PR #256, opened 2022-05-02
Datahub: Fix WMS map initial center

fgravin:
Ensure to correctly extract the WMS layer bbox for the initial view extent.
For that, we check (from an hardcoded list of lon/lat CRS) what is the lon/lat CRS used in the getCapabilities to express the layer extent.

Additionally, the PR improve the initial zoom level, which is no more the nearest but the one which include the expected extent. It improves the rendering.


libs/feature/map/src/lib/map-context/map-context.service.ts
```diff
@@ -123,22 +124,21 @@ export class MapContextService {
...
+    const view = new View({
       center,
       zoom,
       maxZoom,
       extent: maxExtent,
       multiWorld: false,
       constrainResolution: true,
     })
+    if (viewModel.extent && map) {
+      view.fit(viewModel.extent, {
```
jahow, line 137: I feel like we originally used `this.mapUtils.getViewFromExtent` instead of `view.fit` for a reason byt I can't remember why... @tkohr do you remember?

libs/feature/map/src/lib/utils/map-utils-wms.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    )
+  }
+
+  getLayerLonLatBBox(layer: MapContextLayerWmsModel) {
+    return this.getLayerFull(layer).pipe(
+      map((wmsLayerFull) => this.getLonLatBBox(wmsLayerFull))
+    )
+  }
+
+  getLonLatBBox(wmsLayerFull: WMSLayerFull) {
```
jahow, line 34: This utility functions is really nice, and to be honest I think it would fit very well in ogc-client. The philosophy of the library is to save developer from writing this kind of things; for example we do something similar in ogc-client for selecting a mime type that corresponds to geojson for WFS requests.

This can be done later though, I guess

libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -137,13 +136,7 @@ export class MapUtilsService {
...
       )
     } else if (layer && layer.type === 'wms') {
-      geographicExtent = from(
-        new WmsEndpoint(this.proxy.getProxiedUrl(layer.url))
-          .isReady()
-          .then((endpoint) => endpoint.getLayerByName(layer.name))
-      ).pipe(
-        map((layer: any) => layer.boundingBoxes['EPSG:4326']) // eslint-disable-line -- ogc-client lacks typing here
-      )
+      geographicExtent = this.wmsUtils.getLayerLonLatBBox(layer)
```
jahow, line 139: Regarding my other comment, this could become:

```ts
from(
        new WmsEndpoint(this.proxy.getProxiedUrl(layer.url))
          .isReady()
          .then((endpoint) => endpoint.getLayerByName(layer.name))
          .then((layer) => layer.getLonLatBoundingBox())
      )
```

jahow (approved):
Thanks, haven't tested but it makes complete sense.

libs/feature/map/src/lib/utils/map-utils-wms.service.ts
fgravin: Yes I wanted to put it in ogc-client to be honest, but wasn't sure.
Let's do that in a second step.

PR #257, opened 2022-05-03
search: escape special characters

fgravin:
`? + - ( { `and more are escaped in the `query_string` parameter.
This prevent from ES errors.


tkohr (commented):
Thanks @fgravin! I still manage to break the search with an opening double quote, otherwise works as expected.

PR #258, opened 2022-05-03
Datahub: Rework text colors

tkohr:
PR makes various adaptions on text colors to better fit the mockups, in particular to make titles black when using a different `main-color`:

- introduces color `title` and `title-header` classes, makes them configurable and uses them in the components
- renames color `gray-...` to `main-...`
- makes some adaptions on used grays to better fit the mockups

Note: There are a lot of different grays or even derivations of the primary color in the mockups, that might not all be on purpose. The implementation tries to get as close and coherent as possible here. Also, but it's a small detail, personally I preferred the colored API and Links titles, that seemed lighter to me.

dh-search
dh-record

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -77,8 +77,11 @@ secondary_color = "#001638"
 main_color = "#212029" # All-purpose text color
 background_color = "#fdfbff"
 
-# This optional parameter indicates which background should be used for the main header
+# These optional parameters indicate which background should be used for the main header and the text color used 
+# on top of the background. The color should be chosen to contrast well with the background (defaults to white).
+# Note: The search header does not use the header_foreground_color as it allows futher customisation via HTML.
 # header_background = "center /cover url('assets/img/header_bg.webp')" or "var(--color-gray-500)"
+# header_foreground_color = 'white'
 
 # This optional parameter allows to override the fallback image that should be used for thumbnails, 
 # if the metadata record has no thumbnail image url or it is broken.
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.html
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.html
@@ -2,7 +2,7 @@
   <input
     #searchInput
     type="text"
-    class="appearance-none focus:outline-white leading-tight rounded w-full text-main shadow-primary-light focus:shadow-primary"
+    class="appearance-none focus:outline-white leading-tight rounded w-full text-black shadow-primary-light focus:shadow-primary"
     [placeholder]="placeholder"
     [formControl]="control"
     [matAutocomplete]="auto"
--- a/libs/util/app-config/src/lib/app-config.ts
+++ b/libs/util/app-config/src/lib/app-config.ts
@@ -47,6 +47,7 @@ interface ThemeConfig {
   SECONDARY_COLOR: string
   MAIN_COLOR: string
   BACKGROUND_COLOR: string
+  HEADER_FOREGROUND_COLOR: string
   HEADER_BACKGROUND: string
   THUMBNAIL_PLACEHOLDER: string
   MAIN_FONT?: string
@@ -158,6 +159,7 @@ export function loadAppConfig() {
         'theme',
         ['primary_color', 'secondary_color', 'main_color', 'background_color'],
         [
+          'header_foreground_color',
           'main_font',
           'title_font',
           'fonts_stylesheet_url',
@@ -175,6 +177,8 @@ export function loadAppConfig() {
               SECONDARY_COLOR: parsedThemeSection.secondary_color,
               MAIN_COLOR: parsedThemeSection.main_color,
               BACKGROUND_COLOR: parsedThemeSection.background_color,
+              HEADER_FOREGROUND_COLOR:
+                parsedThemeSection.header_foreground_color || '#ffffff',
               THUMBNAIL_PLACEHOLDER:
                 parsedThemeSection.thumbnail_placeholder ||
                 'assets/img/thumb_placeholder.webp',
--- a/tailwind.base.config.js
+++ b/tailwind.base.config.js
@@ -19,6 +19,7 @@ module.exports = {
         'secondary-darker': 'var(--color-secondary-darker)',
         'secondary-darkest': 'var(--color-secondary-darkest)',
         main: 'var(--color-main)',
+        title: 'black',
         background: 'var(--color-background)',
         gray: {
           100: 'var(--color-gray-100)',
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -5,6 +5,7 @@
     <div
       *ngIf="metadata"
       class="font-title text-center px-2 text-[28px] mb-[41px] line-clamp-4 sm:mx-48"
+      [style.color]="foregroundColor"
     >
       {{ metadata.title }}
     </div>
--- a/apps/datahub/src/app/record/header-record/header-record.component.spec.ts
+++ b/apps/datahub/src/app/record/header-record/header-record.component.spec.ts
@@ -8,6 +8,7 @@ import { HeaderRecordComponent } from './header-record.component'
 jest.mock('@geonetwork-ui/util/app-config', () => ({
   getThemeConfig: () => ({
     HEADER_BACKGROUND: 'red',
+    HEADER_FOREGROUND_COLOR: 'white',
   }),
 }))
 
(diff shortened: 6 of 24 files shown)


libs/util/app-config/src/lib/app-config.ts
```diff
@@ -175,6 +179,9 @@ export function loadAppConfig() {
               SECONDARY_COLOR: parsedThemeSection.secondary_color,
               MAIN_COLOR: parsedThemeSection.main_color,
               BACKGROUND_COLOR: parsedThemeSection.background_color,
+              TITLE_COLOR: parsedThemeSection.title_color || 'black',
+              TITLE_HEADER_COLOR:
+                parsedThemeSection.title_header_color || 'black',
```
tkohr, line 184: For the new default theme I should probably set this to white by default.

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.html
```diff
@@ -2,7 +2,7 @@
   <input
     #searchInput
     type="text"
-    class="appearance-none focus:outline-white leading-tight rounded w-full text-main shadow-primary-light focus:shadow-primary"
+    class="appearance-none focus:outline-white leading-tight rounded w-full text-title shadow-primary-light focus:shadow-primary"
```
fgravin, line 5: I would put just `black` here, if text-title is green, not sure we want it green there

libs/feature/record/src/lib/record-metadata/record-metadata.component.css
```diff
@@ -2,6 +2,7 @@
 ::ng-deep .mat-tab-link {
   font-family: var(--font-family-main);
   letter-spacing: 0.88px;
+  color: var(--color-primary-lighter);
```
fgravin, line 5: there is a TW class for that, isn't it ?

fgravin (commented):
Thanks for the color title var.
I wasn't expecting that you adresse the configuration for that, I think we should discuss a bit to not overload the configuration.
Thanks

libs/util/app-config/src/lib/app-config.ts
fgravin: Yes indeed

fgravin: Also update the comment in the config please

libs/feature/record/src/lib/record-metadata/record-metadata.component.css
jahow: `text-primary-lighter` i think

jahow:
Why change the `gray-xxx` variables to `main-xxx`? That would be a huge breaking change for any other application relying on those classes. Remember: the datahub is only one app and the whole config system should not be specifically tailored to it. For example it might break things (even more) in the datafeeder.

tkohr:
> Why change the `gray-xxx` variables to `main-xxx`? That would be a huge breaking change for any other application relying on those classes. Remember: the datahub is only one app and the whole config system should not be specifically tailored to it. For example it might break things (even more) in the datafeeder.

Thanks for pointing this out. I thought `main-xxx` would be more expressive, as the colors depend on the `main-color`. Don't all the `gray-xxx` classes already exist in tailwind?

jahow:
> Thanks for pointing this out. I thought main-xxx would be more expressive, as the colors depend on the main-color. Don't all the gray-xxx classes already exist in tailwind?

Sorry my reaction was probably excessive. Yes currently we pretty much "overwrite" the `gray` color scale built in TW. Originally I think it made sense because `main` was always going to be very close to black anyway (since it's the main text color), so we just rewrote the `gray` scale to match our "definition" of black.

jahow:
Also I was wrong: if someone relies on the `gray-xx` scale it will still work after this PR since it's provided by TW anyways

fgravin:
> Also I was wrong: if someone relies on the `gray-xx` scale it will still work after this PR since it's provided by TW anyways
Yes
https://tailwindcss.com/docs/customizing-colors
I was ok for this move cause to me gray should be gray, if main color is blue and you want to keep some elements in gray, you are screwed. But actually i don't really care, maybe we should have a closer look on the theme and the color, maybe we can elaborate a better strategy, later.

tailwind.base.config.js
```diff
@@ -19,17 +19,18 @@ module.exports = {
...
-        gray: {
-          100: 'var(--color-gray-100)',
-          200: 'var(--color-gray-200)',
-          300: 'var(--color-gray-300)',
-          400: 'var(--color-gray-400)',
-          500: 'var(--color-gray-500)',
-          600: 'var(--color-gray-600)',
-          700: 'var(--color-gray-700)',
-          800: 'var(--color-gray-800)',
-          900: 'var(--color-gray-900)',
```
jahow, line 32: So as discussed I'm not really in favor of this change, I think it's confusing to have both a `text-main` class and a `text-main-xxx` scale; and I think we can safely rely on `gray-xxx` being actually gray since main is supposed to be close to black (or white in a dark theme). But overall it's not breaking anything so I'm leaving this up to you :slightly_smiling_face:

tailwind.base.config.js
```diff
@@ -19,17 +19,18 @@ module.exports = {
         'secondary-darker': 'var(--color-secondary-darker)',
         'secondary-darkest': 'var(--color-secondary-darkest)',
         main: 'var(--color-main)',
+        title: 'black',
```
jahow: We shouldn't need a new color here, as it does not really fit with our concept of theme. A theme is 4 colors: secondary, primary, main (text) and background. Titles have their own font but their color should be picked from the available palette offered by thos 4 colors.

conf/default.toml
```diff
@@ -76,6 +76,10 @@ primary_color = "#c82850"
 secondary_color = "#001638"
 main_color = "#212029" # All-purpose text color
 background_color = "#fdfbff"
+# Title colors can be defined optionally. Make use of title_header_color if a higher contrast to the header background is needed.
+# If not set, both values default to black.
+# title_color = '#000000'
+# title_header_color = '#000000'
```
jahow, line 82: IMO these settings are confusing.
* Title color feels like it does not fit in our concept of theme, as I wrote in an earlier comment
* Title header color is definitely relevant, but it should be placed closer to the `header_background` setting since the reason it exists is to allow users to define an "opposite" or "accent" color alongside a header background; I would call it `header_foreground_color` and put it right underneath the `header_background` setting, specify it is optional (default to white) and that it should be defined according to the header background

libs/util/shared/src/lib/services/theme.service.ts
```diff
@@ -89,6 +91,13 @@ export class ThemeService {
     applyColor('main-800', scale(0.8))
     applyColor('main-900', scale(0.9))
 
+    if (titleColor) {
+      applyColor('title', chroma(titleColor))
+    }
+    if (titleHeaderColor) {
+      applyColor('title-header', chroma(titleHeaderColor))
```
jahow, line 98: There is no need to define a CSS variable for this setting; the header text color can just be read directly from the settings in the header component, thus we end up with a much small scale change

tailwind.base.config.js
tkohr: Hum, I'm wondering if we actually still need these scaled classes, since tailwind provides the `gray` scales and the renamed `main` scales are currently not used. Regarding the colors, neither the tailwind grays nor the custom ones match the mockup's grays RGB a 100%, but we get very close with both. After all, it's a decision if we want the gray scale dependence to our main color or not. If it's not essential, I propose to remove these classes and their calculation for the sake of simplicity.

tailwind.base.config.js
tkohr: Discussing with @fgravin we thought it would be handy to be able to change all the title colors (at least from the tailwind config). Unfortunately, `black` cannot not be inferred from the 4 theme colors and using `text-black` in the templates, would make a change less flexible.

conf/default.toml
tkohr: Good idea, I'll adapt this, thanks!

libs/util/shared/src/lib/services/theme.service.ts
tkohr: Ok, I'll have a look at this.

tailwind.base.config.js
fgravin: Yes `--color-main-900` is not dark enough, so if we put `text-black` it's ok but it means we can't change it.
If we have a class for this specfic elements, it means we can imagine to customize it in the future (via TW config or via css variable or configuration if needed).
Not a big deal for now anyway.

tkohr:
Thanks for the reviews and the eager discussion @jahow @fgravin . I've reworked the rework, being less intrusive on the existing. Hope I didn't forget anything;-)

tailwind.base.config.js
tkohr: After re-discussing with @fgravin I finally, kept it as it was before.

fgravin (approved):
Thanks for the rework, I think it's cleaner now and goes straight to the goal.

tailwind.base.config.js
jahow: I honestly don't see the point of this style instead of `title-black` since it's not customizable any more.

Another solution would be to have `main-light`, `main-lighter`, `main-dark` and `main-darker` classes. But honestly this should not be needed because `main` was originally intended to be very close to black.

PR #259, opened 2022-05-03
ES search: set default operator as AND

fgravin:
instead or OR to have more accurate searches


tkohr (approved):
I guess it makes sense to be more "restrictive" here in order to be more precise. The only drawback that I see is that a typo in the search query might not get a result anymore, that it got with `OR`.

fgravin:
> The only drawback that I see is that a typo in the search query might not get a result anymore, that it got with `OR`

you are right but if you do a typo the word won't be find anyway with a OR or a AND. 
if we wan't to handle typo, we should set up the fuzzy search instead of what we are doing right now. I think the move to the AND is mandatory. Fuzzy search would be better in the future for sure.

PR #261, opened 2022-05-04
search: any is multilingual

fgravin:
On the last GN4 version, `any` is an object, the index is multilingual.
So using `any` in a field should be like `any.*`


jahow (approved):
Makes sense with the latest evolutions on GN4, thanks

PR #263, opened 2022-05-12
Add language setting for elasticsearch queries

tkohr:
PR adds a `metadata_language` setting that allows to indicate in which language queries should be sent to elasticsearch.

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -10,6 +10,10 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 # The actual URL will be appended after this path, e.g. : https://my.proxy/?url=http%3A%2F%2Fencoded.url%2Fows`
 # This is an optional parameter: leave empty to disable proxy usage
 proxy_path = ""
+# This optional parameter defines, in which language metadata should be queried in elasticsearch.
+# Use ISO 639-2/B (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) format to indicate the language of the metadata.
+# If not indicated, a wildcard is used and no language preference is applied for the search.
+# metadata_language = "fre"
 
 ### MAP SETTINGS
 
--- a/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.spec.ts
+++ b/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.spec.ts
@@ -131,7 +131,6 @@ describe('FuzzySearchComponent', () => {
     })
     describe('when output is defined', () => {
       beforeEach(() => {
-        component.inputSubmited.subscribe((event) => (outputValue = event))
         component.handleInputSubmission('blarg')
       })
       it('updates the search filters as well', () => {
--- a/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.spec.ts
+++ b/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.spec.ts
@@ -51,7 +51,7 @@ const searchServiceMock = {
   updateSearch: jest.fn(),
 }
 const esServiceMock = {
-  buildAutocompletePayload: jest.fn(() => of({ fakeQuery: '' })),
+  buildAutocompletePayload: jest.fn(() => ({ fakeQuery: '' })),
 }
 const elasticsearchMapperMock = {
   toRecords: jest.fn(() => [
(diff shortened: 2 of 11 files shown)


conf/default.toml
```diff
@@ -10,6 +10,10 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 # The actual URL will be appended after this path, e.g. : https://my.proxy/?url=http%3A%2F%2Fencoded.url%2Fows`
 # This is an optional parameter: leave empty to disable proxy usage
 proxy_path = ""
+# This optional parameter defines, in which language metadata should be queried in elasticsearch.
```
fvanderbiest, line 13: I would add "IMPORTANT" in the readme :-)

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.spec.ts
```diff
@@ -51,7 +51,7 @@ const searchServiceMock = {
   updateSearch: jest.fn(),
 }
 const esServiceMock = {
-  buildAutocompletePayload: jest.fn(() => of({ fakeQuery: '' })),
+  buildAutocompletePayload: jest.fn(() => ({ fakeQuery: '' })),
```
fgravin, line 54: 2 new warnings that you should fix plz.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -35,18 +35,13 @@ export class FuzzySearchComponent {
   displayWithFn: (MetadataRecord) => string = (record) => record?.title
 
   autoCompleteAction = (query) =>
-    this.esService
-      .buildAutocompletePayload(query)
+    this.searchApiService
```
fgravin, line 38: This is not very secured, but it wasn't the case before either.
Note that if a request takes a long time to respond, the results will erase more recent ones.

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -196,37 +190,59 @@ export class ElasticsearchService {
...
-                },
-                ...template.query.bool.must.filter(
-                  (clause) => !('multi_match' in clause)
+        },
+      },
+      _source: ['resourceTitleObject'],
+      from: 0,
+      size: 20,
+    }
+    return {
```
fgravin, line 217: if there is no more a config, then you can get rid of the cascade spread operator, just build the correct object once.
use the tests to be sure you created exactly the same object as before

fgravin (changes requested):
Very good thanks, tested and works as expected.
Some request changes though, just to be cleaner.

libs/util/app-config/src/lib/app-config.ts
```diff
@@ -96,7 +97,7 @@ export function loadAppConfig() {
         parsed,
         'global',
         ['geonetwork4_api_url'],
-        ['proxy_path'],
+        ['proxy_path', 'metadata_language'],
```
jahow, line 100: Can we implement a check here to make sure that the provided language is valid? So:
1/ if not a valid code, stop the app and warn the user (e.g. `fra` instead of `fre`, `de` instead of `ger`)
2/ if not lowercase, force it to be

jahow: otherwise all search requests will fail silently

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
tkohr: I added `take(1)` here. Does that make sense?

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
tkohr: Good point, completely missed that, should be better now.

libs/util/app-config/src/lib/app-config.ts
tkohr: Good point, thanks, the `lowerCaser()` is added, still have to add the check.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -41,6 +41,7 @@ export class FuzzySearchComponent {
         JSON.stringify(this.esService.buildAutocompletePayload(query))
       )
       .pipe(
+        take(1),
```
fgravin, line 44: useless, we'll discuss about that later.
don't secure it for now, it wasn't before anyway.

tkohr: ok

libs/util/app-config/src/lib/app-config.ts
tkohr: I've started with https://github.com/geonetwork/geonetwork-ui/pull/264 regarding this.

PR #264, opened 2022-05-13
Add a check for metadata_language setting

tkohr:
Just a first draft to add a check for the metadata_language value. Still wondering where this check should be called from, best.


libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -121,3 +123,20 @@ export function parseMultiConfigSection(
...
+  parsedConfigSection: any,
+  outErrors: string[]
+) {
+  if (
+    LANGUAGE_CODES_ISO_3.indexOf(
+      parsedConfigSection.metadata_language.toLowerCase()
+    ) === -1
+  ) {
+    outErrors.push(
+      'In the [global] section: metadata_language is not in ISO 639-2/B format'
```
jahow, line 137: Maybe added the actual value in the error message, to make this even more explicit?

jahow (commented):
Simple and efficient, well done! :+1:

libs/util/app-config/src/lib/parse-utils.ts
tkohr: Ok, added the value to the message and used the warnings instead of errors, since this is rather an "unexpected" value than a "mandatory value missing".

tkohr:
Oups, forgot a little about this open PR,... added two commits.

jahow:
Thanks, merging!

jahow (approved):

PR #265, opened 2022-05-30
Datahub: Add crossorigin attribute for preload links

tkohr:
PR adds `crossorigin` attribute to preload links to make them being used by browser and prevent loading them twice (cf current chromium warning for `default.toml`).

```
polyfills.ed0cba819d7212ef.js:1 A preload for 'https://www.geo2france.fr/datahub/assets/configuration/default.toml' is found,
but is not used because the request credentials mode does not match. Consider taking a look at crossorigin attribute.
(anonymous) @ polyfills.ed0cba819d7212ef.js:1
search:1 The resource https://www.geo2france.fr/datahub/assets/configuration/default.toml was preloaded using link preload
but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is
preloaded intentionally.
```

### DIFF
--- a/apps/datahub/src/index.html
+++ b/apps/datahub/src/index.html
@@ -12,7 +12,12 @@
       href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined"
       rel="stylesheet"
     />
-    <link rel="preload" href="assets/configuration/default.toml" as="fetch" />
+    <link
+      rel="preload"
+      href="assets/configuration/default.toml"
+      as="fetch"
+      crossorigin
+    />
     <!--
       Add more preload links here to reduce the time to initial page render!
       Example:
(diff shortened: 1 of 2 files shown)


fgravin:
Hi @tkohr, thanks for the PR, it makes perfect sense.
What preload were you targeting ? the dataset header ? others ?

Anyway, I think there is an issue with the server response which add `no-cache` headers to all static resources.
All js, css and images should be cached.

jahow (approved):
Great find, thanks!

tkohr:
> Hi @tkohr, thanks for the PR, it makes perfect sense. What preload were you targeting ? the dataset header ? others ?
> 
In the `index.html` I was targeting the `default.toml` which currently causes the warning, but I think it makes sense for any `preload` links we add for the `assets` via the `docker-entrypoint`. Just wondering why they are not triggering warnings right now.

> Anyway, I think there is an issue with the server response which add `no-cache` headers to all static resources. All js, css and images should be cached.

Ok, the warning also occurs locally, btw.

PR #266, opened 2022-06-01
(datahub): Improve download links rendering

akhelouat:
Improve the download list panel
- add colors for each type
- sort types by importance
- add a filter
- improve WFS labels


libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -0,0 +1,74 @@
+import { Component, Input } from '@angular/core';
+import { MetadataLinkValid } from '@geonetwork-ui/util/shared';
+import { getBadgeColor } from '@geonetwork-ui/feature/search';
+
+@Component({
+  selector: 'gn-ui-downloads-list',
+  templateUrl: './downloads-list.component.html',
+  styleUrls: ['./downloads-list.component.css']
+})
```
fgravin, line 9: add `changeDetection` `OnPush` to all dumb components.

libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -0,0 +1,74 @@
...
+      [...this.activeFilterFormats, format]
+
+      this.filteredLinks = this.filterLinks(this.processedLinks);
+  }
+
+  isFilterActive(filter: string): boolean {
+    return this.activeFilterFormats.includes(filter)
+  }
+
+  ngOnChanges(): void {
```
fgravin, line 34: prefer `ngOnInit()` method

libs/feature/search/src/lib/utils/links/link-utils.ts
```diff
@@ -3,26 +3,95 @@ import { marker } from '@biesbjerg/ngx-translate-extract-marker'
...
+    mimeTypes: ['application/zip'],
+  },
 }
 
-export function getWfsFormat(link: MetadataLinkValid): string | void {
+export function sortPriority(link: MetadataLinkValid): number {
   for (const format in FORMATS) {
-    for (const alias of FORMATS[format]) {
+    for (const ext of FORMATS[format].extensions) {
+      if ('format' in link && new RegExp(`${ext}`, 'i').test(link.format)) {
```
jahow, line 78: A general question about this PR: why are we doing all these `for (const format in FORMATS) { ...` loops? Can't we just do something like `FORMATS[link.format]`? (maybe I'm missing something)

libs/feature/search/src/lib/utils/links/link-utils.ts
```diff
@@ -70,3 +130,13 @@ export function checkFileFormat(
...
 }
+
+export function getBadgeColor(linkFormat: string): string | void {
+  for (const format in FORMATS) {
+    for (const alias of FORMATS[format].extensions) {
+      if (new RegExp(`${alias}`, 'i').test(linkFormat))
+        return FORMATS[format].color
+    }
+  }
+  return 'var(--color-primary)' // Default color ?
```
jahow, line 141: it might be better to keep this default value in the CSS, and make sure that specific colors will have precedence over it

jahow (commented):
Thanks! Had a few comments, and also noticed some things while testing:

E.g.:
http://localhost:4200/dataset/102de7ff-4066-4d29-b868-d820f91b7565

image

* GPKG does not show up (white background)
* labels should be uppercase (I think)
* clicking on the div does not do anything (only the cloud icon does something)
* "generated by wfs" should also appear in the main block, not only in the title

libs/feature/search/src/lib/utils/links/link-utils.ts
akhelouat: It's due to the structure of the FORMATS object, it is necessary to iterate on its properties and then on the arrays of extensions.

libs/ui/elements/src/lib/downloads-list/downloads-list.component.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(DownloadsListComponent)
+    component = fixture.componentInstance
+    component.links = []
+    fixture.detectChanges()
+  })
+
+  it('should create', () => {
```
fgravin, line 34: zero test while the logic of the component is important, filtering, WFS specific label etc...

fgravin (approved):
So I have fixed the failing tests, I don't really understand what were the blockers ? The failing test just pointed out that some functions or that some file formats were not good. Nothing tricky in there.
There is no test in the only component that you added as well :/

Anyway we don't have more time for this US so we need to move on.
Please fix the lint, the sonarcloud security things, also be sure there is a color for geopackage and that the WFS message is only present once (take the new text from Francois).

@tkohr I didn't see any issue with the observable stream test, what was the use case which you talked about ? do you have a metadata as an exemple ? Thanks

tkohr:
> @tkohr I didn't see any issue with the observable stream test, what was the use case which you talked about ? do you have a metadata as an exemple ? Thanks

@akhelouat added a test without WFS links, in which the observable returned an empty array (I would have kept the test, it was very useful to detect the problem). Anyways, if you remove the two WFS links from https://github.com/akhelouat/geonetwork-ui/blob/GSBHDF-479/libs/feature/record/src/lib/data-downloads/data-downloads.component.spec.ts#L174 and https://github.com/akhelouat/geonetwork-ui/blob/GSBHDF-479/libs/feature/record/src/lib/data-downloads/data-downloads.component.spec.ts#L180 the test also returns an empty array.

If I understand it correctly, the `catchError` will only return `[...otherLinks, ...esriRestLinks]` if an error occured in the `pipe` but not if `wfsLinks` is empty to begin with. So I guess we'll need to recover the `startWith`.

fgravin:
Thanks @tkohr, @akhelouat did you remove that test ? If yes please add it back and I'll take a look to make it passes after you have finished all the other requirements, thanks.

PR #271, opened 2022-06-10
Improvement for static resources, js and css files cache at nginx config

jemacchi:
Improvements for static resources, js and css files cache applied to nginx configuration file.
Reference: https://faun.pub/my-nginx-configuration-for-angular-6f748a4ff683

### DIFF
--- a/tools/docker/nginx.apps.conf
+++ b/tools/docker/nginx.apps.conf
@@ -1,14 +1,30 @@
 server {
-    listen       80;
-    listen  [::]:80;
-    server_name  localhost;
-    location / {
+        listen       80;
+        listen  [::]:80;
+        server_name  localhost;
+
         root   /usr/share/nginx/html;
-        index  index.html index.htm;
-        try_files $uri $uri/ /APP_NAME/index.html;
-    }
-    error_page   500 502 503 504  /50x.html;
-    location = /50x.html {
-        root   /usr/share/nginx/html;
-    }
+
+        server_tokens off;
+
+        location ~ /index.html|.*\.json$ {
+          expires -1;
+          add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
+        }
+
+        location ~ .*\.css$|.*\.js$ {
+          add_header Cache-Control 'max-age=31449600'; # one year
+        }
+
+        location / {
+          try_files $uri $uri/ /APP_NAME/index.html;
+
+          add_header Cache-Control 'max-age=2592000'; # 30 days
+        }
+
+        error_page   500 502 503 504  /50x.html;
+        location = /50x.html {
+            root   /usr/share/nginx/html;
+        }
 }
+


tools/docker/nginx.apps.conf
```diff
@@ -1,14 +1,30 @@
...
+        }
+
+        location ~ .*\.css$|.*\.js$ {
+          add_header Cache-Control 'max-age=31449600'; # one year
+        }
+
+        location / {
+          try_files $uri $uri/ /APP_NAME/index.html;
+
+          add_header Cache-Control 'max-age=2592000'; # 30 days
```
jahow, line 22: doesn't that mean that the index.html file will be cached for a month, effectively preventing the browser from seeing new versions of the app?

tools/docker/nginx.apps.conf
```diff
@@ -1,14 +1,30 @@
...
-        try_files $uri $uri/ /APP_NAME/index.html;
-    }
-    error_page   500 502 503 504  /50x.html;
-    location = /50x.html {
-        root   /usr/share/nginx/html;
-    }
+
+        server_tokens off;
+
+        location ~ /index.html|.*\.json$ {
```
jahow, line 10: I think `/index.html` does not match anything in our case, the file is in `/APP_NAME/index.html` (`APP_NAME` is replaced at build time with e.g. `datahub`). See my other comment!

tools/docker/nginx.apps.conf
jahow: Also for all the other assets this might not be the best idea to cache them for so long; maybe one day would be better?

jahow (commented):
Thanks! Interesting way to optimize caching. I agree it makes sense to have a tailored configuration for the docker image, and also this can serve as reference for people deploying the apps without docker.

Haven't tested yet, I think some parts of the config might not work with our setup, let me know if that makes sense!

tools/docker/nginx.apps.conf
fgravin: generally, we should add hash to all static resources.
indeed this cannot be done for index.html, but at least for css and js, maybe for other assets as well.

jemacchi:
@fgravin @jahow I'm not really an angular developer at all, but certainly, you are right about hashed filenames, and that was the reason I had to take a look at something more specific (some research about angular, deployment, cache, and nginx). 
Basically, I found this thread:  https://faun.pub/my-nginx-configuration-for-angular-6f748a4ff683  and followed some suggestions.

Probably you should take a look too, and decide which could be the best expiration time and if it's good to add or remove details concerning security (I tailored them at the solution I pushed -but you can see them in the thread-, since it was more than what I expected to solve)

The reason for no caching index.html is this:
`"if the index.html file were to be cached by the browser, then when a new version of the application was deployed, users would still see the old version, unless the cache expired or was explicitly cleared"`
based on thread. And probably will not affect geonetwork-ui (in that case we can remove it then)

tools/docker/nginx.apps.conf
jahow: Actually no, it works fine so forget what I said!

jahow (approved):
Actually it's fine like that, and the long cache duration for assets will not be an issue any more if we add hashes to them! so this can be merged as is, thanks @jemacchi

PR #272, opened 2022-06-13
Fix build errors related to npm 8.11, improve CI

jahow:
This PR does:
* use ranges for engine version in package.json
* force npm 8.5.5 for the gh-pages workflow as well
* rework the CI a bit to make it faster (hopefully), using NX recommended steps as described here: https://nx.dev/ci/monorepo-ci-github-actions

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,7 @@
   "name": "geonetwork-ui",
   "version": "0.0.0",
   "engines": {
-    "node": "^16.15.0",
-    "npm": "^8.5.5"
+    "node": ">=14.17.0 <=16.15.0"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,8 +2,8 @@
   "name": "geonetwork-ui",
   "version": "0.0.0",
   "engines": {
-    "node": "^16.15.0",
-    "npm": "^8.5.5"
+    "node": ">=14.17.0",
```
fgravin, line 5: should be equal or inferior as 16.15.0 as well

fgravin (approved):

PR #275, opened 2022-06-16
Add a new map viewer application, improve error handling

jahow:
This PR introduces a new `map-viewer` application which currently does almost nothing.

This was the opportunity to work on the following improvements:
* introduce constraints for dependencies between libs and apps; this is enforced by ESLint and should help us avoid wrong dependencies (e.g. `ui` libs depending on `feature` ones)
* made sure that no lib was depending on the `util/app-config` one; this means that *it is not mandatory* to use an app configuration (the map viewer does not have one); applications are responsible for using a configuration and providing default values
* added error handling in the fuzzy search component and search results component (both smart ones), in case the API is unavailable or returns an error
* map contexts now have their center and extent expressed in lon/lat instead of the view proj (e.g. web mercator)
* added an icon next to the fuzzy search field

Screenshots:

image

image

There is still a lot to do but that at least demonstrates how smart components can be reused in a simple way across different apps.


fgravin (approved):
I browsed the commits and everything looks all good.
Thanks for the fixes, the cleaning, the error state and error handling + the linting rules.

Therefore, the search application does not work very well, the search input throws an error.
The datahub seems to work well though so we can move forward and merge so you can merge the Angular migration and work on this later on.

jahow:
This is part of #337

PR #277, opened 2022-06-21
Fix CI

fgravin:
The PR means to fix the CI.

- Fix storybook build due to refactoring
- Make CI steps clearer
- lint in quiet mode
- Improve doc

Also the PR address some issues about download-links

### DIFF
--- a/libs/util/shared/src/lib/links/link-utils.ts
+++ b/libs/util/shared/src/lib/links/link-utils.ts
@@ -10,6 +10,15 @@ export const FORMATS = {
     color: '#559d7f',
     mimeTypes: ['text/csv', 'application/csv'],
   },
+  excel: {
+    extensions: ['xls', 'xlsx', 'ms-excel', 'openxmlformats-officedocument'],
+    priority: 2,
+    color: '#0f4395',
+    mimeTypes: [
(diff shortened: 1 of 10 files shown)


libs/util/shared/src/lib/links/link-utils.ts
```diff
@@ -39,37 +48,28 @@ export const FORMATS = {
...
-    color: 'var(--color-primary)',
+    priority: 8,
+    color: '#db544a',
     mimeTypes: ['application/vnd.ms-excel'],
   },
   jpg: {
     extensions: ['jpg', 'jpeg', 'jfif', 'pjpeg', 'pjp'],
-    priority: 0,
-    color: 'var(--color-primary)',
+    priority: 8,
```
tkohr, line 69: Shouldn't either `pdf` or `jpg` have priority `9`?

tkohr (approved):
LGTM and no broken imports in CI anymore, thanks @fgravin!

libs/util/shared/src/lib/links/link-utils.ts
fgravin: it has not been expressed so whatever they have same prio ..

PR #278, opened 2022-06-21
Datahub (record-preview): fix icons

tkohr:
PR fixes a small glitch with the icons in the record preview component :

broken_icons


fgravin (approved):
I've noticed that in the download panel as well... i changed the size of the icon.
Don't know what causes this actually

Thanks for the fix

PR #279, opened 2022-06-21
Fix links rendering

akhelouat:
- Fix link sorting
- Add excel format to sort test
- Display links when format is unknown : [GSHDF-563](https://jira.camptocamp.com/browse/GSHDF-563)


libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -54,6 +54,7 @@ export class DownloadsListComponent implements OnInit {
 
   ngOnInit(): void {
     this.processedLinks = this.formatLinks(this.links)
+    console.log(this.processedLinks)
```
fgravin, line 57: to remove

fgravin (approved):
Thanks for the fixes.
Globally, it's always great to write some tests that fail because of the issue, fix the issue and then the test passes green.
You fixes stuff which are not tested, which is not good.

For instance, you should write a test with an unknown format, and check if it's in the DOM.
Before your change it's red, after it's green.
This decrease the technical depth and ensure better test coverage.
Thanks

PR #281, opened 2022-06-23
Remove crossorigin from preloaded images assets

jemacchi:
Adding the crossorigin attribute to preloaded images was wrongly applied in the context of the request credentials mode since it does not match as expected (no Access-Control-Allow-Origin rule and those resources are not external either).

Probably was added when fixing this issue https://github.com/geonetwork/geonetwork-ui/pull/265 (but when it's a toml file requested as fetch, it changes the behavior). 
Found some good intro on this link: https://beamtic.com/crossorigin-attribute-and-preload-not-used


tkohr (approved):
Thanks for the find @jemacchi, seems to work fine. Was not aware that as `fetch` and `image` made a difference for local resources. (Curious that `fetch` even needs `crossorigin` when it's local).

jemacchi:
Excellent ! and I was curious too about that and the reason behind it. Finally, I decided to cut down research time and just push it

fgravin:
Thanks @jemacchi !

PR #282, opened 2022-06-28
Pagination component

akhelouat:
WIP: Added pagination dump component to UI elements GSHDF-552

image


fgravin:
Hi, please add a minimal description and a screenshot when you publish a new component.
Thanks !

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -0,0 +1,36 @@
+<div style="width: 100%; height: 4rem; display: flex; justify-content: center">
```
fgravin, line 1: no hard coded style when useless, use CSS class for this
`w-full, h-something, d-flex` etc...

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -0,0 +1,36 @@
...
+  <div
+    class="capitalize text-white flex justify-center items-center p-3 rounded-lg mr-3 cursor-pointer"
+    style="background-color: #8cc630"
+    translate
+  >
+    datahub.search.nextPage
+  </div>
+  <div style="display: flex; align-items: center" translate>
```
fgravin, line 9: same in all the template

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -0,0 +1,36 @@
...
+    <span translate>datahub.search.page</span>
+    <span class="border border-gray-500 rounded p-4">{{ currentPage }}</span>
+    <span class="m-3" translate> datahub.search.pageOn </span> {{ nPages }}
+  </div>
+
+  <mat-icon>
+    <button
+      (click)="previousPage()"
+      id="navigate_previous"
```
fgravin, line 18: Don't use id for the tests, at least a class.
But you should use css selector on button and take the correct one in the list.

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -0,0 +1,36 @@
+<div style="width: 100%; height: 4rem; display: flex; justify-content: center">
+  <div
+    class="capitalize text-white flex justify-center items-center p-3 rounded-lg mr-3 cursor-pointer"
+    style="background-color: #8cc630"
```
fgravin, line 4: no it's `text-secondary`

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -0,0 +1,36 @@
+<div style="width: 100%; height: 4rem; display: flex; justify-content: center">
+  <div
```
fgravin, line 2: it should be a button
think of accessibility

what if no next page ?

fgravin (changes requested):
Thanks for the component, looks good.
Some comments about the usage of TW and the colors.

How did you test it ?
You must write a story (storybook) like for other ui components, and run `npm run storybook` which refreshes for testing.

I can't test it.
And plz take a screenshot and add it to the PR.
Thanks

libs/ui/elements/src/lib/pagination/pagination.component.html
akhelouat: text-secondary doesn't work, should I add it to the tailwind config?

fgravin: did you test `bg-secondary` ? should work

fgravin (commented):
Thanks I will take over it.

fgravin (approved):

PR #283, opened 2022-06-30
List organisations

tkohr:
PR adds a smart component to query and filter organisations and display them via the UI component `organisation-preview` implemented in a separate PR.

Note: I was wondering if the `search` libs are the aptest place for these components or if `catalog` was more fitting. Also wondering if the `organisations$` Observable should go in the `ngOnInit`, as I'm struggling once again to test this [here](https://github.com/geonetwork/geonetwork-ui/commit/f124ceb12d4a6a51410387a14fac289180fd2cdd#diff-72f4a1f2ffda8716f12604bccf20f1a982017570fc52ceb2ba6c7d840863e199R93).


CONTRIBUTING.md
```diff
@@ -157,7 +157,7 @@ You can also generate other kind of Angular objects with `ng generate directive|
 Recommended options to generate a component are CSS-based styling and OnPush changes detection, like so:
 
 ```shell script
-ng g c smart/fuzzy-search --project=lib-search --style=css --changeDetection=OnPush -d
+ng g c smart/fuzzy-search --project=feature-search --style=css --changeDetection=OnPush -d
```
fgravin, line 160: thanks

libs/feature/search/src/lib/organisations/organisations.container.component.ts
```diff
@@ -0,0 +1,53 @@
...
+})
+export class OrganisationsContainerComponent implements OnInit {
+  constructor(private searchfacade: SearchFacade) {}
+  organisations$: Observable<CatalogOrganisation[]>
+
+  ngOnInit(): void {
+    this.searchfacade.setConfigAggregations({
+      org: {
+        terms: {
+          size: 1000,
```
fgravin, line 21: do you have empty orgs ?
might need to use `exclude` if yes

libs/util/shared/src/lib/models/search.model.ts
```diff
@@ -13,6 +13,12 @@ export interface StateConfigFilters {
   elastic?: any
 }
 
+export interface CatalogOrganisation {
```
fgravin, line 16: why not just `Organisation` just a question.

libs/feature/search/src/lib/organisations/organisations.container.component.html
```diff
@@ -0,0 +1,3 @@
+<gn-ui-organisations-list
```
fgravin, line 1: not sure that this component is usefull.

libs/feature/search/src/lib/organisations/organisations.container.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    })
+    this.searchfacade.requestMoreResults()
+    this.organisations$ = this.searchfacade.resultsAggregations$.pipe(
+      filter((aggs) => aggs.org),
+      map((aggs) =>
+        aggs.org.buckets
+          .map(
+            (organisation) =>
+              ({
+                name: organisation.key,
```
fgravin, line 37: you should add the count as well

fgravin (commented):
Very good,
I am happy that it is that simple, the architecture work seems to pay off.

libs/feature/search/src/lib/organisations/organisations.container.component.ts
tkohr: Right, I forgot about this. Don't have any empty orgs on current catalog, but will add it in case.

libs/util/shared/src/lib/models/search.model.ts
tkohr: My first thought was to easier differentiate it from `MetadataConctact.organisation` which is a `string` and I was hesitating putting the smart component into the `catalog` lib. Maybe `Organisation` makes more sense in the end.

libs/feature/search/src/lib/organisations/organisations.container.component.html
tkohr: Yes, I added it to display something. So if I get it right, the idea is to use the organisation cards and pagination dumb compnents directly in the smart component and not add an intermediate dumb component for listing.

libs/feature/search/src/lib/organisations/organisations.container.component.ts
tkohr: Ok, I'll add it to the model. Do we get this info from the organisation-group binding?

tkohr: Ah no I just see that there is `doc_count` in the aggregation response :-)

PR #285, opened 2022-07-05
storybook: fix translations in stories

fgravin:
Remove demo assets and copy them on build from `translations`/ folder.

You need to import `UtilI18nModule` in your story to trigger the translation process.


tkohr (approved):
Nice fix, thanks @fgravin !

PR #290, opened 2022-07-21
Bind groups to organisations

tkohr:
This PR 

- refactors the organisations querying, moving it from `feature-search` lib where it was implemented using a directive to `feature-catalog` lib, now implemented via a service. This makes organisations available in other parts of the application.
- binds the georchestra groups to the organisations to enrich their information with a description and logo

Todos:

- [x] prevent importing `ui-search` into `ui-catalog`
- [x] write tests

Note: 
- The binding via a "normalized" `name/label/key` currently works on 14 of 135 organisations on the dev instance. For a better binding, these values should match better, as there are not many alternatives to use as a key.
- Descriptions and logos are not yet mapped automatically from georchestra and need to be input in geonetwork for now


libs/feature/catalog/src/lib/organisations/organisations.service.ts
```diff
@@ -0,0 +1,88 @@
...
+      )
+    )
+  }
+
+  normalizeName(name: string): string {
+    return name
+      .normalize('NFD') // decompose graphemes to remove accents from letters
+      .replace(/[\u0300-\u036f]/g, '') // remove accent characters
+      .toLowerCase()
+      .replace(/[^a-z0-9]/g, '') // replace all except letters & numbers
```
jahow, line 86: I think this step is a bit too much; IMO it makes sense to keep parenthesis when matching for example. Maybe just removing accents and making things lowercase could be enough?

libs/util/shared/src/lib/models/search.model.ts
```diff
@@ -20,6 +20,11 @@ export interface Organisation {
   recordCount?: number
 }
 
+export interface OrganisationApiModel {
+  key: string
+  doc_count: number
+}
```
jahow, line 26: This can probably stay in the scope of the organisations service? it might not be useful to anyone else

jahow (commented):
Looking good! I haven't tested it but I think it's a good direction forward, thanks!

jahow (approved):

libs/feature/catalog/src/lib/organisations/organisations.service.ts
tkohr: As discussed, I kept it as is and added some tests to illustrate which string match and which not

tkohr:
Thanks for the quick review @jahow !

I've tested the group binding on some other instances and the results are not great. This might need refinement in the future.

|gn|matched|not matched|
| ------------- |-------------| -----|
|https://apps.titellus.net/geonetwork/|0|523|
|https://metawal.wallonie.be/geonetwork/|0|77|
|https://sextant.ifremer.fr/geonetwork/ |19|422|

PR #291, opened 2022-08-01
[Datahub] Add pagination for organisations component

tkohr:
The PR implements pagination for the `organisations.component`, using the UI `pagination.component` as illustrated on the screenshot below:

pagination


libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -2,17 +2,17 @@
   <button
     (click)="nextPage()"
     [disabled]="currentPage === nPages"
-    class="uppercase text-white flex justify-center items-center p-3 rounded-lg mr-8 cursor-pointer h-14 mr-12 bg-[#8cc630] disabled:bg-gray-100 disabled:cursor-default"
+    class="uppercase text-white justify-center items-center p-3 rounded-lg cursor-pointer h-14 w-44 bg-secondary hidden sm:flex sm:mr-28 disabled:bg-gray-100 disabled:cursor-default"
```
jahow, line 5: Just a remark on the `sm:` classes: I think this is a bit too "context-dependant", as the screen width is not really a strong indication that there is room for the button. To keep in mind if this component has to be reused somewhere!

libs/ui/elements/src/lib/pagination/pagination.component.html
```diff
@@ -2,17 +2,17 @@
...
   <div class="flex items-center" translate>
-    <span class="m-3" translate> datahub.search.page </span>
-    <span class="border border-gray-300 rounded py-1 px-5 mr-3">{{
+    <span class="m-3" translate> pagination.page </span>
+    <span class="border border-gray-300 rounded w-14 py-1 px-5 mr-3">{{
       currentPage
     }}</span>
-    <span class="mr-3" translate> datahub.search.pageOn </span>
+    <span class="mr-3" translate> pagination.pageOf </span>
     <span class="mr-3">{{ nPages }}</span>
```
jahow, line 16: Would it be possible to combine this into one translation key with the total page count as parameter?

jahow (approved):
Thanks, good work! I've added a couple of remarks, but nothing mandatory for this PR.

libs/ui/elements/src/lib/pagination/pagination.component.html
tkohr: Ok, I see your point. What do you recommend for the case that the screen becomes too narrow to display the "next page" button on the left to the pagination buttons ? Letting the pagination buttons wrap below doesn't really look neat IMO.

libs/ui/elements/src/lib/pagination/pagination.component.html
tkohr: Yes indeed, tested it quickly with `[translateParams]="{ pages: nPages }"`, but will keep as is for now as the count param displays as `undefined` during load time when passing it to the translation.

tkohr:
Thanks for the review @jahow !

libs/ui/elements/src/lib/pagination/pagination.component.html
jahow: An endless problem... soon [CSS will allow handling sizes at the block level](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) and not just the page, but for now there is no straightforward way to do this.

A solution might be to leave the button there but make sure that it correctly repositions if there is not enough room, using flex.

PR #293, opened 2022-08-10
Datahub: Sort organisations

tkohr:
The PR adds a component to sort organisations alphabetically by `name` and the number of `publications`. The component also displays an introduction text regarding organisations.

Todos:

- [x] handle mobile

organisations_sort


translations/en.json
```diff
@@ -110,6 +110,12 @@
   "map.select.layer": "Select the layer to display",
   "nav.back": "Back",
   "next": "next",
+  "organisation.sort.intro": "The organisations are...",
```
jahow, line 113: Maybe having an actual text here would be nice, otherwise a vanilla install of the datahub will look a bit weird :)

translations/en.json
```diff
@@ -110,6 +110,12 @@
...
+  "organisation.sort.sortBy": "Sort by:",
+  "organisations.sortBy.nameAsc": "Name A > Z",
+  "organisations.sortBy.nameDesc": "Name Z > A",
+  "organisations.sortBy.recordCountAsc": "Publications (ASC)",
+  "organisations.sortBy.recordCountDesc": "Publications (DESC)",
```
jahow, line 118: Here I'd rather use "Name (ascending)", "Name (descending)", "Records (ascending)" and "Records (descending)", what do you think?

libs/ui/catalog/src/lib/organisations-sort/organisations-sort.component.html
```diff
@@ -0,0 +1,19 @@
+<div
+  class="flex flex-wrap sm:flex-nowrap bg-primary text-white px-6 py-7 mt-4 mb-8 rounded-lg"
+>
+  <span class="flex-grow mb-4 mr-4 sm:mr-16">
+    <p translate>organisation.sort.intro</p>
```
jahow, line 5: the `p` element has a bottom margin which makes the box look weird with longer text; using `m-0` should fix it

jahow (approved):
Thanks! a few comments but overall very good work!

translations/en.json
tkohr: Good point, what do you think of: `The organisations are the providers of metadata records and their corresponding data. They may be owners of metadata catalogs harvested by this instance or owners of (meta-)data referenced by another catalog.` ?

translations/en.json
tkohr: Hum, I followed `Name A > Z` from the mockups (which also seems a little more intuitive to me), but you are right that `ascending/descending` for all four is more consistent and more intuitive that `ASC/DESC`, I'll adapt that

libs/ui/catalog/src/lib/organisations-sort/organisations-sort.component.html
tkohr: I can't reproduce this and don't have any margin on `<p>` due to tailwind. I'll add a `sm:mb-0` to the <span> above it, though, since it makes the margin grow too much. Maybe this is what you mean?

tkohr:
Thanks for the review and test @jahow !

translations/en.json
tkohr: Hum, when testing I'm not a big fan of the repetitive `ascending/descending`, actually. I propose just to replace the `ASC/DESC` by `ascending/descending`.

jahow: makes sense, why not "A → Z" for names and "0 → 9" for records? or "A ➡ Z"? (unicode is useful for symbols which are not icons)

tkohr: ah, good idea, I'll go for that, thanks!

translations/en.json
jahow: Yes that'll do, we can revisit that later on if we have more ideas!

PR #295, opened 2022-08-11
[WIP] datahub: add news feed with latest datasets

jahow:
Replaces #294 

Still requires some work

image


jahow:
There are still bugs related to the multiple search states, I'm working on this but I think this can go in a separate PR. @akhelouat can we merge this? Thanks!

PR #296, opened 2022-08-15
Improve handling of parallel searches

jahow:
This adds a custom operator for making sure that search requests with the _same search id_ will behave correctly, e.g. new requests will cancel previous ones. Requests with different search ids will not interact.


akhelouat:
Everything is good for me I think we can merge it

akhelouat (approved):

PR #297, opened 2022-08-29
[WIP] ISO Metadata converter service

jahow:
This branch contains an experimental service that converts ISO 19139 metadata as XML to an internal metadata model.

This could be used to convert from/to XML records in a frontend editor made with geonetwork-ui, in a backend node-based API.


jahow:
This is it! the converter works well, at least with the current test cases, and is able to alter an existing XML document while preserving its contents as much as possible. The pivot format has been refined along the way and will hopefully work well for other schemas too (e.g. DCAT).

The `simple-editor` app can be used to test it but currently it is very limited and not user friendly at all. I'm not sure how to showcase this in a simpler way, any ideas welcome :slightly_smiling_face:

PR #299, opened 2022-09-01
Fix downloads block

akhelouat:
**Fixed bugs**
- The download block was visible even when there were no download links. 
- The download button was not disabled when there were no download links.
- Translation did not work on the filter button.

image


fgravin:
@akhelouat please always add a description to the PR, we don't understand what was the issue.

translations/en.json
```diff
@@ -82,6 +82,8 @@
   "datahub.header.title.html": "<div class=\"text-white\">Discover open<br> data from my Organization</div>",
   "datahub.search.back": "Back to results",
   "datahub.search.filter.generatedByWfs": "generated by API (WFS)",
+  "datahub.search.filter.all": "All",
+  "datahub.search.filter.others": "Othes",
```
jahow, line 86: typo

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -89,8 +90,12 @@
...
 </div>
 
-<div id="access" class="container-lg px-4 lg:mx-auto">
-  <div *ngIf="displayDownload$ | async">
+<div
+  id="access"
+  class="container-lg px-4 lg:mx-auto"
+  *ngIf="displayDownload$ | async"
+>
+  <div>
```
jahow, line 98: this div is not usefule anymore!

jahow (approved):
A couple of remarks! You can merge after addressing.

Thanks for splitting the changes into various commits. I'm not sure this is really addressing the issue that was initially reported, but let's revisit this if we reproduce it in the future.

PR #301, opened 2022-09-05
search filters

akhelouat:
Add search filters https://jira.camptocamp.com/browse/GSHDF-570


jahow:
Added a functional dropdown component, reworked layout & rebased on master:

image

- [ ] Sort by component still needs some improvement
- [ ] dropdown filter needs a text filter & badges inside it

jahow:
I think this PR is ready to review.

Things that are not addressed here but would make the PR too big:
* the filter dropdowns do not have a search filter in them
* the sort-by dropdown does not look like the mockup one (shown below):
  image
* when no result is found the filter block layout is broken (because of fixed height):
  image
* the filter block should be part of the sticky header and expand on top of the content; currently it is contained in the flow of the page:
  image

These should be addressed in a followup PR

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
```diff
@@ -0,0 +1,85 @@
...
+  (keydown)="handleTriggerKeydown($event)"
+  cdkOverlayOrigin
+  #overlayOrigin="cdkOverlayOrigin"
+>
+  <div class="flex-grow text-left truncate font-medium mr-2">
+    {{ title }}
+  </div>
+  <div
+    *ngIf="hasSelectedChoices"
+    class="flex-shrink-0 rounded rounded-full bg-primary text-white font-bold text-[13px] w-6 h-6 py-[3px] mr-1 selected-count"
```
tkohr, line 16: `rounded` seems superfluous

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
```diff
@@ -1,20 +1,51 @@
...
+      class="w-full"
+      [fieldName]="'OrgForResource'"
+      [title]="'search.filters.byOrganisation' | translate"
+    ></gn-ui-filter-dropdown>
+    <gn-ui-filter-dropdown
+      class="w-full"
+      [fieldName]="'linkProtocol'"
+      [title]="'search.filters.byFormat' | translate"
+    ></gn-ui-filter-dropdown>
+    <gn-ui-filter-dropdown
```
tkohr, line 22: There seems to be a bug with the dropdown that is initially hidden (not sure if related to this fact). The first value checked in the dropdow is not added to the selected list and thus can not be unchecked afterwards.

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
```diff
@@ -1,20 +1,51 @@
...
+  [ngClass]="
+    isOpen ? 'bg-primary text-white h-[204px]' : 'bg-white text-main h-[98px]'
+  "
+>
+  <div class="col-span-2 grid grid-cols-2 gap-4 grid-rows-auto">
+    <div class="col-span-2 flex flex-row items-baseline" *ngIf="isOpen">
+      <h2 class="text-2xl mr-4" translate>search.filters.title</h2>
+      <gn-ui-results-hits class="text-sm opacity-[60%]"></gn-ui-results-hits>
+    </div>
+    <gn-ui-filter-dropdown
```
tkohr, line 12: Selecting values scrolls up the page, which detaches the dropdown selection position from the input.

apps/datahub/src/app/home/search/search-summary/search-summary.component.ts
```diff
@@ -18,6 +18,16 @@ export class SearchSummaryComponent {
     private searchService: SearchService
   ) {}
 
+  isOpen = false
+
```
tkohr, line 22: Maybe `SearchFilterComponent` would be more fitting for this component now?

tkohr (commented):
Nice !!! Thanks a lot for this great feature @jahow @akhelouat! Code LGTM, mainly left some comments when testing.

Was still wondering if the user experience could be improved in some regards like:
- ordering scales in dropdown
- finding a way to display the entire title of a dropdown (select wider than input?)
- pre-process format/protocol labels to be more reader friendly (never ending story, I know)

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
```diff
@@ -1,20 +1,51 @@
...
+  </div>
+  <div class="flex flex-col justify-between">
+    <div
+      class="flex flex-row"
+      [ngClass]="isOpen ? 'justify-end' : 'justify-between'"
+    >
+      <gn-ui-button *ngIf="!isOpen" (click)="open()" type="outline"
+        >...</gn-ui-button
+      >
+      <gn-ui-sort-by [ngClass]="{ 'text-white': isOpen }"></gn-ui-sort-by>
```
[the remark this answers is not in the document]
jahow: yes the dropdown-selector component can be reworked with the same base as the dropdown-multiselect one; I've tried doing it in this PR but it's too much as well

apps/datahub/src/app/home/search/search-summary/search-summary.component.ts
jahow: Good point!

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
jahow: Hmm, could not reproduce... can we try to see if that happens on DEV? might be related to something in your local environment

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
jahow: Good point... will try to fix this

jahow: Ok so I think we'll need #344 for this because by default the cdk-overlay reposition strategy listens to scroll events on the document, but currently the scroll happens inside the home page. I'm leaving this like that for now.

jahow:
> ordering scales in dropdown
    pre-process format/protocol labels to be more reader friendly (never ending story, I know)

Yes there are definitely improvements to be made. Also the way the options are fetched is also not fitting for now I think, and there is no way to clear all filters at once. We should keep track of all these.

> finding a way to display the entire title of a dropdown (select wider than input?)

Not sure what you mean here, the dropdown title will be cut if it's too long but that should not happen with the current setup I think?

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
tkohr: Hum, weird, it only appears for the geographic scale on my side. Let's test on dev, then :-)

tkohr (approved):
Thanks for the adaptions! LGTM, even if I can't get rid of the small bug on my side.

Let's also add responsiveness to the to-do list of the refinement PR :-)

> > finding a way to display the entire title of a dropdown (select wider than input?)
> 
> Not sure what you mean here, the dropdown title will be cut if it's too long but that should not happen with the current setup I think?

I just meant making the dropdown wider than the original button, not sure if this is pretty.

apps/datahub/src/app/home/search/search-summary/search-summary.component.html
jahow: Ok, I've reproduced and fixed the issue, it was because the possible values were numerical!

PR #304, opened 2022-09-12
Datahub: fix focus on search input after scroll

tkohr:
Currently, the focus on search input is broken when scrolling down (which sets the opacity of the tabs div to a negative value but does not hide it). The click events are thus intercepted by the buttons of the tabs div.

This PR sets the tabs div visibility to hidden if its opacity becomes smaller than 0 and allows focusing the search again after scrolling down.

### DIFF
--- a/apps/datahub/src/app/home/home-header/home-header.component.html
+++ b/apps/datahub/src/app/home/home-header/home-header.component.html
@@ -1,7 +1,7 @@
 <header class="h-full px-5" [style.background]="backgroundCss">
   <div class="container-lg h-full mx-auto flex flex-col justify-end">
     <div
-      class="py-8 relative"
+      class="py-8 relative z-50"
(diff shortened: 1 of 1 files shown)


jahow:
Thanks, looks good but I wonder if setting the search input to a higher z-index might have been simpler? besides this solution will work when the header is completely retracted, but there might still be issues when it's e.g. at 80% contracted.

jahow (approved):
Thanks ;)

tkohr:
Scrolling after autocomplete still breaks the sticky header design. Any idea on how to solve this ?

scroll_autocomplete

jahow:
Maybe by using the `updatePosition` method in https://material.angular.io/components/autocomplete/api ?

tkohr:
Indeed, `updatePosition` seems to work. As discussed, closing the panel might an alternative due to performance. I'll create a separate ticket for this and merge the resolved bug.

PR #305, opened 2022-09-13
Datahub: Search datasets by org when clicking org

tkohr:
PR searches datasets by organisation when clicking on an organisation in the organisation tab.

### DIFF
--- a/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
+++ b/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
@@ -1,4 +1,4 @@
-<div class="">
+<div (click)="clickOrganisation()" class="cursor-pointer">
   <div
     class="overflow-hidden bg-gray-100 rounded-lg w-full border border-gray-300 h-36"
   >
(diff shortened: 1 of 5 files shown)


libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="">
+<div (click)="clickOrganisation()" class="cursor-pointer">
```
jahow, line 1: When an element becomes interactive, giving it a click handler and a cursor is usually not enough. Things to do include:
* use a `<button type="button">` element to make sure the element is accessible by keyboard
* add a `title` attribute to describe the action
* provide visual feedback on hover

tkohr: Thanks for the feedback. I've changed the `div` to an `a` to make it accessible by keyboard and added a shadow on hovering as visual feedback.

jahow (approved):
The link between the orgs and the search is a bit unclear to me, maybe at some point we should provide the orgs smart component with a search state id or something. For now that'll do, thanks :)

PR #306, opened 2022-09-13
Misc layout fixes

jahow:
* Datahub: better layout for record preview
* Do not disable favorite toggle after click
* Avoid infinite loop on image fallback directive if fallback url fails to load
* Make thumbnail placeholder to have `object-fit: scale-down` instead of `cover`

### DIFF
--- a/apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
+++ b/apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
@@ -1,9 +1,0 @@
-@media only screen and (min-width: 640px) {
-    gn-ui-favorite-star {
-        top: -1.125em;
-        right: 0.4em;
-    }
-    .icons {
-        right: 0.4em;
-    }
-}
--- a/apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
+++ b/apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
@@ -1,0 +1,9 @@
+@media only screen and (min-width: 640px) {
+    gn-ui-favorite-star {
+        top: -1.125em;
+        right: 0.4em;
+    }
+    .icons {
+        right: 0.4em;
+    }
+}
--- a/libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
+++ b/libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
@@ -1,10 +1,9 @@
 <div class="h-full w-full relative flex-shrink-0 overflow-hidden">
   <img
-    *ngIf="imgUrl"
-    class="relative h-full w-full object-cover object-center"
+    #imageElement
+    class="relative h-full w-full object-center"
+    [ngStyle]="{ objectFit: isPlaceholder ? 'scale-down' : 'cover' }"
     alt="thumbnail"
     [src]="imgUrl | safe: 'url'"
-    [fallbackUrl]="placeholderUrl"
-    gnUiImageFallback
   />
 </div>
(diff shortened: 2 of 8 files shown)


apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
```diff
@@ -0,0 +1,9 @@
+@media only screen and (min-width: 640px) {
+    gn-ui-favorite-star {
+        top: -1.125em;
+        right: 0.4em;
+    }
+    .icons {
+        right: 0.4em;
+    }
+}
```
tkohr, line 9: Why can't we use tailwind classes for these ?

libs/util/shared/src/lib/image-fallback.directive.ts
```diff
@@ -7,7 +7,13 @@ export class ImageFallbackDirective {
...
   @HostListener('error')
   useFallback() {
-    this.el.nativeElement.src = this.fallbackUrl
+    const fallbackUrl = new URL(
+      this.fallbackUrl,
+      window.location.toString()
+    ).toString()
+    if (this.el.nativeElement.src !== fallbackUrl) {
+      this.el.nativeElement.src = fallbackUrl
+    }
```
tkohr, line 16: Just for my understanding, why do we need to add this logic?

libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
```diff
@@ -1,7 +1,9 @@
 <div class="h-full w-full relative flex-shrink-0 overflow-hidden">
   <img
+    #imageElement
     *ngIf="imgUrl"
-    class="relative h-full w-full object-cover object-center"
+    class="relative h-full w-full object-center"
+    [style.object-fit]="isPlaceholder ? 'scale-down' : 'cover'"
```
tkohr, line 6: I guess you will still handle the `scale-down` for the `fallbackUrl` as well, so that the placeholder in the fallback case appears the same?

tkohr (approved):
LGTM so far. I left some comments for clarification. Thanks, @jahow !

libs/util/shared/src/lib/image-fallback.directive.ts
jahow: if the fallback url gives an error (e.g. 404), then the directive is stuck in an infinite loop because it tries to set the fallback url again and again

libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
jahow: yes, that's what I was struggling with; eventually I just removed the fallback directive and did everything in the thumbnail component which made things simpler (see latest commits)

apps/datahub/src/app/home/search/record-preview-datahub/record-preview-datahub.component.css
jahow: good point! it works with tw classes actually, I was sure it didn't!!

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
```diff
@@ -5,7 +5,7 @@
   <gn-ui-star-toggle
     [toggled]="isFavorite$ | async"
     (newValue)="toggleFavorite($event)"
-    [disabled]="loading || (isAnonymous$ | async)"
```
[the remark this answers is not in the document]
jahow: the problem is that it gives a weird user experience; thanks for pointing out this new problem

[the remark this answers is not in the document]
jahow: I made sure that the star was clickable again after the request ended; this gives a slightly better UX; the whole thing is not working correctly anyway because of auth issues with GN...

PR #308, opened 2022-09-16
Regenerate API clients with new openapi version

jahow:
This is a regeneration of the API clients following #302 

```bash
npm run generate-api:datafeeder
npm run generate-api:gn4
npm run format:write
```

This only contains formatting and cosmetic changes.


jahow:
Not essential but it would be good to have the generate-api work again, if we ever need to update the api specs...

fgravin:
What is the status of this ?

> Not essential but it would be good to have the generate-api work again, if we ever need to update the api specs...

What do you mean, `generate-api` is broken ?

jahow:
`npm run generate-api` succeeds but the output is different from the current generated code; the differences make the CI fail.

This is probably not very hard to fix though!

PR #309, opened 2022-09-16
Datahub / Add favorites toggle icon on record view and news feed

jahow:
Please note:
* CSS variables were introduced to customize the appearance of the star toggle without passing arguments through the smart component; let me know if you think this is a bad pattern, personnally I think this works well
* adding the star in the newsfeed records required defining a template outlet in the presentation component, this was quite cumbersome so again, I'm open to suggestions!

image

image


tkohr (approved):
Indeed, the templating to use the `favorite-star` within the `result-list-container` looks quite complicated, but I don't know a better way. Customizing the colors also seems necessary (as we can notice with the default theme). Everything displays well! Thanks @jahow !

PR #310, opened 2022-09-16
Datahub: list favorites

tkohr:
The PR lists favorites in the search results when clicking a favorite badge.

Todos:

- [x] fix CSS
- [x] fix tests in `home-header.component`
- [x] check broken test for `updateRequestAggregationTerm$`

favoris


libs/feature/search/src/lib/state/effects.spec.ts
```diff
@@ -78,6 +78,19 @@ class FavoritesServiceMock {
...
+      kind: 'N',
+      value: {
+        id: 'default',
+        key: 'abc',
+        payload: { abc: {} },
+        type: '[Search] Patch Results Aggregations',
+      },
+      hasValue: true,
+    },
+  }))
```
jahow, line 93: This changes looks weird. I think you can probably just return an object like
```
{
      aggregations: {},
      size: 0,
      query: {}
}
```

And that should do the trick?

jahow (approved):
Very clean PR, thanks a lot! Impressive work. Not tested but since we did a part of it together I think it's fine. Thanks for the very thorough test coverage as well.

I feel like the `bg-primary-opacity-25` tailwind class will not work the way we would expect it (as discussed before), so there might be some more tweaking on the presentation components to do. Also, having a generic button component with a translucent background in the `ui/input` lib might be good! This can be done later on when refining the UI.

libs/feature/search/src/lib/state/effects.spec.ts
tkohr: Good point, thanks for the catch, I adapted this in a last commit 820433dfb9e4b964efff030ce28b1defbe688f6b along with fixing one import that broke when I rebased.

PR #311, opened 2022-09-20
Address NPM dependencies issues and reduce bundle size

### LINKED ISSUE #307: datahub build failure since #276
dunno if there are specific requirements, but with npm 8.1.2/nodejs v16.13.1 the build fails
```
> nx run demo:build:production datahub
'_' is not found in schema
> nx run map-viewer:build:production datahub
'_' is not found in schema
> nx run search:build:production datahub
'_' is not found in schema
> nx run webcomponents:build:production datahub
'_' is not found in schema
> nx run util-data-fetcher:build datahub
Compiling TypeScript files for project "util-data-fetcher"...
Done compiling TypeScript files for project "util-data-fetcher".
> nx run datahub:build:production datahub
'_' is not found in schema
 
 >  NX   Running target "build" failed
   Failed tasks:
   
   - datafeeder:build:production
   - demo:build:production
   - map-viewer:build:production
   - search:build:production
   - webcomponents:build:production
   - datahub:build:production
```
according to buildbot logs it started failing with #276.

the command run by the bot is `npm run build datahub -- --base-href="/datahub/"` after `npm i`.

after looking at the docs, i changed the command for `npm run build -- datahub --prod` but the failure is the same. And also same thing after ditching `node_modules`.

jahow:
This PR contains several modifications related to the project architecture:
* ran `nx migrate latest`, nx is now 14.7.5
* updated several other dependencies to make most of the peer dependencies errors go away
* removed the upper version limit for node (previously 16.15.0); `npm install` should now work for node LTS, albeit with some not critical warnings
* the `ngx-translate-testing` library was copied inside the `util/18n` lib; this was necessary because this dependency was causing peer dependencies issues that simply could not be resolved (see commits for details)
* fixed some incorrect instructions in the docs
* reworked imports to avoid having too many things from OL in the bundle, as well as having test fixtures in the final bundle
* finally, the `xlsx` library (which is a real monster) is now imported lazily, which means that it is not part of the main datahub bundle anymore; **the main datahub bundle went down from 2.96MB uncompressed to 1.56MB with this change**

Fixes #307


libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.ts
```diff
@@ -25,7 +25,9 @@ import {
 } from '@geonetwork-ui/ui/layout'
 import { FEATURE_COLLECTION_POINT_FIXTURE_4326 } from '@geonetwork-ui/util/shared'
 import type { FeatureCollection } from 'geojson'
-import { Feature, Map, View } from 'ol'
+import Map from 'ol/Map'
```
fgravin, line 26: Great nice catch !

fgravin (approved):
Thanks for the migration and the optimisations.
Seems it was not a piece of cake !
Very good work for the bundle size, we should be more careful about that.

PR #312, opened 2022-09-23
Datahub / show all constraints & resource contact in record view

jahow:
Previously not all constraints were shown, and the contact was the contact for the metadata.

Remaining to do:
- [x] make filter by org work when clicking on the contact

image


libs/feature/search/src/lib/utils/mapper/elasticsearch.mapper.spec.ts
```diff
@@ -237,80 +237,92 @@ describe('ElasticsearchMapper', () => {
...
             {
               description: 'Lieu de surveillance (point)',
+              label: 'Lieu de surveillance (point)',
               name: 'surval_parametre_point',
               protocol: 'OGC:WFS',
               url: 'http://www.ifremer.fr/services/wfs/surveillance_littorale',
             },
             {
               description: "Extraction des données d'observation",
+              label: "Extraction des données d'observation",
```
tkohr, line 276: I tried running `npm run format`, but strangely it does not seem to get rid of double quotes here anymore?

libs/feature/router/src/lib/default/services/router-search.service.spec.ts
```diff
@@ -1,19 +1,24 @@
...
   let service: RouterSearchService
+  let routerFacade: RouterFacadeMock
+  let searchFacade: SearchFacadeMock
 
   beforeEach(() => {
-    service = new RouterSearchService(searchFacadeMock, routerFacadeMock)
+    state = { OrgForResource: { mel: true } }
+    routerFacade = new RouterFacadeMock()
+    searchFacade = new SearchFacadeMock()
+    service = new RouterSearchService(searchFacade as any, routerFacade as any)
```
tkohr, line 21: For my understanding, why was it necessary to change the mocks to classes here?

tkohr (approved):
Thanks @jahow ! I'm really not an expert on this, so it's hard for me to give an opinion on the approach. The result looks fine to me. From a user perspective, I was just a little surprised to not find the `isOpenData` badge in the `Plan IGN` dataset, as the license sounds very open.

Note: I think you still have to run `npm run i18n:extract` or push the translations (for the badge).

libs/feature/router/src/lib/default/services/router-search.service.spec.ts
jahow: I tend to change value mocks to class mocks because value mocks can introduce side-effects between tests, e.g. running tests in a different order might yield different results (since the same object is used multiple times and potentially modified)

jahow:
> Note: I think you still have to run npm run i18n:extract or push the translations (for the badge).

Oh yeah, good catch, sorry!!

jahow:
> From a user perspective, I was just a little surprised to not find the isOpenData badge in the Plan IGN dataset, as the license sounds very open.

I know, I'm relying on the `isOpenData` flag provided by GeoNetwork but it might not be super reliable (especially for harvested records). On the other hand, I don't see how we could "guess" that a record is open data... I think we can move on like that and see how it behaves in the real world?

libs/feature/search/src/lib/utils/mapper/elasticsearch.mapper.spec.ts
tkohr: ah, of course I didn't notice the apostrophe

libs/feature/router/src/lib/default/services/router-search.service.spec.ts
tkohr: ok, thanks for the explanation :-)

fgravin (approved):
Thanks
Great improvement for the constraints and the opendata flag.
For the contact it LG

PR #313, opened 2022-09-27
Document limitations re: authenticated requests from gn-ui

jahow:
This should help people have things like favorites (#300) work on their gn-ui deployments

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ You can either try complete applications or showcases of components using the fo
 
 ## More information
 
-### Running GeoNetwork UI
+### Running GeoNetwork UI applications
 
 To run a specific application using a development server, use:
 
@@ -52,7 +52,7 @@ npx nx serve (app_name)
 
 And navigate to `http://localhost:4200/`.
 
-### Build
+### Build GeoNetwork-UI applications
 
 To build a specific application, use:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,6 +62,47 @@ npx nx build (app_name)
...
+                <value>/[a-zA-Z0-9_\-]+/api/search/.*</value>
+                <value>/[a-zA-Z0-9_\-]+/api/site</value>
+       +        <value>/[a-zA-Z0-9_\-]+/api/userselections.*</value>
+              </set>
+            </constructor-arg>
+          </bean>
+       ```
+
+2. GeoNetwork and GeoNetwork-UI are *not* deployed on the same host, e.g. https://my.host/geonetwork and https://another.org/datahub
+   * In this scenario, even if CORS settings are correctly set up on GeoNetwork side, most authenticated request will probably fail because by default they are not sent with the [`withCredentials: true`](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) option
```
fgravin, line 101: Does it mean we need to add interceptors to add this header ?

fgravin (approved):
Very clear, thanks for making the documentation more accurate.

README.md
jahow: No, we just have to add this option in the API client configuration. Currently it is done only in the datafeeder here:

https://github.com/geonetwork/geonetwork-ui/blob/2a695f52614ab26dd36a2a8ebe4c48e621c5088e/apps/datafeeder/src/app/app.module.ts#L43

Maybe we could do it with a setting, or maybe we can enable it for everyone, I'm not sure yet

jahow: I think we could offer a setting like "allow authenticated cross origin requests"; then it would mean that GN would have to be configured to not send `*` in the `Access-Control-Allow-Origin` header, but the actual origin instead, e.g. `Access-Control-Allow-Origin: my-other-site.org`.

This is an optional behaviour that requires tweaking the backend more; the default behaviour would be to not do authenticated requests in a cross origin scenario (e.g. only public search).

PR #316, opened 2022-09-29
Datahub: Add badges for most recent and popular records

tkohr:
PR is still WIP.

Todos:

- [ ] Make sort by `userSavedCount` work.
- [ ] Tests


tkohr:
Thanks for the review @jahow. According to the latest discussions, I removed the logic to display the badges on the search tab only, and they are now present on all tabs. Let's confirm tomorrow during the daily which alternative to merge for now.

PR #319, opened 2022-09-30
Continue organizations refactoring

jahow:
This is a follow-up of #312 to make sure that contacts for the resource and the metadata are used everywhere consistently:
* fix the filter by organization on the orgs page; also reworked the test a little bit to improve readability
* show organizations correctly in the news feed and datahub search


fgravin (approved):
LG thanks

PR #324, opened 2022-09-30
Fix translation for isOpenData

tkohr:
Use translate pipe for string displayed in badge's ng-content.

### DIFF
--- a/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
+++ b/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
@@ -63,8 +63,8 @@
   [title]="'record.metadata.usage' | translate"
 >
   <div class="py-4 px-6 rounded bg-gray-100 text-gray-700 flex flex-wrap gap-2">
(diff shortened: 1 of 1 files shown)


jahow:
Weird, why is the `translate` directive not working there?

tkohr:
> Weird, why is the `translate` directive not working there?

Good question, I just stumbled on other cases where people had this problem when translating things passed into `ng-content`.

fgravin:
> Good question, I just stumbled on other cases where people had this problem when translating things passed into `ng-content`

maybe because you put it on the component tag ? if you add a div in there what happens ?

tkohr:
> > Good question, I just stumbled on other cases where people had this problem when translating things passed into `ng-content`
> 
> maybe because you put it on the component tag ? if you add a div in there what happens ?

You're right @fgravin, thanks, I adapted accordingly

tkohr:
> In general, using a `<span>` for inline text is better; semantically `<p>` means paragraph :) thanks!

Yep, thanks, I adapted accordingly.

PR #325, opened 2022-10-03
Datahub: Open related records in new tab

tkohr:
PR adds `<a>` tag to `related-record-card` and opens records in new tab:

related

### DIFF
--- a/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
+++ b/libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
@@ -1,6 +1,7 @@
-<div
-  class="w-72 h-96 overflow-hidden rounded-lg shadow-primary-light bg-white cursor-pointer"
+<a
+  class="w-72 h-96 overflow-hidden rounded-lg shadow-primary-light bg-white cursor-pointer block"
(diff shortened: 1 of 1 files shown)


libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
```diff
@@ -1,27 +1,28 @@
...
-      class="max-h-24 font-title text-21 text-black overflow-ellipsis overflow-hidden"
-    >
-      {{ record.title }}
-    </h4>
-    <div>
-      <button
-        mat-raised-button
-        [matTooltip]="'tooltip.url.open' | translate"
-        matTooltipPosition="above"
+  <a [routerLink]="['/dataset', record.uuid]" target="_blank">
```
jahow, line 4: why not make the encompassing `div` an `a` instead? with `display-block` it should give the same result I think

jahow (approved):
Thanks! minor comment, feel free to address!

libs/ui/elements/src/lib/related-record-card/related-record-card.component.html
tkohr: Indeed, thanks for the hint!

PR #327, opened 2022-10-04
Data preview / allow inferring a WFS feature type if none provided

jahow:
This PR adds a heuristics when a record contains a WFS link but with no featureType specified, and that the WFS only advertises one feature type in total. In this case and only in this one, the only feature type available will be used.

Also removed some heuristics related to ESRI records that should not be required any more.

Requires https://github.com/geonetwork/core-geonetwork/pull/6587 to have this work for records harvested from ArcGIS catalogs:

image

### DIFF
--- a/libs/feature/record/src/lib/data-view-map/data-view-map.component.spec.ts
+++ b/libs/feature/record/src/lib/data-view-map/data-view-map.component.spec.ts
@@ -103,7 +103,7 @@ const SAMPLE_GEOJSON = {
 
 class DataServiceMock {
   getGeoJsonDownloadUrlFromWfs = jest.fn((url) => of(url + '?download'))
-  getGeoJsonDownloadUrlFromEsriRest = jest.fn((url) => of(url + '?download'))
+  getGeoJsonDownloadUrlFromEsriRest = jest.fn((url) => url + '?download')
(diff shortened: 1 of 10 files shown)


libs/feature/record/src/lib/data-view-map/data-view-map.component.spec.ts
```diff
@@ -430,32 +430,6 @@ describe('DataViewMapComponent', () => {
...
     })
 
-    describe('with a link using ESRI:REST protocol', () => {
-      beforeEach(fakeAsync(() => {
-        mdViewFacade.mapApiLinks$.next([])
-        mdViewFacade.geoDataLinks$.next([
-          {
-            protocol: 'ESRI:REST',
-            name: 'mes_hdf',
-            url: 'https://services8.arcgis.com/rxZzohbySMKHTNcy/arcgis/rest/services/mes_hdf/WFSServer/0',
```
tkohr, line 440: Maybe it could make sense to just change this test to a `FeatureServer` (surprised I don't see one yet)?

jahow: Yes, this is indeed missing, thanks!

PR #328, opened 2022-10-07
Autocomplete: Clear search results

tkohr:
PR emits an empty string as submitted value when clicking the autocomplete clear button in order to clear the search results.


jahow (approved):
Thanks!

PR #329, opened 2022-10-10
Autocomplete: Close suggestions on page scroll

tkohr:
PR now updates the position of autocomplete suggestions when the page is scrolled. Couldn't add the `window` event listener the "angular" way via `HostListener`, `Renderer2` nor `fromEvent`, wondering why...

Todo:

- [ ] remove event listener
- [ ] add a test


tkohr:
Will close, as the misplaced suggestions have been fixed by #344 and are repositioned now when scrolling. Thanks, @jahow !

PR #343, opened 2022-10-13
Datahub: Fix MapStyleService 

akhelouat:
make the mapStyleServcie return a different style depending on the geometric type

Related to [GSHDF-529](https://jira.camptocamp.com/browse/GSHDF-529)

image


fgravin (changes requested):
Thank you @akhelouat 
The rendering is good and the implementation quite clean !
There are some things to improve though, we will handle them because it concerns implementation details of Openlayers styles, which are not easy to understand.
Thanks !

PR #344, opened 2022-10-17
Datahub / Make the document the scroll parent for home & record views

jahow:
This should give a better behaviour on mobile and also allow for inserting a header on top of the whole page (typical geOrchestra use case) while not breaking the general behaviour, specifically the sticky header.


tkohr (approved):
Works as before and seems to make sense for the header integration. Thanks, @jahow !

jahow:
@tkohr I've reworked that to make sure that the document is actually used as a scroll container. Will deploy this on a test instance to make sure it works fine!

jahow:
The infinite scroll results are broken right now, this needs to be investigated. I'll merge this and create another PR to handle the search logic

PR #345, opened 2022-10-19
Datahub: Improve responsiveness

tkohr:
PR aligns content in the header and news feed to fit mobile devices as well.

Todos:

- [x] implement tab navigation bar as a burger menu for mobile
- [x] make search filter responsive (separate PR?)

Note: It could make sense to also rework the sticky header a bit to allow moving down the search bar. The title font size does not correspond exactly to the mockups, neither, but it might make sense to keep it smaller to handle smaller screens than used in the mockups (having a width of 390px).


jahow (approved):
Great work @tkohr! The end result looks and feels great. Nothing much to add really. We can iterate over this to improve UX, @fgravin had a few suggestions but I think this should be merged and tested first.

By the way the `style():...` prefix in commit messages is meant to be used for code style I think, not CSS :slightly_smiling_face: not a big deal though!

tkohr:
Thanks for the feedback @jahow, indeed I always confused `style` commit usage, haha. Would you use `feat` for CSS changes?

jahow:
>Would you use feat for CSS changes?

yep!

PR #347, opened 2022-10-20
WC: fix multiple searches

fgravin:
The issue was in the facade initialization timing.

Also

- Improve the `gn-results-list-multiple` example.
- Improve the preview layouts
- Fix the body `overflow:scroll` in the datahub `body`

### DIFF
--- a/apps/datahub/src/index.html
+++ b/apps/datahub/src/index.html
@@ -25,7 +25,7 @@
     -->
     <!--%PRELOAD_LINKS%-->
   </head>
-  <body>
+  <body class="overflow-scroll">
     <datahub-root></datahub-root>
   </body>
 </html>
--- a/apps/datahub/src/index.html
+++ b/apps/datahub/src/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="overflow-scroll">
   <head>
     <meta charset="utf-8" />
     <title>Datahub</title>
@@ -25,7 +25,7 @@
     -->
     <!--%PRELOAD_LINKS%-->
   </head>
-  <body class="overflow-scroll">
+  <body>
     <datahub-root></datahub-root>
   </body>
 </html>
--- a/apps/webcomponents/src/app/components/gn-results-list/gn-results-list-multiple.sample.html
+++ b/apps/webcomponents/src/app/components/gn-results-list/gn-results-list-multiple.sample.html
@@ -37,6 +37,12 @@
         background-color: var(--color-primary);
         margin: 30px 16px;
       }
+      .first-list {
+        --color-primary: #c40;
+      }
+      h2 {
+        color: var(--color-primary);
+      }
     </style>
   </head>
   <body>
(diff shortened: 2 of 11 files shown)


apps/webcomponents/src/app/components/gn-results-list/gn-results-list-multiple.sample.html
```diff
@@ -37,6 +37,12 @@
         background-color: var(--color-primary);
         margin: 30px 16px;
       }
+      .first-list {
+        --color-primary: #c40;
+      }
+      h2 {
+        color: var(--color-primary);
+      }
```
jahow, line 45: We agree that this shouldn't be needed and that it should be possible to set two `primary-color` attributes on web components, right?

apps/datahub/src/index.html
```diff
@@ -25,7 +25,7 @@
     -->
     <!--%PRELOAD_LINKS%-->
   </head>
-  <body>
+  <body class="overflow-scroll">
```
jahow, line 28: Not tested but I think this should go on `html`, otherwise this won't work well on mobiles (e.g. the address bar won't go away when scrolling)

tailwind.base.config.js
```diff
@@ -72,6 +72,9 @@ module.exports = {
           },
         ],
       },
+      opacity: {
+        45: 0.45,
+      },
```
jahow, line 77: Not a fan of adding variants here, I mean... what if we need `opacity: 55` now? sometimes CSS values have to be tweaked quite precisely to have a good result. And I don't see the added value of having a predefined tailwind class, in the end the generated CSS will have

```css
opacity-45 {
  opacity: 0.45;
}
```

instead of

```css
opacity-[0.45] {
  opacity: 0.45;
}
```

Not a huge difference I think? Or am I missing something?

jahow (commented):
Thanks, I'm not sure about the change on the scroll container (use to be the document and now it's `body`), I think this will break behaviour on mobile. I'm also not a fan of adding new custom tailwind classes in the config, as I don't really see the added value (but I see the added config complexity).

Letting you judge :sunglasses:

apps/webcomponents/src/app/components/gn-results-list/gn-results-list-multiple.sample.html
fgravin: I just illustrate a way to tweak the actual system.
Hopefully, we should improve the system so for WC, it set the css var not on html but in the current WC IMO.
It's just temporary.
Does it answer your question ?

jahow: yes :slightly_smiling_face:

tailwind.base.config.js
jahow: Yes but the config also has various font sizes for example, and in time it might keep growing. Tailwind now offers the possibility to write custom values in brackets which I think it's awesome; if it wasn't possible I would probably use inline css instead, because sometimes you just need to provide a very specific value.

Consistency is important but in my experience CSS is really about tweaking values to get the result just right. Also sometimes UX designers build pixel-perfect designs that does not necessarily rely on standardized margins, paddings, sizes etc. UX > technical considerations :slightly_smiling_face:

PR #349, opened 2022-10-25
Fix gn-search-input WC

fgravin:
`FuzzySearch`:
move facade usage in `ngInit()` to ensure the facade is available in web components.

In the web component, we can specify as an `Input()` the search id, so we can't instantiate the facade before the `ngInit` method


jahow (approved):
Thanks!

PR #350, opened 2022-10-26
Datahub / improve loading of new results, offer "Show more" button

jahow:
This reworks the infinite scroll system internally and also offers the possibility to use a "Show more results" button to manually load new results.

The `ngx-infinite-scroll` library was removed and a new `ViewportIntersector` component was created to make it easier to automatically trigger the loading of new results when the bottom of the list is reached. This should hopefully simplify the codebase.

In the datahub, a "show more" button is now used both in the newsfeed and in the search results.

image

Also fixes a (very bad) error where the newsfeed and search results were actually using the same search facade instance... :sob:


libs/feature/search/src/lib/results-list/results-list.container.component.html
```diff
@@ -6,6 +6,22 @@
...
   ></gn-ui-results-list>
+
+  <ng-container *ngIf="(facade.isLoading$ | async) === false">
+    <div class="show-more" *ngIf="showMore === 'button'">
+      <gn-ui-button type="primary" (click)="onShowMore()">
+        <span translate>results.showMore</span>
+      </gn-ui-button>
+    </div>
+    <div
+      class="show-more absolute left-[0px] bottom-[0px] right-[0px] height-[2px]"
```
fgravin, line 17: you have `inset` classes for that ! maybe only in v3 idk

libs/feature/search/src/lib/results-list/results-list.container.component.html
```diff
@@ -6,6 +6,22 @@
...
+      *ngIf="showMore === 'auto'"
+    >
+      <gn-ui-viewport-intersector
```
fgravin, line 20: smart !

apps/webcomponents/src/app/components/gn-results-list/gn-results-list.html
```diff
@@ -1,4 +1,5 @@
 <gn-ui-results-list-container
   [layout]="layout"
   (mdSelect)="onMdClick($event)"
+  [showMore]='fixed ? "none" : "auto"'
```
fgravin, line 4: we could change the api and let people choose the `showMore` option

libs/feature/search/src/lib/state/search.facade.ts
```diff
@@ -60,6 +60,11 @@ export class SearchFacade {
   constructor(private store: Store<SearchState>) {}
 
   init(searchId: string = DEFAULT_SEARCH_KEY): void {
+    if (this.searchId)
```
fgravin, line 63: does it really matters ?
I mean, how do you interact with a search state if you are outside of the `search-directive` DOM tree ?

fgravin (approved):
So clean !
Way better than the infinite-scroll library.

Thanks for this addition !

libs/ui/inputs/src/lib/viewport-intersector/viewport-intersector.component.ts
```diff
@@ -0,0 +1,58 @@
...
+@Component({
+  selector: 'gn-ui-viewport-intersector',
+  templateUrl: './viewport-intersector.component.html',
+  styleUrls: ['./viewport-intersector.component.css'],
+})
+export class ViewportIntersectorComponent implements OnInit, OnDestroy {
+  @Output() isInViewport = new EventEmitter<boolean>()
+  @Output() entersViewport: Observable<void> = this.isInViewport.pipe(
+    filter((inViewport) => inViewport),
+    map(() => undefined)
```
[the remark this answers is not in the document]
jahow: so that the `entersViewport` and `exitsViewport` emit empty values instead of a boolean

apps/webcomponents/src/app/components/gn-results-list/gn-results-list.html
jahow: yep that would be nice! didn't want to change the API in this PR though

libs/feature/search/src/lib/state/search.facade.ts
jahow: it would have thrown an error before this PR since we were initializing the same facade multiple times, which could happen in the future (nothing else prevents it)

apps/webcomponents/src/app/components/gn-results-list/gn-results-list.html
jahow: Done in the latest commit!

libs/feature/search/src/lib/state/search.facade.ts
fgravin: still you don't answer my question.
it's not that a big deal (even though it's not clean) to have several facades with the same id, in the end the id is just used to target to correct state. If your facade is encapsulated in a DOM tree by the directive, then you can't interact with the state from outside this DOM section (via the facade at least because you can still dispatch some action with the search id but it's not meant to be like that).

let's discuss that later

jahow: the problem that this change is trying to avoid is that a SearchFacade instance is reused by mistake in several places, even though the developer assumed he/she was using different ones; this is exactly what happened to me in the datahub

libs/feature/search/src/lib/results-list/results-list.container.component.html
jahow: I've added a commit that restores the possibility to do `inset-0` and similar, I've always wondered why we wouldn't be able to do that but it was actually just because of the TW config. So I went ahead and also simplified a lot of places by using TW classes instead of CSS with `left: 0`, `top: 0` etc.

@fgravin can you let me know if you think that's OK?

PR #351, opened 2022-10-28
Improve organisation icons display

tkohr:
The PR makes the `thumbnail.component` `objectFit` configurable which is used to better display the organisations icons via `contain`.

org-icons

### DIFF
--- a/libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
+++ b/libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
@@ -30,11 +30,16 @@ export class ThumbnailComponent implements AfterViewInit, OnDestroy {
     this.imgUrl = url || this.placeholderUrl
     this.isPlaceholder = !url
   }
+  @Input() fit: 'cover' | 'contain' = 'cover'
   @ViewChild('imageElement') imgElement: ElementRef<HTMLImageElement>
   imgUrl: string
   placeholderUrl = this.optionalPlaceholderUrl || DEFAULT_PLACEHOLDER
   isPlaceholder = false
   sub: Subscription
+
+  get objectFit() {
+    return this.isPlaceholder ? 'scale-down' : this.fit
+  }
 
   constructor(
     @Optional()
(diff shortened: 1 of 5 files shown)


libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
```diff
@@ -20,16 +21,23 @@ export const THUMBNAIL_PLACEHOLDER = new InjectionToken<string>(
...
   templateUrl: './thumbnail.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ThumbnailComponent implements AfterViewInit, OnDestroy {
+export class ThumbnailComponent implements OnInit, AfterViewInit, OnDestroy {
   @Input() set thumbnailUrl(url: string) {
     this.imgUrl = url || this.placeholderUrl
     this.isPlaceholder = !url
   }
+  @Input() objectFit = ObjectFit.COVER
```
jahow, line 39: I feel like this could just be a choice such as `@Input() fit: 'cover' | 'contain'`

Unsing string unions instead of enums make for less complexity, no need to import stuff elsewhere etc.

libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
```diff
@@ -43,6 +51,10 @@ export class ThumbnailComponent implements AfterViewInit, OnDestroy {
     private changeDetector: ChangeDetectorRef
   ) {}
 
+  ngOnInit() {
+    this.setPlaceholderObjectFit()
+  }
```
jahow, line 56: I don't think it's a good idea to rely on ngInit for something that is essentially a computed value

Maybe you could do a getter, like `get objectFit_() { return this.isPlaceholder ? 'scale-down' : this.fit }`?

jahow (approved):
Thanks! Some remarks to make things simpler, as always ;) feel free to adapt how you see fit

libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
jahow: Also the `scale-down` value is only used for the placeholder so shouldn't be part of the "API" of this component

libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
tkohr: Ah, great hint, I was just thinking in `setters` and did not want to implement two because of the uncertainty which one would be executed first. The `getter` is the solution :-)

tkohr:
Thanks for the valuable feedback (once again) @jahow ! I adapted accordingly.

PR #352, opened 2022-10-28
Allow to prioritize page navigation over map navigation

tkohr:
PR allows prioritizing page navigation over map navigation via an additional input for the `MapContextComponent`. This makes two finger interaction (on mobile) or ctrl + mousewheel (on desktop) necessary to navigate on the map, in order to facilitate page scrolling. It allows setting the parameter according to the application implemented, when using the `MapContextComponent`.

To-do:
- [x] improve UI of displayed message

### DIFF
--- a/libs/feature/map/src/lib/utils/map-utils.service.ts
+++ b/libs/feature/map/src/lib/utils/map-utils.service.ts
@@ -13,12 +13,15 @@ import ImageWMS from 'ol/source/ImageWMS'
 import TileWMS from 'ol/source/TileWMS'
 import VectorSource from 'ol/source/Vector'
 import { Options, optionsFromCapabilities } from 'ol/source/WMTS'
+import { DragPan, MouseWheelZoom, defaults, Interaction } from 'ol/interaction'
+import { platformModifierKeyOnly } from 'ol/events/condition'
 import WMTSCapabilities from 'ol/format/WMTSCapabilities'
 import { from, Observable, of } from 'rxjs'
 import { map } from 'rxjs/operators'
 import { MapContextLayerModel } from '../..'
 import { MapUtilsWMSService } from './map-utils-wms.service'
 import { MetadataLink } from '@geonetwork-ui/util/shared'
+import Collection from 'ol/Collection'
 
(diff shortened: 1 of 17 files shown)


libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -169,4 +172,26 @@ export class MapUtilsService {
...
+          }),
+        ])
+        .getArray()
+    )
+  }
+}
+
+export const dragPanCondition = (event) => {
+  return (
+    (this as DragPan).getPointerCount() === 2 || platformModifierKeyOnly(event)
```
jahow, line 195: since this is an arrow function, `this` is undefined!

jahow: This works:
```
export function dragPanCondition(this: DragPan, event) {
  return this.getPointerCount() === 2 || platformModifierKeyOnly(event)
}
```

jahow (approved):
The change looks good, but I couldn't test it on mobile. There's an error which should be fixed, I'm approving so that you can merge when you think this is ready.

I wanted to add that I feel the current API of the map-context is a bit weird; ideally the map config would be an injectable object that is uncoupled from the `app-config` system, and then making the scroll/pan allowed only with CTRL would be a part of that (just like base layers etc.). But this can be addressed later, e.g. when working on the map viewer app :)

Lastly, I think we should inform the user that they cannot use the map unless they press CTRL or use two fingers, for example with an overlay (there was something like that in the mockups). Not sure if that should be part of this PR or not.

libs/feature/map/src/lib/utils/map-utils.service.ts
tkohr: Thanks, indeed I missed that in the end.

tkohr:
Thanks for the review @jahow, I'm trying to add the user feedback, but still having some troubles (after our discussion).

jahow (approved):
Very good work! I added a commit for adding an opacity transition to the message & simplifying the logic a bit, hopefully making it more readable.

tkohr:
Nice @jahow, thanks for cleaning up the logic and the template!

PR #353, opened 2022-10-28
Datahub / Improve pagination component

jahow:
In this PR:
* improve the UX of the pagination component by relying more on the `gn-ui-button` component
  * this required adding a `light` button style without outline, so the button component was reworked a bit as well
  * also the button component now has its own `buttonClick` output event instead of relying on the native `click` event, which might have caused side effects sometimes
* makes the pagination layout responsive (and actually center the button)
* allows inputting the page number manually
* shows 12 orgs instead of 6 in the organisations page of the datahub

image

### DIFF
--- a/apps/datafeeder/src/app/presentation/components/upload-data/upload-data.component.html
+++ b/apps/datafeeder/src/app/presentation/components/upload-data/upload-data.component.html
@@ -35,7 +35,7 @@
   >
 </div>
 <gn-ui-button
-  (click)="handleUploadBtnClick()"
+  (buttonClick)="handleUploadBtnClick()"
   type="primary"
   extraClass="rounded-full px-20"
   [disabled]="uploading"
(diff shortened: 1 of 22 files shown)


tkohr (commented):
Thanks, @jahow, changes LGTM, just noticed some minor unexpected styling/behavior when testing:

- The organisations component might be missing some padding, as it seems strange to have no space between the "next page" button and the bottom.
- The multiselect boxes in the search filter are too high now.
- The behavior of the number input for setting a page might be a little strange to the user. It currently sets the value to 1 when the value is deleted. Say, we are on page 4, and we want to navigate to page 8. Deleting the 4 and entering 8 results in 18 (or 14 if this is the max). To make this work, currently the value needs to be deleted twice (or highlighted) to get rid of the 1.

jahow:
Thanks for the feedback @tkohr, I've addressed your various (very good) points! the pagination component has very few tests currently unfortunately

tkohr (approved):
Nice, thanks for the adaptions @jahow !

What's the `!` in front of a tailwind class for, btw? Does it correspond to CSS `!important` ?

jahow:
> What's the `!` in front of a tailwind class for, btw? Does it correspond to CSS `!important` ?

Precisely.

PR #355, opened 2022-11-02
Datahub: Update label of burger menu on route changes

tkohr:
PR updates the active label of the (mobile) navigation menu by listening to route changes, instead of setting the label in the component. This is necessary for cases such as executing a search, where tabs are changed on another action.


jahow (approved):
Makes complete sense, good job! Thanks @tkohr

PR #356, opened 2022-11-02
Datahub: Make download filter badges responsive

tkohr:
PR uses flex layout to prevent breaking the download filter badges on small screens:

download-filter

### DIFF
--- a/libs/ui/elements/src/lib/downloads-list/downloads-list.component.html
+++ b/libs/ui/elements/src/lib/downloads-list/downloads-list.component.html
@@ -8,10 +8,10 @@
   >
     record.metadata.download
   </p>
-  <div class="w-[420px] flex flex-wrap justify-between sm:pb-4">
+  <div class="flex flex-wrap justify-start sm:justify-end sm:pb-4">
     <span
       [class.opacity-100]="isFilterActive(format)"
-      class="bg-gray-200 cursor-pointer p-2 mb-2 rounded opacity-50 text-sm"
+      class="bg-gray-200 cursor-pointer p-2 m-2 rounded opacity-50 text-sm"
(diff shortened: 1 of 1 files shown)


libs/ui/elements/src/lib/downloads-list/downloads-list.component.html
```diff
@@ -1,14 +1,17 @@
...
+>
   <p
-    class="font-title text-[28px] text-title font-medium mt-8 mb-6 text-center sm:mt-12 sm:mb-4 sm:text-left"
+    class="font-title text-[28px] text-title font-medium mr-4 pb-4 text-center sm:text-left"
     translate
   >
     record.metadata.download
   </p>
-  <div class="mt-8 mb-6 sm:mt-12 sm:mb-3">
+  <div class="w-[420px] flex flex-wrap justify-between sm:pb-4">
```
jahow, line 11: I think setting a hardcoded width here is not a good option, as the actual size of the buttons will change depending on the language. You could probably leave it in flex and only set the width to full and justify-between when in mobile?

tkohr: You're right, I didn't think of the different labels. `w-full` is what I tried first, but it looks strange on desktop. I'll have a look again.

tkohr: So I adapted using `justify-start/end` depending on screen size, combined with a margin: https://github.com/geonetwork/geonetwork-ui/pull/356/commits/e6c84a69b3e745f045a2a8ad3407fd42a0d4358e

jahow (approved):
Looking much better now, thanks!

PR #357, opened 2022-11-04
Datahub / misc. fixes

jahow:
- Fix console error when loading a record page first
- Fix scroll padding on record page not working any more (since the scroll is now happening on the html element)
- Append commit hash to master docker images, so as to know precisely which commit is deployed


tkohr (approved):
Nice, thanks for the fixes and the hash @jahow !

PR #358, opened 2022-11-07
Map Viewer / Improve button appearance in "Add Layer" dialog

jahow:
The buttons added in #354 were not working after the merge since the event name for `gn-ui-button` has changed to `(buttonClick)`. This PR also improves the appearance by setting a 2-lines max height:

image


jahow:
This needs a review! thanks :)

fgravin (approved):
Thanks, did we have a warning or an error about that ?

fgravin:
> Thanks, did we have a warning or an error about that ?

I mean for having a custom event named `click` ?

jahow:
> I mean for having a custom event named click ?

No because it is theorically allowed, since the event was not managed at all by the ButtonComponent. Basically we just used the standard click event on the gn-ui-button tag, which was a little bit awkward.

Issue #337, opened 2022-10-13
Add a map viewer application


jahow:
This application will offer a basic map viewer with the following features:
* Add layers from a GeoNetwork catalog or third party sources (e.g. OGC endpoints, etc.)
* Manipulate layers (order, opacity, delete)
* Show legend
* Import and export maps using a file format to be determined
* Print the map using e.g. https://github.com/camptocamp/inkmap

PR #360, opened 2022-11-10
Create a web component similar to the Map Viewer app

jahow:
image

No special functionality for now. Planned functionalities are:
* possibility to specify one or several uuid to load layers from these records at startup
* allow adding layers from search results which are shown in another web component

Part of #337

### DIFF
--- a/apps/webcomponents/src/app/components/gn-map-viewer/gn-map-viewer.component.css
+++ b/apps/webcomponents/src/app/components/gn-map-viewer/gn-map-viewer.component.css
@@ -1,0 +1,1 @@
+@import '../../../styles.css';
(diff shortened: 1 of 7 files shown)


fgravin (commented):
Easy !
Seems not working though :smiley: 
https://geonetwork.github.io/geonetwork-ui/map-viewer-add-wc/demo/webcomponents/gn-map-viewer.sample.html

Nothing happens on layer adding.

jahow:
Yep that's because it needs #358... when doing the pagination PR I introduced the `(buttonClick)` event on the gn-ui-button instead of the native `(click)`, and so it broke the map-viewer "add-layer" buttons.

fgravin:
> I introduced the `(buttonClick)` event on the gn-ui-button instead of the native `(click)`

On the way to `buttonClickAndSpin` then :wink:

fgravin:
Ok so if you checked that the WC is working on main you can merge, thanks

jahow:
confirmed :sunglasses:

PR #362, opened 2022-11-10
Rework default map styles to handle geometry types differently

jahow:
This gives better results for lines.

Replaces https://github.com/geonetwork/geonetwork-ui/pull/343

image


fgravin (approved):
Very good,
Thanks for the refactoring

PR #364, opened 2022-11-11
Use the native-map lib instead of the map-context component

jahow:
This branch is an experiment to use https://github.com/jahow/native-map instead of relying on the map-context component inside geonetwork-ui. My hope is that this will eventually simplify the codebase here and factorize recurrent logic related to map layers etc.

To use it, call:
```shell
$ npm link ../path/to/native-map
```

And on native-map side:
```shell
$ npm run build:lib -- --watch
```


fgravin (commented):
How is it going so far ?
I am a bit scared about removing the `MapManagerService` as it was meant to share the map.
Anyway, maybe we won't need any ref on the map but i am skeptical about this for a project like Lux.
Does not work on the datahub though, I think it would be a good usecase to valid the approach.

jahow:
I'm currently trying to replicate the features querying in the lib but I think it's promising; keeping a ref on the OL map breaks encapsulation so I think we should try not to do it as much as possible, and if we really need it we can still use a map sharing service like the map manager.

jahow:
Feature querying works quite well but now we're getting GeoJSON instead of OL features when clicking the map (again, for encapsulation concerns), so it's trickier to change the style of the selected objects. Also the vector layers are not styled in the map.

Peek 12-11-2022 16-34

fgravin:
> again, for encapsulation concerns

Well, `nativeMap` is not map library agnostic, so why returning geojson instead of Openlayers features ?
You want that the host of the lib does not import `ol` at all ?
I mean it's not crazy but maybe it's going to far.

> Also the vector layers are not styled in the map

To me it's a limit, I don't really think it's relevant to define an agnostic styling system. Do you think GeoStyler can do the thing ? It seems to me that we'll lose the style function benefit.

I feel like the usecase more and more looks like the InkMap one. Like the inkmap config and the map context are alsmot the same...

jahow:
> I feel like the usecase more and more looks like the InkMap one. Like the inkmap config and the map context are alsmot the same...

I agree and maybe it's going too far in encapsulation indeed. Maybe the native-map custom element could just be responsible for initializing an OL map and then the application can do whatever it wants with it? Not sure, we'll have to think about this.

jahow:
I adapted the native-map element to allow passing OL styles for vector layers, and to return OL features on click on the map. The last thing that is still broken in gn-ui is the geo-table editor, which relies on finding a feature by its id in a vector source.

fgravin:
Thanks, does it mean now that the map is not reset when you switch the layers in the datahub ?

jahow:
> Thanks, does it mean now that the map is not reset when you switch the layers in the datahub ?

The view is still reset but the base layer is not

jahow:
Replaced by #949

PR #365, opened 2022-11-15
Datahub: Adapt search filter sort by popularity

tkohr:
PR

- uses `-userSavedCount` when sorting by popularity in search filter (and is thus consistent with badge)
- also renames "News" to "Home" (EN/FR) in navigation menu


jahow (approved):
clean & simple, thanks!

PR #366, opened 2022-11-16
Datahub / add a setting for boosting search results by geometry

jahow:
A geojson geometry can optionnally be specified in the config file:
* all records with an extent contained in this geometry will be boosted to the top of the list
* all records which _do not intersect_ the geometry will be excluded from the search

Remaining to do:
- [ ] handle the case where the geojson provided is a feature or a feature collection (pick the first geometry)
- [ ] remove test value in the config file
- [ ] allow defining a separate geometry/endpoint when the user is authenticated (not sure how??)


libs/feature/search/src/lib/state/effects.ts
```diff
@@ -41,14 +42,18 @@ import {
...
 } from './actions'
 import { SearchState, SearchStateSearch } from './reducer'
 import { getSearchStateSearch } from './selectors'
 import { HttpErrorResponse } from '@angular/common/http'
 import { switchMapWithSearchId } from '../utils/operators/search.operator'
 import { FavoritesService } from '../favorites/favorites.service'
+import { Geometry } from 'geojson'
+
+// this geometry will be used to filter & boost results accordingly
+export const FILTER_GEOMETRY = new InjectionToken<Promise<Geometry>>(
```
fgravin, line 54: wonder if it's good to declare the injection token right here.
Might be better to declare it in the module which the file depends on, it will be more readable about "what you can override as an injection token for this module", wdyt ?

conf/default.toml
```diff
@@ -15,6 +15,43 @@ proxy_path = ""
...
+
+# These optional parameters allow changing fonts used in the app
+# main_font = "'My Custom Font', fallback-font"
+# title_font = "'My Custom Title Font', fallback-font-title"
+# fonts_stylesheet_url = "https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;700&family=Permanent+Marker&display=swap"
+
+### SEARCH SETTINGS
+
+# This section contains settings used for fine-tuning the search experience
+[search]
```
fgravin, line 48: > Also made it so that config files without a search section will not generate
> errors (since it only has optional properties)
> 

It is also the case for `[map]` i guess

fgravin (commented):
Good so far.
Good idea to use `@Optional` injections to deal with the config.
I also like the way it's configurable (url or data) and not binded to geOrchestra.

fgravin:
>  allow defining a separate geometry/endpoint when the user is authenticated (not sure how??)

Maybe just launch the request if there is a config, then if the request fails (cause unregistered), then catch it an return `null` ?

libs/feature/search/src/lib/state/effects.ts
jahow: I agree! good point

conf/default.toml
jahow: yep I think so (should address the minimal config test actually)

fgravin (approved):
Thanks for the work, i've reviewed the last commit and it's clean.
I have tested it with a geometry in the config and it works well. Everything is fine.

Did you effectively tested the boost value ? I mean that the correct bbox are shown on top ?
Also why a `filter` instead of a `must` for the `intersect` ?

jahow:
> Did you effectively tested the boost value ? I mean that the correct bbox are shown on top ?

Definitely!

> Also why a `filter` instead of a `must` for the `intersect` ?

Filters do not impact scores unlike `must` and `should`; ES is able to cache their results. See https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-bool-query.html

jahow:
Thanks for the review @fgravin!

PR #369, opened 2022-11-18
Handle sort param in the router

fgravin:
Add abstraction through the `SearchService` for the sort options.
It means that if you use the router search, the sort options is persisted within the url eg `/home/search?_sort=-createDate`

Sort change will write the sort option in the url.
On URL change, search filters and sort options are both extracted from the URL and set to the search state, which trigger 2 searches (one is canceled by the other).

I think the code base address the need, but it might bring new edge cases, maybe the behavior of the badges would need to be redefined precisely.

Hard to test without https://github.com/geonetwork/geonetwork-ui/pull/367 merged.

Note: a change in Dropdown dumb component seems to break the datafeeder, anyway it needs a complete revamp so I don't care for now.

### DIFF
--- a/apps/datahub/src/app/home/header-badge-button/header-badge-button.component.html
+++ b/apps/datahub/src/app/home/header-badge-button/header-badge-button.component.html
@@ -1,8 +1,8 @@
 <button
-  class="items-center justify-center px-4 py-1 text-white rounded transition-all"
-  [class]="toggled ? 'bg-primary' : 'bg-primary-opacity-25'"
+  class="badge-btn"
+  [class]="toggled ? 'hover:text-secondary' : 'bg-primary-opacity-25'"
   (click)="toggle()"
 >
-  <mat-icon *ngIf="icon" class="align-middle">{{ icon }}</mat-icon>
+  <mat-icon *ngIf="icon" class="align-middle mr-2">{{ icon }}</mat-icon>
   <span class="align-middle text-center">{{ label | translate }}</span>
 </button>
--- a/apps/datahub/src/app/home/home-header/home-header.component.html
+++ b/apps/datahub/src/app/home/home-header/home-header.component.html
@@ -16,7 +16,7 @@
         class="text-[18px]"
         (itemSelected)="onFuzzySearchSelection($event)"
       ></gn-ui-fuzzy-search>
-      <div class="flex h-0 py-5" [style.opacity]="-0.6 + expandRatio * 2">
+      <div class="flex h-0 py-5 gap-3" [style.opacity]="-0.6 + expandRatio * 2">
         <datahub-header-badge-button
           [routerLink]="ROUTE_SEARCH"
           class="mr-3"
@@ -26,22 +26,22 @@
           [toggled]="searchFacade.favoritesOnly$ | async"
           (action)="listFavorites($event)"
         ></datahub-header-badge-button>
-        <datahub-header-badge-button
-          class="mr-3"
-          label="datahub.header.lastRecords"
-          [toggled]="
-            (searchFacade.sortBy$ | async) === SORT_BY_PARAMS.CREATE_DATE
-          "
-          (action)="clearSearchAndSort($event, SORT_BY_PARAMS.CREATE_DATE)"
-        ></datahub-header-badge-button>
-        <datahub-header-badge-button
-          class="mr-3"
-          label="datahub.header.popularRecords"
-          [toggled]="
-            (searchFacade.sortBy$ | async) === SORT_BY_PARAMS.POPULARITY
-          "
-          (action)="clearSearchAndSort($event, SORT_BY_PARAMS.POPULARITY)"
-        ></datahub-header-badge-button>
+        <div>
+          <button
+            class="badge-btn bg-primary-opacity-25"
+            (click)="clearSearchAndSort(SORT_BY_PARAMS.CREATE_DATE)"
+          >
+            <span translate="">datahub.header.lastRecords</span>
+          </button>
+        </div>
+        <div>
+          <button
+            class="badge-btn bg-primary-opacity-25"
+            (click)="clearSearchAndSort(SORT_BY_PARAMS.POPULARITY)"
+          >
+            <span translate="">datahub.header.popularRecords</span>
+          </button>
+        </div>
       </div>
     </div>
     <!--
--- a/apps/datahub/src/app/home/home-header/home-header.component.html
+++ b/apps/datahub/src/app/home/home-header/home-header.component.html
@@ -19,7 +19,6 @@
       <div class="flex h-0 py-5 gap-3" [style.opacity]="-0.6 + expandRatio * 2">
         <datahub-header-badge-button
           [routerLink]="ROUTE_SEARCH"
-          class="mr-3"
           *ngIf="isAuthenticated$ | async"
           icon="star"
           label="datahub.header.myfavorites"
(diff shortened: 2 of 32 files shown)


jahow (commented):
Tried it and the badges do not work when clicking them. To reproduce:
* Load http://localhost:4200/home/search
* click on "latest"
* the badge is enabled and immediately disabled; the sortBy param is also added and then immediately removed from the URL

Same thing for the other badge. Also badges behave in a weird way because they appear toggled when the sortBy param has a certain value (date or relevancy), even when there are filters active. And disabling a badge also clears the search.

I feel like badges should not be toggles at all, this is causing too many confusion and technical issues. Just have them be normal buttons, and when clicked a given search preset is applied to the state and voilà!

fgravin:
> Tried it and the badges do not work when clicking them. To reproduce:

This is due to the `routerLink` param on the badges links which just reset the search. It's fixed on the other PR, that's why i said it was hard to test before it's merged, but it works well if the routerlinks are gone.

Yes I noticed that the badges are toggled if you change the sortby from the search results page but I didn't want to touch that cause to me it's another concern, it's how the badges have been designed to be binded to the sort state. I haven't even looked on how it was done.
cf

> I think the code base address the need, but it might bring new edge cases, maybe the behavior of the badges would need to be redefined precisely.

Thanks for the review though.

Please test again now I've merged https://github.com/geonetwork/geonetwork-ui/pull/367 and rebased.

The only bug I see is that a click on a badge does not clear the search filters.
It seems that the router does not handle properly 2 quick route changes eg. a `go` to reset the filters and a `go` with `mergingParams` to set the sort. It merges the params with the old params (the filters :disappointed:  )

fgravin:
> I feel like badges should not be toggles at all, this is causing too many confusion and technical issues. Just have them be normal buttons, and when clicked a given search preset is applied to the state and voilà!

2 things about this.
1. atm there is a default sort (which is the first value of t

fgravin:
Yet another thing.
Before, the first value of the sortby list (revelancy) was the sort by defaut. Now I removed that so the sort can come from the route. It means that there is no default sort if no sort in the route, it means that relevancy is displayed but not applied.

Before this PR as well, default sort was revelancy, a click on a badge changes the sort and when untoggled, the sort is removed from the search (no more default).

All of this things should be addressed correctly as well. I think that
1. we should set a default sort within the state (url) if no
2. we should never allow to have no sort in the request, otherwise the sort input would be dissync

fgravin:
Updates the PR as mentioned:

- [ ] default `_score` sort if no sort in the route (simple implem)
- [ ] remove sort badges toggle
- [ ] add `sortAndFilter` in `searchService` to avoid simultenaous route changes
- [ ] some refactoring

Ready for another review.

libs/feature/router/src/lib/default/state/router.effects.spec.ts
```diff
@@ -161,14 +164,31 @@ describe('RouterEffects', () => {
...
+      })
+      it('dispatches SetFilters and SortBy actions', () => {
+        const expected = hot('(ab)', {
+          a: new SetFilters({ any: 'any' }, 'main'),
+          b: new SetSortBy('-createDate', 'main'),
+        })
+        expect(effects.syncSearchState$).toBeObservable(expected)
+      })
+    })
+    describe('when a sort value in the route', () => {
```
jahow, line 180:
```suggestion
    describe('when no sort value in the route', () => {
```

apps/datahub/src/styles.css
```diff
@@ -29,3 +29,9 @@ html.record-page-active {
     scroll-padding-top: 80px;
   }
 }
+
+@layer components {
+  .badge-btn {
+    @apply items-center justify-center px-2 py-1 bg-primary hover:text-primary text-white rounded transition-all;
+  }
+}
```
jahow, line 37: global CSS :scream: I thought we weren't doing that any more! why not specify this in the datahub header CSS at least?

apps/datahub/src/app/home/home-header/home-header.component.html
```diff
@@ -16,32 +16,31 @@
...
+          </button>
+        </div>
+        <div>
+          <button
+            class="badge-btn bg-primary-opacity-25"
+            (click)="clearSearchAndSort(SORT_BY_PARAMS.POPULARITY)"
+          >
+            <span translate="">datahub.header.popularRecords</span>
+          </button>
+        </div>
```
jahow, line 43: I'm not comfortable having "low level" buttons in a complex component like that (especially since now their styling is specified in the datahub CSS). I think here it would make sense to use `gn-ui-button` at least.

Also these buttons are missing the `type="button"` attribute as well as a title (good examples on why using a common button component helps with accessibility).

apps/datahub/src/app/home/header-badge-button/header-badge-button.component.html
```diff
@@ -1,8 +1,8 @@
...
-  class="items-center justify-center px-4 py-1 text-white rounded transition-all"
-  [class]="toggled ? 'bg-primary' : 'bg-primary-opacity-25'"
+  class="badge-btn"
+  [class]="toggled ? 'hover:text-secondary' : 'bg-primary-opacity-25'"
   (click)="toggle()"
 >
-  <mat-icon *ngIf="icon" class="align-middle">{{ icon }}</mat-icon>
+  <mat-icon *ngIf="icon" class="align-middle mr-2">{{ icon }}</mat-icon>
   <span class="align-middle text-center">{{ label | translate }}</span>
 </button>
```
jahow, line 8: This component is not used anymore, right?

jahow (approved):
So the changes work really well in the datahub, thanks for the hard work! I think the badges are better now. I'm still a bit sad that they are "custom" buttons and not regular gn-ui-buttons, although I understand that it's always harder to use common components all the time.

Letting you judge on what to do before merging! :slightly_smiling_face:

apps/datahub/src/app/home/header-badge-button/header-badge-button.component.html
fgravin: it is still used for the favorites when logged in

apps/datahub/src/app/home/home-header/home-header.component.html
fgravin: Good point, I completely short cut this part.
I will use a gn-ui-button and see if the style is handled properly.

BTW, I would like to rise a point, TW with apply is also a way to define components, at least ui components. We face the overlap in Rennes3D project, where the question is "shall we use a tailwind component or a Vue component for this purpose". That's why i am not completly unfavored to global css TW component.

To be discussed

apps/datahub/src/styles.css
fgravin: Hold on !

apps/datahub/src/app/home/home-header/home-header.component.html
fgravin: Thinking about it again, it appears to me that `gn-ui-button` is almost useless.
Instead of passing a `type` then coding thousands of stuff to style the button, to propagate the event (which is useless btw) and so on, simple TW components like `btn-primary` would have given the same result in a cleaner way (IMO).
At the very least, the `gn-ui-button` could be a proxy to a button with a TW component class (@apply).
To me, all of this should be revamped.

Also for the datahub, there is absolutely no consistency in the styling of the components (buttons, panels, cards, animations, shadow etc...), and TW components are the answer for this IMO. It's way easier to define them instead of tweaking Angular components for styling purposes.

I would add a task for that within the current contract, and keep this PR as it is for now.
Would it be acceptable ?

jahow: Sure! Indeed there are probably many improvements to be done to the buttons, let's talk about that later on.

PR #370, opened 2022-11-18
Datahub: Group format labels

tkohr:
PR is pure WIP so far and just to validate some points in the approach.


jahow:
I think we can close this for now, the approach won't be useable anyway. See #391 for an actual working improvement

PR #371, opened 2022-11-25
Datahub: Add tooltip for favorite star

tkohr:
PR adds a tooltip to the favorite star that proposes to log in for users that are not authenticated yet.

- [x] test component
- [x] make login url configurable


libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -69,6 +85,22 @@ describe('FavoriteStarComponent', () => {
...
+  describe('default login URL', () => {
+    it('should construct a localised GN login URL by default', () => {
+      expect(component.loginUrl).toEqual(
+        '/geonetwork/srv/fre/catalog.signin?redirect=http://localhost/'
+      )
+    })
+  })
+  describe('config login URL', () => {
+    beforeEach(() => {
+      loginUrl = '/cas/login?service='
```
tkohr, line 97: This test doesn't pass yet, since I did not manage to change the value of the `InjectionToken` from one test to the next. Would greatly appreciate any help on this @jahow @fgravin :pray:

tkohr: Yes, this is what I tried to do. Somehow, I don't manage to change the value returned by the factory.

fgravin: Yes it's because the providers are managed once.
See https://stackoverflow.com/questions/58630884/override-a-testbed-provider-for-specific-test

tkohr: Thanks, I stumbled over this as well :-) So, if understand correctly, once `.compileComponents()` has been called the values are frozen, even when using a factory. I've reorganized the test suite a bit to override the provider before.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -1,21 +1,35 @@
...
 import { MetadataRecord } from '@geonetwork-ui/util/shared'
 import { map } from 'rxjs/operators'
 import { AuthService } from '@geonetwork-ui/feature/auth'
+import tippy from 'tippy.js'
+import { TranslateService } from '@ngx-translate/core'
+import { StarToggleComponent } from '@geonetwork-ui/ui/inputs'
+import { Subscription } from 'rxjs'
+import { LANG_2_TO_3_MAPPER } from '@geonetwork-ui/util/i18n'
+
+export const LOGIN_URL = new InjectionToken<string>('loginUrl')
```
jahow, line 24: The default GN4 login URL should be the default value of this token. You can make token default values depend on injected services with the `inject` function: https://angular.io/api/core/InjectionToken#tree-shakable-injectiontoken

jahow: Also I think this token should go somewhere else; it feels quite generic; IMO it should go in `feature/auth`

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -33,17 +47,55 @@ export class FavoriteStarComponent {
   record_: MetadataRecord
   favoriteCount: number | null
   loading = false
+  DEFAULT_GN4_LOGIN_URL = `/geonetwork/srv/${
+    LANG_2_TO_3_MAPPER[this.translateService.currentLang]
+  }/catalog.signin?redirect=`
+  loginUrl = `${
+    this.optionalLoginUrl ? this.optionalLoginUrl : this.DEFAULT_GN4_LOGIN_URL
+  }${window.location}`
```
jahow, line 55: This could go in AuthService maybe? this feels like too much responsibility for this component alone

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -33,17 +47,55 @@ export class FavoriteStarComponent {
...
     private authService: AuthService,
-    private changeDetector: ChangeDetectorRef
+    private changeDetector: ChangeDetectorRef,
+    private translateService: TranslateService
   ) {}
 
+  ngAfterViewInit(): void {
+    this.subscription = this.isAnonymous$.subscribe((anonymous) => {
+      if (anonymous) {
+        tippy(this.starToggleRef.nativeElement, {
```
jahow, line 83: I think next time we need a tooltip we should be careful to factorize this into a separate directive/component, to keep in mind for later :)

translations/en.json
```diff
@@ -105,7 +105,7 @@
   "facets.block.title.tag": "Keywords",
   "facets.block.title.tag.default": "Tag",
   "facets.block.title.th_regions_tree.default": "Regions",
-  "favorite.not.authenticated": "You must be authenticated to add favorites",
+  "favorite.not.authenticated.tooltip": "<div><a href=' {link} '>Login</a> to access this feature</div>",
```
jahow, line 108: why the spaces around `{link}`?

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -33,17 +47,55 @@ export class FavoriteStarComponent {
...
+  @ViewChild(StarToggleComponent, { read: ElementRef })
+  starToggleRef: ElementRef
+  subscription: Subscription
 
   get hasFavoriteCount() {
     return this.favoriteCount !== null
   }
 
   constructor(
+    @Optional()
```
jahow, line 71: Once the token is given a default value the `@Optional` decorator is not needed any more

apps/datahub/src/styles.css
```diff
@@ -35,3 +35,10 @@ html.record-page-active {
     @apply items-center justify-center px-2 py-1 bg-primary hover:text-primary text-white rounded transition-all;
   }
 }
+.tippy-box a {
+  @apply text-primary-lighter;
+  text-decoration: underline;
+}
+.tippy-box a:hover {
+  @apply text-primary-lightest;
+}
```
jahow, line 44: Once we create a separate directive/component we should remember to put this into its stylesheet! (to be done later)

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
jahow: Also I think simply appending the `window.location` value to the login url here is too "magical". What we need to do is clearly offer to administrators the possibility to put a token inside the url in the app config, such as:

`/cas/login?service=${current_url}` (similar to tokens available in the external viewer settings)

Also the default value should be:

`/geonetwork/srv/${lang3}/catalog.signin?redirect=${current_url}`

so we also need to evaluate this token. Tokens should be evaluated on request, with something like that on AuthService:

```
get loginUrl() {
  return this.baseLoginUrl.replaceAll('${current_url}', window.location.toString()).replaceAll('${lang3}', ...)
}
```

jahow (commented):
Thanks for the great work and clean contribution. I made a few suggestions mainly related to code organization with the objective to keep the favorite toggle component as simple as possible.

Please let me know if you need any clarifications! thanks.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
tkohr: I totally agree and moved the URL logic to the auth service. Couldn't quite get the default value of the token working, but the service decides whether to use the token or a default value.

translations/en.json
tkohr: Somehow, the `translateService.instant()` function with a param does not work without the spaces :thinking:

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
tkohr: Couldn't get the default value to be provided by the token, so I kept it `@Optional`

tkohr:
Thanks for the valuable input @jahow! I moved the url logic to the `auth.service`, which indeed makes the component lighter.

jahow (approved):
Thanks, this is looking really good now! Good job :)

PR #372, opened 2022-11-25
Search / do not offer a "load more" button if end of results is reached

jahow:
This simply relies on the `isEndOfResult$` observable on the facade. Also made the button look better:

image

### DIFF
--- a/libs/feature/search/src/lib/results-list/results-list.container.component.html
+++ b/libs/feature/search/src/lib/results-list/results-list.container.component.html
@@ -8,9 +8,15 @@
   ></gn-ui-results-list>
 
   <ng-container *ngIf="(facade.isLoading$ | async) === false">
-    <div class="show-more" *ngIf="showMore === 'button'">
-      <gn-ui-button type="primary" (buttonClick)="onShowMore()">
-        <span translate>results.showMore</span>
+    <div class="show-more mt-[30px]" *ngIf="showMore === 'button'">
+      <gn-ui-button
+        type="secondary"
+        (buttonClick)="onShowMore()"
+        extraClass="lg:m-auto !p-[22px]"
+      >
+        <span class="uppercase font-medium tracking-widest" translate
+          >results.showMore</span
+        >
       </gn-ui-button>
     </div>
     <div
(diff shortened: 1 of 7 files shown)


libs/feature/search/src/lib/results-list/results-list.container.component.html
```diff
@@ -8,9 +8,15 @@
...
 
   <ng-container *ngIf="(facade.isLoading$ | async) === false">
-    <div class="show-more" *ngIf="showMore === 'button'">
-      <gn-ui-button type="primary" (buttonClick)="onShowMore()">
-        <span translate>results.showMore</span>
+    <div class="show-more mt-[30px]" *ngIf="showMore === 'button'">
+      <gn-ui-button
+        type="secondary"
+        (buttonClick)="onShowMore()"
+        extraClass="lg:m-auto !p-[22px]"
```
tkohr, line 19: Could it be an option to center the button on small screens, as well ? In particular, on the timeline, the button is currently "pressed" to the left border of the screen (without any margin).

tkohr (approved):
Thanks, @jahow ! Button is now only displayed if more results are present :+1: 

The button location might be improved a little for small screens as mentioned in comment. Besides that, I noticed a small detail, that the button is not clickable before another click event has been dispatched, when the result list first includes less than 10 results, due to a filter and then more than 10 results, due to another filter (=> button rendered on existing list). A first click on the button makes it clickable on the second click in this case. Seems to be something related to change detection.

libs/feature/search/src/lib/results-list/results-list.container.component.html
jahow: good point, I thought it would be better to have it on the left but it doesn't look that good actually, centered is fine!

jahow:
Thanks for the review! I could not reproduce the issue with the button being not clickable. Could you maybe capture a video of it? (e.g. using https://github.com/phw/peek)

tkohr:
I hope it becomes cleared with the screen capture. First filter choice less than 10 results, second filter choice more than 10 results => need to click button twice. After all, it's not a huge thing.
load_more

jahow:
oh ok, that's because the filter dropdown is still open so you need to click on the (invisible) backdrop to close it. There is an improvement to do here but it's related to the filter dropdown, so I'm going to merge this PR. Thanks for pointing that out though!

tkohr:
Ah of course, good point, thanks for the explanation

PR #373, opened 2022-11-29
Clarify optional sections in app config

jahow:
This PR fixes an error at startup in the Datahub, if the `[search]` section of the app config is not defined (which is valid). I also clarified which sections were optional, and hopefully this should help us avoid this in the future.


tkohr (approved):
Thanks for the fix / improvement @jahow !

PR #374, opened 2022-11-30
Datahub: Improve map navigation 

tkohr:
The PR allows using the `DragPan` tool via pointerdown events on desktop and thus prevents displaying the "map mute" message when clicking on the map on desktop.

One further improvement could be to prevent, displaying the message (for the initial 2 sec) when using 2 fingers on mobile. This is currently the case, because a two finger touch dispatches a first `pointerdown` event with one `activePointer` and only a second `pointerdown` event with two `activePointers`. => Implemented in e7b02160671af8a7d853993cc30f35be6f706cc7

### DIFF
--- a/libs/feature/map/src/lib/utils/map-utils.service.ts
+++ b/libs/feature/map/src/lib/utils/map-utils.service.ts
@@ -14,7 +14,7 @@ import TileWMS from 'ol/source/TileWMS'
 import VectorSource from 'ol/source/Vector'
 import { Options, optionsFromCapabilities } from 'ol/source/WMTS'
 import { DragPan, MouseWheelZoom, defaults, Interaction } from 'ol/interaction'
-import { platformModifierKeyOnly } from 'ol/events/condition'
+import { mouseOnly, platformModifierKeyOnly } from 'ol/events/condition'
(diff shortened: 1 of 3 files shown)


libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -193,10 +193,10 @@ export class MapUtilsService {
 
 export function dragPanCondition(
   this: DragPan,
-  event: MapBrowserEvent<UIEvent>
+  event: MapBrowserEvent<PointerEvent>
 ) {
   const dragPanCondition =
-    this.getPointerCount() === 2 || platformModifierKeyOnly(event)
+    this.getPointerCount() === 2 || event.originalEvent.pointerType === 'mouse'
```
tkohr, line 199: Wondering how we could only listen to the last event for the 2 finger condition here, as on mobile a two finger touch dispatches a first `pointerdown` event with one `activePointer` and only a second `pointerdown` event with two `activePointers`.

tkohr: Also implemented in the following commit. Thanks for the help @jahow !

jahow (approved):
Well done!

PR #386, opened 2022-12-09
Datahub: Do not display map message on 'singleclick' for mobile

tkohr:
PR cancels displaying the "muted" map message on a `singleclick` for mobile. Just wondering if we could perhaps already exclude this case within the condition in the `map-utils.service`. Raised the delay to 300 ms since the map's `singleclick` has a delay of 250 ms itself.


jahow (approved):
Well done thanks!

PR #387, opened 2022-12-13
Datahub: Ignore search params content after a '?'

tkohr:
Not sure if this is the right way to go. This PR ignores everything after a `?` within a search param. This would allow adding things such as `?login` to the url without breaking the search, but could break search params that intentionally include a `?`, if such exist !? Also adds logic which should perhaps not be handled by the frontend.


jahow:
Hmm, not a big fan, this is adding complexity to the code for something that is very georchestra-specific.

tkohr:
I agree, let's go with #390

PR #388, opened 2022-12-13
filters: fix empty dropdowns

fgravin:
The `filter-dropdown` component updates ES aggregations payload to be hydrated.
But it does not request a new search. So depending on the order of things loading within the app, it may provides empty filter list, until a new search is done.

This PR fixes this issue by request a new search on each aggregation update.

### DIFF
--- a/apps/datahub/src/app/home/search/search-filters/search-filters.component.ts
+++ b/apps/datahub/src/app/home/search/search-filters/search-filters.component.ts
@@ -1,5 +1,5 @@
 import { ChangeDetectionStrategy, Component } from '@angular/core'
-import { SearchFacade, SearchService } from '@geonetwork-ui/feature/search'
+import { SearchFacade } from '@geonetwork-ui/feature/search'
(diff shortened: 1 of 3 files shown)


jahow (approved):
This looks like a temporary fix until we do a separate request for getting all terms; also if many filters are present this will cause one request per filter, which might become too much. For now I think we can live with that

fgravin:
> This looks like a temporary fix until we do a separate request for getting all terms; also if many filters are present this will cause one request per filter, which might become too much. For now I think we can live with that

Maybe it shows off the limit of the actual system.
Actually, a `setAgg` must trigger a new search, as a `setSortBy` or a `setFilters`, it's the same. So we need to improve the system, and not only for the facets, to have a pool of changes and flush them when needed.

As at least aggs will be managed separately soon, i merge this one until the refactor.
Thanks

jahow:
Yes agreed, it feels like aggregations should have their own separate system with batching of requests to optimize traffic

PR #389, opened 2022-12-13
Filters panel: rework layout

fgravin:
- Remove resolution filter from filter panel
- Hide more button if no spatial filter
- Homogenize org & filter primary panels


jahow (approved):
Looks and works great, thanks!

PR #390, opened 2022-12-14
Datahub Login tootip: add a special case for georchestra login URL

jahow:
Login urls in geOrchestra rely on a "login" query param, but appending query params is awkward with the current login_url system; we might typically end up with something like:

https://my.georchestra/datahub/?org=Abc&keyword=Def?login

This PR handles this specific case to obtain:

https://my.georchestra/datahub/?org=Abc&keyword=Def&login


fgravin (approved):

tkohr (approved):
Thanks @jahow, looks like a much better alternative to #387 to me and works fine :-)

PR #391, opened 2022-12-14
Search: use the 'format' field from index instead of 'linkProtocol'

jahow:
The "formats" dropdown in the Datahub now looks like this:

image

It used to be like that:

image


tkohr (approved):
Thanks, already looks better to me :-)

PR #392, opened 2022-12-14
Improve filters panel

fgravin:
- [x] Add a clear button per filter
- [x] Add a clear button (only if the panel is opened) that only clears the filters within the panel
- [x] Improve filter selection count layout (from mockup)
- [x] Improve inner selection badges (alignement)
- [x] Improve result hits rendering

Screenshot from 2022-12-14 18-18-44


tkohr (commented):
Thanks for the UX refinement and adding the clear buttons @fgravin !

When testing, I just noticed some incoherences: 
- on mobile the 'reduire' button is broken
- on mobile the 'réinitialiser' button is also displayed when filters are minimized
- no 'agrandir' button when authenticated and thus no geo checkbox (maybe I missed something), nor 'réinitialiser' button
- badges are missing padding when multiple lines
- counter is missing opacity

fgravin:
Thanks for the review
All the changes i've made may broke the mobile
I think we should do a mobile revisit of the whole ui following all the recent changes

the results hits panel will have a gray background instead (seen with lucile)

tkohr (approved):
Ok, I'm just wondering if it wasn't faster to handle the mobile right away. I let you decide.

Also, I think it's a little shame the 'reinitialiser' button will not be displayed on desktop (I know the mockup is like that).

Maybe the padding is worth fixing here?

fgravin:
yes I won't merge it as it is, i will take your comments, or maybe could you handle them ?

tkohr:
Yep, I can have a look

fgravin:
> * no 'agrandir' button when authenticated and thus no geo checkbox (maybe I missed something), nor 'réinitialiser' button

from lucile:

- reset filter**s** button only when filters are opened, when closed, you can manually reset each filter
- checkbox and advanced menu only appears if you have a geometry, dunno actually how to enable that, haven't checked, but it's not binded to "am i connected or not"

fgravin:
> * on mobile the 'réinitialiser' button is also displayed when filters are minimized

yes but it's good I think, it should be accessible as no other direct way to reset a filter is available

fgravin:
> * on mobile the 'reduire' button is broken

i don't see it broken ?

Screenshot from 2022-12-15 11-30-43

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
```diff
@@ -8,15 +8,22 @@
...
+      *ngIf="hasSelectedChoices"
+      class="shrink-0 rounded-full bg-primary text-white font-bold text-[12px] w-5 h-5 flex items-center justify-center mr-1 selected-count"
+    >
+      {{ selected.length }}
+    </div>
   </div>
-  <div
-    *ngIf="hasSelectedChoices"
-    class="shrink-0 rounded-full bg-primary text-white font-bold text-[13px] w-6 h-6 p-[6px] mr-1 selected-count"
+  <mat-icon
```
jahow, line 20: This won't be accessible, it should be a `<button>`

translations/fr.json
```diff
@@ -95,11 +95,12 @@
...
-  "editor.dashboard.myDraft": "",
-  "editor.dashboard.myOrg": "",
-  "editor.dashboard.myRecords": "",
-  "editor.dashboard.mySpace": "",
+  "editor.dashboard.catalog": "Catalogue",
+  "editor.dashboard.myDraft": "Mes brouillons",
+  "editor.dashboard.myLibrary": "Ma bibliothèque",
+  "editor.dashboard.myOrg": "Mon organisation",
+  "editor.dashboard.myRecords": "Mes fiches publiées",
+  "editor.dashboard.mySpace": "Mon espace",
```
jahow, line 103: Same, not relevant to this PR

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
tkohr: I've added a button tag around it

tkohr:
I've added two commits for the CSS adaptions and accessibility. Left the appearance of 'reinit' button as is, following discussion.

When I re-run `npm run i18n:extract` nothing's modified in the translations, so I guess they were missing from somewhere else.

tkohr:
The problem is actually not due to this PR, but since there is no 'maximize' button on desktop anymore, the filter by geometry checkbox introduced in https://github.com/geonetwork/geonetwork-ui/pull/366#issuecomment-1318559203 is not available on desktop anymore. I guess we'd need to condition "if authenticated and filter_geometry_data config present" display 'maximize' button to fix this.

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
```diff
@@ -64,7 +66,7 @@
         class="max-w-full bg-main text-white rounded pr-[7px] pl-2.5 flex gap-2 items-center opacity-70 hover:opacity-100 focus:opacity-100 transition-opacity"
         (click)="select(selected, false)"
       >
-        <div class="text-sm truncate leading-[19px]">{{ selected.label }}</div>
+        <div class="text-sm truncate leading-[26px]">{{ selected.label }}</div>
```
fgravin, line 69: why ?
to me it's not well aligned with this value
Screenshot from 2022-12-15 14-58-53

tkohr: Try to select enough values to wrap to multiple lines ;-)

fgravin: it's the line-height of the label in the tag.
to me it's not aligned with the cross circle. it's not about the amount of selected value in the input is it ?

tkohr: Maybe it should be fixed differently, but yes if you select a high amount of values for the input, so they are wrapped over multiple lines, the badges were stashed together with smaller line height. As long as they stay in one line I actually do not see a difference in the alignment.

PR #394, opened 2022-12-15
Record: force "application/json" format when loading WFS data

jahow:
This makes sure that we're not using `application/geo+json` which is advertised by GeoServer sometimes but not actually allowed in responses.

Also bumps ogc-client version to include this: https://github.com/camptocamp/ogc-client/pull/3


fgravin (approved):
Not tested

tkohr:
Thanks, @jahow ! Tested it with a bunch of different server (didn't find a mapserver though) and works fine for the map.

Just wondering if we should also remove `geojson` from the download list, as the download from http://localhost:4200/dataset/dc720470-bf16-4740-b665-6b1a7fbb4a33 (cf. ticket) still breaks. That would also mean that working `geojson` downloads that do not provide a `json` download would not be displayed anymore (like http://localhost:4200/dataset/a32f267e-7711-48a7-9426-5db40df64ec9).

jahow:
> Just wondering if we should also remove geojson from the download list, as the download from http://localhost:4200/dataset/dc720470-bf16-4740-b665-6b1a7fbb4a33 (cf. ticket) still breaks. That would also mean that working geojson downloads that do not provide a json download would not be displayed anymore (like http://localhost:4200/dataset/a32f267e-7711-48a7-9426-5db40df64ec9).

Indeed the issue is that a download with that format is still offered to users. To me this is not fixable in the datahub, and the application/geo+json format is wrongly advertised and shouldn't be.

PR #395, opened 2022-12-15
Datahub: use localized sort for organization names

jahow:
This is important in languages with accents, e.g.:

image


fgravin (approved):
Indeed !
And it's in my hiring exercice :D

PR #396, opened 2022-12-19
Filter dropdown: Get aggregations via a service instead of state

tkohr:
PR is just a quick test to detach the filter values from the state to always display the entire list. Currently, the filter dropdown only displays the active filter values (if there are any in the URL, when the search page is loaded).  I think there has already been discussion on this (cf. #388). To be discussed if this is the right approach.


libs/feature/search/src/lib/filter-dropdown/filter-dropdown.component.ts
```diff
@@ -46,10 +35,24 @@ export class FilterDropdownComponent implements OnInit {
 
   constructor(
     private searchFacade: SearchFacade,
-    private searchService: SearchService
+    private searchService: SearchService,
+    private aggregationsService: AggregationsService
   ) {}
 
   ngOnInit() {
+    this.choices$ = this.aggregationsService
```
fgravin, line 43: It would be better to aggregate the aggregations into one api call, but I think this could be good for now.
Good call

libs/feature/search/src/lib/utils/service/aggregations.service.ts
```diff
@@ -0,0 +1,48 @@
...
+                  size: 1000,
+                  field: fieldName,
+                  order: {
+                    _key: 'asc',
+                  },
+                  exclude: '',
+                },
+              },
+            },
+            0,
```
fgravin, line 34: all the parameters are optionnal

libs/feature/search/src/lib/utils/service/aggregations.service.ts
```diff
@@ -0,0 +1,48 @@
...
+    private esService: ElasticsearchService,
+    private searchApiService: SearchApiService
+  ) {}
+
+  getAggregation(fieldName: string): Observable<any> {
+    return this.searchApiService
+      .search(
+        'bucket',
+        JSON.stringify(
+          this.esService.getSearchRequestBody(
```
fgravin, line 21: Seems the only added value of the method `esService.getSearchRequestBody` is to add the `template: 'n'` filter.
I think BTW that this should be refactored cause it shouldn't be hardcoded in this method.
Can be done for the editor search maybe

libs/feature/search/src/lib/utils/service/aggregations.service.ts
```diff
@@ -0,0 +1,48 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class AggregationsService {
+  constructor(
+    private esService: ElasticsearchService,
+    private searchApiService: SearchApiService
+  ) {}
+
+  getAggregation(fieldName: string): Observable<any> {
```
fgravin, line 16: I don't think the name of this method is meaning full. 
There are tons of way to fetch ES aggregations and your payload is hardcoded here.

Your method should be `getFullSearchWithoutTemplatesTermAggregationsOrderedByKey` :)
I don't know what should be the strategy here, you should be able to pass the ordering key at least.

Note that aggregations can be something else than `terms`

libs/feature/search/src/lib/filter-dropdown/filter-dropdown.component.ts
```diff
@@ -46,10 +35,24 @@ export class FilterDropdownComponent implements OnInit {
...
+      .getAggregation(this.fieldName)
+      .pipe(
```
fgravin, line 45: I wonder if the pipe content shouldn't be part of the service method ?

fgravin (changes requested):
Thanks @tkohr  for starting to handle that, I think this is a quick win and we should start with this strategy.
Nevertheless, the implementation needs a bit more care.

libs/feature/search/src/lib/filter-dropdown/filter-dropdown.component.ts
tkohr: I would tend to not put it in the same method, as it could be reused (by the `OrganisationsService` for example). I added a commit for that to prevent code duplication. Then again, I'm not sure if architecturally the `feature/search` dependency in `feature/catalog` is disturbing, but I don't think so.
For the pipe content I could create a separate method in the service, but I'm not convinced it's really the better place as we are building the label of the component here.

libs/feature/search/src/lib/utils/service/aggregations.service.ts
tkohr: Hum, good point that helped my understanding:-) I tried to find a compromise.

tkohr:
I've continued the PR addressing your comments yesterday @fgravin and will let you have a second look. (The red cross just seems due to a cancelled CI run, when I marked the PR ready for review, the CI looks fine)

fgravin (approved):
Thanks for the refactoring.
It works well !
I improved some part of it and added several minor ui fixes at the same time :)

I let you review and merge if it's ok, thanks.

tkohr:
Thanks for the final touch in this @fgravin !

PR #397, opened 2022-12-20
Search filters: Display spatial filter via a new check toggle

tkohr:
~~The goal of this draft PR is to add a condition to the search filters to display the spatial filter only if~~

~~- a `filter_geometry` is present in the config (already implemented before this PR)~~
~~- and the user is connected~~

Edit: After discussion below, I reworked the PR to just add a new `check-toggle` UI component for the toggle.


fgravin:
I don't feel this is the right way to go, but I don't know what was the purpose of @jahow by  adding the ability to configure a static json file for the geometry.

To me

- if there is a static file, or a public url for the geometry configuration, then you don't have to be indentified
- if there is a url specific for the user, then the fetch should fail in anonymous so the toggle geometry should not appear

To me, the `hasGeometryAvailable$` should handle those whole cases and be the value watch to display or not the toggle.

tkohr:
Thanks for the feedback @fgravin! Indeed, I missed that with the current config the feature will even be available to the public. I will remove the logic and just add the `check-toggle` in this PR.

Looking at #366 I see that the silent ignore of a failed request is implemented in https://github.com/geonetwork/geonetwork-ui/pull/366/commits/8fa708f03651caa40248c0e196379a17b1410ad8, but I don't really understand how per user URLs can be configured or if they can.

fgravin:
> Thanks for the feedback @fgravin! Indeed, I missed that with the current config the feature will even be available to the public. I will remove the logic and just add the `check-toggle` in this PR.
> 
> Looking at #366 I see that the silent ignore of a failed request is implemented in [8fa708f](https://github.com/geonetwork/geonetwork-ui/commit/8fa708f03651caa40248c0e196379a17b1410ad8), but I don't really understand how per user URLs can be configured or if they can.

Yes i don't think it has been implemented.

apps/datahub/.eslintrc.json
```diff
@@ -21,7 +21,7 @@
           "error",
           {
             "type": "element",
-            "prefix": "datahub",
+            "prefix": ["datahub", "gn-ui"],
```
tkohr, line 24: Wondering, if this necessary or can be handled elsewhere. Without the "gn-ui" value, I get an eslint error when mocking a gn-ui component in a datahub spec.

tkohr:
Ok, I reworked the PR to just handle the UI part for now.

apps/datahub/.eslintrc.json
fgravin: No this is not the way to go cause you could have component which starts with `gn-ui`
What I did in other tests is just ignore eslint for the line, might be the best way IMO

libs/ui/inputs/src/lib/check-toggle/check-toggle.component.html
```diff
@@ -0,0 +1,18 @@
...
+    <input
+      type="checkbox"
+      class="sr-only peer"
+      [ngModel]="value"
+      (ngModelChange)="toggle($event)"
+    />
+    <div
+      class="w-11 h-6 bg-gray-200 peer-focus:outline-none peer-focus:ring-2 peer-focus:ring-{{
+        color
+      }}-lighter rounded-full peer peer-checked:after:translate-x-full peer-checked:after:border-white after:content-[''] after:absolute after:top-[2px] after:left-[2px] after:bg-white after:border-gray-300 after:border after:rounded-full after:h-5 after:w-5 after:transition-all peer-checked:bg-{{
```
fgravin, line 12: aren't some classes purged from TW ? eg `peer-checked:bg-{{color}}`

also i don't get why you need the `peer` class in the div as well.

libs/ui/inputs/src/lib/check-toggle/check-toggle.component.html
```diff
@@ -0,0 +1,18 @@
+<label class="inline-flex relative items-start cursor-pointer" [title]="title">
+  <span class="shrink-0">
```
fgravin, line 2: In flowbite the span does not nest other elements.
Where did you take your exemple from ?

libs/ui/inputs/src/lib/check-toggle/check-toggle.component.html
tkohr: Good catches, thanks ! Wondering why this was working yesterday. Also, the `peer` can be removed.

libs/ui/inputs/src/lib/check-toggle/check-toggle.component.html
tkohr: I took the first flowbite example as a base, but had to adapt this part a little bit as the example button breaks (especially with long labels) on mobile devices.

apps/datahub/.eslintrc.json
tkohr: Ok

fgravin (approved):

PR #398, opened 2022-12-23
Datahub: Display sort badges only on home tab

tkohr:
PR displays the sort badges only on home tab (home = news feed).


fgravin (approved):
Thanks

PR #400, opened 2023-01-03
Map: Disable rotate interactions

tkohr:
PR removes rotate interactions in `prioritizePageScroll`, which are more disturbing than useful in the current use cases.

The `prioritizePageScroll` input has been removed from `map-context.component` and is now called directly from `data-view-map`


fgravin (approved):
Thanks for the refactoring

PR #402, opened 2023-01-04
Datahub: exclude service and featureCatalog in main search

fgravin:
done in suggestion and main search
will need some refactoring later on

### DIFF
--- a/libs/util/shared/src/lib/models/elasticsearch.model.ts
+++ b/libs/util/shared/src/lib/models/elasticsearch.model.ts
@@ -35,3 +35,18 @@ export type EsSearchResponse = any
 
 export type EsTemplateValues = 'y' | 'n' | 's' | 't'
 export type EsTemplateType = EsTemplateValues | EsTemplateValues[]
+
+export const ES_RESOURCES_VALUES = [
+  'dataset',
+  'series',
+  'publication',
+  'nonGeographicDataset',
+  'map',
+  'map/static',
+  'mapDigital',
+  'service',
+  'featureCatalog',
+] as const
+
+export type EsResourceTypeValues = typeof ES_RESOURCES_VALUES[number]
+export type EsResourceType = EsResourceTypeValues | EsResourceTypeValues[]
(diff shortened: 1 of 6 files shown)


libs/util/shared/src/lib/models/elasticsearch.model.ts
```diff
@@ -35,3 +35,18 @@ export type EsSearchResponse = any
 
 export type EsTemplateValues = 'y' | 'n' | 's' | 't'
 export type EsTemplateType = EsTemplateValues | EsTemplateValues[]
+
+export const ES_RESOURCES_VALUES = [
```
tkohr, line 39: Does the model serve primarily for documentation purposes, so far? I don't see it being used yet. Could an `enum` for the values make sense, to prevent hard coding the strings in the code ?

tailwind.base.css
```diff
@@ -4,7 +4,7 @@
 
 @layer components {
   .card-shadow {
-    @apply shadow-xl hover:shadow-primary-lightest transition-shadow;
+    @apply shadow-xl hover:shadow-xl-hover transition-shadow;
```
tkohr, line 7: Not directly related, but maybe we should also adapt the shadow size and use `hover:shadow-xl` instead of `hover:shadow-2xl` in the `OrganisationPreviewComponent` to be consistent. Not sure about the color, though. I think the final UX recommendation was a border in this case anyway.

tkohr (approved):
Thanks, @fgravin! Code LGTM and the service entries do not show up in the search and autocomplete results anymore. Just noticed that they are still accessible via the URL (which should not be disturbing to the avarage user).

tailwind.base.css
fgravin: There is a story to rework the organisation grid.

libs/util/shared/src/lib/models/elasticsearch.model.ts
fgravin: actually i used them but with the refactoring I don't use the type.
but I hope one day we can have proper typing for ES models so I would still keep them.
unions are preferred than enum

fgravin (commented):
thanks for the review @tkohr

PR #403, opened 2023-01-05
Datahub: Set scrollPositionRestoration to enabled

tkohr:
This handles the scroll position between the home and record pages:

- scrolls to top on record page
- scrolls to preceding position when navigating back to results page (via browser back button)
- scrolls to top when navigating back to results page (via datahub back button) => this could eventually improved for the case that the user has already navigated on the datahub before and did not arrive from and external page on the record


fgravin:
> * scrolls to top when navigating back to results page (via datahub back button) => this could eventually improved for the case that the user has already navigated on the datahub before and did not arrive from and external page on the record

which is weird cause it's the meaning of this option. maybe because the route is persisted.
we can insvestigate in that later

fgravin (approved):

PR #404, opened 2023-01-06
Rework organisation hydration

fgravin:
Organisations are fetched from the metadata (ie from an ES agg search), but lack of informations.
We expect to match them with an organisation wich is own by the catalog

- [x] fetch the organisations of the catalog and try an exact normalized match between agg key and org name
- [x] if no match, check for email equality

Added some ghosting to prevent empty page and layout resizing.
Screenshot from 2023-01-06 17-29-03


libs/feature/catalog/src/lib/organisations/organisations.service.spec.ts
```diff
@@ -57,28 +67,57 @@ describe('OrganisationsService', () => {
...
-          recordCount: 5,
-        },
-        {
-          name: 'Association pour le testing',
-          description: undefined,
-          logoUrl: undefined,
-          recordCount: 3,
-        },
-      ])
+    describe('when users tick', () => {
```
tkohr, line 97: 'when groups tick' ?

libs/feature/catalog/src/lib/organisations/organisations.service.ts
```diff
@@ -1,67 +1,126 @@
 import { Injectable } from '@angular/core'
-import { GroupApiModel, GroupsApiService } from '@geonetwork-ui/data-access/gn4'
-import { ElasticsearchService, Organisation } from '@geonetwork-ui/util/shared'
+import { GroupApiModel, SearchApiService } from '@geonetwork-ui/data-access/gn4'
 import { AggregationsService } from '@geonetwork-ui/feature/search'
```
tkohr, line 3: Can be removed now

libs/feature/catalog/src/lib/organisations/organisations.service.ts
```diff
@@ -109,13 +116,17 @@ export class OrganisationsService {
...
     return organisations.map((organisation) => {
-      const group = groups.find(
-        (group) =>
-          (group.label.eng
-            ? this.normalizeName(group.label.eng)
-            : this.normalizeName(group.name)) ===
-          this.normalizeName(organisation.name)
+      let group = groups.find((group) =>
+        this.compareNormalizedStrings(
+          group.label.eng ? group.label.eng : group.name,
```
fgravin, line 121: Aren't translation just a copy of groupe name ?

libs/feature/catalog/src/lib/organisations/organisations.service.ts
```diff
@@ -109,13 +116,17 @@ export class OrganisationsService {
...
+          organisation.name
+        )
       )
+      if (!group) {
```
fgravin, line 125: why not a `const group = test1 ?? test2`

libs/feature/catalog/src/lib/organisations/organisations.service.ts
```diff
@@ -50,12 +49,30 @@ export class OrganisationsService {
     private groupService: GroupService
   ) {}
 
-  normalizeName(name: string): string {
-    return name
-      .normalize('NFD') // decompose graphemes to remove accents from letters
-      .replace(/[\u0300-\u036f]/g, '') // remove accent characters
-      .toLowerCase()
-      .replace(/[^a-z0-9]/g, '') // replace all except letters & numbers
+  compareNormalizedStrings(
```
fgravin, line 52: compare function often return -1, 0, 1 in order to sort things.
it's more an equal test here

fgravin, line 52: should ideally be in a dedicated shared service, not a big deal here

fgravin (commented):
The logic is good and should work.
Some remarks about the code though.

PR #411, opened 2023-01-18
Elasticsearch: Also add empty array of uuids to search query

tkohr:
PR also adds the `uuid` to the payload if it is an empty array, which is necessary when the favorite filter is active, but no favorites are present (to return 0 results, instead of all).


fgravin (approved):
LG thanks

PR #412, opened 2023-01-27
Dataviz: Set up a UI base to create charts

tkohr:
This PR intends to set a base to create charts in gn-ui and discuss its architecture by adding a:

- `data-view-chart.component` component to integrate the chart and upcoming configuration GUI
- `chart-container.component` to handle chart creation and updating
- `chart.service` to encapsulate chart.js within gn-ui and provide an API for chart manipulation

The PR allows first interaction with the chart, but is still WIP. 

Reflection is still ongoing regarding:
- If the 'state' of the chart should rather go into the service (as for now) or the container
- Creation of an intermediate `data-view.component` for the table and chart (which rely on the same data)
- Refactoring column selection when initializing the chart
- Making the chart more meaningful (by adding grouping capabilities)

Note: Do not forget to adapt chart.js import to allow tree shaking

chart


tkohr:
Will close this PR, as it primarily served for discussion/POC and will be implemented in separate PRs.

PR #414, opened 2023-02-07
gn-search-input datahub-url option

fgravin:
Add a new option to `gn-search-input` web component.
If set, the input triggers datahub page opening
- open datahub search page with any input on search submission
- open datahub record page on suggestion selection


apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.html
```diff
@@ -1,3 +1,8 @@
 <div id="angular-app-root">
-  <gn-ui-fuzzy-search class="text-[18px]"> </gn-ui-fuzzy-search>
+  <gn-ui-fuzzy-search
+    class="text-[18px]"
+    (inputSubmited)="search($event)"
```
jahow, line 4: Hmm, the issue is that if `datahub-url` is not defined, these handlers will still be set which means that the fuzzy search input will not update the search state, so it becomes useless.

See https://geonetwork.github.io/geonetwork-ui/wc-search/demo/webcomponents/gn-search-input-and-results.sample.html

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -76,6 +76,10 @@ export class FuzzySearchComponent implements OnInit {
   }
 
   handleInputSubmission(any: string) {
-    this.searchService.updateFilters({ any })
+    if (this.inputSubmited.observers.length > 0) {
+      this.inputSubmited.emit(any)
+    } else {
+      this.searchService.updateFilters({ any })
```
jahow, line 82: +1 for this change, makes sense!

apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
```diff
@@ -14,4 +16,20 @@ import { BaseComponent } from '../base.component'
...
 })
-export class GnSearchInputComponent extends BaseComponent {}
+export class GnSearchInputComponent extends BaseComponent {
+  @Input() datahubUrl: string
+
+  search(any: string) {
+    if (this.datahubUrl) {
+      const landingPage = `${this.datahubUrl}/home/search/?q=${any}`
+      window.open(landingPage, '_self').focus()
+    }
```
jahow, line 26: Thinking about it more, it feels like this creates a strong dependency on the datahub app, whereas we could just offer more generic parameters like so:
```
open-on-search="../datahub/home/search/?q=${search}"
open-on-select="../datahub/dataset/${uuid}"
```
It's a bit more cumbersome to setup but much more flexible. One downside is that if the URL scheme of the datahub changes, people will have to manually change the configuration of their web components.

What do you think?

jahow:
@fgravin I have rebased and updated this PR according to my suggestions, which seem to be working. Please let me know if you think that would be suitable!

PR #416, opened 2023-02-08
Data fetcher / support number and date column types

jahow:
This PR extends the capabilities of the `util/data-fetcher` lib to offer more information on a dataset and ask for aggregations. This is intended to be used in the chart data preview.

Todo:
- [x] introduce Dataset class in data-fetcher
- [x] allow getting field infos for various types
- [ ] ~introduce a query building system on a dataset~


jahow:
This is ready for review, the query building system will come in another PR

libs/util/data-fetcher/src/lib/data-fetcher.ts
```diff
@@ -64,68 +16,32 @@ class FetchError {
...
+      case 'csv':
+        dataset = new CsvDataset(url)
+        break
+      case 'json':
+        dataset = new JsonDataset(url)
+        break
+      case 'geojson':
+        dataset = new GeojsonDataset(url)
+        break
+      case 'excel':
```
fgravin, line 37: shouldn't you need a `default` as well ?

fgravin (commented):
As discussed, I think duckdb would be a better candidate for this work.
Still need to ensure that it's cached and executed in a worker (for external full file loading, csv, geojson etc...)

libs/util/data-fetcher/src/lib/data-fetcher.ts
jahow: all possible cases are covered here, if the file type cannot be guessed then an exception will be thrown before

tkohr:
As far as I can see, the refactoring looks good to me. Wondering if the `processItemProperties` won't be too 'heavy' on big datasets. Maybe some performance testing could give an answer to this?

PR #417, opened 2023-02-09
Dataviz: Create UI component for chart

tkohr:
PR adds a chart UI component with the following inputs:
* `data`: array of objects
* `xAxis`: name of the object property to be used as x-axis
* `yAxis`: name of the object property to be used as y-axis
* `chartType`:  `column` | `bar` | `line` | `curve` | `scatter` | `pie`

Internally, the chartsjs lib is used for visualization.

TODO:
- [x] tests


libs/ui/layout/src/lib/chart/chart.component.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing'
```
fgravin, line 1: I don't think chart is a `layout` component.
Also, we should really focus on bundle size and dependency tree for `affected:test` duration.

IMO, we should start using standalone component, at the very first.
We must really improve the dependency tree of Angular modules.
Note that almost every app or lib will import layout lib, so it means `chart.js` will be in all bundles, even the ones not using it.

libs/ui/layout/src/lib/chart/chart.component.ts
```diff
@@ -0,0 +1,143 @@
...
+  @Input() set xAxis(value: string) {
+    if (this.xAxisValue) {
+      this.xAxisValue = value
+      this.updateChartData()
+    } else {
+      this.xAxisValue = value
+    }
+  }
+  @Input() set yAxis(value: string) {
+    if (this.yAxisValue) {
```
fgravin, line 67: in what case this would be `null` ?
what happens if you call `updateChartData()` while it's null ?

To me those 2 setters are quite verbose.

fgravin (commented):
Some comments as it goes.

It's quite basic so far, it is completely narrowed to chart.js, I think the chart.js strategy should ben implemented into a service, the only reason why it's a component is the html element reference.
We would also need an abstraction, if we decide in a second run to move to d3 for instance.

libs/ui/layout/src/lib/chart/chart.component.spec.ts
jahow: hmm that's a good point, I was the one suggesting to put it in layout but it's true that it's not a good idea in terms of bundle. Maybe a new dedicated module ui/chart or something like that would make sense?

fgravin: It seems that we are going to get stuck anyway with the `lib` system cause we might need to create way too many libs to keep it consistent.

There are 2 things to keep in mind:
- bundle size for apps depends on the `import` files, which mostly match angular modules dependencies. One lib is a module so we include it or not. This can be addressed using standalone component, it's a must go imo.
- affected targets of Nx only works on lib dependency so far I guess , so here we could gain by creating new libs for more dedicated usecases.

With better module architecture and standalone components, we could event imagine to have both smart/ui logic in the same lib, which would make sense somehow.

libs/ui/layout/src/lib/chart/chart.component.ts
tkohr: On chart creation, these properies are `undefined`, so the `if` is only fulfilled on updates. You're right, it's still quite verbose.

tkohr:
Thanks for pointing out these points @fgravin. I've generated a `ui-dataviz` lib and created the `chart.component` as a standalone component within it. Haven't checked the affected behavior yet, nor its usage within an app. I think it would make sense to also move the `TableComponent` and `FigureComponent` here, maybe also as `standalone` looking at the deps.

Regarding the implementation of a service, I'd suggest keeping everything in the component for now, as the logic is pure visualization, not too complex and not reused anywhere. In my understanding, the component inputs provide a fitting abstraction of the lib we use within the component.

@jahow regarding the data mapping, I'm not a big fan of 2d84a82a5853ed9ffb564b8707d6f72b94596237, as it complexifies the chart options and still leaves a parsing/mapping step, just that chartjs now takes care of it. I'd tend to revert this, unless you have some other idea:-)

jahow:
> @jahow regarding the data mapping, I'm not a big fan of https://github.com/geonetwork/geonetwork-ui/commit/2d84a82a5853ed9ffb564b8707d6f72b94596237, as it complexifies the chart options and still leaves a parsing/mapping step, just that chartjs now takes care of it. I'd tend to revert this, unless you have some other idea:-)

I agree, trying to avoid the parsing step in chartjs is going to be quite tricky.

I think in the future we can probably try using a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to avoid creating a new array every time the input changes, but IMO this can be done later on.

tkohr:
> I agree, trying to avoid the parsing step in chartjs is going to be quite tricky.
> 
> I think in the future we can probably try using a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to avoid creating a new array every time the input changes, but IMO this can be done later on.

I actually, noticed that even with the mapping step chartjs parsing cannot be set to `false`. So I still got rid of the additional mapping step, but made the options a little more readable: 225838c387bf5d23930354779d174674a0bdace0

jahow (approved):
This looks good, thanks for the new ui-dataviz lib and the standalone components! hopefully this way of doing things will let us have less affected changes and smaller bundles. Feel free to merge once the CI is green :slightly_smiling_face:

libs/feature/dataviz/src/lib/figure/figure-container/figure-container.component.ts
```diff
@@ -4,7 +4,7 @@ import {
   Input,
   OnChanges,
 } from '@angular/core'
-import { TableItemModel } from '@geonetwork-ui/ui/layout'
+import { TableItemModel } from '@geonetwork-ui/ui/dataviz'
```
fgravin, line 7: You are not leveraging the capacity of standalone component are you ?

If you do that, you keep passing through modules to import your deps, so you bypass standalone import.

tkohr: I might be missing something, but I thought removing the standalone component from the module exports was sufficient. This allows to import the standalone component without the entire module. The export in the `index.ts` still seems necessary to me to import the component somewhere, isn't it?
I chose to make `TableComponent` and `ChartComponent` standalone as they have various dependencies/imports and keep the `FiguerComponent` a common one as it only imports the generic `MatIconModule` which will likely be used by other dataviz components.

tkohr: Do you think I should also get rid of the standalone components in the `index.ts` exports and import them via relative paths?

jahow: Importing via relative path is forbidden by NX rules so... I guess there's no choice

fgravin: I saw somewhere that last version of Nx was proposing a bootstrap based on standalone components.
We should try to have a look about what they propose soon.

Note that you import `TableComponent` module from `FeatureDataviz` module while you might should  import it from `FigureContainerComponent` instead shouldn't you ?
See https://angular.io/guide/standalone-components#the-standalone-flag-and-component-imports

Anyway, it's quite new so we should be careful about using this. I think it's the good direction but we have to do it the proper way.
You can merge the PR but we should figure out the best practices out of it.

tkohr: Indeed, I still have a doubt if the bootstrapping needs further fine-tuning to minimize bundle size. I'll try to do a test with the `ChartComponent`.

Regarding the import of the `TableComponent` into another component (smart ones in our cases), this enforces the importing component to become a standalone component as well, as only standalone components can import standalone components. Not sure if this is the way to go. It would have a larger impact on the architecture in any case.

fgravin: We'll talk about that later, I think that yes, it should be a standalone component too. We need to define a clear strategy for further additions.
Nvm, all of this won't probably change the size of `gn-wc.js` :disappointed:

PR #418, opened 2023-02-11
WIP / fetch data using duckdb wasm

jahow:
Refs used:
- https://duckdb.org/2021/10/29/duckdb-wasm.html
- https://github.com/duckdb-wasm-examples/react-typescript-shell/blob/5f8d06ccc73aa9fb0d42c0866d4a2c86925d2da7/src/app.tsx

Try for example on http://localhost:4200/dataset/602e73d9f8370885652ab230


fgravin:
We will not go into that direction.
We'll rely on the data APIs later on.

PR #419, opened 2023-02-21
Datahub / add error handling when loading a WMTS layer on the map

jahow:
This makes WMTS layer loading more robust in the record view by handling various errors and making sure that the GetCapabilities request is properly built.

Note that WMTS handling should go in https://github.com/camptocamp/ogc-client eventually to streamline error management, check for CORS, etc.

### DIFF
--- a/libs/feature/map/src/lib/utils/map-utils.service.ts
+++ b/libs/feature/map/src/lib/utils/map-utils.service.ts
@@ -159,17 +159,31 @@ export class MapUtilsService {
   }
 
   getWmtsOptionsFromCapabilities(link: MetadataLink): Observable<Options> {
+    const getCapabilitiesUrl = new URL(link.url, window.location.toString())
+    getCapabilitiesUrl.searchParams.set('SERVICE', 'WMTS')
+    getCapabilitiesUrl.searchParams.set('REQUEST', 'GetCapabilities')
     return from(
-      fetch(link.url)
-        .then(function (response) {
+      fetch(getCapabilitiesUrl.toString())
+        .then(async function (response) {
+          if (!response.ok) {
+            throw new Error(`WMTS GetCapabilities HTTP request failed with code ${
+              response.status
+            } and body:
+${await response.text()}`)
+          }
           return response.text()
         })
         .then(function (text) {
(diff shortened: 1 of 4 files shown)


libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -159,17 +159,31 @@ export class MapUtilsService {
...
     return from(
-      fetch(link.url)
-        .then(function (response) {
+      fetch(getCapabilitiesUrl.toString())
+        .then(async function (response) {
+          if (!response.ok) {
+            throw new Error(`WMTS GetCapabilities HTTP request failed with code ${
+              response.status
+            } and body:
+${await response.text()}`)
```
tkohr, line 172: Weird formatting... also noticing that we haven't translated any error messages yet.

libs/feature/map/src/lib/utils/map-utils.service.ts
```diff
@@ -159,17 +159,31 @@ export class MapUtilsService {
...
-          })
+          try {
+            const result = new WMTSCapabilities().read(text)
+            return optionsFromCapabilities(result, {
+              layer: link.name,
+              matrixSet: 'EPSG:3857',
+            })
+          } catch (e: any) {
+            throw new Error(`WMTS GetCapabilities parsing failed:
+${e.stack || e.message || e}`)
```
tkohr, line 185: Also, strange indentation...

tkohr (approved):
Thanks, @jahow. Code LGTM and works fine for displaying the WMTS that failed before and displaying the error for one with a wrong URL.

libs/feature/map/src/lib/utils/map-utils.service.ts
jahow: it looks weird but the result makes sense since we're in a template string

libs/feature/map/src/lib/utils/map-utils.service.ts
jahow: same as above, we're in a template string so any indentation here will show up in the final message :)

jahow:
Thanks for the review!!

PR #423, opened 2023-03-02
Data platform: add a scaffold for backend services and nx 15.8.1 migration

fgravin:
This PR introduces a new kind of application in the monorepo, the **NestJS** node backend application.
The new application is called `data-platform`.
It requires to setup a postgres database with
```js
type: 'postgres',
host: 'localhost',
port: 5432,
username: 'geonetwork',
password: 'geonetwork',
database: 'gn-data-platform',
```
The schema is automatically deployed from `typeorm` ORM
A swagger runs on http://localhost:3333/api


fgravin:
@jahow @tkohr 
I mixed `nestjs` integration and `nx` migration in the same PR cause both mixed the dependencies and it was a mess.
Automagically, doing both in the same codebase fixed some issues with dep compatibility.

The only drawback was to skip the test of the `dropdown-multiselect:keyboard-event:when-overlay-is-closed`.
It might be an issue with jest and `angular/cdk`, things are not sync, but the behavior looks good in the datahub though.

I would suggest to test the datahub on this PR and if everything is working fine then I would push for a merge, it's a reasonable tradeoff IMO.

jahow:
Rebased and:
* added a commit to remove one warning related to jest config (and also hopefully simplified jest configs by putting more stuff in the common preset)
* added a commit to make the ui-inputs test pass again; not sure why it stopped working, `fakeAsync` is not really doing its job here but the change is really minor so I think it's worth it
* centralised all dependencies changes on package-lock into one single commit

The backend part looks nice, the data model will probably change but at least we have something (did not test on my machine with a database though).

jahow (approved):

fgravin:
Thanks @jahow for the follow up.

PR #425, opened 2023-03-08
Datahub: Use charts from dataviz

tkohr:
PR integrates the dataviz charts into the `feature-record` lib via a `DataViewComponent` that proposes `table` and `chart` modes as input. As a result, the datahub displays the charts in a third preview tab.


libs/feature/record/src/lib/data-view/data-view.component.spec.ts
```diff
@@ -146,4 +159,55 @@ describe('DataViewTableComponent', () => {
       })
     })
   })
+  describe('initial state in chart mode', () => {
```
jahow, line 162: I don't think it's necessary to duplicate the whole test here, this seems like a lot of duplicate code; you could probably just check that the component correctly creates a chart-view or table-view depending on its type

jahow (approved):
Thanks! It works quite well :) I made a comment regarding the test for the data-view component, I think the test ends up being too verbose. apart from that nothing to add!

tkohr:
Thanks for the review @jahow ! I simplified the tests a bit :-)

Issue #415, opened 2023-02-08
provide a geonetwork/geonetwork-ui-datahub:latest tag


fvanderbiest:
https://hub.docker.com/r/geonetwork/geonetwork-ui-datahub/tags?page=1&name=latest

So we can refer to it in the geOrchestra docker composition

tkohr:
FYI, datahub images are now available with latest tag : https://hub.docker.com/r/geonetwork/geonetwork-ui-datahub/tags

PR #426, opened 2023-03-08
Docker: Additionally tag images for all apps as latest

tkohr:
PR additionally tags images when merging on `main` as `latest`.

Note: Latest tags should not be generated when generating images via github checkbox

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -88,6 +88,10 @@ jobs:
         if: github.event_name != 'release'
         run: npx nx affected --target=docker-build
 
+      - name: Tag all docker images on main also as latest
+        if: github.ref_name == 'main'
+        run: docker image ls --format 'docker tag {{.Repository}}:{{.Tag}} {{.Repository}}:latest' --filter=reference='geonetwork/*' | bash -
+
       - name: Login to DockerHub
         uses: docker/login-action@v2
         with:


jahow:
Addresses #415

jahow:
I think you can automatically tag all the docker images with `latest` by simply doing this:

```
docker image ls --format 'docker tag {{.Repository}}:{{.Tag}} {{.Repository}}:latest' --filter=reference='geonetwork/*' | bash -
```

This will call `docker tag` for each image and tag is as `latest` as well. So this can simply be called when on the main branch and done :slightly_smiling_face:

.github/workflows/artifacts.yml
```diff
@@ -84,6 +84,10 @@ jobs:
         if: github.event_name == 'release'
         run: npx nx run-many --projects=${{ env.APP_NAMES }} --target=docker-build
 
+      - name: Build all docker images as latest
+        if: github.event.pull_request.merged == true
```
jahow, line 88: I'm not sure this will trigger (this workflow does not react on pull request events); I think a better condition could be:
```
if: github.ref_name != 'main'
```

jahow (changes requested):
Could be made simpler IMO!

tkohr:
Thanks for your feedback @jahow, I was convinced there was an easier way. Did I get it right this time?

.github/workflows/artifacts.yml
```diff
@@ -88,6 +88,10 @@ jobs:
         if: github.event_name != 'release'
         run: npx nx affected --target=docker-build
 
+      - name: Tag all docker images on main also as latest
+        if: github.ref_name != 'main'
```
jahow, line 92: There is a typo here actually (my error):

```suggestion
        if: github.ref_name == 'main'
```

tkohr: Ah, I should pay more attention when copy/pasting :see_no_evil:

jahow (approved):
Let's merge and si if it works!

PR #431, opened 2023-03-14
Dataviz: Fix chart type bar-horizontal

tkohr:
PR removes scales from options to fix `bar-horizontal` chart type. I chose to remove it, as I have the impression chartjs handles this automatically if not indicated in the options. 

If I missed something and we really need the `scales`, an alternative would be to set it for `y` in the `bar-horizontal` options:
```
scales: {
  y: {
    type: this.handlesSecondaryValue() ? 'linear' : 'category',
    position: 'bottom',
  },
},
```


jahow:
Looks like it's working in storybook, not sure exactly why we used a scales originally!

jahow (approved):

PR #434, opened 2023-03-16
Search: Add filter for data with or without spatial reference

tkohr:
Opened this PR for discussion, as I don't have the impression that updating the `format` filter goes into the right direction. In terms of user experience, this creates a cross filter behavior: Changing values in one filter modifies values in another one, something which should be prevented IMO. Representing the state of the spatial/non-spatial filter in the URL also becomes impossible, this way.


jahow:
Superseded by #441

PR #435, opened 2023-03-17
Datahub (v1.0.1): Fix i18n by adding availabe langs to translate.service

tkohr:
PR attempts to fix a bug where the datahub application falls back to the default translations in `en` even though the browser language is `fr`. Unfortunately, the behaviour does not reproduce systematically.

I noticed that `translate.getLangs()` was empty at the moment we call `translate.use()` in the `app.module`, so I added `translate.addLangs()`. Couldn't really find in the ngx-translate docs if this is mandatory to call, but it's worth a try IMO.

I propose to fix this on a `v1.0.x` branch to create a `v1.0.2` tag we can test on dev/prod.

### DIFF
--- a/apps/datahub/src/app/app.module.ts
+++ b/apps/datahub/src/app/app.module.ts
@@ -27,6 +27,7 @@ import {
   getThemeConfig,
 } from '@geonetwork-ui/util/app-config'
 import {
+  AVAILABLE_LANGS,
   getDefaultLang,
   getLangFromBrowser,
   TRANSLATE_DEFAULT_CONFIG,
@@ -170,6 +171,7 @@ export class AppModule {
     router: Router,
     @Inject(DOCUMENT) private document: Document
   ) {
+    translate.addLangs(AVAILABLE_LANGS)
     translate.setDefaultLang(getDefaultLang())
     translate.use(getLangFromBrowser() || getDefaultLang())
     ThemeService.applyCssVariables(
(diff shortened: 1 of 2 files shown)


jahow:
> I propose to fix this on a v1.0.x branch to create a v1.0.2 tag we can test on dev/prod.

+1

jahow (changes requested):
I think a good first step would be to use the `getBrowserLang()` method from the translation service, see:
https://github.com/ngx-translate/core#methods

And remove our local implementation. Then we can try deploying this on a dev environment and see whether this changes anything.

Then, I think we'll probably need to offer a language switching mechanic, because we can't safely assume that the user really has made sure that their browser language fits their expectations (these settings are often quite hard to find).

jahow:
Something else: in the `Util18nModule` we call the `TranslateService.use` method:
https://github.com/geonetwork/geonetwork-ui/blob/main/libs/util/i18n/src/lib/util-i18n.module.ts#L24

And we also call it in the app code, e.g. for the datahub:
https://github.com/geonetwork/geonetwork-ui/blob/main/apps/datahub/src/app/app.module.ts#L176

I found this issue on ngx-translate:
https://github.com/ngx-translate/core/issues/1246

So we definitely have a potential race condition here. IMO the i18n module could be responsible for deciding which language to use, and apps shouldn't even worry about it.

tkohr:
Thanks for your investigations @jahow ! I opened up a follow-up PR for this (see above).

PR #436, opened 2023-03-18
Remove geojson fixtures from bundle

jahow:
This removed a reference to the fixtures lib in the feature/auth plugin, and magically removes 2MB from most bundles from gn-ui.

Also fixed a minor type issue in test file.

Note: to be more future proof on this we should put some linting rules in place to prevent this.


fgravin (approved):
Yes, it was not a worry by those time but has a huge impact.
Thanks for fixing that

PR #438, opened 2023-03-21
Fix i18n (main): Let util-18n.module handle language detection

tkohr:
PR backports #437 on main branch.


jahow (approved):
Thanks! I'm in favor of not requiring a review on PR which are only backports, as long as the CI is green (unless it's a big change)

PR #440, opened 2023-03-23
Datahub: Filter by publication year

tkohr:
PR adds a filter by publication year of the resource to the advanced search filters.

filter_by_year


jahow (approved):
Good job, thanks! it works great. I'm not sure we need a `histogram` facet here since the interval is one. But we'll have to refactor this eventually so no worry.

tkohr:
Thanks for the review @jahow !

PR #441, opened 2023-03-26
Search / add topic and isSpatial filters

jahow:
This PR refactors the way search filters are handled:
* a new `FieldsService` was introduced: this service handles a set of fields for which it can:
  * retrieve available values (i.e. the ones shown in the dropdown)
  * transforms a list of values to search filters
  * transforms search filters to a list of values
  This service replaces the `AggregationsService` and encapsulates all the logic which has a relationship with Elasticsearch.  
  The router lib also relies on it now to convert from url params to search filters, which makes it a bit simpler (the router mapper has been removed as well).
* the existing fields (`publisher`, `publicationYear`, `format`) have been added to this service; the `q` field used for full text search is also handled by this service
* two new fields were added: `topic` and `isSpatial`
  * both have specific logics for getting the available values and translating the labels in the dropdown
  * the `isSpatial` field relies on a new runtime mapping based on a script

image

This refactoring was made to anticipate a future compatibility with other search backends.


libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  SimpleSearchField,
+  TopicSearchField,
+} from './fields'
+
+@Injectable({
+  providedIn: 'root',
+})
+export class FieldsService {
+  private fields = {
+    publisher: new SimpleSearchField('OrgForResource', 'asc', this.injector),
```
tkohr, line 17: Wondering if we should not rather expose the ordering as an input of the filter-dropdown component? Btw, do you think an `enum` for the orders is too verbose?

tkohr (approved):
Nice, I really like the refactoring for using the filters in the future. Thanks, @jahow ! Only noticed that the button resetting all filters is broken, when testing.

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -42,6 +42,19 @@ export class ElasticsearchService {
         geometry
       ),
       _source: requestFields,
+      runtime_mappings: {
+        isSpatial: {
+          type: 'keyword',
+          script: `
+String result = 'no';
```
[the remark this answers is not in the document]
jahow: it also felt like magic to me! I think we can use runtime mappings in the future for producing much more practical fields.
Theorically you can do whatever you want but the script system is also a bit weird so... we'll have to try

jahow:
> Only noticed that the button resetting all filters is broken, when testing.

Thanks! I fixed it in the latest commits.

libs/feature/search/src/lib/utils/service/fields.service.ts
jahow: If we don't need to customize the order, then we don't have to expose it. Alternatively we could have specific configurations for each search field in the future, but that could be provided in a different way than just an input, because this input will have to be carried across several layers. This all depends on the intended use cases.

As for enums, I think when they are sufficiently explicit strings (`'asc'` and `'desc'` for instance) I think there's really no need in using TS enums, and string unions are just fine. Also enums do not exist in Javascript, they are Typescript constructs and that make them a bit awkward IMO, so I try to avoid using them whenever possible.

jahow:
thanks for the review @tkohr!

PR #442, opened 2023-03-26
Dataviz / fix translation bug in table component

jahow:
Before:
image

After:
image

Because the table component is standalone, it needs to explicitly import the translate module on its own. I feel like this might happen again in the future, since without standalone components these imports were only made once in the lib module.


fgravin (approved):
Yes, moving a comp to a standalone one should import the `Translate` module :+1:

PR #443, opened 2023-03-26
[WIP] Work on project graph, fix small tech debt

jahow:
* Removed all dependencies from `util` libs to `data-access` ones
  * Also made these dependencies forbidden in the future; only `feature` libs should rely on `data-access` ones
* Fixed the failed runs on `main` because of a failure of docker push for data-platform
* Fixed `latest` docker images produced on PRs


fgravin (approved):
Excellent, thanks for the docker fix, I saw that on master CI but forgot to fix it.
Very good cleaning :+1:

PR #445, opened 2023-03-28
Datahub / add license filter

jahow:
Followup of #441 

image

Also made sure the runtime fields in the ES queries are only added when they are really needed (e.g. used as filters or in aggregations).

### DIFF
--- a/libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
+++ b/libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
@@ -15,6 +15,10 @@ export const METADATA_LANGUAGE = new InjectionToken<string>('metadata-language')
   providedIn: 'root',
 })
 export class ElasticsearchService {
+  // runtime fields are computed using a Painless script
+  // see: https://www.elastic.co/guide/en/elasticsearch/reference/current/runtime-mapping-fields.html
+  private runtimeFields: Record<string, string> = {}
+
   constructor(
     @Optional() @Inject(METADATA_LANGUAGE) private metadataLang: string
   ) {}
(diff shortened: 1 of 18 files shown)


libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -42,41 +46,55 @@ export class ElasticsearchService {
...
+            addMapping(runtimeField)
+        }
+      }
+      for (const key in node) {
+        if (typeof node[key] === 'object') lookForField(node[key])
+      }
     }
+    lookForField(payload.aggregations)
+    lookForField(payload.sort)
+    lookForField(payload.query)
```
fgravin, line 90: You can add `_source` I guess.

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
```diff
@@ -42,41 +46,55 @@ export class ElasticsearchService {
...
-        license: {
-          type: 'keyword',
-          script: `
-String raw = '';
-if (doc.containsKey('licenseObject.default.keyword') && doc['licenseObject.default.keyword'].length > 0)
-  raw += doc['licenseObject.default.keyword'].join('|').toLowerCase();
-if (doc.containsKey('MD_LegalConstraintsUseLimitationObject.default.keyword') && doc['MD_LegalConstraintsUseLimitationObject.default.keyword'].length > 0)
-  raw += doc['MD_LegalConstraintsUseLimitationObject.default.keyword'].join('|').toLowerCase();
+    }
+    this.processRuntimeFields(payload)
```
fgravin, line 50: :clap:

fgravin (approved):
Tested and all good.
Thanks for working back on this.

translations/de.json
```diff
@@ -195,6 +195,7 @@
   "search.field.any.placeholder": "",
   "search.field.sortBy": "",
   "search.filters.byFormat": "",
+  "search.filters.byLicense": "",
```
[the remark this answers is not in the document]
jahow: just ask chatgpt ;)

libs/util/shared/src/lib/elasticsearch/elasticsearch.service.ts
jahow: I thought about it, it's actually possible to get the value of runtime fields but not through `_source`, only through the `fields` option which we don't use yet. This will be handy for the metadata quality level.

PR #447, opened 2023-03-29
Mapping from GN4 search response to Pivot format

fgravin:
We currently define our own `MetadataRecord` model to represent a record internally in geonetwork-ui (used by libs and the datahub for instance).
We should use the pivot format as the internal model for metadata record representation & processing.

This PR handles the mapping from GN4 elasticsearch search results metadata to the pivot format.
It is a standalone action.
The mapper is in the `api-metadata-converter` lib


libs/api/metadata-converter/src/lib/common/role.mapper.ts
jahow, line 1: the values in the switch statement are specific to iso codelists so I think it would make sense to keep them in the iso19139 converter (or at least as an iso-specific utility)

libs/api/metadata-converter/src/lib/common/status.mapper.ts
jahow, line 1: same here, these are values expected in iso codelists

libs/api/metadata-converter/src/lib/common/update-frequency.mapper.ts
jahow, line 1: idem

libs/util/types/src/lib/metadata/metadata.model.ts
```diff
@@ -131,8 +139,8 @@ export interface DatasetRecord extends BaseRecord {
   datasetUpdated?: Date
   lineage: string // Explanation of the origin of this record (e.g: how, why)"
   distributions: Array<DatasetDistribution>
-  spatialExtents: Array<DatasetSpatialExtent> // not handled yet
-  temporalExtents: Array<DatasetTemporalExtent> // not handled yet
+  spatialExtents?: Array<DatasetSpatialExtent> // not handled yet
+  temporalExtents?: Array<DatasetTemporalExtent> // not handled yet
```
jahow, line 143: Having less optional keys as possible was intended, since it means we can always expect objects to have roughly the same form; since extents are arrays, empty arrays can be used if no extent is available

libs/util/types/src/lib/metadata/metadata.model.ts
```diff
@@ -59,6 +59,8 @@ export interface BaseRecord {
   useLimitations: Array<string>
   licenses: Array<License>
   overviews: Array<GraphicOverview>
+  extras?: Record<string, unknown>
```
jahow, line 62: even though I agree that we need such a field, I think we should make sure that this does not become an alternate free-form storage  in the long run; I thought that maybe having keys such as `iso19139.mySpecificfield` would make sense for example;
Not saying this means to be modified, just a thought I have

jahow (approved):
Some remarks in the code, thanks for the refactoring!

libs/util/types/src/lib/metadata/metadata.model.ts
fgravin: Yes you are right, you need strictness on how we handle those properties.
I think we should mature it a bit, when we have other mappers like `dcat` and `iso19115-3 `, and see how we can handle those extras in the UI, and adapt the UI if they are not present.

PR #448, opened 2023-03-31
Add ogc api records support

fgravin:
Starts adding OGC API Records support
- api angular client generated
- Record model & fixture have been added

From reference: https://github.com/opengeospatial/ogcapi-records


fgravin:
Stand by on this, we will see if it's opportune to have a model for the GSJON records or not, as the spec is quite "open".
We might use `ogc-client` as well.

fgravin:
Closing, the support should be done in ogc-client

PR #449, opened 2023-03-31
CI: parallelize lint & test

fgravin:
Use nx 15 new abilities to run several targets in parallel.
See https://twitter.com/NxDevTools/status/1634557591308189698?s=20

I don't know if it will really improve perfs though

I have removed the coverage as we are not using it yet, it may slow down a bit the CI.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -45,9 +45,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Workspace lint
-        run: npx nx workspace-lint
-
       - name: Format check
         run: npx nx format:check --base=$NX_BASE --head=$NX_HEAD
 
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -19,69 +19,26 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != 'main' }}
 
 jobs:
-  format-lint-test:
+  format-lint-test-build:
     if: github.event.pull_request.draft == false
-    name: Format check, lint, unit tests
-    runs-on: ubuntu-latest
+    name: Nx Cloud - Main Job
+    uses: nrwl/ci/.github/workflows/nx-cloud-main.yml@v0.11.3
+    with:
+      number-of-agents: 3
+      init-commands: |
+        npx nx-cloud start-ci-run --stop-agents-after="build" --agent-count=3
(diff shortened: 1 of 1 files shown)


fgravin:
I don't know why the CI task `test-build` is separated, I thought it was run on demand, or just for releases but it seems it has the same rules as `format-lint-test`.
So we could parallel these ones as well.

.github/workflows/checks.yml
```diff
@@ -45,17 +45,11 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Workspace lint
-        run: npx nx workspace-lint
```
jahow, line 49: This was part of the recommendations at some point from NX, I think it just makes sure that the project structure is coherent. Might be worth checking if it's still recommended or not.

.github/workflows/checks.yml
```diff
@@ -45,17 +45,11 @@ jobs:
...
       - name: Format check
         run: npx nx format:check --base=$NX_BASE --head=$NX_HEAD
 
-      - name: Lint affected --quiet
-        run: npx nx affected --base=$NX_BASE --head=$NX_HEAD --target=lint --parallel=3 --quiet
-
-      - name: Test affected
-        run: npx nx affected --base=$NX_BASE --head=$NX_HEAD --target=test --parallel=3 --ci --code-coverage
+      - name: Affected lint & tests
+        run: npx nx affected --base=$NX_BASE --head=$NX_HEAD --targets=lint,test --parallel=5 --quiet --cache --ci
```
jahow, line 52: How about including the format as well? this will also have the advantage of having the result of all these tasks at every run instead of just stopping at the first failure (typically formatting) and then discovering other failures little by little...

jahow (commented):
I think this is definitely the right approach. This workflow was heavily inspired from the NX 
recommendations for gh actions, so it might also be worthwhile to see if they improved them as well.

.github/workflows/checks.yml
fgravin: Ok I'll have a look, but if you run it now, it says that it is depecrated

```
 >  NX   "nx workspace-lint" has been deprecated.

   Nx has been reworked to make sure that the issues workspace-lint used to catch are not longer possible.
   Remove "nx workspace-lint" from your CI.
```

jahow (approved):
:+1:

PR #450, opened 2023-04-01
Datafeeer / avoid infinite loop when polling job progress

jahow:


tkohr (approved):
Thanks for the fix and the thorough testing @jahow ! Did not test it, but as far as I can see, code LGTM.

PR #453, opened 2023-04-12
Datafeeder / Fix the list of keywords not appearing any more on step 2

jahow:
Due to a subtle bug the keyword list was not showing up when typing, only when clicking the field. This has been fixed.

image

This PR also contains some small fixes taken from the georchestra fork which has been used in production for several months now.

Note that this component relies on https://github.com/Gbuomprisco/ngx-chips which is both outdated and not maintained any more, so we should move to a more recent library whenever working on new features on this part.


fgravin (approved):
Thanks ..
Hope we can have funding one day to refactor this sh***

PR #454, opened 2023-04-13
Updates to the section 'Generate an application' in the contributing guide

josegar74:
Minor updates to the documentation to create a new application to specify the files to update in each step and add a required step to disable the property `noPropertyAccessFromIndexSignature`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,12 +40,60 @@ You'll need manual configuration to make the application running:
 
 - Add `postcss.config.js` and `tailwind.config.js` at the root of your project if you want to use TailwindCSS.
 
-- If you want to an **application configuration**, follow the following steps:
-
-  1. Add the config file as asset for the application in `angular.json`:
+- Disable the properties `noPropertyAccessFromIndexSignature` / `noImplicitOverride` in `tsconfig.json`:
 
   ```json
-  "architect": {
+  "compilerOptions": {
+    ...
+    "noImplicitOverride": false,
+    "noPropertyAccessFromIndexSignature": false,
+    ... 
+  ```
+
+- Configure the `prefix` value in `project.json`, usually the application name:
+
+  ```json
+  {
+    "name": "my-app",
+    ...
+    "prefix": "my-app",
+    ...
+  ```
+
+- Configure the `prefix` value in `.eslintrc.json`, with the same value as specified in the previous step:
+
+  ```json
+  "@angular-eslint/component-selector": [
+    "error",
+    {
+      "type": "element",
+      "prefix": "my-app",
+      "style": "kebab-case"
+    }
+  ]
+  ```
+
+- Update the `selector` in `app.component.ts` to use the prefix specified in the previous steps:
+
+  ```ts
+  @Component({
+    selector: 'my-app-root',
+  ```
+
+- Update the `selector` in `index.html` to use the selection specified in the previous step:
+
+  ```html
+  <body>
+    <my-app-root></my-app-root>
+  </body>
+  ```
+    
+- If you want to use an **application configuration**, follow the following steps:
+
+  1. Add the config file as asset for the application in `project.json`:
+
+  ```json
+  "targets": {
     "build": {
       ...
       "options": {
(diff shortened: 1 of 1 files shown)


josegar74:
@fgravin I noticed also about this error reported in IntelliJ for some `imports`:

```
Error: A project without tags cannot depend on any libraries (nx-enforce-module-boundaries)
```

The `project.json` generated has the following:

```
  },
  "tags": []
}
```

In the Datahub project it contains:

https://github.com/geonetwork/geonetwork-ui/blob/62b3fe2b640f31b2549d3a537774b2d961e59c19/apps/datahub/project.json#L130

Can you config if that values are both required? I will update the pull request to add this additional step.

fgravin:
Hi @josegar74, welcome on board !
We are busy today with Olivia in a workshop.

> ```
> Error: A project without tags cannot depend on any libraries (nx-enforce-module-boundaries)
> ```

For this, we can secure the dependency direction using the tags.
The rules are defined here https://github.com/geonetwork/geonetwork-ui/blob/25cd675c19d71a1daccd54de55da46b1a8606d16/.eslintrc.json#L14

`type:app` should be required yes.

CONTRIBUTING.md
```diff
@@ -40,12 +40,60 @@ You'll need manual configuration to make the application running:
...
 - Add `postcss.config.js` and `tailwind.config.js` at the root of your project if you want to use TailwindCSS.
 
-- If you want to an **application configuration**, follow the following steps:
+- Disable the properties `noPropertyAccessFromIndexSignature` / `noImplicitOverride` in `tsconfig.json`:
 
-  1. Add the config file as asset for the application in `angular.json`:
+  ```json
+  "compilerOptions": {
+    ...
+    "noImplicitOverride": false,
```
fgravin, line 48: Could you explain why you had to change those values ?
Are there some errors in your app ? Or do you think these options should be a standard ?

Note that all app/lib tsconfig files extend the root one, so it may be appropriated to add this rules in the root one instead (if opportune).

josegar74: For me the `tsconfig.json` contained these 2 values:

```
    "noImplicitOverride": true,
    "noPropertyAccessFromIndexSignature": true,
```

With `"noImplicitOverride": true`, caused these type of errors in the build:

```
Error: libs/feature/search/src/lib/utils/service/fields.ts:96:3 - error TS4114: This member must have an 'override' modifier because it overrides a member in the base class 'SimpleSearchField'.

96   getAvailableValues(): Observable<FieldAvailableValue[]> {
     ~~~~~~~~~~~~~~~~~~
```

With `"noPropertyAccessFromIndexSignature": true`, caused these type of errors in the build:

```
Error: libs/feature/catalog/src/lib/organisations/organisations.service.ts:132:25 - error TS4111: Property 'eng' comes from an index signature, so it must be accessed with ['eng'].

132             group.label.eng ? group.label.eng : group.name,

```
Maybe previous versions didn't add that values to `tsconfig.json`. But at least with my setup were causing issues. I had to set to `false` to work. I see also removing them works, so probably something has change and these properties were not added previously.

fgravin: Thanks @josegar74 for explanations.
Just tested, it adds the following
```json
"compilerOptions": {
  "useDefineForClassFields": false,
  "noImplicitOverride": true,
  "noPropertyAccessFromIndexSignature": true,
},
"angularCompilerOptions": {
  "enableI18nLegacyMessageIdFormat": false,
}
```
You should just remove them instead of setting them to `false`, we should avoid to have useless properties which overrides defaults, for maintenance sake.

Actually, nx generator changes over time, tsconfig, jest config etc... so it's not a big deal. Our work it to try to keep consistency and to have good options for all these configs.
That said, I think that `noImplicitOverride` could be a correct option to add, we can see that later.

Can you update your PR with telling in the doc to remove the options instead of putting them to `false`, thanks !

fgravin:
@josegar74 could you rebase on main please cause I have fixed the CI from fork PR (hopefully :crossed_fingers: ).
Thanks

PR #455, opened 2023-04-13
Datafeeder / fix another infinite loop when polling progress

jahow:
Similar to https://github.com/geonetwork/geonetwork-ui/pull/450 but for the publishing step (at the end of the process).

The issue was exactly the same and so is the fix.


tkohr (approved):
Thanks for the additional fix @jahow !

PR #456, opened 2023-04-14
Datahub: Fix home page figures

fgravin:
Datahub home page figures are both limited
- records to 10K
- orgs to 1K

The PR removes this limit.

It also fixes an issue with nx-agent if the PR is on draft.


jahow (approved):
Thanks!

PR #459, opened 2023-04-14
fix(feed): strip html in abstract

fgravin:
Refactor so stripped abstract is available for every preview

Also test PR from fork

- [x] Works good with `nx-cloud` 
- [x] Skip the `print-affected` task (only READ mode for forks).


jahow (approved):

fgravin:
Please look @jahow that I skip `print-affected` from forks.
Please merge if it's ok for you.

PR #460, opened 2023-04-21
Improvement to license filter

jahow:
The license filter now makes the distinction between Etalab and Etalab v2 licenses, and also outputs an "unknown" value in case no license was detected.

image

Also license labels are translated now.

Inspiration from https://www.data.gouv.fr/api/1/datasets/licenses/


libs/feature/search/src/lib/utils/service/fields.ts
```diff
@@ -220,13 +234,78 @@ if (doc.containsKey('licenseObject.default.keyword') && doc['licenseObject.defau
...
   raw += doc['MD_LegalConstraintsUseLimitationObject.default.keyword'].join('|').toLowerCase();
 
-if (raw.contains('odbl')) emit('ODbL');
-if (raw.contains('pddl')) emit('PDDL');
-if (raw.contains('odc-by')) emit('ODC-By');
-if (raw.contains('cc0') || raw.contains('cc-0')) emit('CC-0');
-if (raw.contains('cc-by') || raw.contains('cc by')) emit('CC BY');
-if (raw.contains('open license')) emit('Open License');
-if (raw.contains('etalab')) emit('Etalab');`
+boolean unknown = true;
```
fgravin, line 237: This file is going to be huge in the long run.
Might be easier to decouple each field in a dedicated file, wdyt ?

libs/feature/search/src/lib/utils/service/fields.ts
```diff
@@ -220,13 +234,78 @@ if (doc.containsKey('licenseObject.default.keyword') && doc['licenseObject.defau
...
+if (raw.contains('pddl') || raw.contains('public domain dedication and licence')) {
```
fgravin, line 238: I also think that this is a code related to licence mapping and should be agnostic to this service.
For instance, if I implement harvesting with geonetwork-ui, I might need the same logic while I wouldn't need the `FieldService` things.
Let's see.

libs/feature/search/src/lib/utils/service/fields.ts
```diff
@@ -220,13 +234,78 @@ if (doc.containsKey('licenseObject.default.keyword') && doc['licenseObject.defau
...
+} else if (raw.contains('open licence') || raw.contains('licence ouverte') || raw.contains('licence_ouverte')) {
+  unknown = false;
+  emit('etalab');
+}
+
+if(unknown) emit('unknown');`
     )
   }
+
+  getAvailableValues(): Observable<FieldAvailableValue[]> {
```
fgravin, line 274: We should decouple the parent method `getAvailableValues` to fetch/process logic cause here you have to copy paste the fetch logic just to change the process one.
Or implement in the parent a `mapResponse()` and a `mapTranslatedResponse()` or something like this

fgravin (approved):
All good.
Some details eventually to think about.

libs/feature/search/src/lib/utils/service/fields.ts
jahow: Definitely, one file per class of field could make sense; I'm going to refactor things a bit to take less space

libs/feature/search/src/lib/utils/service/fields.ts
jahow: In theory yes... but this is the painless language which is ES specific, I'm not sure how we could factorize that. I considered having an object describing for each licence a set of keywords to look for and generate the painless script out of it, but then it would be too complex I think.

libs/feature/search/src/lib/utils/service/fields.ts
jahow: Done in the latest commit

jahow:
Thanks for the review! I've factorized the code to reduce the size of `fields.ts`. As for runtime fields, they have quite a big impact on performance so I think it'd be be able to generate them at indexing time somehow.

PR #464, opened 2023-04-24
[BIG Refactoring] Add abstraction for search backend client

fgravin:
Work in progress for OGC sprint.
Refactor the search client, which is now a full elasticsearch centric service, to be backend agnostic.
Add an abstraction, with clean architecture in mind and first add the gn4 (elasticsearch) implementation. This will prepare the work for adding OGC Records API later on.

We should work with DDD and try to focus on the domain and describe the models and usecases.


libs/feature/search/src/lib/domain/search.client.ts
```diff
@@ -0,0 +1,19 @@
+import { CatalogRecord } from '@geonetwork-ui/util/types/metadata'
+import { FieldFilter } from './filter.model'
+import { FieldSort } from './sort.model'
+import { FieldName } from './field.model'
+import { Observable } from 'rxjs'
+
+type Aggregations = unknown
+
+export abstract class SearchClient {
```
jahow, line 9: I like how simple this interface is!

libs/feature/search/src/lib/constants.ts
```diff
@@ -0,0 +1,35 @@
+import { FieldName } from '@geonetwork-ui/common/domain/search'
+
+export const DEFAULT_PAGE_SIZE = 10
+
+export const FIELDS_SUMMARY: FieldName[] = [
```
fgravin, line 5: This should go in the domain with domain fields

apps/datahub/src/app/home/news-page/last-created/last-created.component.ts
```diff
@@ -16,11 +16,13 @@ export class LastCreatedComponent implements OnInit {
   ) {}
 
   ngOnInit() {
-    this.searchFacade.setConfigRequestFields({
-      includes: [...ES_SOURCE_BRIEF, 'createDate', 'changeDate'],
-    })
+    this.searchFacade.setConfigRequestFields([
+      ...FIELDS_BRIEF,
```
fgravin, line 20: should be pivot format fields

fgravin (commented):
Thanks,

Few remarks
- we should move out gn4 models from the domain
- `atomic.operation` should not depend on gn4 model
- we should only manipulate the pivot format, and remove the old model
- where will you put the mappers ?
- we should spread files in folders for the domain
- `shared/links` should be moved

It's on a good way though, congratz

apps/datahub/src/app/home/news-page/last-created/last-created.component.ts
jahow: This needs a stronger abstraction and will be done later on

fgravin (commented):
Testing
- clicking on header badges does not update the sortby input

jahow:
This is rebased on main and generally finished, we need to do more tests in the various apps to make sure nothing is broken.

I will rework the commit history to make the various steps of the refactoring clearer.

libs/feature/search/src/lib/state/reducer.spec.ts
```diff
@@ -50,13 +53,13 @@ describe('Search Reducer', () => {
...
       const action = new fromActions.SetConfigFilters({
-        custom: { any: 'blah', other: 'Some value' },
-        elastic: {},
+        any: 'blah',
+        other: 'Some value',
       })
       const state = reducerSearch(initialStateSearch, action)
       expect(state.config.filters).toEqual({
-        custom: { any: 'blah', other: 'Some value' },
-        elastic: {},
```
fgravin, line 59: Will break web components
```
filter='{"elastic": [{ "term": { "OrgForResource": "Géo2France" } }]}'
```

fgravin (commented):
Massive work !
Looks good, I noticed a breaking change with the wc elastic configuration. I added a note in the release description.

We can merge and slowly test and fix the side effects.

f-necas:
🥳

jahow:
:tada:

PR #466, opened 2023-04-28
Dataviz: Improve chart

tkohr:
PR
- adds a background and border to the chart
- gives it a fixed height (like map and table) and improves positioning of its elements
- truncates long axis labels to prevent breaking charts
- now also displays loading mask, when loading data


jahow:
> Note: A potential further improvement could be to add a loader when changing chart params.

Definitely! can it be part of this PR?

tkohr:
> Definitely! can it be part of this PR?

Yes, I added a commit that fixes the loading mask, which has already been implemented for loading data. Once the data is present, generating new charts by changing its parameters seems to be so fast (even with big datasets) that the loader doesn't need to be triggered here, I have the impression.

fgravin:
UX experts say that if the loading time is < to 1sec, then we shouldn't overload the informations with a spinner.
¯\_(ツ)_/¯

tkohr:
> UX experts say that if the loading time is < to 1sec, then we shouldn't overload the informations with a spinner. `¯\_(ツ)_/¯`

Ok, this confirms to trigger the loader when loading data (like in table and map) and not when changing chart params.

libs/ui/dataviz/src/lib/chart/chart.component.ts
```diff
@@ -27,6 +27,7 @@ import {
   ScatterController,
   Tooltip,
 } from 'chart.js'
+import { truncateString } from '../utils/utils'
```
jahow, line 30: I don't think this is really necessary to introduce a new function... this util does exactly the same thing as `String.substring()`

jahow (changes requested):
Thanks, it looks much better! I had an error when loading http://localhost:4200/dataset/c1537a99-733a-49c5-b838-b4f4ca5a3e67 and using "gdb_geomattr_data" as X axis.
image.

I think it's because the `getLabelForValue` function from chart.js might return null sometimes.

libs/ui/dataviz/src/lib/chart/chart.component.ts
tkohr: I guess you are right, you'd just truncate without ending the string with `...` then?

tkohr:
Thanks for finding the error, which indeed seems to occur when the selected data column has no data. I'll add a `?` for that.

libs/ui/dataviz/src/lib/chart/chart.component.ts
jahow: oh right sorry, I overlooked the '...' added at the end! still, I think adding a new `utils/utils.ts` file is a bit awkward, I mean this function could just be part of `chart.component.ts` maybe? just to remove a bit of verbosity

jahow: the thing is it's quite specific to chart.js because usually this would be handled in CSS

tkohr: Ok, I think I got your point. I added a commit to move it to the component. Just wondering if handling the scope in the callback could be done in a more elegant way (to access both the function and the parent scope).

jahow (approved):

PR #467, opened 2023-05-02
Fix a bug in the topic facet

jahow:
This fixes a bug introduced in https://github.com/geonetwork/geonetwork-ui/pull/460 which meant that filtering on a topic wouldn't work anymore.


tkohr (approved):
Thanks for the fix @jahow, now works fine!

PR #469, opened 2023-05-03
Add a web component for viewing a dataset as a chart

jahow:
Both the table and chart components now take a `dataset-id` attribute which allows pointing at a metadata record in the api.

Also made the following improvements:
* made the init logic for web components a bit more reliable and easy to use
* made the chart and table ui components able to expand in height
* fixed the table header background color (used to be transparent)

Note: this requires #466


apps/webcomponents/src/app/components/base.component.ts
```diff
@@ -3,21 +3,31 @@ import {
   Injector,
   Input,
   OnChanges,
+  OnInit,
   SimpleChanges,
```
tkohr, line 7: Doesn't seem to be used.

jahow:
Thanks for the review @tkohr!

PR #476, opened 2023-05-11
[WIP] Better tests results & show coverage in PRs

jahow:


jest.setup.ts
```diff
@@ -4,3 +4,8 @@ import { TextEncoder, TextDecoder } from 'util'
 // see https://github.com/jsdom/jsdom/issues/2524
 global.TextEncoder = TextEncoder
 global.TextDecoder = TextDecoder
+
+// hide console output
```
fgravin, line 8: I would rather try to understand why old errors (like `NO_ERROR_SCHEMA` thing) are now piped to the `console.error`.
And I would try to keep the logging to fix them all. It might be silent issues.

fgravin (commented):
Thanks for tackling this.
Could you explain a bit what are you trying to achieve ?
To me, coverages are built under `./coverages` as an HTML website, which is not used what so ever.
I don't know if we should publish it as it is, or just try to have a badge with the total coverage on the github home page.

jahow:
I was desperately trying to have some kind of coverage information show up in pull requests, but this is really hard.

Also made the `console` calls silent when running the tests to avoid having unreadable and very long test outputs; although these should probably only be disabled when in CI mode, as they are useful in development.

jest.setup.ts
jahow: This was probably from a change in the framework. And most console calls during tests are actually expected behaviour, e.g. when testing border cases and error handling. So I think we really need something like this, at least in the CI.

jahow:
@fgravin I've switched to using Coveralls which is supposed to create PR comments (but hasn't so far). Much simpler to use, but I had to go back to a traditional workflow for the checks instead of relying on nx-cloud and agents.

fgravin (approved):
Very good.
I merge it to see the result for main and how it show the delta.

PR #471, opened 2023-05-10
[datahub] extended filtering options and improved organisation and license filters

alexarobu:
This could be an alternative to the filters already in there and would look like this:
image

I think an even better option would be enabling the user to pick from a given list of filters (as part of the `default.toml` config options). I think standards and INSPIRE keywords are quite important and quite commonly used in GeoNetwork as search filters too (from what I could see in our users' use cases).

PR #477, opened 2023-05-12
[datahub] extend search filter options to include INSPIRE keywords and standards

alexarobu:
This is to replace a previous pull request (https://github.com/geonetwork/geonetwork-ui/pull/471).

@fgravin hopefully this is cleaner :sweat_smile: I tried to take on board all your feedback on the previous pull request. I rebased the main branch on my fork and diverged this feature commit into a separate branch as you suggested, so I left my main branch alone.

Thanks for all the feedback and patience and sorry again for the mess in the previous pull request.


fgravin:
Thanks, 
The CI is failing, you should have received a notification.
Usually, when the task `Checks / Nx Cloud - Main Job / Run ` fails it means that the PR is not correct.
You can click on the details to see what's wrong and it says 
```
   Error:
   Command "npx nx format:check" failed.
```
So you must do
```
npm run format
git add -u
git commit --amend
git push -f origin datahub-search-filters
```

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -15,17 +15,35 @@ import {
 })
 export class FieldsService {
   private fields = {
-    publisher: new SimpleSearchField('OrgForResource', 'asc', this.injector),
+    publisher: new SimpleSearchField(
+      'OrgForResourceObject.default',
```
fgravin, line 19: Why this change ? Do you have a multilingual catalog ?
Testing locally the `organisation/publisher` dropdown is empty, could you check that it works ?
Thanks

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -15,17 +15,35 @@ import {
...
+      this.injector
+    ),
     format: new SimpleSearchField('format', 'asc', this.injector),
     publicationYear: new SimpleSearchField(
       'publicationYearForResource',
       'desc',
       this.injector
     ),
-    topic: new TopicSearchField(this.injector),
+    topic: new SimpleSearchField('cl_topic.default', 'asc', this.injector),
```
fgravin, line 29: Why this change ? What was the issue ?

fgravin (commented):
Some questions to better understand some changes.
Also, locally, the publisher dd is empty, not sure if it's related.

libs/feature/search/src/lib/utils/service/fields.service.ts
alexarobu: This dropdown works for me locally when I test
organisation-dropdown
The previous way was showing as blank for me, I assumed there was an issue with the ES aggregation

libs/feature/search/src/lib/utils/service/fields.service.ts
alexarobu: I used this aggregation to stay consistent with the INSPIRE keywords aggregation 
```
    inspireKeyword: new SimpleSearchField(
      'th_httpinspireeceuropaeutheme-theme_tree.default',
      'asc',
      this.injector
    ),
```

libs/feature/search/src/lib/utils/service/fields.service.ts
jahow: There might an issue with the existing aggregation with certain catalogs; I suggest we don't include that change in this PR because it might have unexpected side effects, and open an issue instead which will allow us to investigate more. @alexarobu would you agree with that?

libs/feature/search/src/lib/utils/service/fields.service.ts
jahow: Same as my comment above, I think we should exclude this change from the current PR as it will definitely break things elsewhere.

libs/feature/search/src/lib/utils/service/fields.service.ts
alexarobu: @jahow f course, I can set these to what they were and amend my commit, I don't want to break anything :see_no_evil:

libs/feature/search/src/lib/utils/service/fields.service.ts
alexarobu: as with the previous comment, I can revert this to what it was and amend the commit :smile:

alexarobu:
@jahow @fgravin I removed the changes to Organisation and Topic from the commit :smile:

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -23,9 +23,23 @@ export class FieldsService {
...
+      this.injector
+    ),
     isSpatial: new IsSpatialSearchField(this.injector),
     q: new FullTextSearchField(),
-    license: new LicenseSearchField(this.injector),
+    license: new SimpleSearchField(
+      'licenseObject.default.keyword',
+      'asc',
+      this.injector
+    ),
```
jahow, line 42: Same, this change shouldn't be part of this PR

Note that there might be an issue with your GN index configuration? We should look into this seperately

jahow:
@alexarobu I understand that you have attempted to improve the existing filters, and I think this is very encouraging. Having working filter using GN4 as a backend is always a little bit tricky as the index configuration is sometimes preventing things from working normally. I would advise to bring up any existing issues that you encountered with the existing filters in a GitHub issue so that we can investigate and eventually find what the best course of action should be.

Something very valuable would be if you could provide us with metadata records to let other people reproduce the issues you noticed.

Thanks!

libs/feature/search/src/lib/utils/service/fields.service.ts
alexarobu: @jahow with this in mind we should probably drop the PR entirely, because the added 2 filters are working in a similar manner and so might break other things. 

What do you think?

alexarobu:
> @alexarobu I understand that you have attempted to improve the existing filters, and I think this is very encouraging. Having working filter using GN4 as a backend is always a little bit tricky as the index configuration is sometimes preventing things from working normally. I would advise to bring up any existing issues that you encountered with the existing filters in a GitHub issue so that we can investigate and eventually find what the best course of action should be.
> 
> Something very valuable would be if you could provide us with metadata records to let other people reproduce the issues you noticed.
> 
> Thanks!

@jahow I am not sure if we can actually share the datasets, I will have a chat with @archaeogeek next week and see what we can come up with.

libs/feature/search/src/lib/utils/service/fields.service.ts
jahow: Well the two added filters have been working great in my tests, so I would definitely be in favor or taking them in. Adding new filters will not break anything, but modifying existing ones might.

alexarobu: @jahow fair enough. I removed any changes to the existing filters from the PR.

jahow (approved):

jahow:
Thanks and congratulations for this first contribution @alexarobu! Looking forward to improving the filters further and make them configurable as well :slightly_smiling_face:

alexarobu:
Thanks for approving this @jahow and @fgravin :heart: it's been a bit of a learning curve, but I got there in the end. Thank you for all the patience and help :pray: looking forward to making more useful contributions in the future!

fgravin:
Great, thanks for your contribution.
I still want to be sure that the vanilla code is working for you for all filters (organisation as well), is it the case ?
BTW, does the "organization" tab displays results ?

If you have an issue within your index we should figure that out.
Thanks

fgravin:
If the organization tab does not work well, it means that the records.json of your catalog is not up to date, you should destroy the index and re-index from the admin console. But be sure you have the last records.json in your index folder.

> The ones that show the organizations also pick up logos for some of the organizations. From what I understand, the logo is only picked up if there is a GeoNetwork group with the same name and email address as the organization details found in `PointofContact` in the metadata record, is that still the case?

Yes it should be

fgravin:
I think that the issue it there: https://github.com/geonetwork/core-geonetwork/blame/7fd35f6c2c58a211fe2e10a1b0f466ab8951bcae/web/src/main/webResources/WEB-INF/data/config/index/records.json#L1292

the contact has an `organisation`, not an `org`, which should be a keyword

archaeogeek:
So... in the latest version of `records.json`, we have `organisation`, and that's not working? Should I change it back to `org`?

fgravin:
> So... in the latest version of `records.json`, we have `organisation`, and that's not working? Should I change it back to `org`?

Sorry I am giving a training... no `organisation` is the correct name in the index.
You must have the correct records.json and delete and rewrite the index if it hasn't been done. Take care of your data_dir conf, to be sure to point on the correct records.json....

Eventually, when loading the organisation tab, you can have a look on the error of the ES query that feeds the organizations, it might tells what's wrong.

I think that we are not using the `org` but the `contact.organisation` field instead anyway.

PR #478, opened 2023-05-12
Dataviz: Allow displaying a chart permalink

tkohr:
PR adds the possibility to display a permalink with a chart configuration in a text input. It allows copying the permalink with the config of the current chart to the clipboard and open the chart in a web component via the static wc-embedder.html page. The location of the `web_component_embedder_url` base can be indicated in the config.

TODO:
- [x] tests
- [x] potentially move url to state

permalink


libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -129,10 +154,34 @@ export class ChartViewComponent {
...
+?e=gn-dataset-view-chart
+&a=api-url=${this.config.basePath}
+&a=aggregation=${aggregation}
+&a=x-axis=${xProp}
+&a=y-axis=${yProp}
+&a=chart-type=${chartType}
+&a=primary-color=%230f4395
+&a=secondary-color=%238bc832
+&a=main-color=%23555
+&a=background-color=%23fdfbff`
```
jahow, line 181: I would use the `URL` class here to build the permalink, it's just safer

jahow: Besides then you don't need the logic in `permalinkBaseUrl`.

Basically for me manipulating an url as a string is always a risk!

libs/ui/widgets/src/lib/permalink/permalink.component.html
```diff
@@ -0,0 +1,15 @@
...
+    readonly
+  />
+  <button
+    type="button"
+    class="py-2 px-4 rounded-r-md bg-gray-300 hover:bg-gray-500 focus:bg-gray-700 text-white"
+    (click)="copyPermalink($event)"
+  >
+    {{ 'permalink.button.copy' | translate }}
+  </button>
+</div>
```
jahow, line 15: I feel like this could be generalized to something like a "copyable text field". This is also what we will use for the web components snippet, and we could use it in any place where a "copy" button is offered (for example by offering an option to show/hide the text to copy). This is typically the case on the API links in the datahub.

In the end, it is not really specific to this permalink.

libs/feature/dataviz/src/lib/chart-view/chart-view.component.html
```diff
@@ -54,4 +57,9 @@
       <span translate>{{ error }}</span>
     </gn-ui-popup-alert>
   </div>
+  <gn-ui-permalink
+    class="mt-2"
+    *ngIf="permalinkBaseUrl"
+    [link]="permalink$ | async"
+  ></gn-ui-permalink>
```
jahow, line 64: I think making the permalink part of this component isn't the best option. This means we'll have to duplicate this for every data preview component. I would say this would be better to have in the metadata view, but maybe there are other issues I'm not thinking of.

jahow (commented):
Thanks! A couple of remarks on the organization of the code.

Also, sorry to be nitpicky but I think the name "permalink" is used a lot in this PR while it's not describing properly what the change will offer. A permalink could be anything, a link to a dataset in the datahub for example.

So the associated setting should be called something like "web-components-iframe-url" or something like that. Doesn't sound great but it does describe what this url really is, as not all permalinks in geonetwork-ui will use this system.

What do you think, does that make sense?

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -129,10 +154,34 @@ export class ChartViewComponent {
...
+    this.yProperty$.pipe(filter((value) => value !== undefined)),
+    this.aggregation$,
+    this.chartType$,
+  ]).pipe(
+    map(([xProp, yProp, aggregation, chartType]) => {
+      const link = `${this.permalinkBaseUrl}
+?e=gn-dataset-view-chart
+&a=api-url=${this.config.basePath}
+&a=aggregation=${aggregation}
+&a=x-axis=${xProp}
```
fgravin, line 175: I haven't follow the devs, but one little remark here. I think we have to be consistent in the naming for all our API
- internal API (code)
- iframe API (params)
- web component API (attributes)

Why `xProp` then `x-axis` ?

tkohr:
> Thanks! A couple of remarks on the organization of the code.
> 
> Also, sorry to be nitpicky but I think the name "permalink" is used a lot in this PR while it's not describing properly what the change will offer. A permalink could be anything, a link to a dataset in the datahub for example.
> 
> So the associated setting should be called something like "web-components-iframe-url" or something like that. Doesn't sound great but it does describe what this url really is, as not all permalinks in geonetwork-ui will use this system.
> 
> What do you think, does that make sense?

Thanks for all the feedback ! I agree about the misfitting "permalink" naming in various parts and will adapt this. What do you think of "wc-embedder-url" for the setting @jahow? This would be a little shorter ;-)

jahow:
> Thanks for all the feedback ! I agree about the misfitting "permalink" naming in various parts and will adapt this. What do you think of "wc-embedder-url" for the setting @jahow? This would be a little shorter ;-)

I want to say "yes" because long names are ugly, but on the other hand... will people really know what "wc" means? I feel like it's really not explicit enough.

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
tkohr: Ah, good point, didn't think of `URL` handling absolute and relative urls

libs/ui/widgets/src/lib/permalink/permalink.component.html
tkohr: Absolutely agree and adapted accordingly, moving it to the `inputs` lib as well

libs/feature/dataviz/src/lib/chart-view/chart-view.component.html
tkohr: Good point, I tried to adapt this, but I'm not completely satisfied yet of how the `record-metadata.component` decides whether or not to display the permalink. Maybe this can still be simplified by passing the url to the state (which I finally haven't done yet). What I'm still not seeing really clearly here is the fact that we currently don't have permalinks for table and map. As the three data preview components all exist once `record-metadata.component` is open, I guess that a url state change would have to be triggered upon changing preview tabs (and table/map would have to set the value `null` for now).

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
tkohr: Good reminder, here it was just the local variable in the stream, which I adapted. The exposed component attributes (internal API) are `xAxis`...

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -92,15 +92,23 @@
...
+      [tooltipText]="'tooltip.url.copy' | translate"
+    ></gn-ui-copy-text-button>
+  </div>
+
+  <div
+    id="access"
+    class="container-lg px-4 lg:mx-auto"
+    *ngIf="displayDownload$ | async"
+  >
+    <gn-ui-data-downloads class="block mt-5"></gn-ui-data-downloads>
```
jahow, line 108: This must be a merge error, this block is now duplicated

jahow (commented):
Impressive work! It now looks much better IMO, less complexity in record-metadata component and everything is more decoupled. I haven't done a full review yet but I think you can put this as non-draft.

Also there's one thing left that might be a problem: the name of the `PermalinkComponent` is very 'generic'. Something like `DataViewPermalinkComponent` would be better IMO, and more future-proof. What do you think?

Thanks for all the hard work on this, I'll review and test it ASAP.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
tkohr: Thanks for your attentive eyes! Indeed, had to go through some nice html rebasing :-) fixed it now.

tkohr:
Thanks for your continuous reviews @jahow. Added some tests and PR is now ready for review :-)

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
tkohr: Ok, I renamed `xAxis` to `xProperty` everywhere to be consistent here.

tkohr:
> Also there's one thing left that might be a problem: the name of the `PermalinkComponent` is very 'generic'. Something like `DataViewPermalinkComponent` would be better IMO, and more future-proof. What do you think?

You're right, renamed it.

PR #483, opened 2023-05-15
feat(orgs): Two lines for org name and fading out description

Angi-Kinas:
Screenshot from 2023-05-15 14-24-00

- We are using the line-clamp after two lines of org name
- Since we cannot know if the org name is one or two lines, we are using a fading out gradient for the description, if it's too long


jahow:
The changes are visible here: https://geonetwork.github.io/geonetwork-ui/style-change-organization/storybook/demo/?path=/story/catalog-organisationpreviewcomponent--primary

jahow (approved):
Did some more tests and the new layout works great. Congratulations on your first contribution to the project, looking forward to many more!

PR #493, opened 2023-05-24
Only show formats that are present in the links and make them more accessible

Angi-Kinas:
image

- change each download format to be a button instead of a span to make it more accessible
- refactor the filtering of the links and filtering of the formats
- add tests to test that only the formats that are present are being displayed as a filter
- change style of filter buttons

### DIFF
--- a/libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
+++ b/libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
@@ -1,15 +1,9 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  Input,
-  OnInit,
-} from '@angular/core'
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core'
 import { TranslateService } from '@ngx-translate/core'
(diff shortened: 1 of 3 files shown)


libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -1,98 +1,80 @@
...
 
-  filteredLinks: MetadataLink[] = []
-  filterButtons: FilterButton[]
+  get filteredLinks(): MetadataLink[] {
+    return this.links.filter((link) => {
+      return this.activeFilterFormats.some((format) =>
+        this.isLinkOfFormat(link, format)
+      )
+    })
+  }
```
jahow, line 36:
```suggestion
  get filteredLinks(): MetadataLink[] {
    return this.links.filter((link) =>
      this.activeFilterFormats.some((format) => this.isLinkOfFormat(link, format))
    )
  }
```

jahow (commented):
Looking good!

jahow (approved):
Well done, now we have actual accessible buttons. Thanks!

PR #495, opened 2023-05-24
Add a docker composition to run support services locally

### LINKED ISSUE #334: Include a simple GeoNetwork docker image with sample data
Having a "local" docker image would allow running end-to-end tests without relying on an external deployment of GeoNetwork, as well as running GeoNetwork-UI apps locally

jahow:
Currently this contains a Postgres database, ElasticSearch 7.11, Kibana and GeoNetwork 4.2.2.

Eventually it would make sense to include the datafeeder backend as well. I also included the OGC API Records service but it doesn't seem to work yet.

Here is what the composition provides:
* a preconfigured GN4 instance with 20+ records from various sources, a few of them having corresponding groups created; some records are multilingual
* an initialization step which bootstraps the database, starts the records indexing process and fixes the bug in the records.json file of GN 4.2.2 (see added documentation)

Hence this setup can become the reference configuration for running GeoNetwork-UI correctly. Note that this could be used to troubleshoot the various issues related to Organizations mentioned by @alexarobu in #477, as it looks like many things can break currently if the wrong version of GN4 is used.

Please let me know if that works well for everyone and we can move forward to better support organizations for GN 4.2.3+ and implement E2E testing.

Part of #338 

Fixes #334


alexarobu:
@jahow now that you mentioned the GN version, I think that's why we were seeing different results with the filters, we are running the latest version of GN, which currently is 4.2.3 I think.

alexarobu:
That would be awesome! Thank you so much for pushing this further @jahow much appreciated here. :heart:

support-services/README.md
```diff
@@ -0,0 +1,41 @@
+# Support services
+
+This folder contains a docker composition which can simulate an actual GeoNetwork 4 instance to be used
+for testing and development on GeoNetwork-UI.
+
+The instance is already prepared with a few sample records and offer a reproducible development and testing environment.
```
tkohr, line 6: offer => offers

support-services/docker-entrypoint.d/00-remove-es-watermark.sh
```diff
@@ -0,0 +1,13 @@
...
+{
+  "transient": {
+    "cluster.routing.allocation.disk.watermark.low": "30mb",
+    "cluster.routing.allocation.disk.watermark.high": "20mb",
+    "cluster.routing.allocation.disk.watermark.flood_stage": "10mb",
+    "cluster.info.update.interval": "1m"
+  }
+}'
+
+echo "Removed disk space watermark in ElasticSearch."
```
tkohr, line 13: Does it remove or set `disk space watermark`?

tkohr (approved):
Thanks for this nice and practical addition @jahow ! Starting the docker compo everything seems to run smoothly.

Just noticed console errors related to the facets when accessing [home](http://localhost:8080/geonetwork/srv/fre/catalog.search#/home) in GN. Once the facets are open (on /search), the error disappears.

gn-facet

support-services/docker-entrypoint.d/00-remove-es-watermark.sh
jahow: it essentially removes it by setting it to a very low minimum

jahow:
Thanks for the review @tkohr! By doing `docker compose up init -d` the startup time excluding pull is 100s on my machine, not too bad for e2e testing I think.

fvanderbiest:
Awesome !

PR #496, opened 2023-05-26
Fix(expandable-panel): Prevent truncating content

tkohr:
PR prevents not displaying all content in expanded panel (as it is the case in https://www.geo2france.fr/datahub/dataset/0e864105-6bb9-4c59-aae6-2ccc2e6cdd45 = "D'où viennent ces données ?").

`contentDiv?.scrollHeight` is used to maintain the transition effect, which does not work with other dynamic properties such as `fit-content`. Couldn't get a test running to check the expanded height, though. An easy alternative would be to just raise the value of `max-h-[300px]`, but then again there will still be a limit.

### DIFF
--- a/libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.html
+++ b/libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.html
@@ -14,7 +14,11 @@
 </div>
 <div
   class="content duration-500 overflow-hidden transition-[max-height]"
-  [ngClass]="collapsed ? 'max-h-0 ease-out' : 'ease-in max-h-[300px]'"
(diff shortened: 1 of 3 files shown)


jahow (approved):
Thanks! Tested on storybook and it's all good.

Do you think you could store the `contentDiv.scrollHeight` value on click instead of using it in the `ngStyle` expression? reading `scrollHeight` can trigger a reflow ([see here](https://gist.github.com/paulirish/5d52fb081b3570c81e3a#scroll-stuff)) which can really impact performance.

Nothing else to add :)

tkohr:
Thanks for the review and hint @jahow ! I've added a commit to address the reflow issue.

jahow (approved):

PR #500, opened 2023-06-07
Datahub: Display HTML code for chart web component

tkohr:
PR displays the web component HTML of the currently displayed chart.

To-do:

- [x] refactor `data-view-permalink` and `data-view-web-component` into one component, including the tabs currently in `record-metadata` that only displays if `web_component_embedder_url` setting is set. Or add one intermediate component which only handles the tabs and the display condition (`web_component_embedder_url`).

web-component-html


libs/feature/record/src/lib/data-view-share/data-view-share.component.html
```diff
@@ -0,0 +1,23 @@
+<div *ngIf="wcEmbedderBaseUrl" class="container-lg px-5 my-1 lg:mx-auto">
```
jahow, line 1: If the embedder url is not provided we can still show the webcomponent export, right?

libs/feature/record/src/lib/data-view-web-component/data-view-web-component.component.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+          x-property="${chartConfig1.xProperty}"
+          y-property="${chartConfig1.yProperty}"
+          chart-type="${chartConfig1.chartType}"
+          primary-color="#0f4395"
+          secondary-color="#8bc832"
+          main-color="#555"
+          background-color="#fdfbff"
+          main-font="'Inter', sans-serif"
+          title-font="'DM Serif Display', serif"
+        ></gn-dataset-view-chart>`
```
jahow, line 96: The indentation here makes the end result a little bit confusing:
image

Something like this would be more readable IMO:
```
<script src="https://cdn.jsdelivr.net/gh/geonetwork/geonetwork-ui@wc-dist/gn-wc.js"></script>
<gn-dataset-view-chart
    api-url="http://localhost:4200/geonetwork/srv/api"
    dataset-id="ded8164c-d56b-4ae0-b93d-459fc7c99a01"
    aggregation="sum"
    x-property="insee_com"
    y-property="superf_km"
    ...
></gn-dataset-view-chart>
```

libs/ui/inputs/src/lib/copy-text-button/copy-text-button.component.html
```diff
@@ -1,11 +1,18 @@
...
     type="text"
-    class="grow py-2 px-4 bg-gray-200 opacity-50 rounded-l-md border border-gray-300"
+    class="grow py-2 px-4 bg-gray-200 opacity-50 rounded-l-md border border-gray-300 truncate"
     value="{{ text }}"
     readonly
   />
+  <textarea
+    *ngIf="displayText && rows > 1"
+    [rows]="rows"
+    class="grow py-2 px-4 bg-gray-200 opacity-50 rounded-l-md border border-gray-300 overflow-y-hidden h-max"
```
jahow, line 12:
```suggestion
    class="grow py-2 px-4 bg-gray-200 opacity-50 rounded-l-md border border-gray-300 overflow-y-auto h-max"
```

Otherwise there is no way to read the whole text without copypasting or expanding the text area.

jahow (commented):
Thanks! and sorry for taking so long to review. I thought we could still show the web component export if no embedder url is provided in the config? Not much else, only minor remarks, I've tested it and it works fine, thanks :)

libs/feature/record/src/lib/data-view-share/data-view-share.component.html
tkohr: I was wondering what would make most sense and rather saw the display of both together. If we want the `data-view-web-component` also without the `data-view-permalink` I think I'll move the decision whether to display the `data-view-share` (with both components) or only the `data-view-web-component` to the `record-metadata` based on `*ngIf="wcEmbedderBaseUrl"`. This will prevent to display the "Share" tab for the `data-view-permalink` if it's not activated.

libs/feature/record/src/lib/data-view-web-component/data-view-web-component.component.spec.ts
tkohr: Good point!

libs/ui/inputs/src/lib/copy-text-button/copy-text-button.component.html
tkohr: I prefer the existing option of expanding the text area in the lower right corner. Don't you think it is sufficient? The vertical scrollbar appears pretty ugly to me. We could also raise the number of rows, but this might become too intrusive.

jahow: The scroll bar makes sense I guess, scroll bars are always ugly but not being able to scroll really struck me as an issue. Also think of keyboard accessibility ;)

libs/feature/record/src/lib/data-view-share/data-view-share.component.html
tkohr: So I added 1a15c881395e13a660e456d7a93667f8fbf265cc for this, encapsulating the logic in the `data-view-share` in the end as I think displaying the tab button as a "label" makes sense as well.

libs/ui/inputs/src/lib/copy-text-button/copy-text-button.component.html
tkohr: Ok, I added the scrollbar for the sake of accessibility ;-)

tkohr:
Thanks for your feedback @jahow ! I've addressed your comments in the latest commits.

jahow (approved):
Looking good, thanks!

PR #501, opened 2023-06-09
fix(story): change hovering from opacity to underline

f-necas:
Set underline instead of opacity change when hovering and move website under email with full link and not only icon.

image

contact

### DIFF
--- a/libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
+++ b/libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
@@ -4,19 +4,22 @@
   </p>
   <div class="flex mb-1">
     <div
-      class="text-primary font-title text-21 mr-2 cursor-pointer hover:opacity-50"
+      class="text-primary font-title text-21 mr-2 cursor-pointer hover:underline"
       (click)="onContactClick()"
     >
       {{ shownContact.organisation }}
     </div>
-    <div *ngIf="shownContact.website">
-      <a [href]="shownContact.website" target="_blank"
-        ><mat-icon
-          class="text-primary opacity-25 cursor-pointer hover:opacity-75 transition-all"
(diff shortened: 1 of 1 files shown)


tkohr:
Thanks @f-necas ! The changes LGTM. Just seeing that the CI breaks because of formatting. I recommend setting up your IDE to use the prettier config of the project "on save" to prevent having to run `npm run format` manually.

libs/ui/search/src/lib/record-preview-row/record-preview-row.component.html
```diff
@@ -8,7 +8,8 @@
     >
       <gn-ui-thumbnail
         class="relative h-full w-full object-cover object-left-top"
-        [thumbnailUrl]="record.thumbnailUrl"
+        [thumbnailUrl]="record.thumbnailUrl || record.contact?.logoUrl"
```
tkohr, line 11: Oups, I think this slipped in from your other modifications...

fgravin (commented):
Thanks @f-necas for your work in gn-ui, looks a pretty good start :smiley: 
Just a global remark, you should pay attention on keeping a clean commit history and don't emphase on the struggles you went through during your implementation.

For instance, 'fix', 'revert' or other small mistakes should be amended when possible.
This kind of PR would deserve a single commit.
It's not a big deal at all, but just some advice for better contributions.

On the opposite, note that when you achieve a complex work, it's really better to have several small feature-centric commits rather than a big one, it really helps the review. 

To wrap up, keep a detailed & clear history and avoid noise commits.
Thanks :+1:

f-necas:
Thanks @fgravin and @tkohr for your advices. Next commits  will be better :smile: !

tkohr (approved):
Thanks @f-necas, LGTM now and CI passes well :-)

PR #502, opened 2023-06-09
E2E: Write a first cypress end-to-end test for the Datahub

cmoinier:
### JIRA issue
https://jira.camptocamp.com/browse/GSHDF-701

### To-do

- [x]  Header
- [x] Home page
- [x] Dataset list
- [x] Organizations
- [x] Dataset page (remains : external links)

### Run the tests

`ng e2e datahub-e2e --watch` to see the tests live

or 

`ng e2e datahub-e2e` to run in headless mode


jahow:
Relates to https://github.com/geonetwork/geonetwork-ui/issues/338

fvanderbiest:
Awesome. Love it ! 
Excellent idea.

cmoinier:
> Thanks! it looks like your IDE isn't using prettier for formatting, so you might have changes when running `npm run format:write`. You can probably set up automatic prettier formatting on save which will avoid this kind of issue :)

Thanks for the tip! I set it up and also ran format:write just to be sure. hope it's good this time :)

fgravin (commented):
Thanks for this heavy work.
I think I have some issues running the test, what did I miss ?

I just tried to run the test

- `nx e2e datahub-e2e --port='cypress-auto'` some tests are failing
- `apps/datahub-e2e$ npx cypress run` all tests are failing

README.md
```diff
@@ -175,6 +177,41 @@ npx nx test (lib_name)
...
+- search page
+- organisations page
+- dataset pages
+
+##### To run the tests with the interface :
+
+Start docker from 'support-services', and then :
+
+```shell script
+npx nx e2e datahub-e2e --port='cypress-auto' --watch
```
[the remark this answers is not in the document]
cmoinier: Weird... Just tested and it works on my end. 
Node 16.17 here.

@tkohr do you remember if we had to do another install or something before starting the tests?

README.md
```diff
@@ -175,6 +177,41 @@ npx nx test (lib_name)
...
+Then select the file(s) you want to test in the interface.
+
+##### To run the tests without interface :
+
+Start docker from 'support-services', and then :
+
+--> ALl tests :
+
+```shell script
+npx cypress run
```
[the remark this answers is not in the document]
cmoinier: edited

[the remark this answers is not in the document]
cmoinier: Never looked into it since I only launched the tests with the  `--watch` command. I know that there is a package named `@nx/cypress`

README.md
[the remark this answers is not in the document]
tkohr: Looking at the error I get on my side (`Error: Can't resolve 'cypress-real-events'`), I have the impression you forgot to commit some changes in your `package.json` @cmoinier.

README.md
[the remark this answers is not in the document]
tkohr: Indeed, the `nx` command further up uses `@nrwl/cypress` which is rescoped to `@nx/cypress` in nx version 16. So I guess we should privilege this!?

tkohr:
Thanks for the big work @cmoinier ! I think we can still delete all the e2e apps besides the datahub one, as we had to regenerate it anyway, due to cypress evolutions since their initial generation.

README.md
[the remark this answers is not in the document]
cmoinier: Will look into it.

cmoinier:
> Thanks for the big work @cmoinier ! I think we can still delete all the e2e apps besides the datahub one, as we had to regenerate it anyway, due to cypress evolutions since their initial generation.

Yes indeed! I will remove them with the next commit.

README.md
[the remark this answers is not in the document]
cmoinier: Okay, I just realized that `cypress-real-events` is actually a dependency I installed for some tests. Which is why it won't work on your end. You will need to install it.

README.md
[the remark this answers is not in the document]
cmoinier: installed `@nx/cypress` and edited `README.md`.

README.md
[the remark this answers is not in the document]
tkohr: Could you add it to the `package.json`? This will ensure everyone running an `npm install` on the project to have it.

cmoinier:
> Thanks for this heavy work. I think I have some issues running the test, what did I miss ?
> 
> I just tried to run the test
> 
>     * `nx e2e datahub-e2e --port='cypress-auto'` some tests are failing
> 
>     * `apps/datahub-e2e$ npx cypress run` all tests are failing

Steps I took to make all tests work : 

- Installed `cypress-real-events` dep.
- Installed `@nx/cypress`.
- Ran `ng e2e datahub-e2e --browser chrome`
- I also fixed a few tests that still weren't passing due to timeout and browser permissions in headless mode.
- To fix some tests, I needed to install the dep : `cypress-browser-permissions` with `npm i cypress cypress-browser-permissions --save-dev` .
- Dependencies are now in `package.json`.
- -WARNING : there is ONE test that keeps failing because of the headless mode not finding some elements that `--watch` mode finds. Will fix tomorrow.

I also removed the old e2e apps.

jahow:
For the record, I ran the tests using `ng e2e datahub-e2e` and got this:
image

cmoinier:
> > For the record, I ran the tests using `ng e2e datahub-e2e` and got this: image
> > The differences with the `cypress` binary is most likely due to a difference in options.
> > Edit: same thing with `ng e2e datahub-e2e --headed`
> 
> Yes, most of the failing tests are failing due to the default browser. You will need to add the flag `--browser chrome` to run them properly. This is needed because one of the dependencies only works with chrome. Sorry, I forgot to edit the `README.md` file.
> 
> Edit : Actually, I ran this command just like you did and weirdly enough, all tests passed with the exception of `datasetDetailPage` for which 4 tests fail. Tests that never failed before... Will look into it in the morning.

As of my last commit, by running `ng e2e datahub-e2e`, all tests pass : 

Screenshot from 2023-07-25 08-58-22

README.md
```diff
@@ -175,6 +177,35 @@ npx nx test (lib_name)
...
+- search page
+- organisations page
+- dataset pages
+
+##### To run the tests with the interface :
+
+Start docker from 'support-services', and then in the 'geonetwork-ui' folder :
+
+```shell script
+ng e2e appname --watch
```
fgravin, line 194:
```suggestion
nx e2e appname --watch
```

fgravin: Error while running this command.
BTW, it's better to also have the command to run in the PR description, it's easier to review :pray: 
```
 >  NX   Cannot find configuration for task datahub:e2e
```

fgravin (commented):
I can't run the tests, cf my comments.

README.md
cmoinier: Well, I get `nx command not found` by running `nx e2e appname --watch`. Works with `ng` only for me which is why I commited that command.

fgravin: `nx` and `ng` are the same command, it's weird that you don't have `nx`, but I've installed it globally.
Your could then use `npx nx`, ng should be removed soon (from next nx migration).

With `ng e2e datahub --watch` I have the same error, is `datahub` a correct input ?

cmoinier: No, the right input is `datahub-e2e`. Let me know if that was the issue.

apps/datahub-e2e/src/e2e/header.cy.ts
```diff
@@ -0,0 +1,125 @@
+describe('header', () => {
+  beforeEach(() => cy.visit('/'))
```
fgravin, line 2: You may need to introduce mocks & fixtures
eg.
```
    beforeEach(() => {
        cy.intercept('GET', /your_url/, { fixture: response_fixture }).as('your_alias');
    });
```

apps/datahub-e2e/cypress.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'cypress'
+import { nxE2EPreset } from '@nrwl/cypress/plugins/cypress-preset'
+
+export default defineConfig({
```
fgravin, line 4: I don't think that we the video.

cmoinier: I think you're missing a word here!

apps/datahub-e2e/src/e2e/header.cy.ts
cmoinier: Yes, Olivia and are planning to look into that in the coming days.

apps/datahub-e2e/cypress.config.ts
fgravin: Yes we don't need the video :smile: 
We could disable it.

cmoinier: added the param.

README.md
jahow: Generally we should invite users to use the `nx` cli; to run it, you probably need to do `npx nx e2e datahub-e2e --watch` for example (npx is needed if the binary is not globally installed, which is probably the case)

jahow:
@cmoinier committed some improvements & fixtures as discussed this morning, there are still random failures sometimes but not very often. I think this is good to merge for now, let me know if you think of anything else, thanks!

Note: here are the durations I've got on my latest run:
* dataset details: 17s
* datasets: 51s
* header: 11s
* home: 5s
* orgs: 6s

jahow (approved):
Truly impressive work, congratulations @cmoinier :star_struck:

cmoinier:
Thanks @jahow for the rework :)

I updated the Cypress version since the older one was creating a crash with the Chrome DevTools after running the first file. One test still crashes pretty often.

PR #503, opened 2023-06-09
Update url structure to remove /home

f-necas:
`home/*` is now redirected to `/*`.

- `/ `> `/news`
- `/home` > `/news` (:warning: formerly `/home` was redirected to `/home/search`. Seen with @tkohr ) 
- `home/news` > `/news`
- `home/organisations` > `/organisations`
- `home/search` > `/search`

### DIFF
--- a/apps/datahub/src/app/home/home-header/home-header.component.ts
+++ b/apps/datahub/src/app/home/home-header/home-header.component.ts
@@ -39,7 +39,7 @@ export class HomeHeaderComponent {
   ) {}
 
   displaySortBadges$ = this.routerFacade.currentRoute$.pipe(
-    map((route) => route.url[0].path === ROUTER_ROUTE_NEWS)
+    map((route) => route.url[0]?.path === ROUTER_ROUTE_NEWS)
   )
 
   isAuthenticated$ = this.authService
--- a/apps/datahub/src/app/home/home-header/home-header.component.ts
+++ b/apps/datahub/src/app/home/home-header/home-header.component.ts
@@ -39,7 +39,7 @@ export class HomeHeaderComponent {
   ) {}
 
   displaySortBadges$ = this.routerFacade.currentRoute$.pipe(
-    map((route) => route.url[0]?.path === ROUTER_ROUTE_NEWS)
+    map((route) => route.url[0].path === ROUTER_ROUTE_NEWS)
   )
 
   isAuthenticated$ = this.authService
--- a/apps/datahub/src/app/home/navigation-menu/navigation-menu.component.ts
+++ b/apps/datahub/src/app/home/navigation-menu/navigation-menu.component.ts
@@ -39,7 +39,7 @@ export class NavigationMenuComponent {
   activeLink$ = this.routerFacade.currentRoute$.pipe(
     map(
       (route) =>
-        this.tabLinks.find((tab) => tab.link === route.url[0].path) || {
+        this.tabLinks.find((tab) => tab.link === route.url[0]?.path) || {
           link: '',
           label: '',
         }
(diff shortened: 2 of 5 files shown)


tkohr:
Thanks for the efficient change @f-necas! Works as expected. Don't forget to adapt the tests;-)

apps/datahub/src/app/router/datahub-router.service.ts
```diff
@@ -29,16 +29,21 @@ export class DatahubRouterService {
     return [
       {
         path: '',
-        redirectTo: `${ROUTER_ROUTE_HOME}/${ROUTER_ROUTE_NEWS}`,
+        redirectTo: ROUTER_ROUTE_NEWS,
         pathMatch: 'full',
       },
       {
         path: ROUTER_ROUTE_HOME,
```
fgravin, line 36: There are 2 redirections for `ROUTER_ROUTE_HOME`, one should be useless then.

apps/datahub/src/app/router/datahub-router.service.ts
```diff
@@ -29,16 +29,21 @@ export class DatahubRouterService {
...
+        redirectTo: ``,
+        pathMatch: 'prefix',
+      },
+      {
+        path: '',
```
fgravin, line 46: Does it work if you just have one config for `''` path instead of 2 like here ?

f-necas: It does'nt. If I remove the 'full' one, slash path is broken, and if I remove the 'prefix' it's new/search/org which is broken.

apps/datahub/src/app/router/datahub-router.service.ts
f-necas: The one that redirect to news is, in sort of, useless. It's just in the case if someone hit the `/home/` path.  I don't know if there will be some redirection to this old path from other websites/softwares, but let it in case it does. 
Should I remove it ?

fgravin: IMO yes cause nobody goes to `/home`, either they went to `/` or `/home/news`

f-necas: /Home removed

tkohr (approved):
Retested it and works fine besides the pure `/home` redirect which should not occur. Thanks for the cleanup of the `ROUTER_ROUTE_HOME` leftovers!

apps/datahub/src/app/home/home-header/home-header.component.ts
```diff
@@ -39,7 +39,7 @@ export class HomeHeaderComponent {
   ) {}
 
   displaySortBadges$ = this.routerFacade.currentRoute$.pipe(
-    map((route) => route.url[0].path === ROUTER_ROUTE_NEWS)
+    map((route) => route.url[0]?.path === ROUTER_ROUTE_NEWS)
```
fgravin, line 42: What was the issue about this ?

apps/datahub/src/app/home/navigation-menu/navigation-menu.component.ts
```diff
@@ -39,7 +39,7 @@ export class NavigationMenuComponent {
   activeLink$ = this.routerFacade.currentRoute$.pipe(
     map(
       (route) =>
-        this.tabLinks.find((tab) => tab.link === route.url[0].path) || {
+        this.tabLinks.find((tab) => tab.link === route.url[0]?.path) || {
```
fgravin, line 42: same ?

fgravin (commented):
I've made some improvements, in term of clarity
Please check if it works as expected and adapt the test.
Thanks

apps/datahub/src/app/home/navigation-menu/navigation-menu.component.ts
f-necas: With my fist attempt, when `/home` was hit, some errors were displayed in the console and a blank page appeared.
it doesn't happen with your imrovements. Thansk @fgravin !

fgravin (approved):

PR #507, opened 2023-06-16
feat: can launch custom scripts at startup of container

f-necas:
Custom sh scripts can be deployed into a container by setting the environment variable `CUSTOM_SCRIPTS_DIRECTORY` pointing to a mounted volume.

It can be used to deploy custom header for instance. 

Env variable and volume has to be set and scripts must be shebang sh
Cmd Example :
```bash
$ docker run -p 8080:80 \
             -v /home/user/my-scripts:/some/random/path \
             -e CUSTOM_SCRIPTS_DIRECTORY=/some/random/path \
             geonetwork/geonetwork-ui-datahub
```

@edevosc2c

### DIFF
--- a/apps/datahub/project.json
+++ b/apps/datahub/project.json
@@ -26,11 +26,6 @@
             "glob": "*",
             "input": "conf",
             "output": "assets/configuration/"
-          },
-          {
-            "glob": "*",
-            "input": "apps/datahub/docker-entrypoint.d",
-            "output": "docker-entrypoint.d/"
           }
         ],
         "styles": [
--- a/apps/datahub/project.json
+++ b/apps/datahub/project.json
@@ -26,6 +26,11 @@
             "glob": "*",
             "input": "conf",
             "output": "assets/configuration/"
+          },
+          {
+            "glob": "*",
+            "input": "apps/datahub/docker-entrypoint.d",
+            "output": "docker-entrypoint.d/"
           }
         ],
         "styles": [
--- a/tools/docker/docker-entrypoint.sh
+++ b/tools/docker/docker-entrypoint.sh
@@ -59,8 +59,6 @@ if [[ -z "${CUSTOM_SCRIPTS_DIRECTORY}" ]]; then
   echo "[INFO] No CUSTOM_SCRIPTS_DIRECTORY env variable set"
 else
   echo "[INFO] CUSTOM_SCRIPTS_DIRECTORY env variable set to ${CUSTOM_SCRIPTS_DIRECTORY}"
-  chmod +x ${CUSTOM_SCRIPTS_DIRECTORY}/*.sh  
-  cd ${CUSTOM_SCRIPTS_DIRECTORY}/
   run-parts ${CUSTOM_SCRIPTS_DIRECTORY}/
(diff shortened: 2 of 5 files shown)


fgravin (commented):
Hi @f-necas, could you please give more context in the description of the PR ?
Can you give some use cases for instance.
Thanks

apps/datahub/project.json
```diff
@@ -26,6 +26,11 @@
             "glob": "*",
             "input": "conf",
             "output": "assets/configuration/"
+          },
+          {
```
fgravin, line 30: why do you need that ?

tools/docker/docker-entrypoint.sh
```diff
@@ -54,6 +54,17 @@ else
   echo "[INFO] No custom assets found at ${CUSTOM_ASSETS_PATH}"
 fi
 
+# Executing custom scripts located in CUSTOM_SCRIPTS_DIRECTORY if environment variable is set
+if [[ -z "${CUSTOM_SCRIPTS_DIRECTORY}" ]]; then
+  echo "[INFO] No CUSTOM_SCRIPTS_DIRECTORY env variable set"
+else
+  echo "[INFO] CUSTOM_SCRIPTS_DIRECTORY env variable set to ${CUSTOM_SCRIPTS_DIRECTORY}"
+  chmod +x ${CUSTOM_SCRIPTS_DIRECTORY}/*.sh  
+  cd ${CUSTOM_SCRIPTS_DIRECTORY}/
```
jahow, line 63: I'm not sure if this line is really necessary; it works with it because the directory is an absolute path, but it could be misleading maybe?

tools/docker/docker-entrypoint.sh
```diff
@@ -54,6 +54,17 @@ else
   echo "[INFO] No custom assets found at ${CUSTOM_ASSETS_PATH}"
 fi
 
+# Executing custom scripts located in CUSTOM_SCRIPTS_DIRECTORY if environment variable is set
+if [[ -z "${CUSTOM_SCRIPTS_DIRECTORY}" ]]; then
+  echo "[INFO] No CUSTOM_SCRIPTS_DIRECTORY env variable set"
+else
+  echo "[INFO] CUSTOM_SCRIPTS_DIRECTORY env variable set to ${CUSTOM_SCRIPTS_DIRECTORY}"
+  chmod +x ${CUSTOM_SCRIPTS_DIRECTORY}/*.sh
```
jahow, line 62: I wouldn't do that here, the scripts should be made executable before running the docker image in my opinion. Modifying the file (even just its rights) from a docker run is a little bit risky; what if it fails?

jahow (approved):
Great, clean and simple, great job! I've tested it locally and it works just fine. I just made two comments, feel free to address and merge when you consider it done. Thanks!

tools/docker/docker-entrypoint.sh
edevosc2c: agree here

tools/docker/docker-entrypoint.sh
edevosc2c: I don't think it's necessary either

PR #508, opened 2023-06-18
Add the possibility to load organizations from group instead of metadata

jahow:
This PR refactors significantly the way organizations are handled across the project.

The organizations are handled by the `OrganisationsServiceInterface` which has two implementations: one getting orgs from metadata (default behavior) and one getting orgs from groups.

This service is now used for:
* getting a list of organizations
* creating a search filter for organizations
* read the organization from an ElasticSearch source record at the mapping phase

The general default behavior should not be impacted, except that search results should now have a full organization attached to them (including information from GN groups when available).

To enable the "groups" orgs strategy, use:
```ts
    {
      provide: ORGANIZATIONS_STRATEGY,
      useValue: 'groups',
    },
```


libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -44,19 +50,52 @@ export class FieldsService {
...
+        filters.reduce((prev, curr) => ({ ...prev, ...curr }), {})
+      )
+    )
+  }
+
+  getFieldValuesForFilters(filters: SearchFilters): Observable<FieldValues> {
+    return of(...this.supportedFields).pipe(
+      mergeScan(
+        (prev, curr) =>
+          this.getValuesForFilters(curr, filters).pipe(
```
fgravin, line 90: I noticed that the observable ticks too often for the organisations, it's always checking for the `filterValues` even though there is no organisation in the route.

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -44,19 +50,52 @@ export class FieldsService {
...
+      return this.getFiltersForValues(fieldName, values as FieldValue[])
+    })
+    return combineLatest(filtersByField$).pipe(
+      map((filters) =>
+        filters.reduce((prev, curr) => ({ ...prev, ...curr }), {})
+      )
+    )
+  }
+
+  getFieldValuesForFilters(filters: SearchFilters): Observable<FieldValues> {
```
fgravin, line 86: It was not obvious to me what this method was supposed to do, from its name.
I thought that it fetches all values for a filter (facets).

fgravin (approved):
Huge work :clap: 
Everything works very well.
I've made some comments in the other repository about keeping the idea to centralize all the impacts of the organisations. Anyway we should move forward and check that afterwards after that some other refactoring & features are merged.

fgravin:
Oh, note @jahow that there is an issue with the `group` strategy on geo2france, this should be addressed before merging

jahow:
Thanks for the review @fgravin, I addressed all comments, merge time! :boom:

PR #511, opened 2023-06-22
feat: adding organisations thumbnails to records

f-necas:
Organisations thumbnails are now visible in feed and datasets tabs. 

If the organization strategy is set to metadata (default behaviour), matching is done on the organization name set in the contact property of the record.

Maybe the IMAGE_URL constant could be common in `organisations-from-groups.service.ts` and `organisations-from-metadata.service.ts`.

@fgravin @jahow @tkohr Tell me if this solution would be suitable. 
Thanks :)

### DIFF
--- a/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
+++ b/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
@@ -17,11 +17,9 @@ import {
   selectField,
   SourceWithUnknownProps,
 } from '@geonetwork-ui/util/shared'
-import { combineLatest, forkJoin, Observable, of } from 'rxjs'
-import { filter, map, shareReplay, startWith } from 'rxjs/operators'
+import { combineLatest, Observable, of } from 'rxjs'
+import { filter, map, shareReplay, startWith, takeLast } from 'rxjs/operators'
 import { OrganisationsServiceInterface } from './organisations.service.interface'
-import { TranslateService } from '@ngx-translate/core'
-import { LANG_2_TO_3_MAPPER } from '@geonetwork-ui/util/i18n'
 
 const IMAGE_URL = '/geonetwork/images/harvesting/'
 
@@ -78,8 +76,7 @@ export class OrganisationsFromMetadataService
   constructor(
     private esService: ElasticsearchService,
     private searchApiService: SearchApiService,
-    private groupsApiService: GroupsApiService,
-    private translateService: TranslateService
+    private groupsApiService: GroupsApiService
   ) {}
 
   equalsNormalizedStrings(
@@ -167,13 +164,12 @@ export class OrganisationsFromMetadataService
   }
 
   private mapContactFromOrganisation(
-    organisation: Organisation,
-    lang3: string
+    organisation: Organisation
   ): MetadataContact {
     const logoUrl = getAsUrl(`${organisation.logoUrl}`)
     return {
-      name: organisation.name[lang3],
-      organisation: organisation.name[lang3],
+      name: organisation.name,
+      organisation: organisation.name,
       email: organisation.email,
       ...(organisation.logoUrl && logoUrl && { logoUrl }),
     } as MetadataContact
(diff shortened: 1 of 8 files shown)


libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -184,17 +202,31 @@ export class OrganisationsFromMetadataService
...
+        ],
+        contact: {
+          ...mapContact(getFirstValue(selectField(source, 'contact')), source),
+        },
+      }),
+      this.organisations$,
+    ]).pipe(
+      map(([record, organisations]) => {
+        const org = organisations.filter(
+          (o) => o.name === record.contact.organisation
```
fgravin, line 221: `resourceContact[0]`

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -184,17 +202,31 @@ export class OrganisationsFromMetadataService
...
+        )[0]
+        if (org) {
+          const lang3 = LANG_2_TO_3_MAPPER[this.translateService.currentLang]
+          record.contact = this.mapContactFromOrganisation(org, lang3)
```
fgravin, line 225: avoid object mutation

fgravin (changes requested):
Looks good to me except the mandatory changes mentioned in the comments.

Overall, I would have some comments about the implementation from @jahow 
1. It's disturbing that we add duplicated contacts within the `contacts` & `resourcesContact` properties of the `MetadataRecord` objects. I wonder if we shouldn't clearly separate this concern with an `organization` property in the record.
2. `mapContact()` could take an optional 3rd argument as the `organization`, for the group strategy, the group is mapped to the organisation first.
3. The records should not have the catalog logo by default
4. I think we should try to maintain a cache because the mapping between contacts & organisation + the hydration would be repeated a lot for nothing. Not in this PR though.

Overall the organisation domain object should raise above contacts/groups, we should just treat with organisations; contacts (for metadata) or groups (for group) should be mapped to an organisation and be forgotten.

jahow:
> 1. It's disturbing that we add duplicated contacts within the `contacts` & `resourcesContact` properties of the `MetadataRecord` objects. I wonder if we shouldn't clearly separate this concern with an `organization` property in the record.

Yes clearly the record format is awkward on that part. I think we should address that in the pivot format and not overthink it with the current `MetadataRecord` format though.

> 2. `mapContact()` could take an optional 3rd argument as the `organization`, for the group strategy, the group is mapped to the organisation first.

I feel like simply adding a 3rd optional argument would be a work around for something missing elsewhere, but not really sure about this.

> 3. The records should not have the catalog logo by default

I think that's the point of this PR, display the thumbnail if any and display the contact logo otherwise. Or am I missing something? Maybe we shouldn't overwrite any existing logo on the record?

> 4. I think we should try to maintain a cache because the mapping between contacts & organisation + the hydration would be repeated a lot for nothing. Not in this PR though.

Why not but in my understanding the contacts in the raw metadata can have many different forms and the method used to match with the groups use several fields. We can't just maintain a map between an org email and a group for example; the mapping could use a combination of the org fields (name, email...). But I also think that this concern would be "premature optimization": the hydration happens maybe at most twenty times in a row when showing the records list, and I don't think this is going to be noticeable in any way.

fgravin:
> I feel like simply adding a 3rd optional argument would be a work around for something missing elsewhere, but not really sure about this.

Thinking about this and the issue is more about the confusion between contact & orgs.
The mapper and atomic-operations are just responsible for mapping the index to our model, at least it should be. Once we are hydrating this model from other sources (other than the index, eg. groups/orgs), then we overextend the responsibility of the mapper IMO.
And the question raised again: should we extend metadata informations from other sources (eg contacts from orgs/groups), or shall we add new properties in the record (eg organisations).

f-necas:
> And the question raised again: should we extend metadata informations from other sources (eg contacts from orgs/groups), or shall we add new properties in the record (eg organisations).

For me, I think it would be clearer if we edit the MetadataRecordInterface with something like : 

```ts
export interface MetadataRecord {
  id: string
  ...
  contact?: MetadataContact
  resourceContacts?: MetadataContact[]
  organisation?: Organisation
  ...
}

```

fgravin:
It's not clear in the US, but to me the rules are
1. use record thumbnail if present
2. if not, use organization logo if present
3. if not, use placeholder

f-necas:
> It's not clear in the US, but to me the rules are
> 
> 1. use record thumbnail if present
> 2. if not, use organization logo if present
> 3. if not, use placeholder

It's conflicting what @jahow told me : Keeping catalog logo if org logo is not present for the feed.
For the feed, organisation logo is set with org.logo or record.logo (which is catalog logo in the index).
For the search,  we take record thumbnail or organisation logo ( so it takes catalog logo if none present ). 

@fgravin Can you confirm that we won't use record logo (catalog's logo) for organisation logo ?  It will change actual display of logos in the feed preview

fgravin:
> @fgravin Can you confirm that we won't use record logo (catalog's logo) for organisation logo ? It will change actual display of logos in the feed preview

Ok, to me this is the kind of things to be discussed with Geo2France/MEL, but you can go as @jahow told you for now.
Keep using catalog logo as a fallback.
Thanks

fgravin (approved):
Thanks for adding the missing behavior of hydrating the records with the organisation details.
Works well, but it lacks of tests. You should test the hydration.

I still have a weird behavior (which is expected), i have an org with a logo in the organization tab, when i click on it, the metadata in the result list has the catalog logo instead of the org one (because the organisation is the second `resourceContact`). Actually, having several `resourceContact` breaks a bit the approch we have, because the record does not have only one organisation, anyway.

Feel free to merge after adding a test.
Thanks, great work

PR #512, opened 2023-06-26
feat: Set star animation to primary instead of hardcoded red

f-necas:


jahow (approved):
Thanks! the hardcoded red was chosen to be brighter and more saturated, but this works well with the current theme. Maybe if the need arises again we can add a `--color-primary-saturated` variable for this kind of uses :)

PR #513, opened 2023-06-28
Add ability to predefine specific search in config

Angi-Kinas:
Proposal: (Admin) users will now be able to add predefined searches through the `default.toml` file that will appear under the main search field as a button.
One or several search presets can be defined; every search preset is composed of:
- a name (which can be a translation key)
- a sort criteria: either `createDate`, `userSavedCount` or `_score` (prepend with `-` for descending sort)
- filters which can be expressed like (see `default.toml`)

Screenshot:
image

Todo:

- [x] Add screenshot
- [x] Add a proposal for the configuration schema changes

### DIFF
--- a/libs/util/shared/src/lib/services/proxy.service.spec.ts
+++ b/libs/util/shared/src/lib/services/proxy.service.spec.ts
@@ -2,7 +2,7 @@ import { PROXY_PATH, ProxyService } from './proxy.service'
 import { TestBed } from '@angular/core/testing'
 
 // mock window.location
-;(global as any).window = Object.create(window)
+;(global as any).window ??= Object.create(window)
 Object.defineProperty(window, 'location', {
   value: new URL('http://myhost:1234/my/path?abc'),
 })
--- a/apps/datahub/src/app/app.module.ts
+++ b/apps/datahub/src/app/app.module.ts
@@ -184,7 +184,7 @@ export class AppModule {
     ThemeService.generateBgOpacityClasses(
       'primary',
       getThemeConfig().PRIMARY_COLOR,
-      [10, 25, 75]
+      [10, 25, 50, 75, 100]
     )
   }
 }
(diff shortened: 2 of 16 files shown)


libs/util/shared/src/lib/services/proxy.service.spec.ts
```diff
@@ -2,7 +2,7 @@ import { PROXY_PATH, ProxyService } from './proxy.service'
 import { TestBed } from '@angular/core/testing'
 
 // mock window.location
-;(global as any).window = Object.create(window)
+;(global as any).window ??= Object.create(window)
```
jahow, line 5: Is this change intentional? I've never seen that before :open_mouth:

f-necas: It is ! But it has to be discussed. I've seen it there and in this [PR](https://github.com/geonetwork/geonetwork-ui/pull/505) that `proxy.service.spec.ts` fails with the error [TypeError: Cannot redefine property: window](https://stackoverflow.com/questions/75414699/jest-typeerror-cannot-redefine-property-window). 

I can't understand why because nothing have been edited in proxy file... If you have some hints.

jahow: Oh ok, that's a bug in node 18, see https://github.com/geonetwork/geonetwork-ui/pull/482
which version of node do you use for running the tests?

f-necas: 18.16.00 ... I'll edit those two branches and create an issue in order to find a way to upgrade node version. It seems unfortunate that we can"t upgrade it due to one test. What do you think ?

jahow: Apparently 18.15 does not have the bug so we could use that? but it's also probably a matter of weeks before the backport happens on 18.x and then we can just switch to LTS safely :)

jahow: Ok let's leave it like that, it actually works around the bug in a very simple way. If we add a link to the node bug report then it's all good :)

jahow (approved):
:+1:

PR #515, opened 2023-07-03
Make filter options customisable

Angi-Kinas:
with all filters:
image

customised version:
image


libs/util/app-config/src/lib/app-config.ts
```diff
@@ -195,6 +195,54 @@ export function loadAppConfig() {
           : ({
               FILTER_GEOMETRY_DATA: parsedSearchSection.filter_geometry_data,
               FILTER_GEOMETRY_URL: parsedSearchSection.filter_geometry_url,
+              CUSTOM_FILTERS: (parsedSearchSection.custom_filters as [])?.map(
+                (filter) => {
+                  switch (filter) {
+                    case 'organisation':
+                      return {
+                        fieldName: 'publisher',
```
jahow, line 203: we shouldn't do any mapping with field names here, only the fields service has this power :)

conf/default.toml
```diff
@@ -63,6 +63,8 @@ background_color = "#fdfbff"
 # Note: if the GeoJSON object contains multiple features, only the geometry of the first one will be kept!
 # filter_geometry_url = "https://my.domain.org/assets/boundary.geojson"
 # filter_geometry_data = '{ "coordinates": [...], "type": "Polygon" }'
+# available filters: 'organisation', 'format', 'publicationYear', 'standard', 'inspireKeyword', 'topic', 'spatial', 'license'
+custom_filters = ['organisation', 'format', 'publicationYear', 'standard', 'inspireKeyword', 'topic', 'spatial', 'license']
```
jahow, line 67:
```suggestion
# The advanced search filters available to the user can be customized with this setting.
# The following fields can be used for filtering: 'publisher', 'format', 'publicationYear', 'standard', 'inspireKeyword', 'topic', 'isSpatial', 'license'
# advanced_filters = ['publisher', 'format', 'publicationYear', 'topic', 'spatial', 'license']
```

Note: since this is an optional key, it should be commented out in the toml file and the default value should be written in the app (datahub in this case).

libs/util/app-config/src/lib/app-config.ts
```diff
@@ -195,6 +195,54 @@ export function loadAppConfig() {
...
+                        title: 'search.filters.byOrganisation',
```
jahow, line 204: use `marker` for i18n

jahow (approved):
A few things to address and that's it, very good work!

fgravin (commented):
Just a comment without doing a review.
A proposal of the new configuration inputs should be provided in the description of the PR.
The configuration is quite sensitive as it has to be human manageable.
I suggest that everyone time we touch the configuration schema, we propose the change within the PR, so we can discuss it even though we are not looking at the code.
Thanks :hugs:

Angi-Kinas:
@fgravin I updated the description of the PR. Let me know if there is something to change/add.

fgravin:
> @fgravin I updated the description of the PR. Let me know if there is something to change/add.

Great thanks.
I wonder if a user add a filter which is not in the list, what would happen ?
BTW, I think that a user should be able to add any field of the domain metadataRecord model as a filter, it should perform a basic term aggregation. I don't know how this is handled right now, could be addressed when the need appear later on though.
:+1:

Angi-Kinas:
@fgravin good question. When the user adds a filter that's not in the list, there will be an Error in the console and the app is not usable anymore:
image
@jahow and I decided that this is ok for now. What do you think?

jahow:
> BTW, I think that a user should be able to add any field of the domain metadataRecord model as a filter, it should perform a basic term aggregation. I don't know how this is handled right now, could be addressed when the need appear later on though.

That might be doable yes, if the need arise we could do it! E.g. filter on uuid, title, links etc.

fgravin:
> @jahow and I decided that this is ok for now. What do you think?

I think that it should not happen, shouldn't it ?
An error in the config should not break the application IMO, I don't know what is the behavior for the other fields.
Might not be hard to check for the error and add a warning in the console at least ?

jahow:
> > @jahow and I decided that this is ok for now. What do you think?
> 
> I think that it should not happen, shouldn't it ? An error in the config should not break the application IMO, I don't know what is the behavior for the other fields. Might not be hard to check for the error and add a warning in the console at least ?

Hmm, I would say an error in the config should break everything because the configuration should be valid. At load time we check if every required configuration setting is there, but we cannot know yet if one of the filters is invalid or not. If an invalid field is specified then an error is thrown in the console, the only issue here is that the application still loads and kind of look normal, so an administrator might not realize that they made a mistake in the config.

The best solution IMO would be to have a strong error message showing up like: "WARNING: the configuration is invalid, the application cannot run" or something like that.

PR #516, opened 2023-07-05
feat: update select dropdown to match multiselect one

f-necas:
An overlay has been set to be as closed as the multiselect dropdown. 

Sometimes the selected value is a string or an object. I made the majority of the changes in the dropdown component in order to keep the original behavior of each parent component.

### DIFF
--- a/libs/feature/dataviz/src/lib/chart-view/chart-view.component.html
+++ b/libs/feature/dataviz/src/lib/chart-view/chart-view.component.html
@@ -5,6 +5,7 @@
     <gn-ui-dropdown-selector
       class="basis-1/4"
       [choices]="typeChoices"
+      [extraBtnClass]="'secondary min-w-full'"
       (selectValue)="chartType$.next($event)"
       [selected]="chartType$.value"
       [title]="'chart.dropdown.type' | translate"
@@ -12,6 +13,7 @@
     <gn-ui-dropdown-selector
       class="basis-1/4"
       [choices]="xChoices$ | async"
+      [extraBtnClass]="'secondary min-w-full'"
       (selectValue)="xProperty$.next($event)"
       [selected]="xProperty$.value"
       [title]="'chart.dropdown.xProperty' | translate"
--- a/libs/feature/dataviz/src/lib/chart-view/chart-view.component.stories.ts
+++ b/libs/feature/dataviz/src/lib/chart-view/chart-view.component.stories.ts
@@ -13,22 +13,23 @@ import { ChartViewComponent } from './chart-view.component'
 import { ChartComponent, UiDatavizModule } from '@geonetwork-ui/ui/dataviz'
 import { LoadingMaskComponent } from '@geonetwork-ui/ui/widgets'
 import { importProvidersFrom } from '@angular/core'
-import { DropdownSelectorComponent } from '@geonetwork-ui/ui/inputs'
+import {
+  DropdownSelectorComponent,
+  UiInputsModule,
+} from '@geonetwork-ui/ui/inputs'
 import { MatProgressSpinner } from '@angular/material/progress-spinner'
+import { OverlayModule } from '@angular/cdk/overlay'
 
 export default {
   title: 'Smart/Dataviz/ChartView',
   component: ChartViewComponent,
   decorators: [
     moduleMetadata({
-      declarations: [
-        DropdownSelectorComponent,
-        LoadingMaskComponent,
-        MatProgressSpinner,
-      ],
       imports: [
         ChartComponent,
+        OverlayModule,
         TranslateModule.forRoot(TRANSLATE_DEFAULT_CONFIG),
+        UiInputsModule,
       ],
     }),
     applicationConfig({
--- a/libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
+++ b/libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
@@ -1,27 +1,71 @@
-<div class="flex flex-col items-start sm:flex-row sm:items-center relative">
-  <label
+<div
+  class="flex items-start flex-col sm:flex-row sm:items-center relative w-full"
+>
+  <span
     *ngIf="showTitle"
     class="tracking-wide text-sm mb-2 sm:mb-0 sm:mr-2 whitespace-nowrap"
     [attr.for]="id"
   >
     {{ title }}
-  </label>
-  <select
-    [id]="id"
-    (change)="this.selectValue.emit($event.target.value)"
-    class="w-full min-w-[66px] truncate text-main bg-white border border-gray-300 py-[10px] px-2 rounded-[0.25em] cursor-pointer leading-tight focus:outline-none hover:text-primary-darker hover:border-primary-lighter focus:ring-primary-lightest focus:ring-4"
-    [ngClass]="extraClass"
-    [class.w-full]="!showTitle"
+  </span>
+  <gn-ui-button
+    type="outline"
+    class="grow min-w-0"
+    style="min-width: {{ minWidth }}"
+    extraClass="w-full !p-[8px] !pl-[16px] {{ extraBtnClass }}"
+    [title]="title"
+    [attr.aria-owns]="id"
+    (buttonClick)="openOverlay()"
+    cdkOverlayOrigin
+    #overlayOrigin="cdkOverlayOrigin"
   >
-    <option
-      *ngFor="let choice of choices"
-      [value]="choice.value"
-      [selected]="isSelected(choice)"
-    >
-      {{ choice.label | translate }}
-    </option>
-  </select>
-  <div
-    class="pointer-events-none absolute inset-y-0 right-0 flex items-center px-2 text-gray-800"
-  ></div>
+    <div class="grow flex items-center mr-2 gap-2 overflow-hidden">
+      <div class="text-left font-medium truncate py-1">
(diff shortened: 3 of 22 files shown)


libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.ts
```diff
@@ -18,20 +25,76 @@ export type DDChoices<T> = Array<{
...
+      overlayY: 'bottom',
+      offsetY: -8,
+    },
+  ]
+  get selectedChoice(): Choice {
+    return (
+      this.choices.find((choice) => choice.value === this.selected) ??
+      this.choices[0]
+    )
+  }
```
tkohr, line 63: I think we need a fallback here if `choices` are undefined.

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.ts
```diff
@@ -18,20 +25,76 @@ export type DDChoices<T> = Array<{
...
   get id() {
     return this.title.toLowerCase().replace(/[^a-z]+/g, '-')
   }
 
+  getChoiceLabel(): string {
+    return this.selectedChoice.label
+  }
+
+  ngOnInit(): void {
+    if (!this.maxRows) this.maxRows = 6
```
tkohr, line 74: Maybe set a const with value 6 at top of component, to get the hardcoded value outside the code.

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
```diff
@@ -1,27 +1,68 @@
-<div class="flex flex-col items-start sm:flex-row sm:items-center relative">
-  <label
+<div class="flex items-start sm:flex-row sm:items-center relative w-full">
+  <span
     *ngIf="showTitle"
     class="tracking-wide text-sm mb-2 sm:mb-0 sm:mr-2 whitespace-nowrap"
     [attr.for]="id"
   >
```
tkohr, line 6: I think mobile needs some adaptions on `items-*` and `mr` classes here.

tkohr (commented):
Thanks @f-necas   for the big work! The visual result is really nice and identical to the multi-dropdown.

However, the app currently breaks if `choice` is undefined in the `dropdown-selector`. Can easily be tested with records without data for the previews. Also noticed two minor issues in the UI:
- The values in the orgs sort now wrap. Do you think this could be prevented?
- The sort label is not properly aligned on mobile.

I've left corresponding comments inline.

libs/feature/dataviz/src/lib/feature-dataviz.module.ts
```diff
@@ -25,6 +26,7 @@ import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
     UiWidgetsModule,
     TranslateModule,
     ChartComponent,
+    OverlayModule,
```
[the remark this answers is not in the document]
f-necas: Because they are used in the chart view (like for selecting bar/horizontal bar etc...)

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
```diff
@@ -1,27 +1,68 @@
...
     {{ title }}
-  </label>
-  <select
```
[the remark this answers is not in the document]
f-necas: I've tried but we couldn't put custom mat-icon arrows, neither edit the visual content of the dropdown. The result was always different from the multiselect-dropdown..

f-necas:
Thanks for your review @tkohr. I've updated the code in order to :

1. Display only a part of choice label if it's too long (I've seen in my dataset a long text which overflow the entire page) 
2. Set a min-width in order to have a correct display for sort by. (I could take the longest label but it may be related to  the first point (very long select) and setting a whitespace-nowrap in choices labels overflows the overlay)
3. Set the number of max rows as constant
4. Put a conditional display if choices isn't set.  ( I think we won't need the select if there's no value ?)

libs/feature/dataviz/src/lib/feature-dataviz.module.ts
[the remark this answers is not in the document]
tkohr: Ok, so we need to import the `OverlayModule` everywhere, where we use the `dropdown-selector`?

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
```diff
@@ -8,7 +11,7 @@
   </span>
   <gn-ui-button
     type="outline"
-    class="grow"
+    class="grow min-w-[165px]"
```
tkohr, line 14: Just wondering if a `whitespace-nowrap` somewhere could also do the job to prevent setting a hard value that depends on the content to look good.

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.ts
```diff
@@ -67,11 +69,14 @@ export class DropdownSelectorComponent implements OnInit {
   }
 
   getChoiceLabel(): string {
-    return this.selectedChoice.label
+    return (
+      this.selectedChoice.label.substring(0, 50) +
+      (this.selectedChoice.label.length > 50 ? '...' : '')
+    )
```
tkohr, line 75: Can't we achieve this with a `truncate` via CSS?

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
```diff
@@ -1,4 +1,7 @@
-<div class="flex items-start sm:flex-row sm:items-center relative w-full">
+<div
+  class="flex items-start flex-col sm:flex-row sm:items-center relative w-full"
```
tkohr, line 2: Thanks looks nice on mobile now :-)

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.ts
fgravin: I haven't followed the discussion but TW provide a `truncate` class for ellipsis. It's not about flex.
https://tailwindcss.com/docs/text-overflow#truncate

f-necas: > I haven't followed the discussion but TW provide a `truncate` class for ellipsis. It's not about flex. https://tailwindcss.com/docs/text-overflow#truncate

Yeah I know and use it already, but the flex class on a parent breaks his behaviour.

tkohr: Hum, indeed it doesn't seem simple to make the cascaded `truncate` work within the `flex` here. A `<div class="truncate">` wrapper directly under the flex helps to apply the `truncate` further down, but also cuts borders etc. because of the `overflow-hidden`. 
We really need to keep the `flex justify-end` layout in the `map-view.component`, though, to keep a fitting design for short/normal dropdown entries.

f-necas: @jahow @fgravin What do you think of it ?

libs/feature/dataviz/src/lib/chart-view/chart-view.component.stories.ts
```diff
@@ -21,6 +22,7 @@ export default {
       imports: [
         ChartComponent,
         HttpClientModule,
+        OverlayModule,
```
jahow, line 25: I think we should import the UiInputs module here instead, as OverlayModule is an underlying dependency

libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html
```diff
@@ -1,27 +1,71 @@
...
-  <div
-    class="pointer-events-none absolute inset-y-0 right-0 flex items-center px-2 text-gray-800"
-  ></div>
-</div>
+      <input
+        class="w-[0px] h-[18px] align-text-top shrink-0"
+        type="text"
+        [checked]="isSelected(choice)"
+        (click)="onSelectValue(choice)"
+      />
```
jahow, line 65: this has no reason to be there; instead, of `isSelected(choice)` is true, the following span should be highlighted just like in a native select:
image

libs/feature/dataviz/src/lib/chart-view/chart-view.component.html
```diff
@@ -1,23 +1,25 @@
 <div class="w-full h-full flex flex-col">
   <div
-    class="flex flex-col space-y-2 sm:flex-row sm:space-y-0 sm:space-x-2 justify-between text-[13px]"
+    class="flex flex-col flex-wrap space-y-2 sm:flex-row sm:space-y-0 sm:space-x-2 justify-between text-[13px] gap-4"
```
jahow, line 3: In principle, when modifying a presentation component, we shouldn't have to modify anything that uses this component. Otherwise this indicates that the encapsulation is broken and that our changes are leaking "outside" of the modified component scope.

In that case I think it is acceptable to adjust options because the dropdown is a little bit larger, but modifying the layout (introducing `flex-wrap` etc.) is a too big change in my opinion.

jahow (approved):
I think we're in a pretty good state now. @f-necas you can merge whenever you feel that the PR is ready! Thanks for the continued effort.

PR #523, opened 2023-07-12
feat: Make email in contact block "mailto"

Angi-Kinas:
In order to make the email in the contact block clickable, this PR adds the mailto: functionality. As discussed in our weekly, we don't want to add a copy button next to the email, because this is not consistent with what we see on other websites.

### DIFF
--- a/libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
+++ b/libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
@@ -17,7 +17,7 @@
     <a
       [href]="shownContact.website"
       target="_blank"
-      class="text-primary text-sm cursor-pointer hover:underline transition-all"
+      class="contact-website text-primary text-sm cursor-pointer hover:underline transition-all"
       >{{ shownContact.website }}
       <mat-icon class="!w-[12px] !h-[12px] !text-[12px] opacity-75"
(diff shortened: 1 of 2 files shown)


tkohr (approved):
Thanks @Angi-Kinas ! Works as expected. 

Don't hesitate to add a short description to your PRs. Not sure what has been decided for the ticket. The implementation doesn't exactly correspond to the ticket's description and conversation. Maybe adding a `<gn-ui-copy-text-button>` with the email value on the right to the mailto link would satisfy all demands!?

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -10,12 +10,14 @@
       {{ shownContact.organisation }}
     </div>
   </div>
-  <p class="text-gray-700 text-sm">{{ shownContact.email }}</p>
+  <a [href]="'mailto:' + shownContact.email" class="text-gray-700 text-sm">{{
```
fgravin, line 13: Should open the mail in another tab.

f-necas: Added target _blank attribute.

fgravin (commented):
CI is failing

PR #533, opened 2023-07-13
datahub: Favorite star bug on page change

cmoinier:
Thanks Tobias and Florent for the help!

## JIRA ISSUE

https://jira.camptocamp.com/browse/GSHDF-679

## DESCRIPTION

- [x] Permettre d'ajouter/supprimer un record des favoris
- [x] Permettre de garder ce favori en changeant d'onglet
- [x] Permettre de garder ce favori en allant sur la page du record

### DIFF
--- a/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
+++ b/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
@@ -115,9 +115,7 @@ describe('FavoriteStarComponent', () => {
       describe('if record is not part of favorite', () => {
         beforeEach(() => {
           ;(favoritesService as any).myFavoritesUuid$.next([
-            'aaa',
-            'bbb',
-            'ccc',
+            component.record.uuid,
           ])
           starToggle.newValue.emit(true)
           fixture.detectChanges()
@@ -138,7 +136,6 @@ describe('FavoriteStarComponent', () => {
         beforeEach(() => {
           ;(favoritesService as any).myFavoritesUuid$.next([
             'aaa',
-            'bbb',
             component.record.uuid,
           ])
           starToggle.newValue.emit(false)
@@ -150,6 +147,15 @@ describe('FavoriteStarComponent', () => {
           ])
           expect(favoritesService.addToFavorites).not.toHaveBeenCalled()
         })
+        beforeEach(() => {
+          component.record = {
+            ...RECORDS_SUMMARY_FIXTURE[0],
+            favoriteCount: 42,
+          }
+          ;(favoritesService as any).myFavoritesUuid$.next(['aaa'])
+          starToggle.newValue.emit(false)
+          fixture.detectChanges()
+        })
         it('decrease record favorite count by one', () => {
           expect(favoriteCountEl.textContent).toEqual(
             (component.record.favoriteCount - 1).toString()
@@ -158,11 +164,7 @@ describe('FavoriteStarComponent', () => {
       })
       describe('two subsequent changes', () => {
         beforeEach(() => {
-          ;(favoritesService as any).myFavoritesUuid$.next([
-            'aaa',
-            'bbb',
-            component.record.uuid,
-          ])
+          ;(favoritesService as any).myFavoritesUuid$.next(['aaa'])
           starToggle.newValue.emit(false)
           starToggle.newValue.emit(true)
           fixture.detectChanges()
--- a/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
+++ b/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
@@ -131,6 +131,11 @@ describe('FavoriteStarComponent', () => {
             (component.record.favoriteCount + 1).toString()
           )
         })
+        it('adds the record to myFavoritesUuid$', async () => {
+          await expect(favoritesService.myFavoritesUuid$['_value'][0]).toBe(
+            component.record.uuid
+          )
+        })
       })
       describe('if record is part of favorite', () => {
         beforeEach(() => {
@@ -159,6 +164,12 @@ describe('FavoriteStarComponent', () => {
         it('decrease record favorite count by one', () => {
           expect(favoriteCountEl.textContent).toEqual(
             (component.record.favoriteCount - 1).toString()
+          )
+        })
+        it('removes the record from myFavoritesUuid$', async () => {
+          console.log(favoritesService.myFavoritesUuid$)
+          await expect(favoritesService.myFavoritesUuid$['_value'][0]).not.toBe(
+            component.record.uuid
           )
         })
       })
--- a/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
+++ b/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
@@ -31,7 +31,7 @@ describe('FavoriteStarComponent', () => {
   let fixture: ComponentFixture<FavoriteStarComponent>
   let authService: AuthService
   let favoritesService: FavoritesService
-  let favoriteCountEl: HTMLElement
+  let favoriteCountHTMLEl: HTMLElement
   let starToggle: StarToggleComponent
 
   beforeEach(async () => {
@@ -65,7 +65,6 @@ describe('FavoriteStarComponent', () => {
     favoritesService = TestBed.inject(FavoritesService)
     fixture = TestBed.createComponent(FavoriteStarComponent)
     component = fixture.componentInstance
-    component.record = { ...RECORDS_SUMMARY_FIXTURE[0], favoriteCount: 42 }
     fixture.detectChanges()
     starToggle = fixture.debugElement.query(
       By.directive(StarToggleComponent)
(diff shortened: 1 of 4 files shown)


libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,34 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
...
       }
     })
+    this.countSubscription = this.favoritesService.myFavoritesUuid$
+      .pipe(pairwise())
+      .subscribe((fav) => {
+        const oldFavs = fav[0]
+        const newFavs = fav[1]
+        let editedDs = []
+        if (oldFavs.length < newFavs.length) {
+          editedDs.push(newFavs.slice(-1)[0])
```
fgravin, line 85: We either have to choose if the modified uuids is a `string` or an `array`.
Here `editedDs` is an array but with at most one element, which is confusing.
The datahub does only need it to be a `string`, because we can change the selection just one by one.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,34 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
...
+        editedDs.forEach((ds) => {
+          if (this.hasFavoriteCount && editedDs.includes(this.record.uuid)) {
+            if (newFavs.includes(ds)) {
+              this.favoriteCount += 1
+            } else {
+              this.favoriteCount += -1
+            }
+          }
+        })
+        editedDs = []
```
fgravin, line 99: useless

fgravin (changes requested):
Great work thanks.
I have made some comments to make the logic clearer about what we are doing.

Also, I don't think that you properly test the new behavior.
You should add a couple of tests where the `modifiedFavorites$` ticks 
- contains the component record
- does not contain the component record

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,34 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
         })
       }
     })
+    this.countSubscription = this.favoritesService.myFavoritesUuid$
```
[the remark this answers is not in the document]
cmoinier: Not 100% sure why, but changing it to `modifiedFavorites$` makes it come back to the old behavior and doesn't even update the count on click (on the same page). Also, changing it to `modifiedFavorites$` only means that in the case the count is already X when the user adds it to its favorites, the display count will display '1' since the value of `modifiedFavorites$` is never synchronized with the api response in the first place and starts as an empty array. It also means that when the user adds its first favorite, `modifiedFavorites$` will only contain this record and all other favorites will display a count of '0'. I hope it's kind of clear?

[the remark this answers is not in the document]
cmoinier: > Not 100% sure why, but changing it to `modifiedFavorites$` makes it come back to the old behavior and doesn't even update the count on click (on the same page). Also, changing it to `modifiedFavorites$` only means that in the case the count is already X when the user adds it to its favorites, the display count will display '1' since the value of `modifiedFavorites$` is never synchronized with the api response in the first place and starts as an empty array. It also means that when the user adds its first favorite, `modifiedFavorites$` will only contain this record and all other favorites will display a count of '0'. Let me know if it's not clear :)

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,34 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
...
+              this.favoriteCount += -1
+            }
+          }
+        })
+        editedDs = []
+      })
   }
 
   ngOnDestroy(): void {
     this.subscription.unsubscribe()
```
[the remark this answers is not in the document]
cmoinier: Yes, I did try that. For some reason the subscription isn't working (as in, nothing happens, no update on directly clicked record nor on the other page) unless it's in its own subscription. I agree this is not the best way.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,34 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
         })
       }
     })
+    this.countSubscription = this.favoritesService.myFavoritesUuid$
+      .pipe(pairwise())
+      .subscribe((fav) => {
+        const oldFavs = fav[0]
+        const newFavs = fav[1]
+        let editedDs = []
```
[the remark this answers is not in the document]
cmoinier: It stands for "edited datasets" but I agree this is weird, I will change it to `editedFavs` to match with the other two arrays.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
cmoinier: Okay! Will change it to a string then. It's just that I thought I remembered you or Tobias saying that in some cases, multiple favorites can be added at the same time and so I figured that an array would be best.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
cmoinier: fixed

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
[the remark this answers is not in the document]
fgravin: It should work, what did you try ?

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -150,19 +152,30 @@ describe('FavoriteStarComponent', () => {
...
+          starToggle.newValue.emit(false)
+          fixture.detectChanges()
+        })
         it('decrease record favorite count by one', () => {
           expect(favoriteCountEl.textContent).toEqual(
             (component.record.favoriteCount - 1).toString()
           )
         })
+        it('removes the record from myFavoritesUuid$', async () => {
+          console.log(favoritesService.myFavoritesUuid$)
```
fgravin, line 170: remove
you should be able to run the test in your IDE instead of relying on logs.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -74,10 +75,29 @@ export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
...
+    this.countSubscription = this.favoritesService.myFavoritesUuid$
+      .pipe(pairwise())
+      .subscribe(([oldFavs, newFavs]) => {
+        let editedFavs = ''
+        if (oldFavs.length < newFavs.length) {
+          editedFavs = newFavs.slice(-1)[0]
+        } else {
+          const deletedFav = oldFavs.filter((fav) => !newFavs.includes(fav))[0]
+          editedFavs = deletedFav
+        }
```
fgravin, line 87:
```suggestion
        const editedFavs = (
          oldFavs.length < newFavs.length
            ? newFavs.slice(-1)
            : oldFavs.filter((fav) => !newFavs.includes(fav))
        )[0]
```

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -150,19 +152,30 @@ describe('FavoriteStarComponent', () => {
           ])
           expect(favoritesService.addToFavorites).not.toHaveBeenCalled()
         })
+        beforeEach(() => {
```
fgravin, line 155: You should not add a `beforeEach` like that, always put it in a new `describe` section.
Note the different describe sections, they play a test scenario. As we changed the logic of the component, the scenario should be changed.
There are 2 potential actions
1. component toggle
2. `myFavorites$` tick, which can comes from another component.

Those 2 actions are not binded to the same toggle as before.
Here you are not testing the fact the `myFavorites$` could tick from another component, which wouldn't change the count of this component.

fgravin (changes requested):
It works well like that thanks.
The test should be modified though, to play the new scenario, and test all the possibilities.

Thanks !

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -115,54 +115,90 @@ describe('FavoriteStarComponent', () => {
...
-          expect(favoritesService.addToFavorites).toHaveBeenCalledWith([
-            component.record.uuid,
-          ])
-          expect(favoritesService.removeFromFavorites).not.toHaveBeenCalled()
+        describe('if change is made from this component', () => {
+          beforeEach(() => {
+            starToggle.newValue.emit(true)
+            fixture.detectChanges()
+          })
+          it('adds record to favorites', () => {
```
fgravin, line 127: This test is a consequence of the click

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -115,54 +115,90 @@ describe('FavoriteStarComponent', () => {
...
+            ])
+            expect(favoritesService.removeFromFavorites).not.toHaveBeenCalled()
+          })
+          it('increase record favorite count by one', () => {
+            expect(favoriteCountEl.textContent).toEqual(
+              (component.record.favoriteCount + 1).toString()
+            )
+          })
+          it('adds the record to myFavoritesUuid$', async () => {
+            await expect(favoritesService.myFavoritesUuid$['_value'][0]).toBe(
```
fgravin, line 139: I don't understand what does this test actually test.
BTW, Observables are a flow and should not be tested as an instantaneous value (which is hacky cause it's not the public API).

fgravin (changes requested):

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
cmoinier: Yes, for when the record is added to favorites from the component.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
cmoinier: It tests if the record is present in the `myFavoriteUuid` after the user clicks on the star. It's important especially when it ticks from another component.

cmoinier: Okay, so how do I test if `myFavoriteUuid` properly recorded that the record is in the user's favorite? I didn't have any other idea

fgravin: `component.record.uuid = 'a'`

- `myFavoriteUuid$.next(['b'])` => nothing happens
- `myFavoriteUuid$.next(['a'])` => the component is updated

fgravin: > It tests if the record is present in the `myFavoriteUuid` after the user clicks on the star. It's important especially when it ticks from another component.

yes but you did the `myFavoritesUuid$.next([component.record.uuid])` in the `beforeEach`, which is exactly what you test in there.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.spec.ts
```diff
@@ -115,54 +115,90 @@ describe('FavoriteStarComponent', () => {
...
+            starToggle.newValue.emit(true)
+            fixture.detectChanges()
+          })
+          it('adds record to favorites', () => {
+            expect(favoritesService.addToFavorites).toHaveBeenCalledWith([
+              component.record.uuid,
+            ])
+            expect(favoritesService.removeFromFavorites).not.toHaveBeenCalled()
+          })
+          it('increase record favorite count by one', () => {
```
fgravin, line 133: Damn I forgot to commit one comment

While this test is a consequence of `myFavoritesUuid$` tick.
Both things are decorrelated and should be tested separatly.

When I click on the star => here you test that the functions have been called
  - When the record is already in favorite
  - When the record is not in favorites
When favorites has been updated => here you test that the amount is correct
  - My record have been updated
  - My record haven't been updated

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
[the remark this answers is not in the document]
cmoinier: Tried the normal way : `this.subscription.add()` on `this.favoritesService.myFavoritesUuid$.pipe() ......`

fgravin:
@cmoinier all the tests were there, thanks !
I  have reworked the whole scenario to make it clearer, what we test and how.
There was a lot of useless statements in the `beforeEach()` as well

fgravin (approved):

cmoinier:
Thanks for the rework @fgravin :) 

Looks good to me, merging now.

PR #534, opened 2023-07-17
GML support when geojson not available from server

f-necas:
GML support is added to prevent empty data display in table and chart view. 

WFS has to support EPSG:4326 to be readable.


libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -83,12 +84,26 @@ export class DataService {
...
+                  namespace: featureType.name,
+                }
+              : null,
         }
       })
     )
   }
 
-  private getGeoJsonDownloadUrlFromWfs(
+  /* private getGeoJsonDownloadUrlFromWfs(
```
fgravin, line 106: remove it if not used

libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -83,12 +84,26 @@ export class DataService {
                 outputCrs: 'EPSG:4326',
               })
             : null,
+          gml:
+            featureType.outputFormats.find((f) =>
+              f.toLowerCase().includes('gml')
+            ) && featureType.otherCrs?.includes('EPSG:4326')
```
fgravin, line 90: Why do you use `4326` ? When are data reprojected ?

libs/util/data-fetcher/src/lib/data-fetcher.ts
```diff
@@ -6,10 +6,12 @@ import { ExcelReader } from './readers/excel'
 import { DataItem, DatasetHeaders, FetchError, SupportedType } from './model'
 import { inferDatasetType } from './utils'
 import { BaseReader } from './readers/base'
+import { GmlReader } from './readers/gml'
 
 export async function openDataset(
   url: string,
-  typeHint?: SupportedType
+  typeHint?: SupportedType,
+  options?: any
```
fgravin, line 14: you should definitely type that.

libs/util/data-fetcher/src/lib/readers/gml.ts
```diff
@@ -0,0 +1,50 @@
...
+} {
+  const splittedNamespace = namespace.split(':')
+  const regex = new RegExp(`xmlns:${splittedNamespace[0]}=["']([^'"]*)["']`)
+  const match = regex.exec(text)
+
+  if (match && match.length >= 2) {
+    const wf = new WFS({
+      featureNS: match[1],
+      featureType: splittedNamespace[1],
+      version: '2.0.0',
```
fgravin, line 21: Actually we don't know what is the version here.
You guess that the default version of the map server is 2.0.0, don't you ?
Is the 2.0.0 the default version of ogc-client, or does ogc-client rely on the version of the getCapabilities ? BTW, do you force a version while call the getCapabilities ?

fgravin (approved):
Excellent work, even though the version is forced, it already very good to have it working in most cases.

Did you test it on the geo2france metadata ? I can't test it because of CORS, seems my CORS plugin is not working ..

Does it work as well for the WFS on the map in the metadata page ?

libs/feature/dataviz/src/lib/service/data.service.ts
f-necas: In order to check if 4326 is available and so, be readable by the ol/format/WFS class. The condition would be better : 
```
... (featureType.defaultCrs === 'EPSG:4326' || featureType.otherCrs?.includes('EPSG:4326') )
```

We can even have a look if EPSG:3857 is available too. 

As it is to cover some olds map servers, maybe we can rely only on 4326 ? 
What do you think of it ?

f-necas:
Thanks @fgravin for your review. I've pushed some changes to cover three of your comments. 

You can bypass the CORS by using the `proxy_path` on `default.toml` with this url : `proxy_path = "http://localhost:4200/dev-proxy?"`

libs/feature/dataviz/src/lib/service/data.service.ts
fgravin: Yes your check might be better, it's easy to add a test that confirms that. I think that 4326 should be supported by any map server though..

Anyway, my question was more "_If you get `4326`, and the geojson request might returns `4326` as well, where in the code, in the end, the geojson is projected back to `3857` ?_"

The answer is here https://github.com/geonetwork/geonetwork-ui/blob/main/libs/feature/map/src/lib/map-context/map-context.service.ts#L111
Which transforms the geojson from `4326` to `3857` to be displayed on the map, which means we must request the GML in `4326`. Note that for the table and the dataviz, the projection is useless as the geometry is not used.

So it's all good.

fgravin (commented):
I d'ont know if it's my computer, but testing the datahub on your branch is very slow. There is a big latency while doing a search for instance.
Is it fast on your side ?

f-necas:
> I d'ont know if it's my computer, but testing the datahub on your branch is very slow. There is a big latency while doing a search for instance. Is it fast on your side ?

Pretty much instantaneous, request results takes approx 50ms and renders directly

libs/feature/dataviz/src/lib/service/data.service.ts
f-necas: Alright, I'm updating this and fix CI.

f-necas: @fgravin  if it's ok for you, I can merge it :)

PR #536, opened 2023-07-19
 fix: hiding favorite star if no id and added margin between panels

f-necas:
PR is here to improve UX when no record is found.

Margin has been added to record only not results list (but it wasn't intended to).

The favorite star is hidden in the header if ther's no id (as id is mandatory)

### DIFF
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -3,7 +3,7 @@
     class="h-full container-lg mx-auto flex flex-col justify-center relative"
   >
     <gn-ui-favorite-star
-      *ngIf="metadata && metadata.id"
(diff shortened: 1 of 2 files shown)


apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -3,7 +3,7 @@
     class="h-full container-lg mx-auto flex flex-col justify-center relative"
   >
     <gn-ui-favorite-star
-      *ngIf="metadata"
+      *ngIf="metadata && metadata.id"
```
fgravin, line 6:
```suggestion
      *ngIf="metadata?.id"
```
should work

fgravin (approved):
Thanks for the quick improvement ! :wink:

PR #538, opened 2023-07-20
fix: org thumbnail loading time and contacts overrides

f-necas:
Organisations observables now use shareReplay to avoid multiple call from toRecord method.

Contact is not overrided anymore to prevent wrong display in metadata view. 

Only first resourceContact logoUrl is overrided.

### DIFF
--- a/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
+++ b/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
@@ -317,22 +317,18 @@ describe('OrganisationsFromMetadataService', () => {
         title: 'Surval - Données par paramètre',
         uuid: 'cf5048f6-5bbf-4e44-ba74-e6f429af51ea',
         contact: {
-          name: 'Ifremer',
-          email: 'ifremer.ifremer@ifremer.admin.ch',
-          logoUrl: 'http://localhost/geonetwork/images/harvesting/ifremer.png',
+          name: "Cellule d'administration Quadrige",
+          organisation: 'Ifremer',
+          email: 'q2suppor@ifremer.fr',
+          website: 'https://www.ifremer.fr/',
+          logoUrl:
+            'http://localhost/geonetwork/images/logos/81e8a591-7815-4d2f-a7da-5673192e74c9.png',
         },
         resourceContacts: [
           {
-            email: 'ifremer.ifremer@ifremer.admin.ch',
-            logoUrl:
-              'http://localhost/geonetwork/images/harvesting/ifremer.png',
-            name: 'Ifremer',
-            organisation: 'Ifremer',
-          },
-          {
             email: 'q2_support@ifremer.fr',
             logoUrl:
-              'http://localhost/geonetwork/images/logos/81e8a591-7815-4d2f-a7da-5673192e74c9.png',
+              'http://localhost/geonetwork/images/harvesting/ifremer.png',
             name: "Cellule d'Administration Quadrige",
             organisation: 'Ifremer',
           },
--- a/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
+++ b/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
@@ -326,9 +326,16 @@ describe('OrganisationsFromMetadataService', () => {
         },
         resourceContacts: [
           {
+            email: 'ifremer.ifremer@ifremer.admin.ch',
+            logoUrl:
+              'http://localhost/geonetwork/images/logos/81e8a591-7815-4d2f-a7da-5673192e74c9.png',
+            name: 'Ifremer',
+            organisation: 'Ifremer',
+          },
+          {
             email: 'q2_support@ifremer.fr',
             logoUrl:
-              'http://localhost/geonetwork/images/harvesting/ifremer.png',
+              'http://localhost/geonetwork/images/logos/81e8a591-7815-4d2f-a7da-5673192e74c9.png',
             name: "Cellule d'Administration Quadrige",
             organisation: 'Ifremer',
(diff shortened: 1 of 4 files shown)


libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -220,16 +205,9 @@ export class OrganisationsFromMetadataService
...
-          const contactFromOrg = this.mapContactFromOrganisation(
-            org,
-            metadataRecord.contact
-          )
-          metadataRecord.contact = contactFromOrg
-          metadataRecord.resourceContacts = [
-            contactFromOrg, // FIXME: this should go into an organization field
-            ...metadataRecord.resourceContacts,
-          ]
+        if (org && !record.resourceContacts?.[0].logoUrl && org.logoUrl) {
```
fgravin, line 208: You should take the email if not present.

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -220,16 +205,15 @@ export class OrganisationsFromMetadataService
...
-          ]
+        if (
+          org &&
+          !metadataRecord.resourceContacts?.[0].logoUrl &&
+          org.logoUrl
+        ) {
+          const logoUrl = getAsUrl(`${org.logoUrl}`)
+          metadataRecord.resourceContacts[0].logoUrl = logoUrl // FIXME: this should go into an organization field
+          metadataRecord.resourceContacts[0].email ??= org.email
+          metadataRecord.contact.email ??= org.email
```
fgravin, line 216: The `if` statement is not correct for the email set.

I would suggest to keep your `mapContactFromOrganisation` and change the name to `hydrateContactWithOrg`, and do all the stuff in there.

I would also not touch the `record.contact`, you are guessing `record.contact` and `record.resourceContacts[0]` are the same but it might not be the case.

Note that the `group` strategy seems quite different as it adds new contacts within the metadata, which is a bit weird IMO, anyway, the group has the priority in that case so it's ok.

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -163,6 +164,29 @@ export class OrganisationsFromMetadataService
...
+      (organisation.logoUrl ? getAsUrl(`${organisation.logoUrl}`) : null)
+    const mappedOrg = {
+      name: organisation.name,
+      organisation: organisation.name,
+      email: organisation.email || firstResourceContact.email,
+      logoUrl: logoUrl,
+      website: metadataRecord.resourceContacts[0].website,
+    } as MetadataContact
+
+    metadataRecord.resourceContacts = [
```
fgravin, line 183: I wouldn't add a new contact like for the `group` strategy, that was my point in the previous message.
BTW if you do that the name of the function is not meaningful.

The `group` strategy add a new contact from the group of the record (which could be different from the recource contact), while the `metadata` strategy just hydrates the first existing resource contact.

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
```diff
@@ -317,23 +317,19 @@ describe('OrganisationsFromMetadataService', () => {
...
           {
             email: 'ifremer.ifremer@ifremer.admin.ch',
-            logoUrl:
-              'http://localhost/geonetwork/images/harvesting/ifremer.png',
-            name: 'Ifremer',
-            organisation: 'Ifremer',
-          },
-          {
-            email: 'q2_support@ifremer.fr',
             logoUrl:
```
fgravin, line 330: The test is wrong.
You should write what you expect first, and code the correct behavior, instead of coding and adapting the test so it can pass :wink: 

Here what we said
1. logo of the contact if it exists
2. logo of the organisation
3. logo of the record

In your test, it set the logo of the record while it should use the logo of the organisation instead (because `resourceContact[0].logoUrl = ''`)

Probably, during the mapping, the resource contact logoUrl is set to the record logo if null in the contact, before the mapping with the organisation, which breaks the priority.

libs/util/shared/src/lib/utils/atomic-operations.ts
```diff
@@ -54,12 +54,25 @@ export const mapLogo = (source: SourceWithUnknownProps) => {
   return logo ? getAsUrl(`/geonetwork${logo}`) : null
 }
 
+export const hydrateWithRecordLogo = (
```
fgravin, line 57: I would just call it `hydrateContactsWithRecordLogo`

Do the tests cover this function ? Probably

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -163,19 +165,25 @@ export class OrganisationsFromMetadataService
...
-    return {
+  private hydrateResourceContactWithOrg(
+    metadataRecord: MetadataRecord,
+    organisation: Organisation
+  ): MetadataRecord {
+    const firstResourceContact = metadataRecord.resourceContacts[0]
+    const logoUrl =
+      firstResourceContact.logoUrl ||
+      (organisation.logoUrl ? getAsUrl(`${organisation.logoUrl}`) : null)
+    const mappedOrg = {
```
fgravin, line 176: I think that we said that the organisation props would only hydrate missing contact props, didn't we ? :thinking: 
Would mean that the priority is the contact email, the contact logo, the contact name etc...

libs/util/shared/src/lib/utils/atomic-operations.ts
f-necas: Renamed. 

Yes this function is covered in both org-from-metadata and org from-groups classes.

PR #539, opened 2023-07-24
Feat: improving multiselect dropdown 

f-necas:
Search input in multiselect dropdown is improved by : 

- Removing outlin when input is focused
- Focusing directly this input
- Handle arrow keyboard inputs between input and checkboxes


jahow:
I think this looks very good, thanks!

Using a focus trap might be too radical for the search input since then the tab key no longer allows us to navigate through the options. I guess we could introduce our own `gnUiAutofocus` directive to do that? See this article for example https://davidmcintosh.medium.com/auto-focusing-an-angular-input-the-easy-way-part-2-directives-ac9dc9a7ec73

(there might be other libraries for that as well)

f-necas:
Thanks @jahow for your review. I've updated the code in order to keep keyboard accessibility working. 

Outline is also set to the full search div.

PR #540, opened 2023-07-25
feat: improve image loading and displaying in gn-ui thumbnail

f-necas:
- Using url instead of B64 image
- Set image loading as lazy
- Use JS to set image as cover or contain object-fit


jahow (approved):
Well done, thanks! Feel free to merge when there is a unit test for the cover/scale-down logic :)

PR #543, opened 2023-07-25
fix: predefined search param set value to have closed icon available

f-necas:
Fix behavior of close icon in search input when clicking favorites/last/predefined buttons.

### DIFF
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
@@ -72,6 +72,11 @@ export class AutocompleteComponent
         this.updateInputValue(value.currentValue)
       }
     }
+    if (this.inputRef) {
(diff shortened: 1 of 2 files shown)


libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -72,6 +72,11 @@ export class AutocompleteComponent
         this.updateInputValue(value.currentValue)
       }
     }
+    if (this.inputRef) {
+      this.inputRef.nativeElement.value = value?.currentValue?.title
+        ? value.currentValue.title
+        : ''
+    }
```
fgravin, line 79: to move this section in `updateInputValue` method
```suggestion
    if (this.inputRef) {
      this.inputRef.nativeElement.value = (value as any)?.title || ''
    }
```

fgravin (approved):

PR #544, opened 2023-07-25
fix: autocomplete behavior

f-necas:
Update suggestions to match predefined buttons click


fgravin (approved):
Good thanks !
Maybe adding a test for that would be good :wink:

PR #560, opened 2023-07-28
feat: get logo if available from index

f-necas:
In organisations tab, retrieve the resourceContact logo. This results in a better filled organisations tab (less placholders)

### DIFF
--- a/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
+++ b/libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
@@ -60,6 +60,11 @@ const organisationsAggregationMock = {
                 },
               ],
             },
+            logoUrl: {
+              doc_count_error_upper_bound: 0,
+              sum_other_doc_count: 0,
+              buckets: [],
+            },
           },
           {
             key: 'BAKOM',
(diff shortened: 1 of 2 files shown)


libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.ts
```diff
@@ -58,7 +63,9 @@ export class OrganisationsFromMetadataService
             .filter((mail) => !!mail),
           recordCount: bucket.doc_count,
           description: null,
-          logoUrl: null,
+          logoUrl: bucket.logoUrl.buckets
+            .map((bucket) => bucket.key)
+            .filter((logo) => !!logo)[0],
```
fgravin, line 68: How the bucket key could be null ? To avoid this test, you can add `exclude` in the request.
Are they sorted by occurence ?

This should be enough ?
```suggestion
          logoUrl: bucket.logoUrl.buckets[0]?.key
```

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
```diff
@@ -192,21 +219,21 @@ describe('OrganisationsFromMetadataService', () => {
           {
             description: null,
             emails: ['rolf.giezendanner@are.admin.ch', 'john.doe@are.admin.ch'],
-            logoUrl: null,
+            logoUrl: undefined,
```
fgravin, line 222: I don't know if there was a purpose to be `null` instead of `undefined` ?

fgravin (approved):
Excellent, few remarks then you can merge !

libs/feature/catalog/src/lib/organisations/service/organisations-from-metadata.service.spec.ts
f-necas: I don't think. We just check nullity/emptiness for hydrating.

PR #564, opened 2023-08-02
feat: Add language switcher

Angi-Kinas:
- Makes it possible for an admin user to configure which languages can be selected in the UI in the `default.toml` file.
Default languages are: English and French (to be discussed).

- Adds technical documentation for the internationalization (thanks to @jahow )

There was no mockup for the UI, so it can be improved. It looks currently like this:
image


conf/default.toml
```diff
@@ -152,7 +155,6 @@ background_color = "#fdfbff"
 #
 # [translations.en]
 # results.sortBy.dateStamp = "Last time someone changed something"
-# [translations.fr]
```
Angi-Kinas, line 155: TODO: Undo

conf/default.toml
```diff
@@ -26,6 +26,9 @@ proxy_path = ""
 # The default location in the dockerized datahub app for example is "/datahub/wc-embedder.html".
 # If the URL is not indicated, no permalinks will show up in the UI.
 # web_component_embedder_url = "/datahub/wc-embedder.html"
+# TODO: Description
```
Angi-Kinas, line 29: TODO: Add description, uncomment the languages

conf/default.toml
```diff
@@ -26,6 +26,10 @@ proxy_path = ""
 # The default location in the dockerized datahub app for example is "/datahub/wc-embedder.html".
 # If the URL is not indicated, no permalinks will show up in the UI.
 # web_component_embedder_url = "/datahub/wc-embedder.html"
+# This optional parameter defines the languages that will be provided in a dropdown for the user to translate the UI.
```
jahow, line 29: Having some padding above this block would make the config file more readable I think

libs/ui/inputs/src/lib/ui-inputs.module.ts
```diff
@@ -16,7 +16,6 @@ import { TextInputComponent } from './text-input/text-input.component'
 import { MatAutocompleteModule } from '@angular/material/autocomplete'
 import { MatIconModule } from '@angular/material/icon'
 import { NavigationButtonComponent } from './navigation-button/navigation-button.component'
-import { UiWidgetsModule } from '@geonetwork-ui/ui/widgets'
```
jahow, line 19: I'm sorry but I was wrong here: the `ui-inputs` module indeed relies on the `ui-widgets` module because the autocomplete component relies on an error popup; so we cannot remove this import here

libs/ui/widgets/src/lib/language-switcher/language-switcher.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  Inject,
+  InjectionToken,
+  Input,
+  OnInit,
+  Optional,
+} from '@angular/core'
+import { LANGUAGE_STORAGE_KEY } from '@geonetwork-ui/util/i18n'
+import { TranslateService } from '@ngx-translate/core'
+
+export const LANGUAGE_PLACEHOLDER = new InjectionToken<string[]>(
```
jahow, line 12: I think a better name for this token would be:
```ts
export const LANGUAGES_LIST = new InjectionToken<string[]>(
  'languages-list'
)
```

libs/ui/widgets/src/lib/language-switcher/language-switcher.component.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const DEFAULT_LANGUAGES = ['en', 'fr']
+
+@Component({
+  selector: 'gn-ui-language-switcher',
+  templateUrl: './language-switcher.component.html',
+  styleUrls: ['./language-switcher.component.css'],
+})
+export class LanguageSwitcherComponent implements OnInit {
+  @Input() languageList = this.languagePlaceholder || DEFAULT_LANGUAGES
```
jahow, line 24: This does not need to be an `@Input`

libs/ui/widgets/src/lib/language-switcher/language-switcher.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    return this.translate.currentLang
+  }
+
+  ngOnInit(): void {
+    this.languageList.forEach((language, index) => {
+      this.languageList[index] = {
+        label: `language.${language}`,
+        value: language,
+      }
+    })
```
jahow, line 43: Here you are effectively mutating the `languageList` array, which can in fact be the actual language list that was injected. So you might introduce side effects in completely unrelated modules that would depend on this injection token.

You could probably avoid mutating the injected value without too much changes like so:
```ts
const languages = this.languagePlaceholder || DEFAULT_LANGUAGES;
this.languageList = languages.map((language, index) => ({
  label: `language.${language}`,
  value: language,
}))
```

apps/datahub/src/app/home/home-header/home-header.component.html
```diff
@@ -74,4 +74,8 @@
       [style.opacity]="-0.6 + expandRatio * 5"
     ></datahub-navigation-menu>
   </div>
+  <gn-ui-language-switcher
+    *ngIf="showLanguageSwitcher"
+    class="language-switcher absolute top-0 left-0"
```
jahow, line 79: Making the dropdown a little bit smaller and with a bit of spacing looks a bit better IMO
```
class="language-switcher absolute top-2.5 left-2.5 text-[13px]"
```

jahow (changes requested):
As I said in the comments, the dependency from ui-inputs towards ui-widgets has to be kept. This means that, I think, the language switcher should go in another module. This could be `ui-catalog` for example, and that should solve the circular dependency issue.

Other than that, very good work! Everything works really well. I made several comments in the code, I hope it'll be clear enough.

jahow (approved):
Thanks for addressing everything! I think this contribution looks really good now, cheers!

PR #566, opened 2023-08-04
Web components: Embed translations

tkohr:
PR adds a `EmbeddedTranslateLoader` that allows to embed translations in the built js and uses this loader for web components. This prevents the need to ship translation files along with the web component script. For now, it increases the `gn-wc.js` for about 14 kb per language. At a total size of 4 mb this change seems acceptable to me. Improvements might be made here when we optimize the web component's total bundling size (maybe by including only used translations or lazy loading).

TODO:
- [x] add some doc about web components


jahow (approved):
Thanks @Angi-Kinas, all good!

PR #568, opened 2023-08-07
Nx 16

f-necas:
Migrating Nx and Angular to v16 and storyboook to v7.

Known issues : 
- Angular Components and Webcomponents doesn't work in Storybook. Wasn't working before migration.
- Zone.js blocked to 0.13.0. See https://github.com/angular/angular/issues/50756
- Angular Material stuck at 16.0.4. Search Bar doesn't render as expected in upper versions


jahow (approved):
Impressive work, thanks a lot! Tested locally and everything appears to work fine. The datahub bundle has actually shrunk a little bit! :tada:

PR #569, opened 2023-08-07
GSGCT-54 : Keyword and Usages sections show even when empty

cmoinier:
### JIRA Ticket

https://jira.camptocamp.com/browse/GSGCT-54

### Issue

The "keywords" and "usage conditions" sections on the dataset page show even when empty

### Resolution

- [x] If no keyword is present in the metadata, hide the keywords section altogether
- [x] If no usage conditions were specified in the metadata, keep the "usage conditions" expansible block but simply show "No usage conditions specified for this record." in the block

### To test

- [x] Add doc

### To test

Go to a metadata that doesn't have keywords or usage conditions.
ex : `5d04741e-441e-4cff-84f8-deb819c8a263` or `7bad4fdb-af8f-427a-8f62-5f321d47445f-6571`

### DIFF
--- a/docs/apps/datahub.md
+++ b/docs/apps/datahub.md
@@ -4,20 +4,24 @@ outline: deep
 
 # Datahub
 
-## Chapter 1
-
-## Chapter 2
-
 ## Metadata pages
 
 Each metadata has its own detailed page, made of multiple sections, described as followed.
 
+The following sections are dynamically generated from the [metadata info component](https://github.com/geonetwork/geonetwork-ui/tree/main/libs/ui/elements/src/lib/metadata-info) component.
+
+### Abstract
+
+The abstract section is based on the metadata attribute `abstract`.
+
 ### Keywords
 
-The keyword section is dynamically generated from the `metadata-info` component. The metadata comes from the harvested `record.metadata` object, which contains a `keywords` array or strings. Through a test on the array's length, if the array is empty, the section will not be displayed.
+The keywords section is based on the metadata attribute `keywords`.
+
+### Lineage
+
+The lineage section is based on the metadata attribute `lineage`.
 
 ### Usage and constraints
 
-The usage and constraints section is dynamically generated from the `metadata-info` component. The metadata comes from the harvested `record.metadata` object, which contains a `constraints` array or strings. Both the usage and constraints are stored in this array. Through a test on the array's length (`hasUsage`), if the array is empty, the section will not be displayed.
-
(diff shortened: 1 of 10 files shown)


tkohr (commented):
Nice fix, thanks @cmoinier !

Instead of the merge I'd propose to rebase on main to get rid of this commit and have a cleaner history.

Maybe you could also set up prettier in your IDE to prevent the formatting commits.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
```diff
@@ -27,4 +27,60 @@ describe('MetadataInfoComponent', () => {
...
+        constraints: ['no usage'],
+      }
+      fixture.detectChanges()
+    })
+    it('should not display a message for no usage or constraints', () => {
+      // Use waitForAsync to handle asynchronous changes in the DOM.
+      fixture.whenStable().then(() => {
+        const displayedElement =
+          fixture.nativeElement.getElementsByClassName('noUsage')
+        expect(displayedElement.length).toBe(0)
```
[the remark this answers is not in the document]
cmoinier: Yep, totally

docs/apps/datahub.md
```diff
@@ -7,3 +7,17 @@ outline: deep
...
+
+### Keywords
+
+The keyword section is dynamically generated from the `metadata-info` component. The metadata comes from the harvested `record.metadata` object, which contains a `keywords` array or strings. Through a test on the array's length, if the array is empty, the section will not be displayed.
+
+### Usage and constraints
+
+The usage and constraints section is dynamically generated from the `metadata-info` component. The metadata comes from the harvested `record.metadata` object, which contains a `constraints` array or strings. Both the usage and constraints are stored in this array. Through a test on the array's length (`hasUsage`), if the array is empty, the section will not be displayed.
+
+A part of the section can however, in some cases, be displayed despite an empty array. In this scenario, a message will be displayed `record.metadata.noUsage` to warn the user.
```
tkohr, line 23: Thanks for adding some doc. I think we don't need to go so much into implementation details here and name variables, for example. This will be hard to keep up to date and if developers want to know such detail, they will take a look at the code IMO.

Maybe we could just list the different sections for now or possibly give some information on what metadata attributes they are based on.

If we want to point to some code, the https://github.com/geonetwork/geonetwork-ui/tree/main/libs/ui/elements/src/lib/metadata-info component might be worth linking.

What do you think?

cmoinier: Yes, I agree. I wasn't really sure of what to write, but your idea seems more concise. Will edit.

PR #570, opened 2023-08-09
Fix: Thumbnail multiple images fallback when 404

Angi-Kinas:
For the Thumbnail image we want to display (with priority):
1. the thumbnail image
2. (if no thumbnail image) the org logo
3. (if no org logo) the placeholder

The bug exists because there is a thumbnail image but it returns a 404. In this case we should show the org logo. If the org logo also return a 404 it should show the placeholder. This is fixed in this PR.


f-necas:
Thanks @Angi-Kinas for this nice feature ! 

It's just a proposition : It may be cleaner if html files doesn't handles arrays. Instead of `[images]="[{ url: record.thumbnailUrl }]"` we could handle them with some getters ?

Coverage has barely decreased. Covering the existing getter should work for re-increasing it.

I've let some comments. 
Nice work 👍🏼

PR #574, opened 2023-08-11
Web components: Introduce versioning

tkohr:
The goal of this PR is to introduce versioning to gn-ui web components. It should publish the `gn-wc.js` script to a `wc-dist-main` branch upon merges on `main` and to `wc-dist-[tagname]` branches when creating version tags complying to `'v*.*.*'`.

Potential to-dos:

- [x] There might be an additional step necessary to create the `publish_branch`, if it does not exist yet.
- [x] The [data-view-web-component.component](https://github.com/geonetwork/geonetwork-ui/blob/web-components-versioning/libs/feature/record/src/lib/data-view-web-component/data-view-web-component.component.ts#L20) does not use the correct version of the web component script yet. Not sure how to solve this best. => solved via new approach
- [x] Adapt/fix tests

### DIFF
--- a/.github/workflows/webcomponents.yml
+++ b/.github/workflows/webcomponents.yml
@@ -7,7 +7,7 @@ on:
     branches:
       - main
     tags:
-      - '*.*.*'
+      - 'v*.*.*'
   release:
     types: [published]
 
--- a/.github/workflows/webcomponents.yml
+++ b/.github/workflows/webcomponents.yml
@@ -6,6 +6,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - '*.*.*'
   release:
     types: [published]
 
@@ -54,10 +56,10 @@ jobs:
           tag: ${{ github.ref }}
           overwrite: true
 
-      - name: Publish web component to ${{ env.PUBLISH_BRANCH }} branch
+      - name: Publish web component to ${{ env.PUBLISH_BRANCH }}-${{ github.ref_name }} branch
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           force_orphan: true
           publish_dir: ./wc-dist
-          publish_branch: ${{ env.PUBLISH_BRANCH }}
+          publish_branch: ${{ env.PUBLISH_BRANCH }}-${{ github.ref_name }}
--- a/apps/datahub/src/environments/environment.ts
+++ b/apps/datahub/src/environments/environment.ts
@@ -1,9 +1,14 @@
+import packageJson from '../../../../package.json'
 // This file can be replaced during build by using the `fileReplacements` array.
 // `ng build` replaces `environment.ts` with `environment.prod.ts`.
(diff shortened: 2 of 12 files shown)


jahow (commented):
Thanks! I feel like this currently brings a lot of complexity in the docker build process just for the WC embedder to use the correct `gn-wc.js` version.

Instead maybe we could somehow make the current gn-ui version available in the angular app. Then the version could be used in the `data-view` component both for the share url and the web component snippet.

Also the wc-embedder.html script could take the version of the `gn-wc.js` as an URL argument? but that could hurt the loading time significantly, not sure about that.

tkohr:
Thanks for the feedback @jahow ! I've reworked the approach how to use the fitting version of the web component within the datahub components as well as the `wc-embedder` following your suggestion. This seems to work fine, also when I test the `wc-embedder` within the docker container.

Still have to check if the gh workflow works without adding a step to create the necessary dist branches.

fgravin (commented):
Looking forward to have this, very good and mandatory improvement indeed !

.github/workflows/webcomponents.yml
```diff
@@ -6,6 +6,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - '*.*.*'
```
jahow, line 10:
```suggestion
      - 'v*.*.*'
```

That might make things a bit safer?

apps/datahub/src/environments/environment.ts
```diff
@@ -1,9 +1,14 @@
+import packageJson from '../../../../package.json'
```
jahow, line 1: I think it would be a better practice to import only the version: `import { version } from '...'`

Otherwise the complete package.json content might end up in the bundle which might be a security issue.

jahow (approved):
Very well done! :clap:  I think the solution looks much better now. Maybe grabbing the version from the package.json should be done in another place as the `environments.ts` files are not really meant to hold any logic. But that can be done later on IMO.

See if you can address my two comments before merging, that would be great! Thanks again.

apps/datahub/src/environments/environment.ts
tkohr: Hum, importing only the version does not seem to work. Following the discussion [here](https://stackoverflow.com/questions/34907682/how-to-display-the-app-version-in-angular) I decided to put the logic into the `environment.ts` to prevent exposing the entire `package.json` within the code. Checking the produced `main.js` in production this indeed seems to work, while in dev the `package.json` content is visible in the debugger.

tkohr:
Thanks for your feedback @jahow ! If the last adaptions are ok to you, PR is ready from my side. I adapted the tag version in the doc, as I guess `2.0.0` will be the first version where the versioned wc script will be generated. 

Also, if I understand correctly, `peaceiris/actions-gh-pages@v3` should take care itself of creating the `publish_branch` and this does not need to be done in a separate step.

jahow:
Yes, that should work :) all good, thanks!

PR #575, opened 2023-08-11
[Datahub] Rework the metadata content page

f-necas:
PR is intented to add some UI improvements for records. It adds :

- More contact information (location, phone)
- Record thumbnail on each records with a lightbox
- Record status information on search page and record page

Peek 2023-08-16 08-45


Angi-Kinas:
I think it might be easier to see the changes if there is a screenshot. Could you add one of this section? :)

apps/datafeeder/project.json
```diff
@@ -19,7 +19,7 @@
           "apps/datafeeder/src/favicon.ico",
           "apps/datafeeder/src/assets",
           {
-            "glob": "*",
+            "glob": "*.json",
```
Angi-Kinas, line 22: What does this do? Do we need it?

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
```diff
@@ -417,11 +423,9 @@ describe('dataset pages', () => {
           .find('gn-ui-copy-text-button')
           .find('button')
           .first()
-          .click({ force: true })
+          .realClick()
```
Angi-Kinas, line 426: What's a `realClick`? :D

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -26,6 +26,27 @@
         </gn-ui-metadata-info>
       </div>
       <div>
+        <div
+          *ngIf="lightbox$ | async as lightbox"
+          data-cy="record-thumbnail"
```
Angi-Kinas, line 31: What is `data-cy`?  Is it something for cypress? If yes I cannot find it being used in the test file.

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
```diff
@@ -55,6 +56,11 @@ export class RecordMetadataComponent {
   errorTypes = ErrorType
   selectedTabIndex$ = new BehaviorSubject(0)
 
+  lightbox$ = this.facade.metadata$.pipe(
+    map((metadata) => metadata?.thumbnailUrl)
+  )
+  lightboxInstance: any
```
Angi-Kinas, line 62: Should this stay `any`?

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -27,4 +27,37 @@
       >
     </a>
   </div>
+  <div
+    class="flex mt-2"
+    *ngIf="shownContact.phone && shownContact.name"
+    data-cy="contact-phone"
```
Angi-Kinas, line 33: Same question as a above.

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -27,4 +27,37 @@
...
+      <mat-icon>place</mat-icon>
+    </div>
+    <div>
+      <p>{{ 'record.metadata.implantation' | translate }}</p>
+      <p class="text-gray-700">{{ shownContact.address }}</p>
+    </div>
+  </div>
+  <div
+    *ngIf="shownContact.logoUrl"
+    data-cy="contact-logo"
```
Angi-Kinas, line 54: Here as well.

translations/addTranslation.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
Angi-Kinas, line 1: I don't really understand what is happening here. Maybe a comment would be helpful for the future? Or add it to the technical documentationn?

Angi-Kinas (commented):
Well done :) ! It looks really nice on the screen record. And thanks for including that into the PR!
I added some comments, mostly things that I don't know about. We can also have a call about it.

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
f-necas: Instead of calling a js click function, cypress fires a native click.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
f-necas: It's indeed for cypress. I've directly added them in order to test but didn't provide test cases for logo and phone, but I created a test for address

translations/addTranslation.sh
f-necas: Pure dev script. I've added description and improved usage in the file.

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
f-necas: Fixed ! Removed it and set it as const in method.

apps/datafeeder/project.json
f-necas: Fixed ! I've moved the script. See last comment.

addTranslation.sh
```diff
@@ -0,0 +1,24 @@
...
+   sed -i '/}/i\
+  "'$1'": ""' $entry
+  else
+   sed -i '/'$2'/a\
+  "'$1'": "",' $entry
+  fi
+done
+
+
+
```
jahow, line 24: I didn't do a full review yet but we shouldn't need this kind of script: all translation keys should be discoverable statically from the code.

See: https://geonetwork.github.io/geonetwork-ui/main/docs/reference/i18n.html#how-to-include-translated-labels-in-geonetwork-ui

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
```diff
@@ -77,4 +82,11 @@ export class RecordMetadataComponent {
       .getFiltersForOrgs([{ name: contact.organisation }])
       .subscribe((filters) => this.searchService.updateFilters(filters))
   }
+
+  openLightbox(src: string) {
+    const lightboxInstance = basicLightbox.create(`<img src="${src}"/>`)
+    lightboxInstance.show()
+    //returns for test only
+    return lightboxInstance
```
Angi-Kinas, line 90: Could you maybe add a class name to the `.create` function instead of returning only to test it? WDYT?

Angi-Kinas (commented):
LGTM! :-) 
Thanks for addressing my comments! I'm leaving you another comment, but that's optional.

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
Angi-Kinas: I like the approach with the lightboxInstance on the component level better :-)
Thanks for addressing this!
I will resolve the commment.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
```diff
@@ -26,6 +26,27 @@
...
+        >
+          <gn-ui-thumbnail
+            class="relative h-full w-full"
+            [thumbnailUrl]="(facade.metadata$ | async).thumbnailUrl"
+            fit="cover"
+          ></gn-ui-thumbnail>
+          <div class="relative">
+            <gn-ui-button
+              class="absolute bottom-2 right-2"
+              [type]="'outline'"
```
fgravin, line 42: probably wrong type compare to the mockup (should be secondary bg).

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
```diff
@@ -77,4 +83,9 @@ export class RecordMetadataComponent {
       .getFiltersForOrgs([{ name: contact.organisation }])
       .subscribe((filters) => this.searchService.updateFilters(filters))
   }
+
+  openLightbox(src: string) {
+    this.lightboxInstance = basicLightbox.create(`<img src="${src}"/>`)
```
fgravin, line 88: Is it cleaned somehow if you click several times ?

fgravin (changes requested):
Thanks @f-necas 

I compared to the mockup and many things remain different:
- height/ratio of the image
- button hover background
- top margin of the 'contact' label
- the contact information font size is too big
- the contact logo too big
- the "contact" label, should it be "owner" instead ? idk

I am really not a big fan of the status badge, and I don't think that other people would want it, it's geocat specific. BTW the text is not vertically aligned. I will talked with Lucile and customers about that.

Thanks for the PR though, it's just like we should not tweak too much the metadata presentation for the sake of geocat, to me all of this just bring noise into the view... anyway

libs/feature/record/src/lib/record-metadata/record-metadata.component.ts
f-necas: Yes, the dom element is removed each time a click on background overlay is done.

libs/feature/record/src/lib/record-metadata/record-metadata.component.html
f-necas: https://geonetwork.github.io/geonetwork-ui/main/storybook/demo/?path=/story/inputs-buttoncomponent--primary&args=type:secondary

Out of our 'default' buttons. In this case, is it better to follow the mockup or to implement one of our button ?

fgravin: You are right, and I think we need more convention about button styling. We should not have so many different button types...
Anyway, I think that `gnui-btn` should be replaced by a TW component, keep it like this we'll talk about that later, thanks.

jahow: we can simply add a new button type, like `secondary-outline`. I think it was the right choice to use the `gn-ui-button` component, our strategy is to rely as much as possible on reusable low-level presentation components to avoid having too many TW classes in the HTML.

fgravin: I do prefer

`<button class="btn-secondary">` over `<gnui-button type="secondary">`,  which just manages CSS.
It might be a matter of taste, but TW component is
- no code
- no maintenance
- easily overridden in the app

f-necas: It could be great if we switch to scss, it could allow to write our buttons this way (with tailwind classes) for example: 
```scss
.btn {
  @apply px-2 py-1;
  .btn-secondary {
    @apply bg-secondary;
  }
  ...
}
```

```html 
<button class="btn btn-secondary" ...
```

f-necas: As far as I know, you can't for `@apply`. 
And for sure, you cannot do nested classes, which is quite useful.

fgravin: This is what I tested in `button.component.css`

```css
.btn-text-white {
  @apply text-white hover:text-primary-lighter;
}
.btn-text-white-a {
  @apply btn-text-white;
}
```

f-necas: Strange, we must have something else which preprocess our css ? 
https://caniuse.com/sr_css-apply

fgravin: https://tailwindcss.com/docs/reusing-styles#extracting-classes-with-apply

f-necas:
Close it, as it is outdated and will be reworked.

PR #576, opened 2023-08-14
Fix: Autocomplete Dropdown not closing correctly on click

Angi-Kinas:
- Makes sure that the click event is being dispatched to the document
- Adds documentation for it


libs/util/shared/src/lib/utils/event.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+    document.body.appendChild(parentComponent)
+    document.addEventListener('click', documentClickHandler)
+
+    event = new Event('click')
+    childComponent.dispatchEvent(event)
+
+    setTimeout(done)
+  })
+
+  it('doess not propagate event to parent', () => {
```
jahow, line 30:
```suggestion
  it('does not propagate event to parent', () => {
```

PR #582, opened 2023-08-17
feat: add http servers routing doc

f-necas:


docs/guide/deploy.md
```diff
@@ -4,6 +4,47 @@ outline: deep
 
 # Deploy
 
-## Chapter 1
+After building your app, you can deploy it in any HTTP server (e.g Nginx, Apache, Azure Static Website with Blob storage ...).
 
-## Chapter 2
+Move the content of `dist/` folder into your server and adjust your configuration file (if needed).  
+
+## Reverse Proxy
```
jahow, line 11: I think the term "reverse proxy" here is not quite right. NGINX and apache are more than reverse proxies. Calling the chapter "Web Server" would be more clear IMO.

jahow (approved):
This is great, thanks a lot! Just added a small comment on the chapter name. :slightly_smiling_face:

PR #588, opened 2023-08-23
[Datahub] Make abstract links as hyperlinks

cmoinier:
Relates to #571 

### JIRA Ticket

https://jira.camptocamp.com/browse/GSGCT-58

### Issue

In dataset view, in the `abstract` section, the text may contain hyperlinks, that weren't clickable nor displayed as such. This is different than the case in #571 since the text in `abstract` needs to use the `safe` pipe.

### Resolution

Created a function that creates the hyperlinks before sending the text to the HTML to pass the `safe` pipe. It is not allowed to use a directive if the text is in the `innerHTML` input, and the `safe` pipe has to be used in it.

### To test

Go to `dataset/ee965118-2416-4d48-b07e-bbc696f002c2`.

### DIFF
--- a/libs/ui/elements/src/lib/metadata-info/linkify.directive.ts
+++ b/libs/ui/elements/src/lib/metadata-info/linkify.directive.ts
@@ -1,16 +1,13 @@
 /* eslint-disable @angular-eslint/directive-selector */
-import { Directive, ElementRef, Renderer2, OnInit } from '@angular/core'
+import { Directive, ElementRef, Renderer2, AfterViewInit } from '@angular/core'
 
 @Directive({
   selector: '[gnUiLinkify]',
 })
-export class GnUiLinkifyDirective implements OnInit {
+export class GnUiLinkifyDirective implements AfterViewInit {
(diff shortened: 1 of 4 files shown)


fgravin:
What is the status of this PR, is it ready for a review ?
@cmoinier did you work on a similar task before ? How related is it with https://github.com/geonetwork/geonetwork-ui/pull/571/files ?
Could we factorize

cmoinier:
@fgravin  As far as I know, back when I opened this, it was working correctly (meaning that all the links in the abstract section were working). It's related to #571 because it's using the same directive I created. I opened this ticket for the abstract section since it was mentioned in the lineage and usages ticket that it would have to be done eventually.

So IMO, yes, it's ready for review.

fgravin:
> It's related to #571 because it's using the same directive I created.

But you are not using your directive.
The safe pipe has gone BTW, but the abstract is in an `innerHtml` binding, so I guess we would need a pipe ?

I think that @Angi-Kinas is taking over it anyway.
Thanks

fgravin:
Overseded by https://github.com/geonetwork/geonetwork-ui/pull/739

PR #591, opened 2023-08-25
docs: add a datafeeder docker comp + edit documentation

f-necas:


support-services/datafeeder/docker-compose.yml
jahow, line 1: I think you could make this file into an "extension" of the original file:
* call it `docker-compose.datafeeder.yml`
* put it in the same folder as the main one
* to use it, call `docker compose -f docker-compose.yml docker-compose.datafeeder.yml up`

This then means that you don't have to duplicate the `database`, `geonetwork` and `elasticsearch` services.

proxy-config-dev.js
```diff
@@ -0,0 +1,39 @@
...
+      'sec-roles':
+        'ROLE_SUPERUSER;ROLE_GN_ADMIN;ROLE_ADMINISTRATOR;ROLE_USER;ROLE_MAPSTORE_ADMIN',
+      'sec-org': 'camptocamp',
+      'sec-username': 'testadmin',
+      'sec-fetch-user': '?1',
+      'sec-proxy': 'true',
+      referer: 'https://georchestra-127-0-1-1.traefik.me/',
+      'sec-orgname': 'camptocamp',
+    },
+  },
```
jahow, line 38: I think you can just put this in the `proxy-config.js` file instead?

support-services/datafeeder/datadir/console/cities.geojson
jahow, line 1: Do we really need this file? or at least can we make it smaller?

proxy-config-dev.js
```diff
@@ -0,0 +1,39 @@
...
+    secure: true,
+    logLevel: 'debug',
+    changeOrigin: true,
+    cookiePathRewrite: {
+      '/datafeeder': '/',
+    },
+    headers: {
+      'sec-roles':
+        'ROLE_SUPERUSER;ROLE_GN_ADMIN;ROLE_ADMINISTRATOR;ROLE_USER;ROLE_MAPSTORE_ADMIN',
+      'sec-org': 'camptocamp',
```
jahow, line 31: Maybe use a more neutral org here :wink:

proxy-config-dev.js
```diff
@@ -0,0 +1,39 @@
...
+      'sec-username': 'testadmin',
+      'sec-fetch-user': '?1',
+      'sec-proxy': 'true',
+      referer: 'https://georchestra-127-0-1-1.traefik.me/',
+      'sec-orgname': 'camptocamp',
```
jahow, line 36: see above

jahow (changes requested):
Thanks, great work! I think you can organize things to make it simpler and avoid repetitions, see my comments.

I've tried locally and the datafeeder actually runs, well done for setting up a minimal datadir! Also thanks for adding documentation :pinched_fingers: 

Last thing, I had an error at the end of the submit process:
```
Error publishing the following datasets:
aerodromePolygon: java.net.UnknownHostException: geonetwork
```

I haven't looked further, if you can reproduce as well then maybe we should investigate more. But this is already very good.
```

File: docs/apps/datafeeder.md [hunk 4]
```diff
@@ -4,6 +4,31 @@ outline: deep
 
 # Datafeeder
 
-## Chapter 1
+"Datafeeder" is a geOrchestra's backend RESTful service to upload file based datasets and publish them to GeoServer and GeoNetwork in one shot.
```
fvanderbiest, line 7: "Datafeeder" is geOrchestra's backend RESTful service

support-services/datafeeder/datadir/console/cities.geojson
fvanderbiest: Indeed, a single feature will be enough for the purpose

support-services/datafeeder/datadir/datafeeder/frontend-config.json
```diff
@@ -0,0 +1,27 @@
...
+    {
+      "label": "Lambert 93",
+      "value": "EPSG:2154"
+    },
+    {
+      "label": "Web Mercator",
+      "value": "EPSG:3857"
+    }
+  ],
+  "thesaurusUrl": "https://dev.geo2france.fr/geonetwork/srv/api/registries/vocabularies/search?type=CONTAINS&thesaurus=external.theme.httpinspireeceuropaeutheme-theme&rows=200&q=${q}&uri=**&lang=${lang}"
```
fvanderbiest, line 26: do we really want to rely on dev.geo2france.fr ? rather a local service

support-services/datafeeder/datadir/default.properties
```diff
@@ -0,0 +1,27 @@
+domainName=georchestra.test.org
```
fvanderbiest, line 1: I think you should keep the original contents of defaults.properties, as it will be easier in the future to keep them in sync.

jahow (approved):
Thanks for the work! if you can make sure that cities.geojson is smaller then it's all good!!

Issue #556, opened 2023-07-27
[Datahub]: Multilingual search


fgravin:
ATM the Datahub can be configured to be used for one lang only (or `*`). This will search only in specific languages fields from the index (eg. `default, eng, fre, ita`).

The search is then not optimized if the metadata are multilingual.
We will add this support.

**Sponsored by SwissTopo geocat.ch**

jahow:
Let's keep this issue to track the work left on multilingual search :)

PR #593, opened 2023-08-25
Adding i18n to records

f-necas:
This PR is intended to get user current locale in records instead of the default one.

If a translation is available in the record for his language, it uses this one.

Issue #556

### DIFF
--- a/libs/api/metadata-converter/src/lib/gn4/atomic-operations.ts
+++ b/libs/api/metadata-converter/src/lib/gn4/atomic-operations.ts
@@ -24,7 +24,8 @@ export const selectFallback = <T, U>(field: T, fallback: U): T | U =>
 export const selectTranslatedValue = <T>(
   source: SourceWithUnknownProps,
   lang3: string
-): T | null => selectField(source, lang3) || selectField(source, 'default')
+): T | null =>
+  selectFallback(selectField(source, lang3), selectField(source, 'default'))
 
 export const selectTranslatedField = <T>(
   source: SourceWithUnknownProps,
@@ -61,13 +62,14 @@ export const mapLogo = (source: SourceWithUnknownProps) => {
 }
 
 export const mapOrganization = (
-  sourceContact: SourceWithUnknownProps
+  sourceContact: SourceWithUnknownProps,
+  lang3: string
 ): Organization => {
   const website = getAsUrl(selectField<string>(sourceContact, 'website'))
   const logoUrl = getAsUrl(selectField<string>(sourceContact, 'logo'))
   return {
     name: selectFallback(
-      selectTranslatedField<string>(sourceContact, 'organisationObject'),
+      selectTranslatedField<string>(sourceContact, 'organisationObject', lang3),
       selectField<string>(sourceContact, 'organisation')
     ),
     ...(logoUrl && { logoUrl }),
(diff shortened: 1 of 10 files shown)


libs/feature/search/src/lib/utils/mapper/elasticsearch.field.mapper.ts
```diff
@@ -31,9 +33,14 @@ type EsFieldMapperFn = (
 export class ElasticsearchFieldMapper {
   constructor(
     private metadataUrlService: MetadataUrlService,
-    private linkClassifier: LinkClassifierService
+    private linkClassifier: LinkClassifierService,
+    private translateService: TranslateService
   ) {}
 
+  private readonly lang3 = LANG_2_TO_3_MAPPER[this.translateService.currentLang]
```
fgravin, line 40: Is it a fixed variable ? With the language switcher, I expect people to dynamically change the language of the UI.

libs/util/shared/src/lib/fixtures/elasticsearch/full-response.ts
```diff
@@ -1269,6 +1269,8 @@ export const ES_FIXTURE_FULL_RESPONSE = deepFreeze({
             {
               default: 'Restriction lié à l’exercice du droit moral',
               langfre: 'Restriction lié à l’exercice du droit moral',
+              langger:
+                'Einschränkung im Zusammenhang mit der Ausübung moralischer Rechte',
```
fgravin, line 1273: :astonished:

fgravin (commented):
Very nice start !
We have to be careful about what fields are multilingual and what are not.
Also, it depends on the versions of GN... for instance, `organizations` & `contacts` have been converted to multilingual very shortly.
It might miss the `contacts` for the last version of GN `4.2.5`.

Also, the hard work, which could be led in another PR, is about the way we run the search.
The index is multilingual, there is a setting to set the language of the catalog (which is more the language of the records of the catalog though). If set to `fre` the search is led on `any.fre` instead of `any.default` (`resourceTitleObject.fre` and all other multilingual fields). You can also have `any.*` etc...

We might need to work with Christophe on this cause he studied that for geocat and see how it's handle in geonetwork vanilla as well. It's not an easy task.
:raised_hands:

libs/util/shared/src/lib/fixtures/elasticsearch/full-response.ts
f-necas: I added a test 😄

libs/feature/search/src/lib/utils/mapper/elasticsearch.field.mapper.ts
f-necas: I don't know what's the actual behavior  when we change locale.Is it reloading the page ? in this case it should work. 

If it doesn't reload, I'll get a bigger look on onLangChange to edit this variable.

libs/api/metadata-converter/src/lib/gn4/atomic-operations.ts
```diff
@@ -22,13 +22,15 @@ export const selectFallback = <T, U>(field: T, fallback: U): T | U =>
   field === null ? fallback : field
 
 export const selectTranslatedValue = <T>(
-  source: SourceWithUnknownProps
-): T | null => selectField(source, 'default')
+  source: SourceWithUnknownProps,
+  lang3: string
+): T | null => selectField(source, lang3) || selectField(source, 'default')
```
jahow, line 27:
```suggestion
): T | null => selectFallback(selectField(source, lang3), selectField(source, 'default'))
```

jahow (approved):
Great work!!

f-necas:
Thanks for your reviews ! 
I've updated the fallback thing and also updated code to get translation of contacts and orgs.

f-necas:
Thanks for your reviews ! 
I've updated the fallback thing and also updated code to get translation of contacts and orgs.

jahow (approved):
Thanks!

PR #596, opened 2023-08-28
Metadata-editor: organization service interface error

f-necas:
Fixes app which doesn't start correctly.

For the name error displayed in console. User must be logged in geonetwork `locahost:4200/geonetwork`


apps/metadata-editor/src/app/app.module.ts
```diff
@@ -42,7 +52,19 @@ import { SignInPageComponent } from './sign-in/sign-in-page.component'
     TranslateModule.forRoot(TRANSLATE_DEFAULT_CONFIG),
     MatProgressSpinnerModule,
   ],
-  providers: [],
+  providers: [
```
jahow, line 55: I think you just have to import the `FeatureCatalogModule` here, and it should take care of the rest

jahow (approved):
So simple, I love it :heart_eyes: thanks!

PR #597, opened 2023-08-28
Fix: Simplify link classifier service

Angi-Kinas:
Bug: If the format is 'csv', in some cases the format will not be recognized.

This PR fixes this bug.


jahow (approved):
well done :)

PR #603, opened 2023-08-31
Metadata-editor/ Pagination

cmoinier:
### Issue

As a user, I want to be able to look through all available search results by selecting a different page.

### Rules:

- [x] The pagination component should be visible at the bottom of the search results list
- [x] The pagination component should be composed of a list of clickable page numbers:
       - First page
        - "..." separator
        - Two previous pages
        - Current page
        - Two following pages
        - "..." separator
        - Final page
- [x] Left and right arrow should be shown as buttons before and after the page numbers, and respectively change the current page by substracting/adding 1
- [x] These arrow should appear as disabled if their respective action is not possible
- [x] The current page number should be highlighted and not clickable
- [x] Clicking on a page number immediately refresh the results list

### Screenshot

Screenshot from 2023-08-31 12-27-26


apps/metadata-editor/src/app/dashboard/dashboard-page.component.html
```diff
@@ -43,12 +43,13 @@ <h1 class="text-[56px] capitalize font-title grow" translate="">
             results.records.hits.displayedOn
           </div>
           <div class="grow">
-            <gn-ui-pagination
-              hideButton="true"
+            <!-- <gn-ui-pagination hideButton="true" [currentPage]="searchFacade.currentPage$ | async"
+              [nPages]="searchFacade.totalPages$ | async" (newCurrentPageEvent)="paginate($event)"></gn-ui-pagination> -->
```
jahow, line 47: This can be removed

libs/ui/elements/src/lib/pagination-buttons/pagination-buttons.component.css
jahow, line 1: I think all of these styles can be safely removed once you replace the `.pagination-separator` style with tailwind classes

libs/ui/elements/src/lib/pagination-buttons/pagination-buttons.component.html
```diff
@@ -0,0 +1,31 @@
...
+    <gn-ui-button
+      type="light"
+      [disabled]="currentPage === 1"
+      (buttonClick)="previousPage()"
+    >
+      <mat-icon>chevron_left</mat-icon>
+    </gn-ui-button>
+    <ng-container *ngFor="let page of visiblePages">
+      <ng-container *ngIf="page === '...'">
+        <span class="pagination-separator">{{ page }}</span>
```
jahow, line 12:
```suggestion
        <span class="mx-[5px]">{{ page }}</span>
```

libs/ui/elements/src/lib/pagination-buttons/pagination-buttons.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  ngOnChanges(): void {
+    this.calculateVisiblePages()
+  }
+
+  calculateVisiblePages(): void {
+    const maxVisiblePages = 5
+    const halfVisible = Math.floor(maxVisiblePages / 2)
+    const startPage = Math.max(this.currentPage - halfVisible, 1)
+    const endPage = Math.min(startPage + maxVisiblePages - 1, this.totalPages)
```
jahow, line 28:
```suggestion
    const endPage = Math.min(this.currentPage + halfVisible, this.totalPages)
```

Otherwise when on page 1, the pages 1 to 5 are shown :)

jahow (commented):
Very good work! I noticed that when on the first page, too many pages were shown:
image

Instead on the last page it only shows the 3 last pages:
image

Also added a few comments, and I think you can get rid of all the CSS classes. :slightly_smiling_face:

apps/metadata-editor/src/app/dashboard/dashboard-page.component.html
cmoinier: I overlooked it! thanks :)

libs/ui/elements/src/lib/pagination-buttons/pagination-buttons.component.ts
cmoinier: Thanks, that's a good idea!

libs/ui/elements/src/lib/pagination-buttons/pagination-buttons.component.html
cmoinier: thank you :)

cmoinier:
Thanks for the review :) I made the changes and added some e2e tests.

jahow (approved):
Well done! thanks for addressing everything :slightly_smiling_face: and for the e2e tests!

PR #605, opened 2023-09-01
Editor / Implement search bar and associated route

### LINKED ISSUE #385: Improve internal search state logic
Planned improvements:
* better handling of search reset when doing pagination/filtering
* isolate implementation related to GN4/ElasticSearch to allow support for other backends in the future

jahow:
This PR contains a refactoring on the search and router states:
* search state now supports paginated results and setting the page number using a dedicated action
* router state now supports the `_page` param in the url and also better handle param changes

**There is still an issue remaining**: in the [dashboard page component](https://github.com/geonetwork/geonetwork-ui/blob/da7fb686c53555e4d4019830693eda1c58bd4fca/apps/metadata-editor/src/app/dashboard/dashboard-page.component.ts#L28-L31) we initialize the search params, but that happens _after_ the url params have been parsed. This means that the pagination will be set to whatever is asked in the URL, but will be cleared right after because of the filters change.

I think this requires adding the possibility to set initial params on a search facade without calling its methods.

Also added a search results page in the editor app:
image

Closes #385


Angi-Kinas (approved):
Thank you and great work!!
I only found a minor thing.

PR #607, opened 2023-09-04
[editor] Change URL query for pagination

cmoinier:
### Issue

The current page should appear in the current route as a query param like "&page=X"

### Resolution

Most changes were adressed in #605 , e2e test is added here.

### To test

Go to `/records/all` in metadata-editor app and switch pages around.

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
@@ -1,66 +1,58 @@
 describe('dashboard', () => {
+  let pageOne
   describe('pagination', () => {
-    let originalList
-    let newList
     it('should display different results on click on arrow', () => {
-      cy.visit('/')
+      cy.visit('/records/all')
       cy.get('gn-ui-record-table')
         .find('.record-table-col')
         .first()
-        .as('pageOne')
-
-      cy.get('@pageOne')
         .invoke('text')
-        .then((list) => {
-          originalList = list.trim()
-          cy.get('gn-ui-pagination-buttons').find('gn-ui-button').last().click()
-          cy.get('gn-ui-record-table')
-            .find('.record-table-col')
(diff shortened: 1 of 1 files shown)


apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
```diff
@@ -1,38 +1,36 @@
 describe('dashboard', () => {
-  let originalList
-  let newList
+  let pageOne
   describe('pagination', () => {
-    it('should display different results on click on arrow', () => {
-      cy.visit('/')
+    it('should display different results on click on arrow and change url', () => {
+      cy.visit('/records/all')
+      cy.wait('@setPageSizeRequest')
```
jahow, line 6: This appears to be undefined, did you forget to commit something?

jahow (commented):
Thanks, the test fails because apparently there's an undefined alias in there, that looks unintentional I think? Anyways the structure is good :slightly_smiling_face: We have the issue now that the local docker composition contains less than 15 datasets and so there is only one page available, but this will change soon I think so we can leave the test like that.

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
cmoinier: Uh, that was from an attempt to change the table size to 5 instead of 15, so that the test would not fail.

cmoinier: Anyway I rebased and it looks like the pagination e2e test has some issues. Taking this opportunity to fix it.

cmoinier:
Thanks for the review. It should be working now (with enough records) !

jahow (approved):
Thanks! At some point we'll have to show the service records as well, so that will fix the issue of the page count :)

PR #608, opened 2023-09-05
Metadata-editor : Sorting catalog

cmoinier:
### Issue

Sort different columns of the catalog.

### Rules:

- [x] By clicking on a column header, records can be sorted in ascending/descending order; the title header should appear with an up/down arrow if sorting is active
- [x] The records can be sorted up/down by title, author (alphabetically) and last update time by clicking on the respective header
- [x] Style the button

### Screenshot

image


jahow:
Thanks for all the improvements @cmoinier. I've added a commit to better handle icons in buttons (so that they don't grow vertically), and also used that opportunity to migrate to symbols outlined. I've also added a story for the RecordTable component, so that we can test it in storybook.

jahow (approved):
Approving so that you can merge whenever you feel it's ready :slightly_smiling_face:

PR #609, opened 2023-09-06
[editor]: Add formats to table

Angi-Kinas:
This PR adds the formats with its corresponding color badges to the table and sorts them by priority. On hover over the number you can see the remaining formats.

image

TODO:
- Styling for colum (make badges same size)

### DIFF
--- a/libs/ui/search/src/lib/record-table/record-table.component.ts
+++ b/libs/ui/search/src/lib/record-table/record-table.component.ts
@@ -65,13 +65,4 @@ export class RecordTableComponent {
   getBadgeColor(format: FileFormat): string {
     return getBadgeColor(format)
   }
-
-  firstFormat(record: CatalogRecord): FileFormat {
-    return this.getRecordFormats(record)[0]
-  }
-
-  secondToLastFormat(record: CatalogRecord): FileFormat[] {
-    const formats = this.getRecordFormats(record)
-    return formats.slice(formats.length - 2)
-  }
 }
--- a/libs/ui/search/src/lib/record-table/record-table.component.ts
+++ b/libs/ui/search/src/lib/record-table/record-table.component.ts
@@ -1,13 +1,10 @@
 import { Component, EventEmitter, Input, Output } from '@angular/core'
-import {
-  CatalogRecord,
-  DatasetDistribution,
-} from '@geonetwork-ui/common/domain/record'
+import { CatalogRecord } from '@geonetwork-ui/common/domain/record'
 import {
   FileFormat,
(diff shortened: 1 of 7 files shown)


libs/ui/search/src/lib/record-table/record-table.component.ts
```diff
@@ -32,4 +40,38 @@ export class RecordTableComponent {
...
+  }
+
+  firstFormat(record: CatalogRecord): FileFormat {
+    return this.getRecordFormats(record)[0]
+  }
+
+  secondToLastFormat(record: CatalogRecord): FileFormat[] {
+    const formats = this.getRecordFormats(record)
+    return formats.slice(formats.length - 2)
+  }
```
jahow, line 76: There is I think an issue with these functions: using either of them will recompute the formats array each time, and this will happen many times because they are used several times in the template. I usually don't worry too much about recomputing values on the fly, but in that case it feels a little too much.

One solution would be to only keep the `getRecordFormats` function, and store its result in the template using the `ngIf` directive, e.g.:

```html
<div class="record-table-col text-16 basis-1/2">
  <div *ngIf="getRecordFormats(record) as formats">
    <span
        class="flex badge-btn text-sm text-white"
        [style.background-color]="getBadgeColor(formats[0])"
    >{{ records[0] }}</span>
  </div>
</div>
```

Then the equivalent of `secondToLastFormat` would simply be `formats.slice(1)`.

libs/ui/search/src/lib/record-table/record-table.component.ts
```diff
@@ -32,4 +40,38 @@ export class RecordTableComponent {
...
+    return getBadgeColor(format)
+  }
+
+  firstFormat(record: CatalogRecord): FileFormat {
+    return this.getRecordFormats(record)[0]
+  }
+
+  secondToLastFormat(record: CatalogRecord): FileFormat[] {
+    const formats = this.getRecordFormats(record)
+    return formats.slice(formats.length - 2)
```
jahow, line 75: I believe there is a logic error here, to get the content of an array _without_ the first element you need to do `formats.slice(1)`; the associated test seems to be passing by coincidence I think :slightly_smiling_face:

jahow (commented):
This is looking quite good and simple, thanks! The formats computation might have a little bit of an unnecessary overhead in its current form, see my comments below.

jahow (approved):
I've added a commit that fixes the format order (it was upside down), and adjusted the badge size a little bit to make them more consistent:
image

Thanks for all the work on this @Angi-Kinas!!

PR #612, opened 2023-09-08
Fix web components which rely on API version

fgravin:
Angular Services should never call a `gn4API` function on Service initialization, because for web component, the `API_URL` is not defined yet.

One should use cold Observable instead, to be sure that the API is fetched only if subscribed.

This fixes the Web components runtime.


jahow (approved):
Ok, I understand; the HTTP calls are already cold observables, but since the functions are called right away they use the API url before initialization. Thanks for the simple fix!

PR #615, opened 2023-09-11
feat(editor): Filter by current org

Angi-Kinas:
With this PR, we will be able to filter the records by the current organisation of the authenticated user.
Additionally the highlighting of the active menu item is re-implemented here.

Right now the table is empty, because the organisation name is empty:
image

### DIFF
--- a/apps/metadata-editor/src/app/dashboard/dashboard-menu/dashboard-menu.component.html
+++ b/apps/metadata-editor/src/app/dashboard/dashboard-menu/dashboard-menu.component.html
@@ -2,26 +2,28 @@
   <a
     class="menu-item"
     routerLink="/records/my-org"
-    routerLinkActive
+    routerLinkActive="btn-active"
     #rlaMyOrg="routerLinkActive"
   >
     <mat-icon class="material-icons-outlined">home</mat-icon>
     <span translate="">dashboard.records.myOrg</span>
   </a>
+
   <a
     class="menu-item"
     routerLink="/records/all"
-    routerLinkActive
+    routerLinkActive="btn-active"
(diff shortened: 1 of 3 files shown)


apps/metadata-editor/src/app/dashboard/dashboard-menu/dashboard-menu.component.html
```diff
@@ -39,7 +41,7 @@
   <a
     class="menu-item"
     routerLink="/records/my-library"
-    routerLinkActive
+    routerLinkActive="btn-active"
```
cmoinier, line 44: Very good idea, navigation looks clearer now :+1:

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.spec.ts
```diff
@@ -31,6 +52,19 @@ describe('MyOrgRecordsComponent', () => {
...
+          useClass: OrganisationsServiceMock,
+        },
+        {
+          provide: SearchFacade,
+          useClass: SearchFacadeMock,
+        },
+        {
+          provide: SearchService,
+          useClass: searchServiceMock,
+        },
```
cmoinier, line 67: I think we need a test for this new functionnality!

cmoinier (changes requested):
Thanks for this work! LGTM overall, but I think that since it's hard to demo without an organization for the users we have in hand, it needs a UT to test the new function :slightly_smiling_face:

cmoinier (approved):
Great work, thanks! :slightly_smiling_face:

PR #616, opened 2023-09-11
Metadata-editor: sorting url 

cmoinier:
### Issue

Adding a sorting parameter to the URL.

### Resolution

This was more of a fix than a dev, and took some investigation. Turns out the Search Facade was called instead of the Search Service in `record-list-component`, preventing the URL from being updated on sorting (including when the catalog is initialized, was not working beforehand).


jahow (approved):
Looking good, thanks! That system is error-prone...

cmoinier:
> Looking good, thanks! That system is error-prone...

Thanks @jahow ! However, @Angi-Kinas reported an error on loading the app, that I now reproduce. It seems that the app isn't too happy with the initialization of the catalog with `this.searchService.setSortBy(['desc', 'changeDate'])` in the constructor, making the URL flicker. I'm looking into it now. Did everything work when you tested?

PR #617, opened 2023-09-13
GN4 repository: fix conversion to ES filters

jahow:


jahow:
Closing this as it is not the correct fix and can be misleading

PR #618, opened 2023-09-14
Datahub : change updatedOn labels

cmoinier:
In this PR the label for `updatedOn` was changed to be more specific about what was updated.

eg: 'Updated On' becomes 'Last update of data information'.


Angi-Kinas (approved):
I think you can run the command:
```npm run i18n:extract```
to extract the translation keys for all the other languages as well. Otherwise looks good to me . Thank you!

PR #619, opened 2023-09-14
Datahub : Custom data toggle broken

cmoinier:
### Issue

The custom filter buttons of the home header were broken because of a missing `sort` argument expected in their properties. However, such a property is not asked in the TOML's instructions. Fixed by getting the current sorting parameter as a `sort` argument.

### To test

Add a custom button to the TOML file, for instance : 

`[[search_preset]]
name = 'filterByLicense'
filters.q = 'Full text search'
filters.license = ['unknown']
sort = 'createDate'

### DIFF
--- a/apps/datahub/src/app/home/home-header/home-header.component.ts
+++ b/apps/datahub/src/app/home/home-header/home-header.component.ts
@@ -22,7 +22,7 @@ import { map } from 'rxjs/operators'
 import { ROUTER_ROUTE_NEWS } from '../../router/constants'
 import { firstValueFrom, lastValueFrom } from 'rxjs'
 import { CatalogRecord } from '@geonetwork-ui/common/domain/record'
-import { sortByToString } from '@geonetwork-ui/util/shared'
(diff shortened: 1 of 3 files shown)


apps/datahub/src/app/home/home-header/home-header.component.ts
```diff
@@ -84,9 +84,7 @@ export class HomeHeaderComponent {
         customSearchParameters.filters
       )
     )
-    this.searchService.setSortAndFilters(
-      searchFilters,
-      sortByFromString(customSearchParameters.sort)
-    )
+    const sortBy = await firstValueFrom(this.searchFacade.sortBy$)
```
jahow, line 87: Could e we make the `sort` parameter in the TOML configuration at least optional? not being able to specify a sort order sounds quite limiting for this feature

cmoinier: Sure, good idea :slightly_smiling_face:

PR #620, opened 2023-09-18
fix(editor): Unsubscribe from observable subscription

Angi-Kinas:
In this PR (https://github.com/geonetwork/geonetwork-ui/pull/615) I added a subscription to Observables without unsubscribing. This will be fixed in this PR.


jahow (approved):
Thanks! FYI you can group subscriptions together using `subscription.add()`, see https://rxjs.dev/guide/subscription

Then you only have to call `subscription.unsubscribe()` once :)

PR #621, opened 2023-09-18
feat(datahub): Sort formats in filter dropdown by format priority

Angi-Kinas:
With this PR the  data-view dropdown sorts the formats by their priority. This is important because the first entry is the selected one (csv) and will be shown as the default in the data-view.

Screenshot: The order is now 1) csv 2) excel. When taking the screenshot the dropdown closes:
image


jahow (approved):
Looking good, thanks! Could you please update the screenshot in the PR and write an explanation of the reason for this change? thanks!

PR #622, opened 2023-09-19
Enlarge advanced filters behavior

fgravin:
This PR introduces some changes to the **advanced filters** behavior & configuration.
The goal behind it (and the following work) is to

1. simplify the configuration of the filters
2. introduce many other fields
3. make the behavior more homogeneous
4. avoid having to code to add a new filter

I would like to try to factorize a bit the different kinds of filters, not to have one class per field. For instance we have
- code list fields
- therausus fields
- nested fields
- translated object fields
and so on.

I also introduce the fact that the translation keys are managed at the top of the filter component, and fetched automatically from the domain field name (eg. `search.filters.contact`)

This introduces some questions though
- the `SimpleSearchField.getAggregation()` is a domain definition or a ES definition ?
- how to specify the sort if the results are translated afterwards ?
- how to sort by `count` ?
- I see weird behavior in the `topic` field with results in german, french etc..

This PR can initiate incremental work afterwards, to be sure that all kinds of fields are handled correctly
- the good label is displayed
- the good key is used for the search


apps/datahub/src/app/home/search/search-filters/search-filters.component.ts
```diff
@@ -14,6 +14,16 @@ import {
 } from '@geonetwork-ui/feature/search'
 import { getOptionalSearchConfig } from '@geonetwork-ui/util/app-config'
```
jahow, line 16: I feel like this could even go in the fields.service.ts file, so that there is only one file to change when adding a new field

jahow:
>     * the `SimpleSearchField.getAggregation()` is a domain definition or a ES definition ?

It returns a domain aggregation object.

>     * how to specify the sort if the results are translated afterwards ?

Maybe have another generic class that does translating on the frontend?

>     * how to sort by `count` ?

I think you can specify either `key` or `count` for the sort

>     * I see weird behavior in the `topic` field with results in german, french etc..

If it's not in the UI language then it means that the translated label was already in the index I guess?

fgravin:
@jahow do you think we can merge this first set of enhancements ?
Does they make sense to you ?

fgravin:
> I think you can specify either `key` or `count` for the sort

Yes but there is no option for that in the `SimpleSearchField`. I think that the `order` argument could be switched to a `AggregationSort` type, and then, if there is `i18n` afterwards, use the sort argument to sort back the items (either by name or by count).

PR #623, opened 2023-09-19
Reactivate filtering by favorites & geometry

### LINKED ISSUE #590: Reimplement favorites & geometry filtering following the internal data format change
Following #464 filtering on favorites and/or geometry has been disabled in the search state effects. See:
* https://github.com/geonetwork/geonetwork-ui/blob/88f022588755ecc34c85f70fc39582718b09284e/libs/feature/search/src/lib/state/effects.ts#L135
* https://github.com/geonetwork/geonetwork-ui/blob/88f022588755ecc34c85f70fc39582718b09284e/libs/feature/search/src/lib/state/effects.spec.ts#L285
* https://github.com/geonetwork/geonetwork-ui/blob/88f022588755ecc34c85f70fc39582718b09284e/libs/feature/search/src/lib/state/effects.spec.ts#L315

This should be reactivated properly.

jahow:
These features were disabled in #464 .

Filtering by favorites and geometry was reimplemented in a very similar way, with a few structural changes:
* The favorites service is now part of the GN4 repository lib
* The auth service is also part of the GN4 lib
  * This is because the favorites service depends on current user info; besides, user info is also something provided by the GN4 backend; in general I think it makes sense to avoid calling GN4 APIs outside of the `@geonetwork-ui/api/repository/gn4` lib
  * There is still a `feature/auth` lib that simply provides the services for showing avatars; we might put other things there in the future so I didn't get rid of it

Fixes #590 

Added E2E tests for those two filtering options.

Also fixes a bug where, because now we're "grouping" actions before triggering a new search, actions targeting different search ids could end up being list.

Also fixes a bug where icons in buttons were not correctly sized (missed from https://github.com/geonetwork/geonetwork-ui/pull/608/commits/4d7654a36d55d4e20f67b30a1ae885b5c31386f1)
image


apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -385,6 +385,33 @@ describe('datasets', () => {
         })
       })
     })
+
+    describe('filter by geometry', () => {
+      beforeEach(() => {
+        // this will enable spatial filtering
+        cy.intercept('GET', '/assets/configuration/default.toml', {
+          fixture: 'config-with-geometry.toml',
```
tkohr, line 393: Really handy!

apps/datahub-e2e/src/support/commands.ts
```diff
@@ -29,6 +30,48 @@ Cypress.Commands.add('login', () => {
   cy.get('[name="gnSigninForm"]').submit()
 })
 
+Cypress.Commands.add('clearFavorites', () => {
```
tkohr, line 33: Nice, also really handy! I did not know this.

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -72,20 +78,31 @@ export class SearchEffects {
...
-        SET_FAVORITES_ONLY,
-        SET_SPATIAL_FILTER_ENABLED,
-        PAGINATE,
-        SET_PAGE_SIZE
-      ),
-      debounceTime(0),
-      map((action: SearchActions) => new RequestNewResults(action.id))
+    this.actionsWithNewResults$.pipe(
+      // this will aggregate actions until the debounceTime ticks
+      buffer(this.actionsWithNewResults$.pipe(debounceTime(0))),
```
tkohr, line 97: Why do we need this `buffer/debounceTime`?

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -72,20 +78,31 @@ export class SearchEffects {
...
+      switchMap((actions: SearchActions[]) => {
+        // once we have a list of actions emitted since last time, we can split them by search id
+        const requestNewResults = actions
+          .map((action) => action.id)
+          .filter((value, index, array) => array.indexOf(value) === index)
```
tkohr, line 102: Waoh, pretty tricky we need to think of handling this.

tkohr (approved):
Thanks for this recovery @jahow ! Code LGTM and features worked fine when testing them.

Regarding our comment in the `default.toml` "...all records which _do not_ intersect with the geometry will be excluded", we should maybe change "excluded" to something like "ranked lower" as the results still show up in the search, shouldn't we?

libs/feature/search/src/lib/state/effects.ts
jahow: the idea here is that if you call for instance `searchFacade.orderBy('createDate').setPageSize(10)` etc., each call to the facade will trigger an action that will eventually trigger a `RequestNewResults` action. So you end up having many of these, many search requests and potential conflicts.

But because there are several search ids, the grouping is actually quite tricky indeed... I've tried to make this as clear as possible :grimacing:

PR #624, opened 2023-09-19
e2e(datahub): Add e2e test for data distribution priority in data view

Angi-Kinas:
This PR adds an e2e test for the priority order of the data distributions in the data view and changes one record in the data dump.


jahow (approved):
good job! ;)

PR #630, opened 2023-09-22
fix(favorites): do not crash search if favorites cannot be fetched

### LINKED ISSUE #626: Datahub: no result if userSelection API returns an error
If the userselection API is not working on the backend, then the search is broken in the datahub.

We should catch the error and disable the favorite feature if the user selection return an error, but the search should keep working.

jahow:
The search will still work but favorite records will not appear as such. The error received when fetching the favorites will be printed to the console instead.

Fixes #626


fgravin (approved):
All good !

PR #631, opened 2023-09-22
Datahub: Hide nav menu on scroll down

tkohr:
Small fix to hide nav menu completely and make its buttons unclickable when scrolling down.


jahow (approved):
Thanks, that works flawlessly!!

PR #632, opened 2023-09-22
Datahub: show the correct amount of records on the news page

### LINKED ISSUE #627: Datahub: Regression: search count blocked to 10K
Fixed in https://github.com/geonetwork/geonetwork-ui/pull/458/files#diff-0c2d669bc2b22ce8a4a84b9ce7d13f163b9c43724dbf6628400961164f6cea9fR48

Regression, introduced by some change in the state, maybe the pagination.

- Working in https://demo.georchestra.org/datahub/home/news
- Failing in https://geocat-dev.dev.bgdi.ch/datahub/

Previously, the ES request to fetch the amount of records used the bucket `records-count`, it's not the case anymore.

jahow:
A new `getMatchesCount` operation was added to the records repository, and this is used to show the total amount of records.

Before that we used to do a `search()` with a result amount of 0 to get the results count, but having a separate operation makes sense IMO since it allows finer performance tuning. A later optimization might be to not return the total results count with all search queries, so that `track_total_hits` is not used all the time for nothing.

image

Followup of #458 

Fixes #627

### DIFF
--- a/libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
+++ b/libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
@@ -38,7 +38,7 @@ class ElasticsearchServiceMock {
 class SearchApiServiceMock {
   search = jest.fn((bucket, payload) => {
     const body = JSON.parse(payload)
-    const count = body.size ?? 20
+    const count = body.size || 1234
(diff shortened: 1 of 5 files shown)


fgravin (approved):
:+1:

PR #635, opened 2023-09-25
fix(datahub): Console warn when filter fails to load the geometry

Angi-Kinas:
There is an error displayed in the console when the geometry filter fails to load its geometry. This will be changed now to be a console warning with a message.

image

### DIFF
--- a/apps/datahub/src/app/app.module.ts
+++ b/apps/datahub/src/app/app.module.ts
@@ -176,6 +176,11 @@ export const metaReducers: MetaReducer[] = !environment.production ? [] : []
           return fetch(getOptionalSearchConfig().FILTER_GEOMETRY_URL)
             .then((resp) => resp.json())
             .then(getGeometryFromGeoJSON)
+            .catch(() =>
+              console.log(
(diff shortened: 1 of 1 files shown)


apps/datahub/src/app/app.module.ts
```diff
@@ -176,6 +176,9 @@ export const metaReducers: MetaReducer[] = !environment.production ? [] : []
           return fetch(getOptionalSearchConfig().FILTER_GEOMETRY_URL)
             .then((resp) => resp.json())
             .then(getGeometryFromGeoJSON)
+            .catch(() =>
+              console.warn('The geometry filter failed to load its geometry')
```
jahow, line 180: I think we can even make this a console log saying "No spatial filter was applied since a valid geometry could not be found". This is going to be the nominal situation for anonymous users on geOrchestra, so outputting a warning might make it seem like there's a problem to be fixed.

tkohr (approved):
Thanks a lot for the catch @Angi-Kinas ! Works as expected. I just changed it to a log and reformulated as proposed by @jahow :-)

PR #636, opened 2023-09-26
Datahub: Fix thumbnail background for org logos

tkohr:
PR fixes the background color of logos, which is currently gray when loading the datahub on the organisations tab or when turning pages in the orgs tab.

logos

### DIFF
--- a/libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
+++ b/libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
@@ -1,9 +1,7 @@
 <div
   #containerElement
-  class="h-full w-full relative shrink-0 overflow-hidden flex items-center bg-gray-100"
-  [ngClass]="{
-    'bg-white': !isPlaceholder
-  }"
+  class="h-full w-full relative shrink-0 overflow-hidden flex items-center"
+  [ngClass]="isPlaceholder ? 'bg-gray-100' : 'bg-white'"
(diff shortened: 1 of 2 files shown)


jahow (approved):
Looking good, thanks a lot!

libs/ui/elements/src/lib/thumbnail/thumbnail.component.html
```diff
@@ -1,9 +1,7 @@
 <div
   #containerElement
-  class="h-full w-full relative shrink-0 overflow-hidden flex items-center bg-gray-100"
-  [ngClass]="{
-    'bg-white': !isPlaceholder
-  }"
+  class="h-full w-full relative shrink-0 overflow-hidden flex items-center"
+  [ngClass]="isPlaceholder ? 'bg-gray-100' : 'bg-white'"
```
jahow, line 4: Very good catch here, we basically had no guarantee that `bg-white` would have priority over `bg-gray-100` prior to this change

tkohr: The more crucial thing is actually setting the `isPlaceholder` to `false` when `updateImageList` is triggered via `ngOnChanges()` with a logo, I have the feeling. Not sure if the added tests really cover this well. But yes the html should be cleaner this way, too :-)

jahow: yes indeed, the `bg-` classes weren't the core problem but they could have become a random and hard to track bug later on.

Issue #629, opened 2023-09-22
Filter records by owner in the datahub


jahow:
This feature includes:
* support for `owner=1234` in the url query; this points to a user id in the GN4 backend
* when logged in, add a toggle "Show only my records" in the advanced filters
* if a filter for a person other than the current user is set, the toggle above will be replaced by a X button with the label "Showing records from another person"; this will allow the user to clear the filter, while not disclosing the name of the user specified in the filter

PR #638, opened 2023-09-29
Me/dh/records published by me

Angi-Kinas:
## Metadata-editor

PR adds `My records` page and a link to datahub which filters by owner's records. 

Link to datahub is configurable with the [default.toml](https://github.com/geonetwork/geonetwork-ui/blob/fca93603edfae8028a0bbea7671adfcc6151911a/conf/default.toml#L9) file.

## Datahub

Datahubs implement owner filter feature.

- If user is connected, a toggle is available to switch filter.
- If user is looking at another user's records, a button to clear filter is available, see issue #629 

## Testing
⚠️  Url to datahub need to be set in [`default.toml`](https://github.com/geonetwork/geonetwork-ui/blob/fca93603edfae8028a0bbea7671adfcc6151911a/conf/default.toml#L9) (e.g. http://localhost:4200)

### TODO:

- [ ] styling 
- [ ] E2E


f-necas:
Added feature for issue #629 
Peek 2023-10-03 12-10

apps/datahub/src/app/home/search/search-filters/search-filters.component.html
```diff
@@ -63,6 +63,30 @@ <h2 class="text-xl mr-4 font-title" translate>
         (toggled)="toggleSpatialFilter($event)"
       ></gn-ui-check-toggle>
     </div>
+    <div class="sm:col-span-2" *ngIf="userId">
```
cmoinier, line 66: I believe that this filter is supposed to be shown only in the advanced filters (cf #629 ), therefore is should have `isOpen` as a condition.

```suggestion
    <div class="sm:col-span-2" *ngIf="userId && isOpen">
```

cmoinier (changes requested):
Thanks @Angi-Kinas and @f-necas for this big work! It works well overall.

My main issue is that the datahub and editor do not show the same records when the filter is activated (with my docker compo, I get 8 records in ME but 7 in DH. Some are the same, some aren't), which is problematic, but I didn't find a lead about why it's happening. Let me know if you want to look into it together ! :slightly_smiling_face: 

Finally, as I see written on the PR, styling still needs to be done, I'm thinking especially for : 

- [ ] the record count displayed to the right on the editor (maybe add some text for context + make it look like a link, so that users might know they can click).
- [ ] the style of the toogle in DH has a weird behavior when it's focused on (in FF al least). It looks fine when it's not on focus. [insert screencast later]

apps/datahub/src/app/home/search/search-filters/search-filters.component.ts
```diff
@@ -24,14 +27,17 @@ export class SearchFiltersComponent implements OnInit {
...
   constructor(
     public searchFacade: SearchFacade,
     private searchService: SearchService,
-    private fieldsService: FieldsService
+    private fieldsService: FieldsService,
+    private authService: AuthService
   ) {}
 
   ngOnInit(): void {
+    this.authService.user$.subscribe((user) => (this.userId = user?.id))
```
jahow, line 40: Storing the value of an observable in a local state is kind of an anti-pattern, and it can lead to nasty race conditions. In this case I think it's fine actually, so this can stay like that, but in an ideal world where we would have as much time as we want, this should have been a `userId$` observable 🙂 (and yes this makes everything more verbose and harder to read)

apps/datahub/src/app/home/search/search-filters/search-filters.component.ts
```diff
@@ -70,6 +76,26 @@ export class SearchFiltersComponent implements OnInit {
     this.searchFacade.setSpatialFilterEnabled(enabled)
   }
 
+  toggleMyRecordsFilter(enabled: boolean) {
+    this.fieldsService
+      .buildFiltersFromFieldValues({ owner: enabled ? this.userId : [] })
+      .subscribe((filters) => this.searchService.updateFilters(filters))
+  }
+
+  myRecordsFilterEnabled() {
```
jahow, line 85: This shouldn't be a function but a simple observable on the component, like `myRecordsFiltersEnabled$`

apps/metadata-editor/src/app/records/my-records/my-records.component.ts
```diff
@@ -11,8 +14,29 @@ import { SearchFacade } from '@geonetwork-ui/feature/search'
...
+  ngOnInit() {
     this.searchFacade.resetSearch()
+    this.sub = this.authService.user$.subscribe((user) => {
+      this.ownerId = user.id
+      this.searchFacade.updateFilters({ owner: user.id })
+    })
+  }
+
+  getDatahubUrl(): string {
+    return `${this.router.getDatahubSearchRoute()}?owner=${this.ownerId}`
```
jahow, line 36: This assumes that the datahub URL doesn't contain any query param. It's generally safer to use the `URL` class and its API to modify urls

jahow (commented):
Left a few comments before doing a full review and testing 🙂

cmoinier (approved):
Thanks for all the changes and fixes @f-necas :slightly_smiling_face: LGTM!

PR #639, opened 2023-10-02
Datahub: Fix thumbnail background for org logos

f-necas:
PR fixes the background color of logos, which is currently gray when loading the datahub on the organisations tab or when turning pages in the orgs tab.


tkohr (approved):
Good idea, thanks @f-necas! Changes LGTM. Wondering if we can still manage to reflect this change in the tests, which is not simple regarding the change detection I think.

PR #640, opened 2023-10-02
 Me/dh/records published by my org

cmoinier:
### TO-DO

In the "My Organization" menu item:

- [x] The title should be the logo + name of the organization
- [x] Right of the title, the amount of published records should be shown in a numerical figure; when clicked, the user is taken to the datahub with the corresponding filter (amount of records should match!)
- [x] Right of the title, the amount of users should be shown in a numerical figure; when clicked, the user is taken to the "/users/my-org" route
- [x] On this route, a list of users in the organization is shown (using the same layout as the records list); fields shown are name, username, email)

### How to test

For testing purposes, I left commented lines at : 
- [my-org.service](https://github.com/geonetwork/geonetwork-ui/blob/9e960a4860541b5890931e8c67e6c17fe3299394/libs/feature/catalog/src/lib/my-org/my-org.service.ts#L41) (lines 41 and 59)

You might need to restore the commented line depending on the user you're using/your docker compo.
For instance, the user `Barbie` has the organization name `Barbie Inc.` which doesn't match any record or user organization and prevents testing here. 
The comments will be removed before merging.

- In [default.toml,](https://github.com/geonetwork/geonetwork-ui/blob/9e960a4860541b5890931e8c67e6c17fe3299394/conf/default.toml#L9) (line 27) the link to the datahub is to be set to your localhost where datahub is running.


apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.ts
```diff
@@ -15,30 +15,42 @@ import { AuthService } from '@geonetwork-ui/api/repository/gn4'
...
+      const org = orgs.filter((org) => org.name === this.orgName)[0]
+      if (org) {
+        this.logoUrl = org.logoUrl?.href.toString()
+        this.recordCount = org.recordCount
+      }
+    })
 
-    this.subscriptionAuthService = this.authService.user$.subscribe((user) => {
-      this.searchByOrganisation({ name: user?.organisation })
+    this.authService.allUsers$.subscribe((users) => {
```
f-necas, line 43: Those subscribes seems to be duplicated ( with some tweeks) between and `my-org-users`. I don't know if it could  be a good thing to have a `my-org.service` with some generic methods. 

WDYT ?

libs/ui/search/src/lib/record-table/record-table.component.ts
```diff
@@ -14,7 +14,7 @@ import { SortByField } from '@geonetwork-ui/common/domain/search'
   styleUrls: ['./record-table.component.css'],
 })
 export class RecordTableComponent {
-  @Input() records: CatalogRecord[] = []
+  @Input() records: any[] = []
```
f-necas, line 17: Is the `any` mandatory ?

f-necas (changes requested):
Nicely done @cmoinier for this huge work. 
I've let some comments, some debatable. Feel free to remove the comments either. 

Maybe a nice things to have is a 'no record for this organisation/ no user for this organisation' placeholder instead of blank page.

libs/common/domain/src/lib/record/metadata.model.ts
```diff
@@ -132,6 +132,21 @@ export interface DatasetTemporalExtent {
   description?: string
 }
 
+export interface Users {
```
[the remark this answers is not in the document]
cmoinier: It absolutely is. I didn't know about this one. But I ended up not using this interface so I'll delete it anyway.

libs/ui/search/src/lib/record-table/record-table.component.ts
cmoinier: I know it's not ideal, but I ended up using `any` because `records` can actually have two types. It can be a user record or a catalog record (hence why I created the `User` interface you commented on). I tried to manage to have `records` to be any of these types, but couldn't. Do you have a suggestion ? I would be using `UserApiModel` in that case.

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.ts
cmoinier: I guess that would make sense, since it's done twice the exact same way! Will try it.

apps/metadata-editor/src/app/my-org-users/my-org-users.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  ) {
+    this.searchFacade.resetSearch()
+    this.authService.user$.subscribe((user) => {
+      // this.orgName = 'Géo2France'
+      this.orgName = user?.organisation
+    })
+    this.orgService.organisations$.subscribe((orgs) => {
+      const org = orgs.filter((org) => org.name === this.orgName)[0]
+      if (org) {
+        this.logoUrl = org.logoUrl?.href.toString()
```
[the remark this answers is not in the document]
cmoinier: Seems to work just fine with that edit, thanks :slightly_smiling_face:

libs/feature/catalog/src/lib/my-org/my-org.service.ts
```diff
@@ -0,0 +1,66 @@
...
+    })
+
+    this.authService.allUsers$.subscribe((users) => {
+      const orgName = this.myOrgDataSubject.value.orgName
+      // const orgName = 'Barbie Inc.'
+      const userList = users.filter((user) => user.organisation === orgName)
+      const userCount = userList.length
+      this.myOrgDataSubject.next({ ...this.myOrgDataSubject.value, userList })
+      this.myOrgDataSubject.next({ ...this.myOrgDataSubject.value, userCount })
+    })
```
f-necas, line 64: I am skeptical for this. I may need @jahow advice there ... 

Instead of subscribing to 3 observables, it may be good to do it with rxjs with some kind of (with combineLatest or zip, or something else if it exists.) : 
```typescript
const orgData$ = combineLatest([users$, allUsers$, organisations]).pipe(
  map(([a,b,c]) => resolve resolve),
);
```

and in html using an `async` pipe to get datas.

f-necas (commented):
Thanks Camille for those nice improvements ! If @jahow can tell us what she thinks about it.

Otherwise, it seems to be good :)

libs/feature/catalog/src/lib/my-org/my-org.service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor(
+    private authService: AuthService,
+    private orgService: OrganizationsServiceInterface
+  ) {
+    this.myOrgData$ = this.myOrgDataSubject.asObservable()
+
+    this.authService.user$.subscribe((user) => {
+      const orgName = user?.organisation
+      // const orgName = 'Géo2France'
```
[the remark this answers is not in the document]
cmoinier: Yes don't worry, it's just to make it easier for everyone to test!

libs/ui/search/src/lib/record-table/record-table.component.ts
cmoinier: Tried it. Then for some reason the HTML thinks ` CatalogRecord[]` is the only type and doesn't recognize the `UserApiModel` variables... Same with `CatalogRecord[] && UserApiModel[]`, where the HTML believes it's only `UserApiModel` types :frowning_face:

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.spec.ts
```diff
@@ -1,108 +1,164 @@
...
+    authService = authServiceMock as any
+    searchFacade = searchFacadeMock as any
+    routerService = routeServiceMock as any
+
+    component = new MyOrgRecordsComponent(
+      myOrgService,
+      searchFacade,
+      orgServiceInterface,
+      routerService
+    )
```
jahow, line 96: I was wondering what was a reason here for not using the `TestBed` class?

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.ts
```diff
@@ -16,29 +17,50 @@ import { AuthService } from '@geonetwork-ui/api/repository/gn4'
...
       .getFiltersForOrgs([organisation])
       .subscribe((filters) => this.searchFacade.setFilters(filters))
   }
 
-  ngOnDestroy(): void {
-    this.subscriptionAuthService.unsubscribe()
-    this.subscriptionOrgService.unsubscribe()
+  getDatahubUrl(): string {
+    const url = new URL(
+      `${this.router.getDatahubSearchRoute()}`,
```
jahow, line 53:
```suggestion
      this.router.getDatahubSearchRoute(),
```

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.ts
```diff
@@ -16,29 +17,50 @@ import { AuthService } from '@geonetwork-ui/api/repository/gn4'
...
+      this.router.getDatahubSearchRoute().startsWith('http')
+        ? this.router.getDatahubSearchRoute()
+        : window.location.toString()
```
jahow, line 56: Here you can simply use `window.location.toString()`. If a host is already present in the datahub URL it should be used.

libs/api/repository/src/lib/gn4/auth/auth.service.ts
```diff
@@ -19,6 +20,7 @@ export const LOGIN_URL = new InjectionToken<string>('loginUrl')
 export class AuthService {
   authReady$: Observable<UserModel>
   user$: Observable<UserModel>
+  allUsers$: Observable<any[]>
```
jahow, line 23: Do you think you can make this be `Observable<UserModel[]>` instead? `any` should really be avoided, it means everything downstream will not be correctly typechecked.

libs/feature/catalog/src/lib/my-org/my-org.service.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        MyOrgService,
+        {
+          provide: TranslateService,
+          useValue: translateServiceMock,
+        },
+        {
+          provide: AvatarServiceInterface,
+          useClass: AvatarServiceInterfaceMock,
+        },
+        { provide: OrganizationsServiceInterface, useValue: orgServiceMock },
```
jahow, line 47: It would be nice if you could make `orgServiceMock` a class instead of a value, as it would be more consistent with other places where we inject mock services

jahow (commented):
Thanks for the extensive work! It feels like tackled quite a complex task with this story. I've added some comments, and I think it could be great if you could address them.

The metadata-editor components will still need some work for the future and that's fine, let's keep this an iterative process. I hope my comments are clear enough :slightly_smiling_face:

apps/metadata-editor/src/app/records/my-org-records/my-org-records.component.spec.ts
cmoinier: Oh, I find it easier and lighter without a `TestBed` actually. Since these tests don't require interactions with the DOM (as all UT should in fact, since we have integration tests) and it was a brand new test file I thought of giving it a goo :slightly_smiling_face:  But I can change it to align with the rest of the app.

libs/api/repository/src/lib/gn4/auth/auth.service.ts
cmoinier: Sure, can do :slightly_smiling_face:

libs/feature/catalog/src/lib/my-org/my-org.service.spec.ts
cmoinier: Done!

libs/feature/catalog/src/lib/my-org/my-org.service.ts
cmoinier: Got it, thanks for the extensive explanation :slightly_smiling_face: 
Following this guidance, should I change the way `my-org-records.component.html` receives its data? 
Go from : `orgData?.logoUrl` to something like `myOrgRecordsService.myOrgData$.logoUrl | async` ?

libs/feature/catalog/src/lib/my-org/my-org.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    recordCount: number
+    userCount: number
+    userList: UserApiModel[]
+  }>({
+    orgName: '',
+    logoUrl: '',
+    recordCount: 0,
+    userCount: 0,
+    userList: [],
+  })
```
jahow, line 31: this is not needed anymore I think?

jahow (approved):
Thanks for getting my comments addressed, I think it can be merged now :slightly_smiling_face:

Could you please rebase the branch on `develop` before merging and remove the merge commit in the PR history? We generally try to avoid merge commits in Pull Requests as much as possible, thanks!

libs/feature/catalog/src/lib/my-org/my-org.service.ts
cmoinier: Yes, that's true!

PR #641, opened 2023-10-04
Update default.toml to correct default.toml for filtering on metadata standard field

alexarobu:
The correct filter to use seems to be `documentStandard`, not just `standard` for filtering based on metadata standard.


fgravin (approved):
Nice catch.
Note that we are working on improving the way of customizing the filters, to be more permissive.

BTW, for geocat.ch, we added new allowed fields, we are using those one for the filters:
```
'topic', 'publisher', 'inspireKeyword', 'resourceType', 'representationType', 'format'
```

alexarobu:
Hey Florent! :smiley: 

Those sound like really useful things to include!

PR #643, opened 2023-10-04
Datahub: Fix filter clicks close to header

tkohr:
PR fixes click events close under the header (like on the filter dropdowns), by making the collapsed header around the search unclickable.

header_filter


f-necas (approved):
Works exactly as expected !  Thanks @tkohr for this nice and clean solution :)

PR #644, opened 2023-10-06
Improve autocomplete input rendering

fgravin:
- Set an input `padding-right` so the input text does not go under the absolute buttons.
image

- Add ellipsis on placeholder, works on page load but disappear on focus on Chrome
image


jahow (approved):
Perfect, thanks!

PR #645, opened 2023-10-06
Fixes and improvements to e2e tests

jahow:
E2E tests have been made to pass again; some of them were failing following the dropdown refactoring. Also some of them were made quicker.

Also did some UI improvements while working on E2E tests:
* add missing filter labels
* made the pagination buttons the correct size
* made the star toggle hitbox bigger


tkohr (approved):
Thanks, @jahow, especially for the last commit making e2e tests independent of the system language.

I still have one test failing, but this might be on my side ("home" => "my favorites button") and should not be blocking.

PR #650, opened 2023-10-14
i18n - remove tech words

fvanderbiest:
The target audience being the general public, we cannot afford tech words like "metadata", "WFS" or even "GeoNetwork" here.

Also removing the placeholder text in the search bar. 
Do we see any such placeholder text on https://www.google.fr/ ?


fgravin (commented):
Makes sense for the vocabulary, thanks.

However, for the placeholder, it should be discussed with Lucile and with the customers.
I have to admit that such a big empty input is a bit weird, maybe because I am used to see a placeholder...

image

fvanderbiest:
> However, for the placeholder, it should be discussed with Lucile and with the customers.

Restored the placeholder with d5b82a3 so the PR can be merged without waiting for a discussion.

fgravin (approved):
Thanks !

PR #651, opened 2023-10-18
Multilingual search

fgravin:
Introduces a new option for the `metadata_language` configuration property: `current`.
This is mostly useful for **multilingual catalogs**.

If `metadata_language == current`, then the Elasticsearch search is queried in the current language of the application (the UI language).

The other rules remains
- `metadata_language` is `undefined`: we use the `.*` wildcard
- `metadata_language` is set to one language: we use the that language for the search eg. `.langeng`

Here what the search looks like with the UI in english and the property set to `current`
```json
{
  "query_string": {
    "query": "carte établie",
    "default_operator": "AND",
    "fields": [
      "resourceTitleObject.langeng^5",
      "tag.langeng^4",
      "resourceAbstractObject.langeng^3",
      "lineageObject.langeng^2",
      "any.langeng",
      "uuid"
    ]
  }
} 
```

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -12,8 +12,9 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 proxy_path = ""
 # This optional parameter defines, in which language metadata should be queried in elasticsearch.
 # Use ISO 639-2/B (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) format to indicate the language of the metadata.
+# Setting to "current" will use the current language of the Datahub, the UI langage.
 # If not indicated, a wildcard is used and no language preference is applied for the search.
-# metadata_language = "fre"
+# metadata_language = "fre" or "current"
 # This optional URL should point to the login page that allows authentication to the datahub.
 # If not indicated, the default geonetwork login page is used.
 # The following three placeholders can be part of this URL:
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -12,9 +12,9 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 proxy_path = ""
 # This optional parameter defines, in which language metadata should be queried in elasticsearch.
 # Use ISO 639-2/B (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) format to indicate the language of the metadata.
-# Setting to "current" will use the current language of the Datahub, the UI langage.
+# Setting to "current" will use the current language of the User Interface.
 # If not indicated, a wildcard is used and no language preference is applied for the search.
-# metadata_language = "fre" or "current"
+# metadata_language = "current"
 # This optional URL should point to the login page that allows authentication to the datahub.
 # If not indicated, the default geonetwork login page is used.
 # The following three placeholders can be part of this URL:
--- a/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.spec.ts
+++ b/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.spec.ts
@@ -362,7 +362,7 @@ describe('ElasticsearchService', () => {
   })
 
   describe('#injectLangInQueryStringFields - Search language', () => {
-    const queryStringFields = ['resourceTitleObject.${searchLang}']
+    let queryStringFields = ['resourceTitleObject.${searchLang}']
     describe('When no lang from config', () => {
       beforeEach(() => {
         service['metadataLang'] = undefined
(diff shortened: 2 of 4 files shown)


benoitregamey:
Hi, here's my 2 cents

I'm wondering if it would be useful to implement kind of a fallback mechanism to `any.*` in case the `metadata_language` is set to `current`. 

I can imagine the usecase where a user with e.g. an english configured browser lands on the catalogue and try to search something. If the catalogue contains very little english records, he will find almost nothing even if the words used for the search are technical (often the same in all languages) or proper nouns.

Can't we imagine something like 
```json
{
  "query_string": {
    "query": "carte établie",
    "default_operator": "AND",
    "fields": [
      "resourceTitleObject.langeng^6",
      "tag.langeng^5",
      "resourceAbstractObject.langeng^4",
      "lineageObject.langeng^3",
      "any.langeng^2",
      "any.*",
      "uuid"
    ]
  }
} 
```

conf/default.toml
```diff
@@ -12,8 +12,9 @@ geonetwork4_api_url = "/geonetwork/srv/api"
 proxy_path = ""
 # This optional parameter defines, in which language metadata should be queried in elasticsearch.
 # Use ISO 639-2/B (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) format to indicate the language of the metadata.
+# Setting to "current" will use the current language of the Datahub, the UI langage.
```
jahow, line 15:
```suggestion
# Setting to "current" will use the current language of the User Interface.
```

conf/default.toml
```diff
@@ -12,8 +12,9 @@ geonetwork4_api_url = "/geonetwork/srv/api"
...
 # If not indicated, a wildcard is used and no language preference is applied for the search.
-# metadata_language = "fre"
+# metadata_language = "fre" or "current"
```
jahow, line 17:
```suggestion
# metadata_language = "current"
```

This is a more sensible default I believe

jahow (commented):
I agree with @benoitregamey here, also does the request fail if for instance the field `.langesp` is used, but there's absolutely no spanish translations in the catalog?

fgravin:
> I agree with @benoitregamey here, also does the request fail if for instance the field `.langesp` is used, but there's absolutely no spanish translations in the catalog?

Yes, we actually discussed that internally, we should have a fallback on `*` while boosting on the current language.
I think that it's very hard to have something working well though, there are so many possible combinations of things about the languages... :upside_down_face: 

I'll adapt the query

fgravin:
Ok, I've made another adaption where the priority are kept in the fallback as well.
The original priorities are increased by 10
eg:
```js
[
  'resourceTitleObject.langeng^15',
  'resourceTitleObject.*^5',
  'tag.langeng^14',
  'tag.*^4',
  'resourceAbstractObject.langeng^13',
  'resourceAbstractObject.*^3',
  'lineageObject.langeng^12',
  'lineageObject.*^2',
  'any.langeng^11',
  'any.*',
  'uuid',
]
```

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -168,13 +175,40 @@ export class ElasticsearchService {
...
+      ]
+    }, [])
+  }
+
+  private getQueryLang(): string {
+    return this.metadataLang
+      ? this.isCurrentSearchLang()
+        ? `lang${this.lang3}`
+        : `lang${this.metadataLang}`
+      : `*`
```
jahow, line 208: nested ternary operators are hard to read IMO

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.spec.ts
```diff
@@ -1,13 +1,18 @@
...
+  iso3: 'eng',
+} as LangService
 
 describe('ElasticsearchService', () => {
   let service: ElasticsearchService
   let searchFilters
 
   beforeEach(() => {
-    service = new ElasticsearchService('fre')
+    service = new ElasticsearchService(langServiceMock, 'fre')
```
jahow, line 15: Using a testBed here would make the test more readable I think, as it would be clear what are the injected dependencies (currently the test manipulates the service properties directly inc. private ones)

jahow (approved):
Thanks! I've tested it with a multilingual config and it seems to be producing good results. I made a couple of comments about readability, feel free to merge when you think this is ready!

PR #652, opened 2023-10-18
Datahub: Fix display of ESRI data on map

tkohr:
PR 

- should fix display of ESRI data on map by correctly encoding the attached query params (currently broken in https://www.geo2france.fr/datahub/dataset/atmo-hdf::emissions-de-polluants-atmosph%C3%A9riques-par-epci-dans-les-hauts-de-france-ann%C3%A9e-2010-1)
- only displays md abstract, if it is defined (see: https://www.geo2france.fr/datahub/dataset/atmo-hdf::mod%C3%A9lisation-r%C3%A9gionale-des-concentrations-en-no2-en-2019-%C2%B5g-m%C2%B3)

### DIFF
--- a/libs/feature/dataviz/src/lib/service/data.service.spec.ts
+++ b/libs/feature/dataviz/src/lib/service/data.service.spec.ts
@@ -593,11 +593,22 @@ describe('DataService', () => {
       it('builds a proxied url', () => {
         expect(
           service.getDownloadUrlFromEsriRest(
-            'http://esri.rest/local/',
+            'http://esri.rest/local',
             'geojson'
           )
         ).toBe(
-          'http://proxy.local/?url=http%3A%2F%2Fesri.rest%2Flocal%2F%2Fquery%3Ff%3Dgeojson%26where%3D1%3D1%26outFields%3D*'
+          'http://proxy.local/?url=http%3A%2F%2Fesri.rest%2Flocal%2Fquery%3Ff%3Dgeojson%26where%3D1%3D1%26outFields%3D*'
+        )
+      })
+      it('calls DataFetcher.openDataset with a proxied url', () => {
+        service.getDataset({
+          url: new URL('http://esri.rest/local'),
+          accessServiceProtocol: 'esriRest',
+          type: 'service',
+        })
+        expect(openDataset).toHaveBeenCalledWith(
+          'http://proxy.local/?url=http%3A%2F%2Fesri.rest%2Flocal%2Fquery%3Ff%3Dgeojson%26where%3D1%3D1%26outFields%3D*',
+          'geojson'
         )
       })
     })
--- a/libs/feature/dataviz/src/lib/service/data.service.ts
+++ b/libs/feature/dataviz/src/lib/service/data.service.ts
@@ -168,9 +168,8 @@ export class DataService {
   }
 
   getDataset(link: DatasetDistribution): Observable<BaseReader> {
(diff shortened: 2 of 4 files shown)


libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -197,7 +197,10 @@ export class DataService {
       link.type === 'service' &&
       link.accessServiceProtocol === 'esriRest'
     ) {
-      const url = this.getDownloadUrlFromEsriRest(linkUrl, 'geojson')
```
fgravin, line 200: Maybe the `linkUrl` was not explicit enough...
Should be `linkProxifiedUrl`, could have help to know to not use it, as `getDownloadUrlFromEsriRest` also proxifies the url.

By the way, I also see that in the same function, the call of `getDownloadUrlsFromWfs` also proxifies the URL twice, so to me, the codeflow within this method is confusing.

fgravin (commented):
Thanks Tobias, it looks good but I think the method `getDataset` could be reviewed and improved a bit.
Thanks

libs/feature/dataviz/src/lib/service/data.service.spec.ts
```diff
@@ -593,11 +593,22 @@ describe('DataService', () => {
       it('builds a proxied url', () => {
         expect(
           service.getDownloadUrlFromEsriRest(
-            'http://esri.rest/local/',
+            'http://esri.rest/local',
             'geojson'
           )
         ).toBe(
-          'http://proxy.local/?url=http%3A%2F%2Fesri.rest%2Flocal%2F%2Fquery%3Ff%3Dgeojson%26where%3D1%3D1%26outFields%3D*'
+          'http://proxy.local/?url=http%3A%2F%2Fesri.rest%2Flocal%2Fquery%3Ff%3Dgeojson%26where%3D1%3D1%26outFields%3D*'
```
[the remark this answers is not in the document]
tkohr: Because it only calls `getDownloadUrlFromEsriRest` once manually and thus is not bothered by the first call from `getDataset`.

libs/feature/dataviz/src/lib/service/data.service.ts
tkohr: Yep, true I'll rename it and see if I can improve more, like removing the double proxying for WFS, which does not cause errors right now thanks to [ogc-client](https://github.com/camptocamp/ogc-client/blob/main/src/shared/http-utils.ts#L80).

libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -197,7 +197,10 @@ export class DataService {
...
+      const url = this.getDownloadUrlFromEsriRest(
+        link.url.toString(), //let getDownloadUrlFromEsriRest() proxy url with params
```
jahow, line 201: Comment to remove

libs/feature/dataviz/src/lib/service/data.service.ts
tkohr: Ok, I've adapted my previous commit accordingly.

libs/feature/dataviz/src/lib/service/data.service.ts
tkohr: Ok, code should be clearer now

fgravin (approved):
Thanks @tkohr

PR #653, opened 2023-10-18
Fix favorite star tooltip translation

fgravin:
Ensure that the translation is loaded when the tooltip is created.
The `i18n` service loader is aynch, keep that in mind.


jahow (approved):
Thanks, indeed using `.instant()` should be avoided as much as possible!

PR #655, opened 2023-10-19
Fix CORS issues in webcomponent demo

fgravin:
The web components related to data visualization (table & chart) had CORS issue because of the Data API they used (wrong ODS API with CORS restriction).

So I just changed the dataset to have other API calls without CORS restriction.

I also removed the **facet** example as it is not working.


jahow (approved):
thanks!

PR #656, opened 2023-10-20
Fix [datahub]: Display external viewer button correctly

tkohr:
Fixes the external viewer button display while allowing truncating long values in the dropdown select.

button
no-button

### DIFF
--- a/libs/feature/record/src/lib/map-view/map-view.component.html
+++ b/libs/feature/record/src/lib/map-view/map-view.component.html
@@ -1,13 +1,14 @@
 <div class="w-full h-full flex flex-col p-1">
-  <div class="w-full mb-7 mt-1">
+  <div class="w-full flex justify-end mb-7 mt-1">
     <gn-ui-dropdown-selector
+      class="truncate p-1 -mx-1"
       extraBtnClass="!text-primary font-sans font-medium"
       [title]="'map.select.layer' | translate"
       [choices]="dropdownChoices$ | async"
       (selectValue)="selectLinkToDisplay($event)"
     ></gn-ui-dropdown-selector>
     <gn-ui-external-viewer-button
-      class="shrink-0"
+      class="shrink-0 py-1 place-self-end"
       [link]="selectedLink$ | async"
       [mapConfig]="mapConfig"
     >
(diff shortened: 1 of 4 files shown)


fgravin (approved):
Does it works well on mobile too ?

Thanks you :pray:

tkohr:
> Does it works well on mobile too ?

Ah, I forgot mobile, I'll have a look

tkohr:
I've added the `place-self-end` for correct positioning of the button on mobile. 

Independent of this PR the `truncate` within the `dropdown-selector` is broken when switching to mobile layout and applying `flex-col` [here](https://github.com/geonetwork/geonetwork-ui/blob/main/libs/ui/inputs/src/lib/dropdown-selector/dropdown-selector.component.html#L2). I remember we already had troubles making `truncate` work within `flex` in #516. 

For `flex-row` [this](https://github.com/geonetwork/geonetwork-ui/pull/656/files#diff-ab83ef264c8f460aa159ed9d48db47e4d8696a042dc3309384c48bd4d93e0e6fR4) works, but apparently not when the child switches to `flex-col`.

libs/feature/record/src/lib/map-view/map-view.component.html
```diff
@@ -1,13 +1,14 @@
 <div class="w-full h-full flex flex-col p-1">
-  <div class="w-full mb-7 mt-1">
+  <div class="w-full flex justify-end mb-7 mt-1">
     <gn-ui-dropdown-selector
+      class="truncate p-1 -mx-1"
```
jahow, line 4: I'm surprised that this does anything; for me the dropdown button already has a truncate built-in if the width is too low for the text to show

tkohr: Yes, strangely this is necessary, because of the flex I added in the parent.

tkohr:
I've added two more commits to fix the dropdown `truncate` when the title is displayed on top of the dropdown, as well the currently broken `truncate` in the `data-view.component` due to the `flex` around the dropdown.

mobile-no-button mobile-button


Everything displays well on desktop and mobile on my side now.

jahow (approved):
It looks good, I wish we could understand why this kind of class is needed outside of the component (it should be able to decide its own size correctly), but I looked and couldn't find any reason so, let's go with this.

Thanks @tkohr!

PR #657, opened 2023-10-20
Add E2E tests to workflow

f-necas:
🔧 E2E tests now works with Github actions.

⚠️ The default dump has been modified 

- New group with a logo in Geonetwork : `Barbie Inc.`
- A record has been modified to get this group as organization.

💢 Known problems

- Tests who access to Geonetwork (login/editing parameters) takes a lot of time (~15/20 sec).


support-services/docker-entrypoint-initdb.d/dump
jahow, line 1: Are you sure that adding the "barbie inc." organization was the only change? +1.26MB seems like a little too much for that, and it would be great to keep the size of this file low :)

conf/default.toml
```diff
@@ -6,7 +6,7 @@
 [global]
 # This URL (relative or absolute) must point to the API endpoint of a GeoNetwork4 instance
 geonetwork4_api_url = "/geonetwork/srv/api"
-datahub_url = "/datahub"
+datahub_url = "http://myhost/datahub"
```
jahow, line 9: this doesn't look like a good default for everyone; why not keep `/datahub`? btw one of the datahub e2e tests uses a custom configuration file

apps/metadata-editor-e2e/src/support/commands.ts
```diff
@@ -27,18 +27,14 @@ declare namespace Cypress {
 Cypress.Commands.add(
   'loginGN',
   (username: string, password: string, redirect = true) => {
-    cy.visit('http://localhost:8080/geonetwork/srv/eng/catalog.search#/home')
-    cy.get('.cookie-warning-actions').then(($cookie) => {
-      if ($cookie.is(':visible')) {
-        $cookie.find('button').eq(0).click()
-        cy.scrollTo('top')
-      }
+    Cypress.on('uncaught:exception', (err) => {
```
jahow, line 30: I was hoping we could have a central place to put Cypress custom commands, instead of having to duplicate these in the different apps; or maybe from here you could import the commands from the datahub-e2e app?

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
```diff
@@ -46,6 +37,9 @@ describe('dashboard', () => {
         .then((list) => {
           originalFirstItem = list.trim()
           cy.get('.record-table-header').first().click()
+          // Takes time to refresh results
+          // eslint-disable-next-line cypress/no-unnecessary-waiting
+          cy.wait(500)
```
jahow, line 42: My understanding is that this kind of `wait` should not be necessary because before declaring that an assertion fails, Cypress will "retry" several times for 2 seconds; what happens if you remove it?

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
```diff
@@ -10,15 +10,6 @@ describe('dashboard', () => {
...
         })
-      cy.get('gn-ui-pagination-buttons').find('gn-ui-button').last().click()
-      cy.get('gn-ui-record-table')
-        .find('.record-table-col')
-        .first()
-        .invoke('text')
-        .then((text) => {
-          expect(text).not.to.equal(pageOne)
-          cy.url().should('include', 'page=2')
-        })
```
jahow, line 21: Why remove that test?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches:
+      - main
+      - develop
+  pull_request:
+    types: [opened, synchronize, ready_for_review]
+
+env:
+  # a list of apps to build and publish on releases
```
jahow, line 11: I don't think you need to specify that actually; only these two apps have the `e2e` target defined, so you can probably just run `npx nx run-many --target=e2e` and it should do the same

apps/metadata-editor-e2e/src/e2e/my-org.cy.ts
```diff
@@ -1,9 +1,14 @@
...
   beforeEach(() => {
     cy.loginGN('barbie', 'p4ssworD_', false)
+    cy.intercept({
+      method: 'GET',
+      url: 'http://localhost:4200/geonetwork/srv/api/userselections/0/101',
+    }).as('dataGetFirst')
     cy.visit(`/records/my-org`)
     cy.get('md-editor-dashboard-menu').find('a').first().click()
-    cy.get('main').children('div').first().children('div').eq(1).as('linkGroup')
+    cy.wait('@dataGetFirst').its('response.statusCode').should('equal', 200)
```
jahow, line 10: I'm not sure I understand this assertion; first, it's counter-intuitive to have one in a `beforeEach` block, and then it looks like it's checking that favorites are correctly queried for the current user: why is it done, and why here?

jahow (commented):
Thanks so much for tackling this! that must have been a lot of work. Having e2e tests running on the CI will add more constraints but hopefully it will help us keep a higher level of quality. :sunglasses: 

I think it might be good to add a `npm run e2e` script (if it's not already there) and mention that having it pass is mandatory in the contributing guidelines.

Other than that I had several questions on your changes, see below!

support-services/docker-entrypoint-initdb.d/dump
f-necas: I started a fresh new docker comp (after removing the volumes) and just added a group, and edited a record to point to this group. 

Then I used the command provided in the Readme to generate the dump only ...

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
f-necas: As page number is hardcoded and the dump contains less records than 15, test fails. 
https://github.com/geonetwork/geonetwork-ui/blob/3c899e0f7ebd167a5cacb15c94aa7c91dc5ac88d/apps/metadata-editor/src/app/records/records-list.component.ts#L37

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
f-necas: On my computer it fails one out of five times. I remove to see how it works with gh.

conf/default.toml
f-necas: This method crashes if url is not  https://github.com/geonetwork/geonetwork-ui/blob/a12b13c742694b340efea9d196d625fc2a505bb9/apps/metadata-editor/src/app/router.service.ts#L21

support-services/docker-entrypoint-initdb.d/dump
jahow: I tried doing that locally and the file size almost didn't change (it actually went from 405K to 404K). What geonetwork version are you using? a version above 4.2.2 will trigger automatic migrations on startup and might make the database larger.

apps/metadata-editor-e2e/src/e2e/my-org.cy.ts
f-necas: The beforeEach sets a block as linkGroup, but sometimes, the xhr response hasn't finished and the test was failing on this line `cy.get('main').children('div').first().children('div').eq(1).as('linkGroup')`

Before XHR : 
Screenshot from 2023-10-23 11-33-08

After XHR : 
image

conf/default.toml
jahow: Right, this is a common problem. That's because the URL constructor will throw if it's provided with a relative url and no second argument; in that case the typical solution is to use:
```
new URL(`${getGlobalConfig().DATAHUB_URL}/search`, window.location.toString())
```
This will make a relative URL use the current host.

apps/metadata-editor-e2e/src/support/commands.ts
f-necas: I didn't find a simple and clean way for the moment so I didn't implement it by now.

jahow: this can be done later, no worries

PR #658, opened 2023-10-20
WebComponents / fix dropdown-selector not showing

jahow:
This PR brings several fixes to the web components:
* Rework the logic for providing a custom OverlayContainer in web components, by looking at the first ShadowDOM found and putting the container in it; this will make any custom overlays (e.g. dropdowns) work correctly
* Use a different event for closing the overlays in the dropdown components
* Fixes the font URL used for icons
* Add a logic on web components init to copy all `@font-face` definitions to the host document; this is necessary because `@font-face` rules in ShadowDOM will not trigger the download of the font, thus icons will never show in web components without that


apps/webcomponents/src/app/components/base.component.ts
```diff
@@ -68,13 +83,35 @@ export class BaseComponent implements OnChanges, OnInit {
       this.titleFont
     )
     this.facade.init(this.searchId)
+    this.copyFontFacesToDocument()
```
tkohr, line 86: Wow, good finding, weird that I have the feeling to already have seen icons in the web components in the past.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.html
```diff
@@ -4,6 +4,7 @@
   [action]="autoCompleteAction"
   (itemSelected)="handleItemSelection($event)"
   (inputSubmitted)="handleInputSubmission($event)"
+  (inputCleared)="handleInputCleared()"
```
tkohr, line 7: Good catch!

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
```diff
@@ -43,7 +43,7 @@
   [cdkConnectedOverlayPositions]="overlayPositions"
   [cdkConnectedOverlayScrollStrategy]="scrollStrategy"
   [cdkConnectedOverlayFlexibleDimensions]="true"
-  (backdropClick)="closeOverlay()"
+  (overlayOutsideClick)="closeOverlay()"
```
tkohr, line 46: I think there is still a `closeOverlay()` call missing when clicking the select box (with its arrow down) itself.

tkohr (commented):
Thanks for the various fixes @jahow !

Just noticed that the dropdown does not close when clicking the select box itself.

apps/webcomponents/src/app/components/base.component.ts
jahow: it used to work in the demo app because the icons font was  present in the host page

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
jahow: right, thanks!!

tkohr: Thanks, indeed works flawlessly now :-) 

Do I understand correctly that the pointer events on the select box were intercepted by the backdrop that might have been in front of the select box before?

tkohr (approved):
Thanks, @jahow, working very smoothly now!

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.html
jahow: I think it's just that the backdrop didn't have any style applied, so it wasn't positioned correctly and capturing anything. In the web components bundle we don't put the full Material theme CSS because it's too big, so we have to extract CSS classes each by each when using a Material component.

Issue #531, opened 2023-07-12
[Datahub] Simplity content of "format" dropdown


cmoinier:
Currently, the content of the formats dropdown is identical to what is found in the metadata.
Expected: a "human-compatible" presentation.

PR #659, opened 2023-10-20
Add an ElasticSearch pipeline to make formats human-readable

jahow:
This PR adds a script that will register a pipeline specific to GeoNetwork-UI on the ES index. This pipeline does:
- compute the quality score (taken from https://github.com/geonetwork/geonetwork-ui/discussions/446#discussioncomment-6035489)
- process the formats to make them more consistent & easier to read

The script needs to be run against the desired ES instance, and then a reindexation should be started.

image

Addresses #531 

To do:
- [x] make the ES host and index name configurable in the script
- [x] put the script somewhere else
- [x] document
- [x] make it easy to use in a docker context

### DIFF
--- a/docs/guide/deploy.md
+++ b/docs/guide/deploy.md
@@ -141,3 +141,45 @@ As such, **authenticated requests are not yet supported in GeoNetwork-UI in the 
 Lastly, even if authenticated requests were cleared regarding CORS rules, it would still be needed to disable the XSRF mechanism for the endpoints that GeoNetwork-UI relies on; XSRF protections works by making the client read the content of an HTTP cookie, and that is forbidden in a cross-origin context
 
 :::
+
+## Enabling improved search fields
+
+ElasticSearch offers the possibility to preprocess the records of a catalog, and this can be leveraged to **improve the search experience in GeoNetwork-UI**. This is done by registering so-called _ingest pipelines_.
+
+GeoNetwork-UI provides several pipelines:
+
+- Enable the Metadata Quality Score
+- Show better, human-readable data formats
+
+There are two options to register these pipelines:
+
+### Option A: Executing a Node script
+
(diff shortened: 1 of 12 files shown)


jahow:
I have added documentation on how to register pipelines provided by GeoNetwork-UI; hopefully this system can be used more in the future to preprocess records in various ways and optimize search.

Also added a docker image to register pipelines on startup in a docker/k8s environment.

This is ready for review!

f-necas (commented):
Thanks @jahow for this nice tool ! 

I just had to reindex records manually using geonetwork to have the correct values. I think we also have to mention that if we delete the index and we reindex records (using geonetwork), the pipelines is also dropped. and needs to be re-registered. 

 I didn't see any impact if we register multiple pipelines. But I don't know if it's better to clear pipeline before registering it. 

I'm also thinking about how we could implement it even more easily at starup. Could it be a script in the es container ? Just food for thought so.

jahow:
> I'm also thinking about how we could implement it even more easily at starup. Could it be a script in the es container ? Just food for thought so.

This can be discussed further with people doing devops, but I felt it was in our interest to keep the base images (geonetwork, ES...) identical and simply add this (optional) system alongside the rest.

jahow:
> I just had to reindex records manually using geonetwork to have the correct values. I think we also have to mention that if we delete the index and we reindex records (using geonetwork), the pipelines is also dropped. and needs to be re-registered.

Good point, I didn't mention that in the docs, thanks! this can be counter-intuitive.

> I didn't see any impact if we register multiple pipelines. But I don't know if it's better to clear pipeline before registering it.

I don't think it makes sense to register multiple pipelines. One pipeline can contain as many scripts as we want. We can even version it to make sure that we're not overwriting it with an older version, but I didn't go that far for now.

Also we don't need to clean it beforehand, although the CI gives the option to do it. The pipeline will always be overwritten. Tried and tested :)

f-necas (approved):
Nice improvements and documentation ! LGTM, thanks @jahow !

PR #660, opened 2023-10-24
fix(DH): Make font-sizes and colors same across the expandable panel

Angi-Kinas:
Changed the font-size and color as proposed to:
image

Also made opening of the panel faster.

### DIFF
--- a/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
+++ b/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
@@ -63,7 +63,11 @@
     <gn-ui-badge *ngIf="metadata.extras?.isOpenData">
       <span translate>record.metadata.isOpenData</span>
     </gn-ui-badge>
-    <span *ngFor="let usage of usages" gnUiLinkify>
+    <span
+      class="text-primary font-medium mt-2"
+      *ngFor="let usage of usages"
+      gnUiLinkify
+    >
       {{ usage }}
     </span>
     <span class="noUsage" *ngIf="!hasUsage">
--- a/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
+++ b/libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
@@ -4,7 +4,10 @@
 >
   record.metadata.about
 </p>
-<div class="mb-6 md-description sm:mb-4 sm:pr-16" *ngIf="metadata.abstract">
+<div
+  class="text-gray-900 mb-6 md-description sm:mb-4 sm:pr-16"
+  *ngIf="metadata.abstract"
+>
   <gn-ui-content-ghost ghostClass="h-32" [showContent]="fieldReady('abstract')">
     <p
       class="whitespace-pre-line break-words"
@@ -32,7 +35,10 @@
   *ngIf="metadata.lineage"
   [title]="'record.metadata.origin' | translate"
 >
-  <p class="mb-5 pt-4 whitespace-pre-line break-words" gnUiLinkify>
+  <p
+    class="text-gray-900 mb-5 pt-4 whitespace-pre-line break-words"
+    gnUiLinkify
+  >
     {{ metadata.lineage }}
   </p>
   <div
@@ -40,19 +46,19 @@
   >
     <div *ngIf="metadata.recordUpdated">
       <p class="text-sm" translate>record.metadata.updatedOn</p>
-      <p class="text-primary font-medium mt-2">
+      <p class="text-primary font-medium mt-[4px]">
         {{ metadata.recordUpdated && metadata.recordUpdated.toLocaleString() }}
       </p>
     </div>
     <div *ngIf="metadata.updateFrequency">
       <p class="text-sm" translate>record.metadata.updateFrequency</p>
-      <p class="text-primary font-medium mt-2" translate>
+      <p class="text-primary font-medium mt-[4px]" translate>
         domain.record.updateFrequency.{{ metadata.updateFrequency }}
       </p>
     </div>
     <div *ngIf="metadata.status">
       <p class="text-sm" translate>record.metadata.updateStatus</p>
-      <p class="text-primary font-medium mt-2" translate>
+      <p class="text-primary font-medium mt-[4px]" translate>
         domain.record.status.{{ metadata.status }}
       </p>
     </div>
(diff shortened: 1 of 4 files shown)


fgravin:
Thanks @Angi-Kinas 
Some remarks regarding the mockup:

**1. D’où viennent ces données ?**

- Margin between labels and values are too big
- The text, before the grid is not black
image

**2. Licence ou conditions d’utilisation ?**
- Text should be small
- Would be good to have the label stuff, if possible. Meaning we should be able to categorize the licence into a label.
image

**3. Détails**
The right margin after the icon seems too big



Overall, fonts are 14px & 16px, but it's hard to understand strictly the mockup as the font is completely different :smile:

tkohr:
Thanks @Angi-Kinas ! I'm wondering if the font sizes are not ok in this PR though @fgravin...

> 
> **1. D’où viennent ces données ?**
> 
>     * The text, before the grid is not black

I think this will be handled by the theming config, no?
> 
> 
> **2. Licence ou conditions d’utilisation ?**
> 
>     * Text should be small

Lucile lately proposed in a JIRA comment to put it in primary color (and then same size as primary text in other panels, I guess). It would look more consistent if there was also a label inside the panel I guess. But primary in 14px won't look consistent neither I think.

jahow (approved):
All good on my side, thanks!

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -32,27 +35,30 @@
   *ngIf="metadata.lineage"
   [title]="'record.metadata.origin' | translate"
 >
-  <p class="mb-5 pt-4 whitespace-pre-line break-words" gnUiLinkify>
+  <p
+    class="text-gray-900 mb-5 pt-4 whitespace-pre-line break-words"
```
tkohr, line 37: I thought, we would let the config set the `main` color here (as well as in the abstract above), but maybe I'm wrong (cc @fgravin @jahow)

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -4,11 +4,12 @@
 >
   record.metadata.about
 </p>
-<div class="mb-6 md-description sm:mb-4 sm:pr-16" *ngIf="metadata.abstract">
+<div class="text-gray-900 mb-6 md-description sm:mb-4 sm:pr-16">
```
fgravin, line 7: The text should be the `main` text color, I think it means that no color class is required.
We should not use `text-gray` things I guess.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -32,27 +35,30 @@
...
+  >
     {{ metadata.lineage }}
   </p>
   <div
     class="py-4 px-6 rounded bg-gray-100 grid grid-cols-2 gap-y-[10px] gap-x-[20px] text-gray-700 info-grid"
   >
     <div *ngIf="metadata.recordUpdated">
       <p class="text-sm" translate>record.metadata.updatedOn</p>
-      <p class="text-primary font-medium mt-2">
+      <p class="text-primary font-medium mt-[4px]">
```
tkohr, line 47: `mt-1` should be the same I think

Angi-Kinas:
@fgravin and @tkohr Thanks for the review. 
@fgravin I added most of your suggestions (see screenshot).

To point 2:
- As Lucile mentioned in the ticket that she wanted to change the way this is being displayed so it will be aligned with the other panels (in blue).
- the badge is only set if the flag is true. It would be too hacky to retrieve it from the text as it is right now.

Also:
Olivia and I discovered that there was another bug. The content ghost for the abstract was not shown. We fixed this as well.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
Angi-Kinas: > The text should be the `main` text color, I think it means that no color class is required. We should not use `text-gray` things I guess.

the gray-900 is derived from the main color, so that should not be a problem. the main color is almost black and here we wanted a lighter color (as you also mentioned in your review).
cc @tkohr

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
Angi-Kinas: I think it is more explixit to put it in px, because it reflects what is in the mockup. In terms of generated css it's almost the same.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
tkohr: Ok, I thought using the `main` color would allow more consistence with other components, but looking at the card on the home page we also use the derived `text-gray-900`, so it seems ok to me, as long as we try to stay consistent with the derived grays ;-)

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
tkohr: Ok, so far we've privileged the non interpolated values (also for the lightest CSS possible, I think). I don't have a strong opinion on this, but maybe we should agree on a best practice.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
fgravin: Why do you remove `*ngIf="metadata.abstract"` ? Is it the abstract container or not ?

And yes, the abstract should have the `main` color, and not `text-gray-900` I guess.

I think the colors are:
- Abstract title: **black**
Abstract content: **main**

- D’où viennent ces données ? title: **black**
Text underneath: **main**
Property label: **grayer**
Property color: **primary**

Angi-Kinas: The ngIf was not removed but just moved to the paragraph. Otherwise the content ghost will not be shown.

Regarding the colors:
main is almost black, so compared to the mockups it would be too dark. 
image

Maybe we shouldn't use pure `black` but `main` because [pure black is apparently bad practice](https://uxplanet.org/basicdesign-never-use-pure-black-in-typography-36138a3327a6) and it's not dependent on the theme.

PR #661, opened 2023-10-24
ME: Add checkbox for records table

Angi-Kinas:
image


jahow (approved):
Thanks a lot! looking good :)

BTW we should add some e2e tests for that but that can be done in https://github.com/geonetwork/geonetwork-ui/pull/666

PR #662, opened 2023-10-24
fix(DH): Change font style in expandable-panel

Angi-Kinas:
image


jahow (approved):
we missed that earlier

PR #663, opened 2023-10-24
Datahub: fine tuning

fgravin:
- Colors, margins, markup, transition, hover, borders, duration etc...

See commit messages for more information.

**For the color debate**, I don't really care. The only thing that I would like is that the geo2france design is respected.
There are way many other inconsistencies in the geo2france page, like too lighter grays all over the place. I don't know what is the best strategy, if there is a contrast issue, we could change the main color for the default theme ? or ask a UI designer to work on a default theme maybe.
The last commit can be removed anyway.

### DIFF
--- a/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
+++ b/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
@@ -15,7 +15,7 @@
     </div>
     <div class="px-3 pb-2 capitalize flex flex-col flex-grow overflow-hidden">
       <span
-        class="flex-shrink-0 mb-3 mt-5 font-title text-21 text-title group-hover:text-primary line-clamp-2 sm:mt-2"
+        class="flex-shrink-0 mb-3 mt-5 font-title text-21 text-title group-hover:text-primary line-clamp-2 sm:mt-2 transition-colors"
         data-cy="organizationName"
       >
         {{ organisation.name }}</span
--- a/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
+++ b/libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
@@ -4,7 +4,7 @@
     [title]="organisation.name"
   >
     <div
-      class="flex-shrink-0 bg-gray-100 rounded-lg overflow-hidden w-full border border-gray-300 h-36 group-hover:shadow-xl group-hover:border-0"
+      class="flex-shrink-0 bg-gray-100 rounded-lg overflow-hidden w-full border border-gray-300 h-36"
     >
       <gn-ui-thumbnail
         class="relative h-full w-full"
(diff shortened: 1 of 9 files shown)


Angi-Kinas (commented):
Thanks for making those changes. I just found a minor thing.

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.html
```diff
@@ -15,7 +15,7 @@
   </div>
 </div>
 <div
-  class="content duration-250 overflow-hidden transition-[max-height]"
+  class="content duration-250 overflow-hidden transition-[max-height] duration-300"
```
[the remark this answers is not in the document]
fgravin: Nice catch, actually the `duration-250` does not exist, that's why the animation was not working :smile:

fgravin:
The branch have been rebased and is ready to merge.
It does not change the gray color behavior, the tuning on the gray is just about the figure panel and the feed record preview, which were not consistent with the UI and the rest of the application.

libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
```diff
@@ -4,7 +4,7 @@
     [title]="organisation.name"
   >
     <div
-      class="flex-shrink-0 bg-gray-100 rounded-lg overflow-hidden w-full border border-gray-300 h-36 group-hover:shadow-xl group-hover:border-0"
```
tkohr, line 7: Are we going to add some other animation here, besides the title color transition?

libs/ui/dataviz/src/lib/figure/figure.component.spec.ts
```diff
@@ -62,7 +62,7 @@ describe('FigureComponent', () => {
   })
   it('label is main text color', () => {
```
tkohr, line 63: 'label is **black**...'

libs/ui/search/src/lib/record-preview-row/record-preview-row.component.html
```diff
@@ -20,7 +20,7 @@
   >
     <div class="mb-3 mt-5 sm:mt-2">
       <div
-        class="font-title text-21 text-title line-clamp-2 col-start-1 col-span-2 sm:line-clamp-1 group-hover:text-primary"
+        class="font-title text-21 text-title line-clamp-2 col-start-1 col-span-2 sm:line-clamp-1 group-hover:text-primary transition-colors"
```
tkohr, line 23: The transition has not been applied to the feed preview, but I guess this is on purpose (as it already has the shadow transition and primary org title).

tkohr (approved):
Thanks for fixing all these visual details @fgravin !

libs/ui/catalog/src/lib/organisation-preview/organisation-preview.component.html
fgravin: I don't think that it's expected, it should be the same as result search items.

libs/ui/dataviz/src/lib/figure/figure.component.spec.ts
fgravin, line 63:
```suggestion
  it('label is black text color', () => {
```

PR #667, opened 2023-10-30
ME: checkbox color options

Angi-Kinas:
The 'default' color for the checkboxes looks like this (it relies on --color-main variable):
image


jahow (approved):
Looking good, clean and simple!

PR #668, opened 2023-10-30
DH/Bug : Links format filter filters out everything

cmoinier:
### Summary

This PR fixes the behavior of the downloads filter. 

#### Issue
When clicking on one filter and clicking again on it, the download list becomes empty.

#### Fix
When clicking on one filter and clicking again on it, the download list falls back to 'All' and shows all download links.

### How to test

Go to a dataset with download links with multiple formats, eg : `/dataset/ee965118-2416-4d48-b07e-bbc696f002c2` , and click twice on the same format filter in the downloads section.


jahow (commented):
Thanks for the fix! I see two things here:
1. The unit tests (`downloads-list.component.spec.ts`) would probably be a better place to test this behaviour; it looks like the `toggleFilterFormat` method is not actually tested there, so this looks like a good time to add tests for it :slightly_smiling_face: Kudos for adding an associated test though!  
  (Note that you can of course keep the e2e test as well)
3. Nested ternary operations usually only make sense for their author; it's surprisingly hard to read for anyone else. I for one would really appreciate if you could instead define an intermediate variable like `bool noActiveFormat` :wink:

jahow:
I added a commit addressing my own comments, will merge once everything is green

jahow:
backported in https://github.com/geonetwork/geonetwork-ui/commit/67c862ff10a39ae011d56c3aa1dc1716eb41ce38

PR #669, opened 2023-11-02
[libs, ME]: persist records selection

Angi-Kinas:
- Adds selection service (select records, deselect records, clear selection)
- Uses selection service in record-table


libs/ui/search/src/lib/record-table/record-table.component.ts
```diff
@@ -14,12 +14,13 @@ import { SortByField } from '@geonetwork-ui/common/domain/search'
...
 export class RecordTableComponent {
-  selectedRecords: string[] = []
-
+  @Input() selectedRecords: string[] = []
   @Input() records: any[] = []
   @Input() totalHits?: number
   @Input() sortBy?: SortByField
   @Output() recordSelect = new EventEmitter<CatalogRecord>()
+  @Output() recordsSelection = new EventEmitter<CatalogRecord[]>()
+  @Output() recordsDeselection = new EventEmitter<CatalogRecord[]>()
```
jahow, line 23: I feel like the names of these events will bring confusion. How about:
* `recordClick`
* `recordsSelect`
* `recordsDeselect`?

apps/metadata-editor/src/app/records/records-list.component.ts
```diff
@@ -72,4 +77,27 @@ export class RecordsListComponent {
...
   }
+
+  getSelectedRecords() {
+    return this.selectionService.selectedRecordsIdentifiers$
+  }
+
+  handleRecordsSelection(records: CatalogRecord[]) {
+    this.selectionService
+      .selectRecords(records)
+      .pipe(takeUntil(this.onDestroy$))
```
jahow, line 88: Interesting, that looks like a valid approach to avoid having long-running observables stopping after a component is destroyed. Here though we shouldn't need it as the `selectRecords` and `deselectRecords` of the service are built on top of HTTP calls, which means they are short-lived observables and will complete after their first emission.

apps/metadata-editor/src/app/records/records-list.component.spec.ts
```diff
@@ -23,6 +24,7 @@ export class RecordTableComponent {
   @Input() records: CatalogRecord[]
   @Input() totalHits: number
   @Output() recordSelect = new EventEmitter<CatalogRecord>()
+  @Output() recordsSelection = new EventEmitter<CatalogRecord[]>()
```
jahow, line 27: Is this output really needed? I didn't see any use of it

apps/metadata-editor/src/app/records/records-list.component.spec.ts
```diff
@@ -23,6 +24,7 @@ export class RecordTableComponent {
   @Input() records: CatalogRecord[]
   @Input() totalHits: number
   @Output() recordSelect = new EventEmitter<CatalogRecord>()
```
jahow, line 26: See my other comment about event names, this might deserve a clarification as well :)

jahow (commented):
Thanks, this works really well. I made a few comments on the code, I think especially the event names deserve a little bit of clarification. Well done :)

jahow (approved):
Looking good, thanks!

PR #670, opened 2023-11-02
Fix [DH]: Display record's update status correctly

tkohr:
Currently, the update status coming from `cl_status` always displays the (translated) fallback value of `under_development`, because the mapper uses the `cl_status` object instead of its key.

under_dev


jahow (approved):
Thanks, good catch!!

jahow:
backported on 2.0.x by https://github.com/geonetwork/geonetwork-ui/commit/56412037f9c0de61b00de6cf5633644cf1d56a5e

PR #672, opened 2023-11-03
Datahub: Fix abstract display in record preview

tkohr:
Remove unnecessary whitespaces from html which can break line-clamp behaviour. [Example](https://www.geo2france.fr/geonetwork/srv/api/records/atmo-hdf::mes-hdf-journalier-poll-princ-1/formatters/xml?approved=true) abstract. Has only been reproduced on firefox.

line-clamp

### DIFF
--- a/libs/ui/search/src/lib/record-preview/record-preview.component.ts
+++ b/libs/ui/search/src/lib/record-preview/record-preview.component.ts
@@ -8,7 +8,11 @@ import {
   Output,
   TemplateRef,
 } from '@angular/core'
-import { propagateToDocumentOnly, stripHtml } from '@geonetwork-ui/util/shared'
+import {
+  propagateToDocumentOnly,
+  stripHtml,
+  removeWhitespace,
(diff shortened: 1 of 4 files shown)


jahow (commented):
I think this is taking the `stripHtml` function too far. Keeping whitespace and at the very least line breaks can be considered a feature, although I'm not sure if it would break anything.

Maybe adding another util `removeWhitespace` would make sense? And then the two can be combined for the results abstract.

tkohr:
Good point, I completely agree (although the `stripHtml` function is currently only used for the preview abstract). This should make things much more clear. I've adapted accordingly.

jahow (approved):
Thanks a lot @tkohr, even though it works exactly like before I believe you still improved the readability for future contributors ;)

jahow:
backported in https://github.com/geonetwork/geonetwork-ui/commit/a230b2c1e6216f6ac93eec832eed4929dc0d7e5a

PR #673, opened 2023-11-07
Fix [Datahub]: Do not point to future web component version

tkohr:
PR sets same `env.version` for prod as for dev as prod env is also used for latest builds (and not only local dev). This should fix the datahub latest build to point to the main web component dist branch instead of a version that does not exist yet.


jahow (approved):
thanks!

PR #675, opened 2023-11-07
fix(libs): Show metadata origin

Angi-Kinas:
Now the metadata origin will be displayed if there is data present, even if there is no lineage text:

image


libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -30,7 +30,13 @@
...
+  class="metadata-origin"
+  *ngIf="
+    metadata.lineage ||
+    metadata.recordUpdated ||
+    metadata.updateFrequency ||
+    metadata.status
+  "
   [title]="'record.metadata.origin' | translate"
 >
   <p class="mb-5 pt-4 whitespace-pre-line break-words" gnUiLinkify>
```
jahow, line 42: I think you need to also hide that block if there's no lineage, otherwise it makes a lot of whitespace

jahow (approved):
Looking good, thanks! let's wait for the backport bot to be operational before merging this :)

PR #592, opened 2023-08-25
Publish @gnui npm packages

fgravin:
Attempt to publish `geonetwork-ui` librarues as npm packages using Nx publishable targets.
eg. `nx g @nrwl/angular:library ui/blabla --publishable=true --importPath=@gnui/ui-blabla`

In this PR, 2 new libs

- `npm-publishable` which includes `npm-publishable-dep`
- which includes `ui-widgets` (renamed from `@geonetwork-ui/ui/widgets` to `@geonetwork-ui/ui-widgets`)
- which includes `util-shared` and third party libraries (`ngx-translate`, `@angular/material` ...)

So `npm-publishable` library is the entry point in this case for the third party angular application.
To make it `buildable` & `publishable`, all dependencies must be `publishable` as well. So `util-shared` and `ui-widgets` configuration files have been modified too (`project.json`, `package.json` ....).

`geonetwork-ui` internal libraries depends on third party libraries that **must** be included in the package dependencies. I choose to add them as `dependencies` instead of `peerDependencies`.

- ui-widgets/package.json
```json
  "dependencies": {
    "tslib": "^2.3.0",
    "@angular/material": "16.0.4",
    "@angular/material-moment-adapter": "16.0.4",
    "ngx-chips": "3.0.0",
    "@ngx-translate/core": "~14.0.0",
    "@ngx-translate/http-loader": "~7.0.0",
    "ngx-dropzone": "^3.0.0",
    "@types/geojson": "^7946.0.7",
    "@bartholomej/ngx-translate-extract": "^8.0.2",
    "@biesbjerg/ngx-translate-extract-marker": "^1.0.0",
    "chroma-js": "^2.1.2"
  },
```
- ui-widgets/ng-package.json
```json
  "allowedNonPeerDependencies": [
    "@angular/material",
    "@angular/material-moment-adapter",
    "ngx-chips",
    "@ngx-translate/core",
    "@ngx-translate/http-loader",
    "ngx-dropzone",
    "@types/geojson",
    "@bartholomej/ngx-translate-extract",
    "@biesbjerg/ngx-translate-extract-marker",
    "chroma-js"
  ]
```

**Build libraries**
`nx build npm-publishable`

**Publish libraries**
`npm i -g yalc`
from all library build package folders (eg. `dist/libs/npm/publishable`) run `yalc publish`

**Create the host app**
nx ng new gnui-host

**add dependencies from host**
[...]

PR #676, opened 2023-11-08
Export all libs as NPM package

jahow:
Similar to #592.

This PR introduces a `package` directory with:
* a `generate-package.js` node script
* an `index.ts` file which serves as an entrypoint for the package compilation
* a `ng-package.json` file which is used by [`ng-packagr`](https://github.com/ng-packagr/ng-packagr)
* a `package.json` describing the npm package
* a `tsconfig.json` file used for the package compilation

The `generate-package.js` file does:
* copy the contents of the `libs` folder in the `package` directory, keeping only relevant files (ts, css, html...)
* transform the import aliases in the TS files back to a relative path using the `paths` property of `tsconfig.base.json`; for instance:
  ```ts
  import { UiLayoutModule } from "@geonetwork-ui/ui/layout"
  ```
  becomes:
  ```ts
  import { UiLayoutModule } from "../../../libs/ui/layout"
  ```
  (this is needed because [`ng-packagr` does not support Typescript path aliases](https://github.com/ng-packagr/ng-packagr/pull/1502#issuecomment-572079617))
* launch a compilation of the copied sources using [`ng-packagr`](https://github.com/ng-packagr/ng-packagr); this is necessary because Angular libraries have to be published according to the [Angular Package Format](https://angular.io/guide/angular-package-format)
* the compilation output is going to `package/dist`; then the script also copies the source `libs` folder there as well so that consumers of the package can be redirected to the source file when inspecting it

An example usage of the generated package can be seen here: https://github.com/jahow/geonetwork-ui-custom-app
E.g.
```ts
import {UiElementsModule, UiInputsModule, UiWidgetsModule} from 'geonetwork-ui'
```

Notes:
* To make the compilation work I had to ensure that _all components were exported through their respective lib `index.ts` file_
* Some cyclic dependencies had also to be resolved

To do:
- [x] more tests on the consumer app side
- [ ] maybe add separate entrypoints, e.g. `geonetwork-ui/ui-layout`?
- [x] handle material theme and icon font
- [x] document


libs/common/domain/src/lib/record/contact.model.ts
```diff
@@ -1,5 +1,4 @@
 import { Organization } from './organization.model'
-import _default from 'chart.js/dist/plugins/plugin.tooltip'
```
jahow, line 2: No idea what this line was doing here to be honest!!

apps/map-viewer/src/app/app.module.ts
```diff
@@ -31,15 +31,7 @@ export const metaReducers: MetaReducer<any>[] = !environment.production
...
-  providers: [
-    {
-      provide: Configuration,
-      useFactory: () =>
-        new Configuration({
-          basePath: '/geonetwork/srv/api',
-        }),
-    },
-  ],
+  providers: [provideRepositoryUrl('/geonetwork/srv/api')],
```
tkohr, line 34: Why does the `map-viewer` not use this value from the config?

docs/guide/custom-app.md
```diff
@@ -0,0 +1,225 @@
...
+In order to discourage forking, GeoNetwork-UI is made to be publishable as an NPM package. This means that
+**a Custom Application should be a separate project in a separate repository**, and that it should simply list
+GeoNetwork-UI as a regular dependency. No forking, no git submodules, just NPM.
+
+This also means that whenever new functionalities from GeoNetwork-UI are needed the version of the dependency can be
+bumped accordingly.
+
+> **Note:** for now, the aforementioned NPM package isn't published on any repository.
+
+## What does the NPM package for GeoNetwork-UI contains?
```
tkohr, line 25:
```suggestion
## What does the NPM package for GeoNetwork-UI contain?
```

docs/guide/custom-app.md
```diff
@@ -0,0 +1,225 @@
...
+@import 'geonetwork-ui/style.css';
+```
+
+### Step 9: Initialize the color theme
+
+GeoNetwork-UI lets users define their own theme based on primary and secondary colors for instance. By default, no theme is specified
+so UI components might not render properly. To define a theme, call the following function on the application module constructor:
+
+```ts
+export class AppModule {
```
tkohr, line 205: When generating an angular app with the CLI 17 I noticed, that by default the app is generated as standalone without any `AppModule`. Maybe the doc should mention where the color theme should be initialized for standalone apps or that the app should be generated with the `--no-standalone` flag?

package.json
```diff
@@ -157,6 +157,7 @@
     "jest-canvas-mock": "2.5.2",
     "jest-environment-jsdom": "29.6.2",
     "jest-preset-angular": "13.1.1",
+    "ng-packagr": "^16.2.3",
```
tkohr, line 160: Maybe we could also add a script command to the `package.json`? Btw when running `generate-package.js` I noticed that it needs to be called from inside `package` and not from root.

package/generate-package.js
```diff
@@ -0,0 +1,211 @@
...
+const CURRENT_DIR_PATH = path.dirname(fileURLToPath(import.meta.url))
+const PROJECT_ROOT_PATH = path.join(CURRENT_DIR_PATH, '..')
+const PACKAGE_DIST_PATH = path.join(CURRENT_DIR_PATH, './dist')
+const LIBS_SOURCE_PATH = path.join(PROJECT_ROOT_PATH, './libs')
+const LIBS_DEST_PATH = path.join(CURRENT_DIR_PATH, './libs')
+const TRANSLATIONS_SOURCE_PATH = path.join(PROJECT_ROOT_PATH, './translations')
+const TRANSLATIONS_DEST_PATH = path.join(CURRENT_DIR_PATH, './translations')
+
+/**
+ * @param {number} depth Depth value from the `libs` folder (0 means in the libs folder)
```
tkohr, line 26: One `depth` too much :-)

package/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "geonetwork-ui",
+  "version": "2.1.0-dev",
```
tkohr, line 3: Will this be updated automatically, when releasing a new version, like in the root `package.json`?

tkohr (commented):
Thanks for this nice PR @jahow with a great description and extensive documentation!

Maybe the PR description could also go into a `package/readme.md`?

FYI, when testing, I needed to force `npm install` because of peer dependency conflicts with gn-ui still in angular 16 and your demo app in angular 17. Nevertheless, I ended up with your demo app running and calling GN behind it, nice :-)

Generating a new app following the doc I still run into some error when trying to start it: `⠸ Building...✘ [ERROR] Could not resolve "react-native-fs" node_modules/alasql/dist/alasql.fs.js:26676:25:`. My app is a standalone app without any `AppModule` for now. Not sure if this is related.

I've used `yalc` to test this PR:
`yalc publish` inside gn-ui's `package/dist`
`yalc add geonetwork-ui@2.1.0-dev` inside the angular app

apps/map-viewer/src/app/app.module.ts
jahow: the map viewer doesn't use the app config system yet, it just has a hardcoded value here; I would say it's intentional until we have more budget to improve this app

package.json
jahow: good point, this should be improved

docs/guide/custom-app.md
jahow: Very good point, I'll mention this in the doc; I don't think we can support standalone apps?

package/generate-package.js
jahow: it's normal actually, the first `depth` is the parameter name and afterwards is the parameter description. JSDoc weirdness

package.json
jahow: As for adding a script to publish the package in the root `package.json`, this could definitely be useful; I think we can do this later on, if e.g. we want to publish the package automatically on every commit/tag.

package/package.json
jahow: Done in https://github.com/geonetwork/geonetwork-ui/pull/676/commits/1a381ea23cdfd1f40bd7759f82bf315fb17e6b43

jahow:
I've added a README for the package folder as well as a README for the published package; it's now publicly accessible here: https://www.npmjs.com/package/geonetwork-ui?activeTab=readme

Also added some doc for development workflow in a custom app.

tkohr (approved):
Thanks for the additional doc and its latest adaptions! Following it, I now get to use the package without forcing `npm install`s. Really nice :smiley:

PR #677, opened 2023-11-08
feat(libs): Focus search field on app load

Angi-Kinas:
Add manual change detection as described [here](https://angular.io/errors/NG0100)

### DIFF
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
@@ -1,6 +1,7 @@
 import {
   AfterViewInit,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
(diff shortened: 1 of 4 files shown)


libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -109,6 +109,7 @@ export class AutocompleteComponent
   }
 
   ngAfterViewInit(): void {
+    this.inputRef.nativeElement.focus()
```
jahow: I think you removed this line in the other commit, but you still need it otherwise the field is not focused

jahow (commented):
Thanks! It looks like you unintentionally removed the line that was doing the focus, but I tested with the line re-added and it works well. Also I think this behavior should be configurable and disabled by default, otherwise it might disrupt the experience in other contexts e.g. the web component or the map viewer.

jahow (approved):
Great, thanks!

PR #678, opened 2023-11-10
DH/Improve navigation button style

cmoinier:
### Description

This PR addresses the request to have a return button (in record page) that is easier to find. Some users struggle to find the current return button with a complex background image : 

image(1)

After consulting Lucile it was decided to invert the button style, and have a white and transparent background and the text in back + a shadow.

image

### DIFF
--- a/libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
+++ b/libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
@@ -1,12 +1,6 @@
 <button
-  type="button"
-  class="group flex items-center justify-center p-1 bg-white rounded opacity-80 box-shadow-xl shadow-[4.0px_8.0px_8.0px_rgba(0,0,0,0.38)] mix-blend-luminosity text-black rounded hover:opacity-100"
+  class="group flex items-center justify-center backdrop-blur p-1 bg-primary-opacity-50 text-white rounded"
 >
-  <div
-    class="absolute -z-0 inset-0 bg-white opacity-80 group-hover:opacity-100 rounded transition duration-[20ms]"
-  ></div>
-  <mat-icon class="z-0 material-symbols-outlined align-middle">{{
-    icon
-  }}</mat-icon>
-  <span class="z-0 mx-2 hidden group-hover:inline">{{ label }}</span>
+  <mat-icon class="material-symbols-outlined align-middle">{{ icon }}</mat-icon>
+  <span class="mx-2 hidden group-hover:inline">{{ label }}</span>
 </button>
--- a/libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
+++ b/libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
@@ -1,9 +1,9 @@
 <button
   type="button"
-  class="relative backdrop-blur group flex items-center justify-center p-1 text-main rounded overflow-hidden shadow-2xl"
+  class="group flex items-center justify-center p-1 bg-white rounded opacity-80 box-shadow-xl shadow-[4.0px_8.0px_8.0px_rgba(0,0,0,0.38)] mix-blend-luminosity text-black rounded hover:opacity-100"
 >
   <div
-    class="absolute -z-0 inset-0 bg-white opacity-60 group-hover:opacity-100 transition duration-[20ms]"
+    class="absolute -z-0 inset-0 bg-white opacity-80 group-hover:opacity-100 rounded transition duration-[20ms]"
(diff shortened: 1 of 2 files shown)


jahow:
There's also this which can be very useful: https://developer.mozilla.org/en-US/docs/Web/CSS/backdrop-filter

Wide spread support is still quite fresh but e.g. blurring the backdrop of an element is a great effect for readability.

cmoinier:
@jahow Thanks for the suggestions! Last commit adds a shadow (it couldnt render in first commit) + the best `mix-blend` option I could find.

Backdrop-filter also looks interesting, but I don't think we can keep the shadow if we use this (both together look weird IMO). Having a shadow is quite important in case of very light/white backgrounds. What do you think?

jahow:
I tried using the backdrop blur and reworked the markup a bit:
Peek 10-11-2023 16-14

This is the markup that I used:
```html
<button
  type="button"
  class="relative backdrop-blur group flex items-center justify-center p-1 text-white rounded overflow-hidden"
>
  <div
    class="absolute -z-0 inset-0 bg-primary opacity-40 group-hover:opacity-100 transition duration-[20ms]"
  ></div>
  <mat-icon class="z-0 material-symbols-outlined align-middle">{{
    icon
  }}</mat-icon>
  <span class="z-0 mx-2 hidden group-hover:inline">{{ label }}</span>
</button>
```

The backdrop that changes opacity is a separate element, the rest is always at full opacity. What do you think?

PS: I tried using the blend modes but it won't work all the time un less we do `invert` or something (which looks weird).

jahow:
Also I tried using the same technique on the badges below the search input and it looks ok I think?
image

BTW we should make sure to use the same component everywhere, currently there are three different kinds

cmoinier:
With a pink-ish image in the background, it's hard to find though : 
image

jahow:
Yes that's true.

libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
```diff
@@ -1,5 +1,5 @@
 <button
-  class="group flex items-center justify-center p-1 bg-primary-opacity-25 text-white rounded"
+  class="group flex items-center justify-center p-1 bg-white opacity-80 box-shadow-xl shadow-[4.0px_8.0px_8.0px_rgba(0,0,0,0.38)] mix-blend-luminosity text-black rounded hover:opacity-100"
```
jahow, line 2: It's better use `text-main` instead of `text-black` here

Also I don't think we should customize the box shadow, it will end up not being consistent with others I think? Although I agree that the box shadows in tailwind are quite faint. I would suggest doing what I did in my comment, which is have a separate element for the background color and only change the opacity of this element; the rest can stay at 100% opacity and thus the shadow will be stronger. But I may be wrong, to be confirmed with trials :)

fgravin:
Definitely backdrop-filter is great, it's what is used in Felt for the popups over the map, and the rendering is way better than just transparent.
`backdrop-filter: blur(10px);`

fgravin:
BTW, if anyone could animate the back to search button with a transition, it will be a nice :gift:
 :star_struck:

libs/ui/inputs/src/lib/navigation-button/navigation-button.component.html
```diff
@@ -1,6 +1,12 @@
 <button
-  class="group flex items-center justify-center p-1 bg-white opacity-80 box-shadow-xl shadow-[4.0px_8.0px_8.0px_rgba(0,0,0,0.38)] mix-blend-luminosity text-black rounded hover:opacity-100"
+  type="button"
+  class="relative backdrop-blur group flex items-center justify-center p-1 text-main rounded overflow-hidden shadow-2xl"
 >
-  <mat-icon class="material-symbols-outlined align-middle">{{ icon }}</mat-icon>
-  <span class="mx-2 hidden group-hover:inline">{{ label }}</span>
+  <div
+    class="absolute -z-0 inset-0 bg-white opacity-60 group-hover:opacity-100 transition duration-[20ms]"
```
cmoinier, line 6: @fgravin Can do, but there's already a transition in the back to search button. Do you have something else in mind?

fgravin: The label extension, but forget it we will do that one day maybe :smile:

fgravin:
I've made it simple, works well for Saint Dizier. I did not want to have a complete different button than the others (shadow, color etc...)
I've homogeneized transparent button
- opacity 50
- blur(8px)

image
image
image

PR #679, opened 2023-11-10
Datahub (fix): Display UpdateFrequencyCustom correctly

tkohr:
The PR fixes `UpdateFrequency`s (like `annualy` from GN4) that are handled as objects in gn-ui.

before:
updateFrequency

after:
updateFrequency_after


translations/fr.json
```diff
@@ -121,14 +121,14 @@
...
   "domain.record.updateFrequency.irregular": "Irrégulière",
-  "domain.record.updateFrequency.month": "mois",
+  "domain.record.updateFrequency.month": "{count} fois par mois",
   "domain.record.updateFrequency.notPlanned": "Non planifiée",
   "domain.record.updateFrequency.periodic": "Récurrente",
   "domain.record.updateFrequency.unknown": "Inconnue",
-  "domain.record.updateFrequency.week": "semaine",
-  "domain.record.updateFrequency.year": "an",
+  "domain.record.updateFrequency.week": "{count} fois par semaine",
+  "domain.record.updateFrequency.year": "{count} fois par an",
```
tkohr, line 131: Looking at the (after) screenshot again, I'm telling myself we should probably translate the displayed numbers into words as well. Can we assume from [here](https://github.com/geonetwork/geonetwork-ui/blob/main/libs/api/metadata-converter/src/lib/iso19139/codelists/update-frequency.mapper.ts#L46) that 4 will stay the highest number?

jahow: I was thinking the same! I would only maybe translate up to 2 or 3. Theoretically it could be 12 times per year but it's fine not to translate the number into a word.

fgravin:
Thanks @tkohr, the status of the data has changed between the 2 screenshots :smile: 

Small detail but I find the "d'ou viennent ces données" layout a bit weird. Can you give the uuid of the record please ?
We should not return to the line in the middle of a link, start at the beginning of the line and use ellipsis if need, and it's weird the `/` at the end and the `BDTOPO.2020-11`

tkohr:
> Thanks @tkohr, the status of the data has changed between the 2 screenshots 😄
> 
> Small detail but I find the "d'ou viennent ces données" layout a bit weird. Can you give the uuid of the record please ? We should not return to the line in the middle of a link, start at the beginning of the line and use ellipsis if need, and it's weird the `/` at the end and the `BDTOPO.2020-11`

[Here](https://www.geo2france.fr/datahub/dataset/5be389bb-3e35-41ce-b638-145b53dd1ff7)'s the record. The different status in the screnshots is related to this [fix](https://github.com/geonetwork/geonetwork-ui/pull/670) ;-)

Regarding "d'ou viennent ces données", for me the main issue is that the `/` is not recognized of being part of the URL and the icon gets squeezed in between. The line break doesn't shock me and allows reading the link.

translations/fr.json
tkohr: So, finally I just put `1` into words, as the `messageformat` only distinguishes singular and plural out of the box.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -37,6 +42,24 @@ export class MetadataInfoComponent {
     return array
   }
 
+  get updateFrequency(): string {
+    if (
+      (this.metadata.updateFrequency as UpdateFrequencyCustom).updatedTimes !==
```
jahow, line 47: I think the assertion here could be:
```
if (this.metadata.updateFrequency instanceof Object)
```

Then TS should infer that we're indeed in the case of an `UpdateFrequencyCustom` ad you shouldn't have to cast it again :)

jahow (approved):
Thanks!! I made a suggestion to simplify the code, adapt as you see fit!

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
tkohr: You're, right, much more readable... and typescript is pretty smart, actually.

jahow: annoying but smart ;)

PR #681, opened 2023-11-11
Reload the application on language switch

fgravin:
For multilingual concerns, after that the user switches the language, the application should
- translate all the UI
- translate all the multingual content of the metadata
- use the language as a boost factor within the search

The easiest way to achieve all of that is to **reload the page** for now.

**Note**: the `language-switcher` is a UI component and should not interact with the `localStorage` IMO. I didn't want to refactor so I added the reload along the storage process.

The PR also modifies the content of the language switcher, to translate the entries from the targeted language.


conf/default.toml
```diff
@@ -31,7 +31,7 @@ proxy_path = ""
 # This optional parameter defines the languages that will be provided in a dropdown for the user to translate the UI.
 # Available languages are listed here: (https://github.com/geonetwork/geonetwork-ui/blob/main/libs/util/i18n/src/lib/i18n.constants.ts).
 # More information about the translation can be found in the docs (https://geonetwork.github.io/geonetwork-ui/main/docs/reference/i18n.html)
-# languages = ['en', 'fr', 'de']
+languages = ['en', 'fr', 'de']
```
jahow, line 34: This should stay commented I think

conf/default.toml
```diff
@@ -14,7 +14,7 @@ proxy_path = ""
 # Use ISO 639-2/B (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) format to indicate the language of the metadata.
 # Setting to "current" will use the current language of the User Interface.
 # If not indicated, a wildcard is used and no language preference is applied for the search.
-# metadata_language = "current"
+metadata_language = "current"
```
jahow, line 17: I would leave this commented; even though it's a logical behaviour for multilingual catalogs I think the default conf should only contain mandatory keys

jahow (commented):
Looking good, thanks for changing the translations of the language switcher option, it's indeed more logical now. I think the changes to the default config should be reverted, was that intentional?

The language switcher is a very simple component so it wasn't made into a smart/presentation couple. On the other hand it might be useful to abstract the logic for reading/writing the current language into a separate service.

fgravin:
Thanks @jahow yes I unintentionally pushed the configuration.

PR #686, opened 2023-11-13
feat(DH): Improve plurals in text

Angi-Kinas:
image
image


tkohr (approved):
Thanks, @Angi-Kinas ! Maybe you could adapt de `de.json` accordingly as well ?

PR #687, opened 2023-11-13
[Datahub] Add an API form generator for OGC API Records

cmoinier:
This PR creates an API form component to handle customization of OGC API urls.

It also adds a date-range picker component to gn-ui (because such an input was in the original mockup, and will be used in gn-editor in the future).

**Funded by Métropole de Lille**

image


apps/datahub/src/app/record/custom-api/custom-api.component.html
fgravin, line 1: IMO
- the should be a shared feature component
- it's not custom-api, for now it's OGC API Features

apps/datahub/src/app/record/custom-api/custom-api.component.html
```diff
@@ -0,0 +1,47 @@
+<div
+  class="bg-primary-opacity-10 pt-8 overflow-visible sm:pt-5"
+  style="height: 512px"
```
fgravin, line 3: Why a height ? Is it supposed to be scrollable ?
Otherwise the height is built from the content.

BTW, you should use `class="h-[512px]"` instead

apps/datahub/src/app/record/custom-api/custom-api.component.html
```diff
@@ -0,0 +1,47 @@
...
+      </div>
+      <div class="flex flex-col">
+        <p>Intervalle de temps</p>
+      </div>
+    </div>
+  </div>
+  <div class="text-[23px]" translate>URL personnalisée</div>
+  <gn-ui-text-input [value]="'Your URL here'" [required]="false" hint="">
+  </gn-ui-text-input>
+  <div class="text-[23px]" translate>URL racine</div>
```
fgravin, line 36: Does not seem optimzed to switch from 23 to 20 like this.
BTW, do we have a zeplin or figma ?

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -120,9 +120,13 @@
     </div>
 
     <div *ngIf="displayApi$ | async">
-      <datahub-record-apis class="block mt-6"></datahub-record-apis>
+      <datahub-record-apis
+        (displayCustomPanel)="isCustom($event)"
```
fgravin, line 124: Note that in the long run, there will be several kinds of APIs forms. For instance
- OGC API Features
- WFS
- IGN download service
- IGN extraction service

fgravin (commented):
Hi @cmoinier, not a real review but some thoughts from here to there :relaxed:

apps/datahub/src/app/record/custom-api/custom-api.component.html
cmoinier: Okay for the shared feature component point, but I named it like that in prevision of the future, to avoid having to rename it. Since there will be numerous kinds of APIs forms, doesn't it make sense to have a generic name?

apps/datahub/src/app/record/custom-api/custom-api.component.html
cmoinier: This is a WIP, the style is not done at all, I barely did the structure here ^^. This `div` was copied from the `data view` part of the record page (which yes, has a height for some reason), just to have a `div` that would go across the screen. I'm planning on doing styling after everything else.

apps/datahub/src/app/record/custom-api/custom-api.component.html
cmoinier: There is a figma, I got the link to it after committing everything I've done here. I will use it to have better styling values from now on. Styling is not done/approximative right now, I'm focusing on the functionalities first.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
cmoinier: Noted. This output is here to allow the custom API cards to open the panel to customize the URL, which isn't available for pre-defined API cards.

apps/datahub/src/app/record/custom-api/custom-api.component.html
cmoinier: How about : `api-custom-url` ? Any other idea?

fgravin: will it handle all kinds of APIs ? eg OGC, WFS, STAC and so on ? Or is it just for this concern (OGC API Features) ?

cmoinier: Yes, this component will be used in the future for all kinds of APIs. For now, only OGC though

fgravin: `record-api-form` ?

cmoinier: Sounds good!

apps/datahub/src/app/record/record-apis/record-apis.component.html
```diff
@@ -1,25 +1,72 @@
...
+  </p>
+  <mat-tab-group
+    (selectedIndexChange)="onTabIndexChange($event)"
+    animationDuration="200ms"
+    mat-stretch-tabs="false"
+    mat-align-tabs="start"
+  >
+    <mat-tab>
+      <ng-template mat-tab-label>
+        <span class="tab-header-label sm:ml-32" translate
```
fgravin, line 16: translation keys must be normalized.

libs/ui/elements/src/lib/record-api-form/record-api-form.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  styleUrls: ['./record-api-form.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RecordApiFormComponent {
+  @Input() set apiLink(value: DatasetServiceDistribution) {
+    this.linkHref = new URL('/data/swagger-ui/index.html', value.url.origin)
+    this.apiBaseUrl = value.url.href
+    this.resetUrl()
+  }
+  offset$ = new BehaviorSubject('')
```
fgravin, line 22: Why do you use Observables everywhere ?
Classic change detection is meant for this.
BTW, we could take the opportunity to test signals !

apps/datahub/src/app/record/record-apis/record-apis.component.html
```diff
@@ -1,25 +1,72 @@
...
+            [ngClass]="{
+              'mr-[var(--container-outside-width)]': last,
+              'ml-[var(--container-outside-width)]': first
+            }"
+          >
+          </gn-ui-api-card>
+        </gn-ui-carousel>
+      </div>
+    </mat-tab>
+    <mat-tab>
```
fgravin, line 43: Too much duplicated code.
Probably a way to handle that with template outlet + context.
Anyway, don't jump into that yet cause I don't see the point of those 2 tabs, wait Lucile to give feedbacks.

apps/datahub/src/app/record/record-apis/record-apis.component.html
```diff
@@ -1,25 +1,72 @@
...
+              let last = last
+            "
+            [custom]="true"
+            [link]="link"
+            class="w-80"
+            [ngClass]="{
+              'mr-[var(--container-outside-width)]': last,
+              'ml-[var(--container-outside-width)]': first
+            }"
+            (isCustom)="openCustomPanel($event)"
```
fgravin, line 65: The name `isCustom` for the event is confusing.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -120,9 +120,35 @@
...
     </div>
   </div>
+  <div
+    class="bg-primary-opacity-10 pt-8 overflow-visible sm:pt-5"
+    *ngIf="selectedCard$ | async"
+  >
+    <div class="flex flex-col container-lg px-4 lg:mx-auto gap-6">
+      <div class="flex flex-wrap justify-between mb-6 sm:mb-2 ng-star-inserted">
+        <p class="text-[28px] text-title font-title" translate>
+          Générer une URL personnalisée
```
fgravin, line 136: You should take the habit to not write french text in the template, it's a source for missing some translation :upside_down_face:

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -120,9 +120,35 @@
     </div>
 
     <div *ngIf="displayApi$ | async">
-      <datahub-record-apis class="block mt-6"></datahub-record-apis>
+      <datahub-record-apis
+        (selectedLink)="selectCard($event)"
```
fgravin, line 124: wondering if it could go in mdview state .. any thought @tkohr @jahow ?

fgravin (commented):
Thanks @cmoinier for your progress on this.
I've made some general remarks regarding code practice.
Let me know if it make sense for you :relaxed:

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
cmoinier: Translations are not done yet :slightly_smiling_face:

apps/datahub/src/app/record/record-apis/record-apis.component.html
cmoinier: Good catch, we reused this output for something else and forgot the change the name

apps/datahub/src/app/record/record-apis/record-apis.component.html
cmoinier: Translations are not done yet (I was updating the todo list when you reviewed :slightly_smiling_face: )

libs/ui/elements/src/lib/record-api-form/record-api-form.component.ts
cmoinier: Actually, the other solutions we tried to get the different inputs to react to changes made from the code (eg : reset to the initial values), were not working with classic change detection, hence the Observables. But let me know if you have other suggestions, happy to hear them.

fgravin: Did you try `ChangeDetectorRef` ?
Actually I don't know what is the best practice about that.

cmoinier: I believe we did, right @tkohr ?

tkohr: Indeed, we had some change detection problems when switching cards which we were not able to resolve (even by manually calling `ChangeDetectorRef.detectChanges()`)

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
tkohr: I think I'd rather keep it in the `record-metadata` "state" like all the `displaySomething$` observables. But maybe you see some advantage in the facade for later needs?

fgravin:
Could we see a screenshot ?? :relaxed:

apps/datahub-e2e/src/e2e/home.cy.ts
```diff
@@ -78,7 +78,7 @@ describe('header', () => {
     })
   })
 
-  describe('my favorites button', () => {
+  describe.only('my favorites button', () => {
```
tkohr, line 81: Don't forget to remove the `only` here.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -120,7 +120,35 @@
...
+        <button type="button" class="text-main" (click)="closeForm()">
+          <span class="text-sm text-primary" translate
+            >record.metadata.api.form.closeButton</span
+          >
+          <mat-icon
+            class="material-symbols-outlined ml-1 align-middle leading-none text-primary"
+            >close</mat-icon
+          >
+        </button>
+      </div>
```
tkohr, line 148: I think it'd be nice to move this html to the form component.

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -16,6 +25,7 @@ import { MdViewFacade } from '@geonetwork-ui/feature/record'
 })
 export class RecordMetadataComponent {
   @Input() metadataQualityDisplay: MetadataQualityDisplay
+  @Output() apiLink = new EventEmitter<DatasetServiceDistribution>()
```
tkohr, line 28: Do we still need this `Output`?

libs/ui/elements/src/lib/api-card/api-card.component.ts
```diff
@@ -1,12 +1,41 @@
...
+export class ApiCardComponent implements OnInit, OnChanges {
   @Input() link: DatasetServiceDistribution
+  @Input() currentLink: DatasetServiceDistribution
+  displayApiFormButton: boolean
+  currentlyActive = false
+  @Output() openRecordApiForm: EventEmitter<DatasetServiceDistribution> =
+    new EventEmitter<DatasetServiceDistribution>()
+
+  ngOnInit() {
+    this.displayApiFormButton =
```
tkohr, line 28: This could probably also go into a setter, but both works I guess.

libs/ui/elements/src/lib/api-card/api-card.component.ts
```diff
@@ -1,12 +1,41 @@
...
+      this.link.accessServiceProtocol === 'ogcFeatures' ? true : false
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    this.currentlyActive =
+      changes.currentLink.currentValue === this.link ? true : false
+  }
+
+  openPanel() {
```
tkohr, line 37: We could maybe still name the method a little more explicitly (eg. `openRecordApiFormPanel` even if it's long?)

libs/ui/elements/src/lib/api-card/api-card.component.ts
```diff
@@ -1,12 +1,41 @@
...
+    this.currentlyActive = !this.currentlyActive
+    this.openRecordApiForm.emit(this.currentlyActive ? this.link : undefined)
```
tkohr, line 39: I don't think we have a test for this ternary open/close logic, not sure...

tkohr (commented):
Great work @cmoinier ! Thanks for your continued effort on this. I've left some minor comments. Besides that, I still see two points to address:

- Currently, the form breaks on mobile. I guess we can just hide the three dots button on mobile or replace it with the copy url button. Not sure if this has been discussed.
- Comparing the form with the mockup I still see quite some differences, like font sizes, paddings and margins, line breaks of the labels.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -120,7 +120,35 @@
...
     <div *ngIf="displayApi$ | async">
-      <datahub-record-apis class="block mt-6"></datahub-record-apis>
+      <datahub-record-apis
+        [isPanelOpen]="isPanelOpen"
+        (selectedLink)="selectApiLink($event)"
+        class="block mt-6"
+      ></datahub-record-apis>
+    </div>
+  </div>
+  <div
```
fgravin, line 130: I think that this section should be within the `datahub-record-apis` component, it's purely related.
WSYT @tkohr ?

fgravin (commented):
Very good work, you've added a bunch of e2e & unit tests so it's very complete :clap: 

Just yes like @tkohr says, there are many slight differences between the mockup and your design (fonts, sizes, margin etc..). Please try to replicate the mockup with accuracy, it's all about details :smile: 

Thanks

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
tkohr: Ah right, I missed that too. Ideally, I'd see the lines I commented below in the `record-api-form.component` and then the `record-api-form.component` go into the `record-metadata.comonent`. WDYT @cmoinier @fgravin ?

cmoinier:
Thanks @tkohr @fgravin for your reviews. I have corrected most of the things you mentionned as per the last commit. I still need to improve the design before it's considered done. :slightly_smiling_face:

fgravin:
> Thanks @tkohr @fgravin for your reviews. I have corrected most of the things you mentionned as per the last commit. I still need to improve the design before it's considered done. 🙂

Thanks @cmoinier we will have a look.
BTW I am wondering how do you test ? I think that having a metadata which references an OGC API Features resources in the `support-services` compo would help.
I don't know how hard it is though.

Thanks

cmoinier:
@fgravin I added a dump with a metadata that has two OGC API Features resources.

fgravin (dismissed):
Still some issues with the rendering.
You don't handle the width of your form correctly, see the "preview" container which behave the same.
- on full screen, the form is too large
- on mobile, it's too narrow.

Also, the style of the copy text input is not like the mockup.

fgravin:
On my tests, the cross of the close button is bigger than the text.
image

fgravin:
Using the support-services locally, the dataset "Accroche velo" with the ogc api features has lost the "similar records" panel.
Is it related to the PR ?

cmoinier:
Noted for the rendering issues, I'm on it. 

For the "similar records" panel, yes, it is related to the PR. I think I have accidentally deleted the similar record link and therefore, the panel isn't displaying anymore. I will add it again and push another dump.

tkohr (approved):
Thanks, @cmoinier, everything looks good on my side! I just added a last commit to add some opacity to the transition and remove some unnecessary code, I still came across.

I let you have a look at it and wait for the CI to pass, before merging :-)

PR #689, opened 2023-11-14
fix(DH): Link detection with unknown mime type

Angi-Kinas:
The link detection of a file format is now working even if the mime type is unknown

image


tkohr (approved):
Good catch! Thanks, @Angi-Kinas!

PR #698, opened 2023-11-20
Fix(DH): Center bbox extent with different coordinate reference system

Angi-Kinas:
This PR fixes the bug that occurs if the extent of the feature is in a different coordinate system than WGS84 (EPSG: 4326). In this case the entire world was used as extent. Now we transform the coordinate system and use the bbox of the feature.
In order to do this, we had to update openlayers to version 8.2.0 and adapt a few types.


apps/datafeeder/src/app/presentation/components/data-import-validation-map-panel/data-import-validation-map-panel.component.ts
```diff
@@ -10,19 +10,20 @@ import {
...
-import type { Feature } from 'geojson'
+import type { Feature as FeatureType } from 'geojson'
 import { asArray, asString } from 'ol/color'
 import { isEmpty } from 'ol/extent'
 import GeoJSON from 'ol/format/GeoJSON'
 import { Geometry } from 'ol/geom'
 import TileLayer from 'ol/layer/Tile'
 import VectorLayer from 'ol/layer/Vector'
-import Map from 'ol/Map'
+import { Map, Feature } from 'ol'
```
jahow, line 20: It's generally safer to import things from more "precise" entrypoints in ol, to avoid bringing too much of the library. E.g.:

```ts
import Map from 'ol'
import Feature from 'ol/Feature'
```

libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.spec.ts
```diff
@@ -12,7 +12,7 @@ import {
   MapManagerService,
 } from '@geonetwork-ui/feature/map'
 import { FEATURE_COLLECTION_POINT_FIXTURE_4326 } from '@geonetwork-ui/common/fixtures'
-import { Map } from 'ol'
+import { Feature, Map } from 'ol'
```
jahow, line 15: Same as above, it would be great if you could turn this into more precise imports!

libs/feature/dataviz/src/lib/geo-table-view/geo-table-view.component.spec.ts
```diff
@@ -21,6 +21,14 @@ import XYZ from 'ol/source/XYZ'
 import { Subject } from 'rxjs'
 
 import { GeoTableViewComponent } from './geo-table-view.component'
+import { Geometry } from 'ol/geom'
+
+class ResizeObserverMock {
```
jahow, line 26: This type of generic mock can go in the `test-setup` files in each lib, in order to avoid cluttering each separate test file

libs/feature/map/src/lib/utils/map-utils-wms.service.ts
```diff
@@ -24,17 +28,35 @@ export class MapUtilsWMSService {
...
     return this.getLayerFull(layer).pipe(
-      map((wmsLayerFull) => this.getLonLatBBox(wmsLayerFull))
+      switchMap((wmsLayerFull) => from(this.getLonLatBBox(wmsLayerFull)))
     )
   }
 
-  getLonLatBBox(wmsLayerFull: WmsLayerFull) {
+  async getProjFromEPSG(EPSGCode) {
+    return fromEPSGCode(EPSGCode)
+  }
```
jahow, line 37: Is this function really needed? Calling `fromEPSGCode` directly elsewhere should work

jahow (approved):
Looking good, well done!! I've made a few suggestions to improve the OL imports and simplify the code, address as you see fit!

PR #699, opened 2023-11-20
Fix scale value to avoid nullity

f-necas:
# Datafeeder Scale fix

In order to avoid a `null` value sent to the server while a default value is the present in the UI. 

This PR fixes the next issue :
- #685

❗  Also fix an error in the console mentioning  that 'BrowserAnimationsModule' wasn't present.

### DIFF
--- a/libs/feature/editor/src/lib/components/wizard-field/wizard-field.component.ts
+++ b/libs/feature/editor/src/lib/components/wizard-field/wizard-field.component.ts
@@ -1,5 +1,6 @@
 import {
   AfterViewInit,
+  ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
   Input,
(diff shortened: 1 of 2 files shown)


libs/feature/editor/src/lib/components/wizard-field/wizard-field.component.ts
```diff
@@ -103,15 +104,20 @@ export class WizardFieldComponent implements AfterViewInit, OnDestroy {
         return data ? new Date(Number(data)) : new Date()
       }
       case WizardFieldType.DROPDOWN: {
+        //TODO called continuously
```
jahow, line 107: Could you maybe make that clearer? does that mean that this code is executed a lot and that it's undesirable?

jahow (approved):
Thanks, not tested but it makes sense

Issue #694, opened 2023-11-18
Datahub: geometry boost error if area of competence is empty


fgravin:
**Issue**
If the geometry boost is enabled in the configuration, and it returns an empty geometry
```js
coordinates
: 
[[]]
type
: 
"Polygon"
```
Then main search returns an error
```
Error is: Bad Request.
Request:
{"aggregations":{},"from":0,"size":10,"sort":[{"_score":"desc"}],"query":{"bool":{"must":[{"terms":
{"isTemplate":["n"]}}],"must_not":{"terms":{"resourceType":["service","map","map/static","mapDigital"]}},
"should":[{"geo_shape":{"geom":{"shape":{"type":"Polygon","coordinates":[[]]},"relation":"within"},"boost":10}},
{"geo_shape":{"geom":{"shape":{"type":"Polygon","coordinates":[[]]},"relation":"intersects"},"boost":7}}],
"filter":[{"query_string":{"query":"*:* AND (draft:n OR draft:e)"}}]}},"track_total_hits":true,"_source":["uuid","id","title","resource*","resourceTitleObject","resourceAbstractObject","overview","logo",
"codelist_status_text","linkProtocol","contactForResource*.organisation*","contact*.organisation*",
"userSavedCount","op*","groupOwner","owner","id"]}
.
Error:
{"error":{"root_cause":[{"type":"exception","reason":"not enough points to build a line"}],
"type":"x_content_parse_exception","reason":"[1:281] [bool] failed to parse field [should]","caused_by":
{"type":"x_content_parse_exception","reason":"Failed to build [geojson] after last required field arrived","caused_by":
{"type":"exception","reason":"not enough points to build a line"}}},"status":400}.
```
**Expected**
This case is handled and does not break the search

jahow:
The geometry given in example is invalid I think? We should better handle this case yes

fgravin:
> The geometry given in example is invalid I think? We should better handle this case yes

I don't know, but when the organization has no area of competence, the geOrchestra console returns that. Maybe we should fix the console as well.

PR #700, opened 2023-11-21
Datahub: handle filter geometry error

fgravin:
#694

The validation can
- throw an error is the GeoJson is malformed
- return false if the GeoJson is not valid.

Both cases are supported.
This ensure the search runs even with a wrong geometry filter.

I'll provide test if you approve the approach.

### DIFF
--- a/libs/feature/search/src/lib/state/effects.ts
+++ b/libs/feature/search/src/lib/state/effects.ts
@@ -146,7 +146,7 @@ export class SearchEffects {
               map((geom) => [state, favorites, geom]),
               catchError((e) => {
                 return of([state, favorites, null])
-              }) // silently opt out of spatial filter if an error happens
+              })
             )
           }),
           switchMap(
--- a/libs/feature/search/src/lib/state/effects.ts
+++ b/libs/feature/search/src/lib/state/effects.ts
@@ -1,7 +1,7 @@
 import { Inject, Injectable, Optional } from '@angular/core'
 import { Actions, createEffect, ofType } from '@ngrx/effects'
 import { select, Store } from '@ngrx/store'
-import { buffer, combineLatestWith, debounceTime, from, of } from 'rxjs'
+import { buffer, combineLatestWith, debounceTime, from, of, tap } from 'rxjs'
 import {
   catchError,
   map,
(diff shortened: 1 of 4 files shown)


fgravin:
@jahow could you check this one too ?
Thanks

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -130,8 +131,17 @@ export class SearchEffects {
               return of([state, favorites, null])
             }
             return this.filterGeometry$.pipe(
+              tap((geom) => {
+                const isValid = validGeoJson(geom)
+                if (!isValid) {
+                  throw '\nFilter geometry is not a valid GeoJson'
```
jahow, line 137: It's better practice to throw `Error`s, not strings.

Although in that case I would throw an empty error and output the warning in the console.

Also I think `geojson-validation` provides traces for the error it found in the object, maybe that could be interesting to print as well to help debug an invalid geom? See https://gitlab.com/mjbecze/GeoJSON-Validation/-/tree/master?ref_type=heads#usage

```js
//test to see if `invalidFeature` is valid and return a "trace" which contains the error
const trace = gjv.isFeature(invalidFeature, true)
console.log(trace)
```

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -130,8 +131,17 @@ export class SearchEffects {
...
+                }
+              }),
               map((geom) => [state, favorites, geom]),
-              catchError(() => of([state, favorites, null])) // silently opt out of spatial filter if an error happens
+              catchError((e) => {
+                console.warn('The filter geometry cannot be used', e)
```
jahow, line 142: This is not silent any more since we're outputting something in the console now. I would remove this `console.warn` call and instead print to the console if the geojson is invalid.

jahow (commented):
See comments on how to handle the invalid geom case, thanks!

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -128,8 +129,24 @@ export class SearchEffects {
...
+                    throw new Error(trace.join('\n'))
+                  }
+                } catch (error) {
+                  console.warn(
+                    'Error while parsing the geometry filter\n',
+                    error
+                  )
+                  throw new Error()
+                }
+              }),
```
jahow, line 145: :ok_hand:

PR #703, opened 2023-11-24
Add a Platform service

fgravin:
Add a platform service interface + a gn4 implementation to centralize
- user repository
- auth
- organization repository
- api compatibility & version

Please read commit history for more information.
Only some parts of the code use this platform, mostly auth Observables, users and version things.

Incrementally, we could move the organizations repositories into it.


libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  readonly apiVersion$ = this.settings$.pipe(
+    map((info) => info['system/platform/version'] as string),
+    shareReplay(1)
+  )
+
+  readonly isApiCompatible$ = this.apiVersion$.pipe(
+    tap(
+      (version) =>
+        version < minApiVersion &&
```
jahow, line 37: Does that really work for semver? e.g.  `4.10.0` might be considered "lower" than `4.2.0`

libs/common/domain/src/lib/model/user/user.model.ts
```diff
@@ -1,4 +1,4 @@
-export interface UserModel {
+export interface MeUserModel {
```
jahow, line 1: Why rename this?

jahow (commented):
Thanks :v:  I think this is going in the right direction. I would have liked the auth service to be gone completely though, but maybe this will take too much refactoring.

libs/common/domain/src/lib/platform.service.interface.ts
```diff
@@ -0,0 +1,16 @@
+import { Observable } from 'rxjs'
+import { MeUserModel, UserModel } from './model/user/user.model'
+import { Organization } from './model/record/organization.model'
+
+export abstract class PlatformServiceInterface {
+  abstract readonly apiVersion$: Observable<string>
+  abstract readonly isApiCompatible$: Observable<boolean>
```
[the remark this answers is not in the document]
fgravin: Ok

libs/common/domain/src/lib/platform.service.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  abstract getMe(): Observable<MeUserModel>
```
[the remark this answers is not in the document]
fgravin: I figured out that the `Me` response is quite different than the `User` response object from the User api of GN4.
The User API is more meant to return `UserModel`, while the Me api returns something quite specific for the user who is logged in.  Also, I would like that no lib depend son gn4 api data access objects, so the `Gn4ApiUserReponse` should not be the model used internally in gnui IMO.
I am opened to discussion though. 

As we discussed, it's hard to set a border between behind a bit backend agnostic, and having an editor which is purely stuck with GN concepts.

I wanted to create a pure domain model, trying to merge both GN4 models (Me & User), but wasn't sure the way to go, as I don't have enough background on how the users are used.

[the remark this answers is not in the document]
fgravin: Yes I do agree, I wasn't sure about the scope of both models.
Anyway, I've merged them into one `UserModel` for now.

libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
fgravin: Yes this is not correct, I will add the test of the version in another PR and keep this one as just refactoring.
I don't want that some heavy work is started and brings conflicts because of this refactoring.

libs/common/domain/src/lib/platform.service.interface.ts
```diff
@@ -0,0 +1,17 @@
+import { Observable } from 'rxjs'
+import { UserModel } from './model/user/user.model'
+import { Organization } from './model/record/organization.model'
+
+export abstract class PlatformServiceInterface {
+  abstract getTye(): string
```
jahow, line 6: typo ?

libs/common/domain/src/lib/platform.service.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+  abstract getApiVersion(): Observable<string>
+  abstract isApiCompatible(): Observable<boolean>
```
jahow, line 8: I'm wondering if this is really needed? In what scenario could we possibly want to still run the application with a non-compatible backend? Maybe we should just throw a "incompatible backend" error and stop running the app

libs/common/domain/src/lib/model/user/user.model.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface UserModela {
+  enabled?: boolean
+  emailAddresses?: Set<string>
+  organisation?: string
+  kind?: string
+  lastLoginDate?: string
+  accountNonExpired?: boolean
+  accountNonLocked?: boolean
+  credentialsNonExpired?: boolean
+}
```
jahow, line 21: Modela? Por favor?

libs/api/repository/src/lib/gn4/platform/gn4-platform.service.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  hash: 'girafe',
+  organisation: null,
+  admin: true,
+  groupsWithRegisteredUser: [],
+  groupsWithEditor: [],
+  groupsWithReviewer: [],
+  groupsWithUserAdmin: [],
+}
+
+const me$ = new Subject()
```
jahow, line 30: again, this pattern introduces the risk of side-effects between tests

libs/api/repository/src/lib/gn4/gn4.provider.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Gn4PlatformService } from './platform/gn4-platform.service'
+import { Gn4PlatformMapper } from './platform/gn4-platform.mapper'
+
+export function provideGn4(): Provider[] {
+  return [
+    {
+      provide: PlatformServiceInterface,
+      useClass: Gn4PlatformService,
+    },
+    Gn4PlatformMapper,
```
jahow, line 12: I don't see why this should be provided? Isn't it an explicit dependency of Gn4PlatformService?

libs/api/repository/src/lib/gn4/platform/gn4-platform.mapper.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  MeResponseApiModel,
+  UserApiModel,
+} from '@geonetwork-ui/data-access/gn4'
+import { UserModel } from '@geonetwork-ui/common/domain/model/user/user.model'
+import { Injectable } from '@angular/core'
+import { AvatarServiceInterface } from '../auth/avatar.service.interface'
+
+@Injectable()
+export class Gn4PlatformMapper {
```
jahow, line 10: I feel like these mappers could be simple pure functions exported as is, instead of an Angular service.

libs/api/repository/src/lib/gn4/platform/gn4-platform.mapper.ts
```diff
@@ -0,0 +1,42 @@
...
+      organisation,
+      kind,
+      lastLoginDate,
+      accountNonExpired,
+      accountNonLocked,
+      id,
+      credentialsNonExpired,
+      ...user
+    } = apiUser
+    return { ...apiUser, id: id + '' } as UserModel
```
jahow, line 40:
```suggestion
    return { ...apiUser, id: id.toString() } as UserModel
```

jahow (commented):
I did a first pass on the changes, most of it looks OK but I still stumbled on some possible improvements. I feel like it would have been a great opportunity to get rid of the AuthService as well, but maybe that was taking the refactoring too far?

libs/common/domain/src/lib/model/user/user.model.ts
fgravin: was to remove the dependence on the model :laughing:

libs/common/domain/src/lib/platform.service.interface.ts
fgravin: I'll treat that in the next PR where I will effectively test for the compatibility

libs/api/repository/src/lib/gn4/platform/gn4-platform.mapper.ts
fgravin: Yes I actually wanted to create a static class, but the mappers need for instance the `AvatarService` which is injected.

libs/api/repository/src/lib/gn4/gn4.provider.ts
fgravin: Because it's not injected in root, so it has to be injected somehow.
There is no `providers` option from services.

jahow (approved):
Let's go!

PR #705, opened 2023-11-28
Web component: provide gn4 platform service

fgravin:
Injection issues in web component demos


jahow:
We need E2E tests for these so much :sob:

PR #706, opened 2023-11-29
fix(DH): Prevent XSS attacks

Angi-Kinas:
This PR removes the SafePipe that bypasses angulars sanitization, that prevents JS injection that would execute code.

We think that we implented it in the first place to make the html tags be rendered correctly. But we cannot see any difference of the rendering behaviour without the SafePipe.

This should make the datahub more secure against XSS-Attacks.


fgravin (approved):
Perfect, thanks

PR #707, opened 2023-11-30
fix(DH): Change section titles

Angi-Kinas:
To make section titles more specific to what they show:

image


fgravin (approved):
If Lucile said that, then I guess it's ok.
Thanks

PR #708, opened 2023-11-30
Attempt to fix E2E flaky tests

jahow:
This PR adds a waiting time in the docker composition to make sure that records show up in GN before triggering the rest. 

Previously we were only waiting for GN to declare that indexing was finished, but it turns out there can be a few seconds between end of the indexation and the records being visible in the catalog. I suspect that was the cause of the failing tests


tkohr (approved):
Great find, @jahow, thanks ! I'm not a shell expert, but the idea LGTM and the e2e tests seem to run really well in the CI :-)

PR #711, opened 2023-12-04
Datahub: Fix display of news org thumbnail

tkohr:
PR display news org thumbnail as contain, which should give better results here, especially with wide logos:

before:
news_logo

after:
news_logo_contain


jahow (approved):
Thanks! no change to when no logo is available right?

PR #713, opened 2023-12-05
Set favicon default url to /favicon.ico and allow to customize it

f-necas:
The default favicon points to `/assets/favicon.ico`. As favicon is often provided in root, it's useful to have it there.

### DIFF
--- a/apps/datahub/src/app/app.component.ts
+++ b/apps/datahub/src/app/app.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit } from '@angular/core'
-import { getGlobalConfig } from '@geonetwork-ui/util/app-config'
+import { getThemeConfig } from '@geonetwork-ui/util/app-config'
+import { ThemeService } from '@geonetwork-ui/util/shared'
 
 @Component({
   selector: 'datahub-root',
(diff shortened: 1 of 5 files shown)


fgravin (commented):
Could you give more context ?
Who can't put the favicon in the assets ?
Note that it will break all actual configs.

jahow:
This is unfortunately susceptible to break all existing favicons. I would rather look for a way to customize the favicon path, for instance on the docker image startup why rewriting parts of the HTML, like we do for adding preload links here:
https://github.com/geonetwork/geonetwork-ui/blob/f46b998c18fc5db63319f25a497ae05b708ac49f/tools/docker/docker-entrypoint.sh#L38-L55

Outside of a docker usage, I feel like having a favicon point to the correct place should be easier.

f-necas:
@jahow Is it ok, if we remove it like this ?

jahow:
I'm sorry for all the back and forth on this issue, but I think we shouldn't merge it like that.

The way I see it currently, we should:
1. remove the favicon like in this PR
2. add an option in the config file to point to a place for a custom favicon if none is available at `/favicon.ico`; this is e.g. the case for www.geocat.ch, which gives a 404 for https://www.geocat.ch/favicon.ico

Otherwise we're losing functionality

f-necas:
@jahow I've added the possibility to implement it from toml file.

jahow (commented):
This looks better now, thanks! I've left some comments, please wait for 2.1 to be out before merging this.

I added a "breaking change" label so that we can remember that this new setting will be mandatory for people that don't have a favicon at their website's root. Please also update the PR title accordingly, so that it makes sense when it shows up in the changelog :slightly_smiling_face:  thanks!

jahow (approved):
All good thanks!

PR #717, opened 2023-12-08
[Datahub] Add search filter for organisations

tkohr:
PR adds a search filter for the organisations:

Todos: 

- [x] finish search-input: display clear only if value, no button for search icon, remove red background color, etc.
- [x] rename sort to filter component ?
- [x] i18n (lots of translations seem out of sync)
- [x] search user experience: should we really always apply OR filter on inputs? => impossible to narrow search down to one result => changed to AND after discussion
- [x] clarify why emails of orgs are erased by groups mapping => not sure if my fix is correct here => removed search in emails after discussion
- [x] tests

search_filter_orgs


libs/ui/catalog/src/lib/organisations-filter/organisations-filter.component.ts
```diff
@@ -33,8 +33,14 @@ export class OrganisationsSortComponent {
...
   ]
   @Output() sortBy = new EventEmitter<SortByField>()
+  @Output() filterByValueChange = new EventEmitter<string>()
+  @Output() filterBy = this.filterByValueChange.pipe(debounceTime(300))
 
   selectOrderToDisplay(selectValue: string) {
     this.sortBy.emit(selectValue.split(',') as SortByField)
   }
+
+  filterOrganisations(inputValue: string) {
```
fgravin, line 43: Note that the children output is an Observable, which you could reach with a `@ChildView` component ref.
Parent output would just be `childrenOutput.pipe(debounce(300)`, no event handling with the component, no event emission, just propagation.

fgravin (approved):
Excellent, I start to like the dynamic, clear UX and sharp implementation.
I am wondering if we should have a number of char limit, like 2 or 3 chars.

You just missed one thing, the margin between the search info and the results. It should be bigger, and aligned with the dataset tab.

:ok_hand:

libs/ui/catalog/src/lib/organisations-filter/organisations-filter.component.ts
```diff
@@ -33,8 +33,14 @@ export class OrganisationsSortComponent {
     },
   ]
   @Output() sortBy = new EventEmitter<SortByField>()
+  @Output() filterByValueChange = new EventEmitter<string>()
```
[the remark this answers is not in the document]
tkohr: Ah good catch, completely forgot about it when refactoring

libs/ui/catalog/src/lib/organisations-filter/organisations-filter.component.ts
tkohr: Interesting idea! I guess I'd need an `ngOnInit` for this to update the ouput, though, no?

As doing the following, the `@Output()`'s initialization accesses too early the `searchInput`:
```
@ViewChild('search') searchInput
@Output() filterBy = this.searchInput.valueChange.pipe(debounceTime(300))
```

tkohr:
Thanks for the review @fgravin ! I did some testing introducing a minimum number of chars, but as we don't launch requests here, I finally decided against it. This way the user immediately gets a feedback that something is happening, when he/she stops typing, which seems more intuitive to me.

libs/ui/catalog/src/lib/organisations-filter/organisations-filter.component.ts
fgravin: Yes, the `ViewChild` are only accessible `afterViewInit`.

PR #718, opened 2023-12-09
Update gn4 API client to 4.2.7

fgravin:
Chores

https://github.com/geonetwork/geonetwork-ui/blob/main/libs/data-access/gn4/src/README.md


libs/data-access/gn4/src/openapi/api/records.api.service.ts
```diff
@@ -23,7 +23,10 @@ import {
 import { CustomHttpParameterCodec } from '../encoder'
 import { Observable } from 'rxjs'
 
-import { BatchEditParameterApiModel } from '../model/models'
+import {
+  AssociatedRecordApiModel,
+  BatchEditParameterApiModel,
+} from '../model/models'
```
jahow, line 29: why is that change necessary? Having manual changes in there is going to bite us later if we keep updating the client automatically.

jahow (commented):
Thanks! it'd be better if there was no manual change required in there, is it because the openapi generator is not working correctly? Did you try updating it maybe? (theee's a new minor version available)

Also the e2e tests failure should be fixed before merging, it doesn't look like the usual flaky one.

fgravin:
> Also the e2e tests failure should be fixed before merging, it doesn't look like the usual flaky one.

Fixed in the related PR https://github.com/geonetwork/geonetwork-ui/pull/719/commits/7d3e1e112480501b3c57c7473ae4b6429d923f22

fgravin:
> Fixed in the related PR [7d3e1e1](https://github.com/geonetwork/geonetwork-ui/commit/7d3e1e112480501b3c57c7473ae4b6429d923f22)

Dependency PR merged and code rebased.
Ready for merge into `main` after that the CI passes.

jahow (approved):
thanks!

PR #719, opened 2023-12-09
Datahub: support thesaurus based advanced fields

fgravin:
Depends on https://github.com/geonetwork/geonetwork-ui/pull/718

- [x] Fix translations
- [x] Remove the dependency from Fields to `Gn4ToolApi` (for translation package)
- [x] Add thesaurus based Field
- [x] Add inspire theme field
- [ ] Add a resource contact field

The contact field will be handled in another PR, as I figured out that it is the same as the `OrganizationField` with the metadata strategy.


Angi-Kinas (approved):
Thank you for your help with this task, your solution looks a lot cleaner.
The PR looks good to me :+1:

apps/datahub/src/app/home/news-page/key-figures/key-figures.component.html
```diff
@@ -4,7 +4,7 @@
       class="py-[37px] pl-[47px] rounded-lg border bg-white mb-5 card-shadow cursor-pointer"
       [figure]="recordsCount$ | async"
       [icon]="'folder_open'"
-      [title]="'catalog.figures.datasets'"
+      title="catalog.figures.datasets"
```
jahow, line 7: why not use the `translate` pipe here, instead of `marker`? I now it's translated too in the gn-ui-figure component but amaybe that's the thing that doesn't make sense (translating inputs in presentation components mean the translation keys will not be discoverable)

libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
```diff
@@ -85,4 +100,19 @@ export class Gn4PlatformService implements PlatformServiceInterface {
   getUsers(): Observable<UserModel[]> {
     return this.users$
   }
+
+  translateKey(key: string): Observable<string> {
+    return this.keyTranslations$.pipe(map((translations) => translations[key]))
+  }
```
jahow, line 106: :+1:

PR #729, opened 2023-12-13
[UI]: Add component max-lines with toggle

Angi-Kinas:
This PR adds a new UI component "MaxLinesComponent" that takes the maximum number of lines that should be displayed before a "Read more" button shows up. If the content is less than the maximum number of lines, there will be no button.
This UI component also can be found in the Storybook.

- [x] Add component to Storybook
- [x] Add correct translation for "Read more" / "Read less"
- [x] Add tests
- [x] Add Screenshot in PR

image

image


fgravin (changes requested):
Excellent, your component looks great !
I've tested it and it works well.

However, there was a little misunderstanding in the specification (probably my fault): on the mockup, I realized that the keywords should be part of the extensible panel as well.
Do you think that your component could work with the keywords inside ?
I figured that out while I was looking for the margins below the "readMore" button.

Also,
- The text color of this button is `secondary`.
- If you animate when it expands, I guess you should animate when it collapse too.

Small changes hopefully :crossed_fingers: 
Thanks

fgravin (approved):
LGTM great work !
I've made some comment to Lucile about the hover color and the margin, but it can wait.
Please fix the CI and merge.
Thanks

PR #730, opened 2023-12-13
Attempt at fixing flaky E2E tests

jahow:
* Download file on click: the targeted file is now mocked so we won't be relying on an actual third-party endpoint
* WMS query params order was changed since the update of OpenLayers in #698, this PR fixes the interception pattern
* Sorting by date changes the results order: this test was failing due to a race condition, fixed in this PR
* Improved the login command to avoid race conditions (now everything coming after it will be chained instead of run in parallel)

I also simplified the JSON fixture which was quite big (hence the big diff)


apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -2,7 +2,6 @@ import 'cypress-real-events'
 
 describe('datasets', () => {
   beforeEach(() => {
-    cy.clearCookies()
```
tkohr, line 5: Where are the cookies cleaned now?

apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -445,15 +444,15 @@ describe('datasets', () => {
...
-            .should('not.eql', initialResultTitles)
-        })
+      it('changes the results order', function () {
+        cy.get('@results')
+          .find('[data-cy="recordTitle"]')
+          .then(($titles) => {
+            const titles = $titles
+              .toArray()
+              .map((title) => title.innerText.trim())
+            assert.notEqual(titles, this.initialResultTitles) // @initialResultTitles
```
tkohr, line 454: Why do we need to access `initialResultTitles` in the `this` scope here? I guess `.should('not.eql', @initialResultTitles)` is not the same?

tkohr (approved):
Thanks a lot for the various tweaks @jahow!

apps/datahub-e2e/src/e2e/datasets.cy.ts
jahow: cypress clears cookies between each run by default: https://docs.cypress.io/api/commands/clearcookies

apps/datahub-e2e/src/e2e/datasets.cy.ts
jahow: This is a subtlety of cypress: doing `cy.get('@myAlias')` will recompute the chain for the alias, whereas doing `this.myAlias` will return the value that was computed earlier. See:
https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Avoiding-the-use-of-this

tkohr: Ah, thanks for the explanations :-)

PR #735, opened 2023-12-19
Map context: fix WMTS layer extent

fgravin:
A the moment, the extent used to zoom on a WMTS layer is the tilegrid extent, extracted by Openlayers from the `getCapabilities` informations. It's generally the projection world extent.

The idea is to get the accurate extent of the layer from the capabilities itself, and use it if any.

**Sponsored by SwissTopo geocat.ch**


jahow (approved):
looks good enough for now, thanks :)

PR #736, opened 2023-12-19
Datahub: Propose external viewer button for geojson files

tkohr:
The PR proposes the external viewer button for geojson files as seen on the screenshot.

Can be tested on http://localhost:4200/dataset/accroche_velos with the following config:

```
geonetwork4_api_url = "https://mel.integration.apps.gs-fr-prod.camptocamp.com/geonetwork/srv/api"
proxy_path = "/dev-proxy?"
external_viewer_url_template = 'https://mel.integration.apps.gs-fr-prod.camptocamp.com/mapstore/#/?actions=[{"type":"CATALOG:ADD_LAYERS_FROM_CATALOGS","layers":["${layer_name}"],"sources":[{"url":"${service_url}","type":"${service_type}"}]}]'
```

Note that the deployed mapstore version needs to support this feature and the link url indicated in the metadata for the geojson file needs to allow CORS requests.

external_geojson

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -143,8 +143,8 @@ background_color = "#fdfbff"
 
 # Optional; URL template enabling to open map layers in an external viewer; if set, displays a button next to the map's layer drop down
 # The template must include the following placeholders, which allow the datahub to inject the correct values when adding a layer to a viewer:
-# ${service_url}: URL of the OWS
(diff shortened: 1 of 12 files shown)


fgravin (approved):
Excellent, thanks for adding those tests.
I've tested it and it works well !

PR #739, opened 2023-12-21
feat(ui): Adjust linkify directive

Angi-Kinas:
Adjust linkify directive to be able to process HTML elements and Text.
Will be used additionally in the abstract and landing page.

Replaces this PR https://github.com/geonetwork/geonetwork-ui/pull/588

/dataset/a8b5e6c0-c21d-4c32-b8f9-10830215890a (landing page link):
image

/dataset/n_tri_lill_inondable_s_059 (link in abstract with `<strong>`-tag):
image

/dataset/ee965118-2416-4d48-b07e-bbc696f002c2 (multiple links in abstract):
image

### DIFF
--- a/libs/ui/elements/src/lib/metadata-info/linkify.directive.spec.ts
+++ b/libs/ui/elements/src/lib/metadata-info/linkify.directive.spec.ts
@@ -52,11 +52,35 @@ const testingUrls = [
     'http://foo.com/(something)?after=parens',
   ],
 ]
+
+const testWithMultipleUrls = {
+  input:
+    'Fourteenth links http://foo.com/(something)?after=parens with multiple links http://foo.com/(something)?before=multiple',
+  output: [
+    'http://foo.com/(something)?after=parens',
+    'http://foo.com/(something)?before=multiple',
+  ],
+}
+
+const testWithHTML = {
+  input:
+    '<p>Fourteenth link with html input <a href="http://foo.com/(something)?after=before">This is the display text</a> query params</p>',
+  output: 'http://foo.com/(something)?after=before',
+}
+
 @Component({
-  template: `<div [gnUiLinkify]>{{ text }}</div>`,
+  template: `<div
+      *ngIf="customInnerHTML"
+      [innerHTML]="customInnerHTML"
+      [gnUiLinkify]
+    ></div>
+    <div *ngIf="!customInnerHTML" [gnUiLinkify]>
+      {{ text }}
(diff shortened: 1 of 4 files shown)


tkohr (approved):
Really nice improvement, thanks @Angi-Kinas ! The only thing I see, is that the link color in the abstract should likely also be primary, shouldn't it? Maybe we could also add a test for some HTML input with an existing anchor element.

jahow:
> The only thing I see, is that the link color in the abstract should likely also be primary, shouldn't it?

Really what we're missing is a presentation component for links. We have been copy-pasting tailwind style for these many times now, and this should be factorized in a component.

tkohr:
> > The only thing I see, is that the link color in the abstract should likely also be primary, shouldn't it?
> 
> Really what we're missing is a presentation component for links. We have been copy-pasting tailwind style for these many times now, and this should be factorized in a component.

Shouldn't [this](https://github.com/geonetwork/geonetwork-ui/pull/739/files#diff-53dad72f2fd52572dd8a39e040b3d58dd4df018362a1d97541fea7e7c2474635R78) be the central place where the style should be defined? The only reason why it's displayed differently in the abstract, is because the style is overwritten [here](https://github.com/geonetwork/geonetwork-ui/blob/DH-make-hyperlinks-clickable/libs/ui/elements/src/lib/metadata-info/metadata-info.component.css#L2-L4), I think.

Maybe I'm missing your point for the need of a component though. Do you propose to use the UI component within the directive?

jahow:
> Shouldn't [this](https://github.com/geonetwork/geonetwork-ui/pull/739/files#diff-53dad72f2fd52572dd8a39e040b3d58dd4df018362a1d97541fea7e7c2474635R78) be the central place where the style should be defined? The only reason why it's displayed differently in the abstract, is because the style is overwritten [here](https://github.com/geonetwork/geonetwork-ui/blob/DH-make-hyperlinks-clickable/libs/ui/elements/src/lib/metadata-info/metadata-info.component.css#L2-L4), I think.

Hmm, I think it'd be good to have a common style for all links in the app, not just the ones created by the linkify directive. But it's also true that it might not necessary be a presentation component, it could simply be a tailwind class, otherwise it'll be hard to generate it on the fly in the linkify directive.

jahow:
I've reworked this PR and now the tests should be passing :) great work @Angi-Kinas, it was only a minor detail (see the template of component used in the unit test).

PR #740, opened 2023-12-22
[Storybooj][Fix] : Add accessServiceProtocol to link fixture

cmoinier:
Adding missing property for `ogcApiFormat` element in link fixtures.


tkohr (approved):
Thanks for the fix @cmoinier!

PR #741, opened 2023-12-22
[DH] Metadata-Quality widget update

f-necas:
# Metadata Quality widget simplifier

Removes config parameters for Metadata Quality widget.

## Changes 

- Removes `SORTABLE`,`DISPLAY_WIDGET_IN_DETAIL`,`DISPLAY_WIDGET_IN_SEARCH`,`DISPLAY_TITLE`,`DISPLAY_DESCRIPTION`,`DISPLAY_TOPIC`,`DISPLAY_KEYWORDS`,`DISPLAY_LEGAL_CONSTRAINTS`,`DISPLAY_CONTACT`,`DISPLAY_UPDATE_FREQUENCY`,`DISPLAY_ORGANISATION` from Toml configuration file
- Hide widget if pipeline is not installed and display warning in console
- Update icons

## TODO

- [ ] Documentation

### DIFF
--- a/apps/datahub-e2e/src/e2e/datasets.cy.ts
+++ b/apps/datahub-e2e/src/e2e/datasets.cy.ts
@@ -470,43 +470,24 @@ describe('datasets', () => {
 
     describe('metadata quality widget enabled', () => {
       beforeEach(() => {
-        // this will enable spatial filtering
+        // this will enable metadata quality widget
         cy.intercept('GET', '/assets/configuration/default.toml', {
           fixture: 'config-with-metadata-quality.toml',
         })
         cy.visit('/search')
       })
-      it('should not display widget', () => {
-        cy.get('@results')
-          .eq(2)
-          .get('gn-ui-metadata-quality')
-          .should('not.exist')
-      })
-
-      it('should reindex records', () => {
-        cy.login('admin', 'admin', false)
-        cy.visit(
-          `http://localhost:8080/geonetwork/srv/eng/admin.console#/tools`
-        )
-        cy.intercept({
-          method: 'PUT',
-          url: 'http://localhost:8080/geonetwork/srv/api/site/index?reset=false&asynchronous=true',
-        }).as('indexingRecordsXHR')
-        cy.get('[class=panel-body]').find('button').first().click()
-        cy.wait('@indexingRecordsXHR')
-      })
 
       it('should display quality widget', () => {
-        cy.visit('/search')
+        cy.get('@sortBy').selectDropdownOption('desc,createDate')
         cy.get('gn-ui-progress-bar')
-          .eq(2)
+          .eq(0)
           .should('have.attr', 'ng-reflect-value', 87)
       })
 
       it('should display results sorted by quality score', () => {
         cy.get('@sortBy').selectDropdownOption('desc,qualityScore')
         cy.get('gn-ui-progress-bar')
-          .eq(2)
+          .eq(0)
           .should('have.attr', 'ng-reflect-value', 100)
       })
     })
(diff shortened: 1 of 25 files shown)


jahow (approved):
Thanks!! I think the customization options for the quality will be brought back later on but in a more flexible way. For now this should be more than enough.

Do you think it's possible to add an end-to-end test for the quality indicator? There's already a mechanism in the datahub e2e tests to mock a configuration, it could be worthwile to mock a config with the md quality enabled and check whether the widgets show up and the sorting work. This might necessitate running the pipelines on the support services docker compo as well.

This can be done in another PR though.

support-services/docker-compose.yml
```diff
@@ -140,3 +140,12 @@ services:
...
     ports:
       - '8081:8080'
+
+  tools-pipelines:
+    image: geonetwork/geonetwork-ui-tools-pipelines:latest
+    environment:
+      ES_HOST: http://elasticsearch:9200
+      RECORDS_INDEX: gn-records
+    depends_on:
+      init:
```
jahow, line 150: Do you think you could invert the dependency, and make `init` depend on this? `init` will do a reindexation of the catalog so that would mean the md quality widget would show up right away without needing a manual indexation

apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -456,4 +456,59 @@ describe('datasets', () => {
...
+        cy.get('.cdk-overlay-container')
+          .find('[role=listbox]')
+          .find('button')
+          .should('have.length', 3)
+      })
+    })
+
+    describe('metadata quality widget enabled', () => {
+      beforeEach(() => {
+        // this will enable spatial filtering
```
jahow, line 473:
```suggestion
        // this will enable metadata quality
```

apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -456,4 +456,59 @@ describe('datasets', () => {
...
+        cy.visit(
+          `http://localhost:8080/geonetwork/srv/eng/admin.console#/tools`
+        )
+        cy.intercept({
+          method: 'PUT',
+          url: 'http://localhost:8080/geonetwork/srv/api/site/index?reset=false&asynchronous=true',
+        }).as('indexingRecordsXHR')
+        cy.get('[class=panel-body]').find('button').first().click()
+        cy.wait('@indexingRecordsXHR')
+      })
```
jahow, line 497: See my comment on the docker composition; also this check here is unfortunately not enough to guarantee that all records are indeed indexed correctly, which is why I added a more robust logic in the init container. See:
https://github.com/geonetwork/geonetwork-ui/blob/ab94f4f9791994a26bea36bbfb7a4903a0357290/support-services/docker-entrypoint.d/10-index-records.sh#L50-L65

jahow (commented):
Added a comment on the docker composition, thanks for adding an e2e test!!

apps/datahub-e2e/src/e2e/datasets.cy.ts
f-necas: I don't succeed to make it working that way. The `/index` curl drops index and re-index by default : 
image

If I set a `reset=false` query param, it doesn't work (400 on datahub). Any idea ?

jahow: > If I set a `reset=false` query param, it doesn't work (400 on datahub). Any idea ?

Maybe that's because the index wasn't created yet? I'm not sure. You can try using `reset=false&asynchronous=true` since is what you did in the e2e test.

jahow: I pushed a commit trying to address that, let's see if the CI succeeds

Issue #733, opened 2023-12-18
[IGN] [Datahub] Add markdown support on metadata-info component


cdebarros:
Hi, another feature i'd like to introduce.

We are going to use markdown as our metadata formatted text language (only the descriptive part). 
So we need somehow to implement a markdown parser like [marked](https://marked.js.org/) to render markdown content as html. 

Here is the kind of result i achieved to get.

Before: 
![Capture d’écran du 2023-12-18 17-48-09](https://github.com/geonetwork/geonetwork-ui/assets/45257663/2c9a1f70-e8e4-4853-b9b4-d983bb103cdd)

Before: 
![Capture d’écran du 2023-12-18 17-48-36](https://github.com/geonetwork/geonetwork-ui/assets/45257663/26618000-c343-4562-aa86-28b41e66483a)

It can be an optional setting in the toml file settings. And maybe extend it to others formatted text language ?

jahow:
I completely agree on using Markdown as the format for rich text fields (abstract, lineage etc.).

Currently the Datahub has a very simplistic approach, it will render the content of the abstract as HTML, so if there are HTML tags in it they will show up correctly. One thing that I was considering was to convert all HTML tags to Markdown at some point in the process, and have the metadata page simply render Markdown and not worry about the rest.

Other formatting languages can include RST but this is too elaborate, and I think Markdown is the right compromise as a common language.

cdebarros:
I did this feature, if i can contribute somehow.

I created a new ui-elements component to parse markdown content.
I used [marked](https://marked.js.org/) to parse the content and [github-markdown-css](https://github.com/sindresorhus/github-markdown-css) to style elements.

Do i need some specific rights to push a new branch ? (or maybe should i fork the project ?)

jahow:
Please fork the project and open a Pull Request through Github, then a reviewer can examine your contribution and give you feedback. Thank you very much!

Also, please note that this change is not trivial since simply switching to MD instead of rendering HTML will break the appearance of many records that rely on HTML-based content (which the datahub currently handles).

cdebarros:
Thanks for your feedback, i will fork the project and open a pull request soon (today or in january after holidays).

From what i've tested, [marked](https://marked.js.org/) can take markdown and html content as input: 
- html content will render as it is, with no changes
- markdown content is converted and rendered as html

The thing that can change, is the way we handle styling. 
[...]

PR #746, opened 2024-01-02
add ui element markdown-parser and markdown support on metadata…

cdebarros:
Hi, first of all happy new year to everyone seeing this. Here is my first contribution to the project.

**This pull request makes the following changes :**
* Fixes isssue #733 

I used `marked` to parse markdown content to html and then i defined some basic style for the following elements as suggested by @jahow  :

- h1 to h6
- ul, ol and li
- p
- a
- blockquote
- pre - code
- hr
- table
- img

I tried to use tailwind base colors to make it close to the default style. 
I'm not sure how should i render links, blockquotes or table. I'll wait for your feedbacks to make some changes if needed.

### DIFF
--- a/libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
+++ b/libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
@@ -1,0 +1,271 @@
+/** Body **/
+.markdown-body {
+  -ms-text-size-adjust: 100%;
+  -webkit-text-size-adjust: 100%;
+  margin: 0px 0px 1.5rem 0px;
+  color: var(--color-main);
+  font-size: 16px;
+  line-height: 1.5;
+  word-wrap: break-word;
+}
+
+/** Emphasis **/
+
+.markdown-body strong {
+  font-weight: 600;
+  color: var(--color-secondary-darker);
+}
+
+/** Headings **/
+
+.markdown-body h1,
+.markdown-body h2,
+.markdown-body h3,
+.markdown-body h4,
+.markdown-body h5,
+.markdown-body h6 {
+  margin-top: 24px;
+  margin-bottom: 16px;
+  font-weight: 600;
+  line-height: 1.25;
+}
+
+.markdown-body h1 {
+  margin: 0.67em 0;
+  font-weight: 600;
+  padding-bottom: 0.3em;
+  font-size: 2em;
+  color: var(--color-primary);
+}
+
+.markdown-body h2 {
+  font-weight: 600;
+  padding-bottom: 0.3em;
+  font-size: 1.5em;
+  color: var(--color-secondary);
+}
+
+.markdown-body h3 {
+  font-weight: 600;
+  font-size: 1.25em;
+  color: var(--color-secondary);
+}
+
+.markdown-body h4 {
+  font-weight: 600;
+  font-size: 1em;
+  color: var(--color-secondary);
+}
+
+.markdown-body h5 {
+  font-weight: 600;
+  font-size: 0.875em;
+  color: var(--color-secondary);
+}
+
+.markdown-body h6 {
+  font-weight: 600;
+  font-size: 0.85em;
+  color: var(--color-secondary-lighter);
+}
+
+/** Paragraphs **/
+
+.markdown-body p {
+  margin-top: 0;
+  margin-bottom: 10px;
+}
+
+/** Links **/
+
+.markdown-body p > a {
+  margin-top: 0;
+  margin-bottom: 10px;
+  color: var(--color-primary) !important;
+  text-decoration: none !important;
+  font-weight: 600;
+}
+
+.markdown-body p > a:hover {
+  color: var(--color-primary-darker) !important;
+}
+
+/** Blockquotes **/
+
+.markdown-body blockquote {
+  margin: 0;
+  padding: 0 1em;
+  color: var(--color-secondary-lighter);
+  border-left: 0.25em solid var(--color-primary-lighter);
+}
+
+/** Code **/
+
+.markdown-body pre {
+  margin-top: 0;
+  margin-bottom: 0;
+  font-size: 12px;
+  background-color: var(--color-gray-100);
+  word-wrap: normal;
+}
+
+.markdown-body pre {
+  padding: 16px;
+  overflow: auto;
+  font-size: 85%;
+  line-height: 1.45;
+  color: var(--color-secondary);
+  border-radius: 6px;
+}
+
+.markdown-body code {
+  padding: 0.2em 0.4em;
+  margin: 0;
+  font-size: 85%;
+  white-space: break-spaces;
+  border-radius: 6px;
+}
+
+.markdown-body pre code {
+  display: inline;
+  max-width: auto;
+  padding: 0;
+  margin: 0;
+  overflow: visible;
+  line-height: inherit;
+  word-wrap: normal;
+  border: 0;
+}
+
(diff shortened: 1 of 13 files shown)


fgravin (commented):
Thanks @cdebarros for this first PR.

I need to wrap up the approach a bit with @jahow but as a first review, I would just point out that a component seems not appropriate for such a process.

I would have more seen a Pipe and/or a directive.

Also, I wonder if the css could be encapsulated somehow instead of being spread via the `encapsulation: ViewEncapsulation.None` option.

jahow:
> I need to wrap up the approach a bit with @jahow but as a first review, I would just point out that a component seems not appropriate for such a process.
> 
> I would have more seen a Pipe and/or a directive.

What would be an argument for using a pipe or directive instead of a component? I'm still unsure which solution is the best personally. Using a component has the advantage that we can use an encapsulated style inside it.

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
```diff
@@ -0,0 +1,271 @@
...
+  line-height: 1.25;
+}
+
+.markdown-body h1 {
+  margin: 0.67em 0;
+  font-weight: 600;
+  padding-bottom: 0.3em;
+  font-size: 2em;
+  color: var(--color-primary);
+}
```
jahow, line 39: The title styles currently are inconsistent with other titles, see:
image

I would suggest using `@apply text-title font-title` for all `hx` classes above

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core'
+import { marked } from 'marked'
+
+@Component({
+  selector: 'gn-ui-markdown-parser',
+  templateUrl: './markdown-parser.component.html',
+  styleUrls: ['./markdown-parser.component.css'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class MarkdownParserComponent {
```
jahow, line 10: We usually try to add presentation components to storybook, do you think you could do the same with this one?
I suggest you take inspiration from other components such as this one: https://github.com/geonetwork/geonetwork-ui/blob/main/libs/ui/elements/src/lib/metadata-info/linkify.directive.stories.ts

The goal is to let people easily "try" the component and see its behaviour with varying inputs, and also making sure that it doesn't break after modifications.

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core'
+import { marked } from 'marked'
+
+@Component({
+  selector: 'gn-ui-markdown-parser',
+  templateUrl: './markdown-parser.component.html',
+  styleUrls: ['./markdown-parser.component.css'],
+  encapsulation: ViewEncapsulation.None,
```
jahow, line 8: you can get rid of this and leave the default view encapsulation by adding `:host /deep/` before all classes in the CSS file; this is a bit more verbose but IMO essential since disabling view encapsulation is risky in the long run

jahow (commented):
Thanks for this contribution! I think this looks really good. I added a few comments, specifically to avoid disabling the view encapsulation; once addressed I think this can be merged. Please let me know if you need any more information!

@fgravin I've thought about it again and using a component seems right to me; please let us know if you have arguments against it

jahow:
Note: Snyk workflow always fails for forks; please disregard that

cdebarros:
Thanks for your reviews ! 
i will try to make these changes when i have time, probably before the end of this week.

cdebarros:
I just updated my contribution. Added css encapsulation and storybook for the component.

On the default storybook content, i explain which markdown elements are supported for now.

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    }).compileComponents()
+
+    fixture = TestBed.createComponent(MarkdownParserComponent)
+    component = fixture.componentInstance
+    fixture.detectChanges()
+  })
+
+  it('should create', () => {
+    expect(component).toBeTruthy()
+  })
```
jahow, line 20: Oh sorry, I didn't realize that in the first review but we we need at least one test here, showing that a markdown input will be correctly transformed to HTML; this doesn't have to be exhaustive as we're not testing the underlying library, but simply making sure that the test is processed.

Could you please add that?

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.ts
```diff
@@ -0,0 +1,15 @@
+import { Component, Input } from '@angular/core'
+import { marked } from 'marked'
+
+@Component({
+  selector: 'gn-ui-markdown-parser',
+  templateUrl: './markdown-parser.component.html',
+  styleUrls: ['./markdown-parser.component.css'],
```
jahow, line 7: I also forgot to point out that we should always use `changeDetection: ChangeDetectionStrategy.OnPush,` to make sure that the change detection does not run too often

jahow (commented):
Thanks for the adaptations, I think this looks better now. There are choices in the styling of the output that might not fit for every occasion and theme, but for now we can keep it like that and see how it behaves in the "wild".

I realized there was no unit test, I missed that on the first review. Could you please add one that ensures that the input is processed into the div inner HTML?

Also the branch needs to be rebased; there will likely be a conflict because we added `gnUiLinkify` on the abstract in the meantime; this directive can be removed from the abstract since `marked` does the same thing :)

Thank you for your continued efforts

libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.spec.ts
cdebarros: Oh, i totally forgot about that.

I created 3 tests described like this :
- should render markdown as HTML
- should render HTML as HTML
- should render text as HTML 

Is it too much ? Maybe the first test is enough.

jahow: That does look like a perfect amount of tests :slightly_smiling_face:

cdebarros:
I added tests and the changeDetectionStrategy to the component. Also i rebased the branch.
Let me know if it needs other changes.

jahow (approved):
Thanks for this great contribution! There is a failure on the build attempts in the CI, to fix it you can bump the CSS budgets for all apps like so:
```diff
diff --git a/apps/datafeeder/project.json b/apps/datafeeder/project.json
index 67998c55..b22bc758 100644
--- a/apps/datafeeder/project.json
+++ b/apps/datafeeder/project.json
@@ -40,8 +40,8 @@
             },
             {
               "type": "anyComponentStyle",
-              "maximumWarning": "2kb",
-              "maximumError": "4kb"
+              "maximumWarning": "10kb",
+              "maximumError": "20kb"
             }
           ],
           "fileReplacements": [
```

cdebarros:
I updated it for the 3 projects that have failed, it should be fine now.

jahow:
Don't worry about the CSS warnings, I'll fix them in another PR.

Thank you again for this awesome contribution, congratulations on it and we're looking forward to more of this!! :tada:

PR #747, opened 2024-01-02
Add contact block to news page and improve buttons

f-necas:
# Contact block & buttons
## Contact

Add a contact block with a configurable mail in `default.toml`
image
 
### How-to

Add mail in `default.toml`

## Buttons improvement

As `gn-ui-button` isn't really necessary, this PR also propose to improve them with tailwind classes in `tailwind.base.css`.

This is the first (without breaking changes) step to get rid of `gn-ui-button`.


libs/ui/elements/src/lib/downloads-list/downloads-list.component.spec.ts
```diff
@@ -132,9 +132,6 @@ describe('DownloadsListComponent', () => {
         LINK_FIXTURES.geodataShpWithMimeType
       )
       expect(items[0].componentInstance.format).toEqual('shp')
-      expect(items[0].componentInstance.color).toEqual(
```
Angi-Kinas, line 135: Is this change necessary in this PR? I'm just wondering why this got removed.

Angi-Kinas (commented):
Great work! I like the refactor of the button logic! Thanks @f-necas

libs/ui/elements/src/lib/downloads-list/downloads-list.component.spec.ts
f-necas: It is because the color getter is deleted. So we can't call it anymore :  https://github.com/geonetwork/geonetwork-ui/pull/747/files#diff-5cfed937e30ec743c98e5e024350945ee6d4d0b6e7e34dabac9f3fb580fa0ad4L27-L28

conf/default.toml
```diff
@@ -34,6 +34,9 @@ proxy_path = ""
 # More information about the translation can be found in the docs (https://geonetwork.github.io/geonetwork-ui/main/docs/reference/i18n.html)
 # languages = ['en', 'fr', 'de']
 
+# Allows displaying a contact email on news page for missing data
```
jahow, line 37:
```suggestion
# Enables displaying a "contact block" wherever relevant in applications
```

libs/ui/elements/src/lib/downloads-list/downloads-list.component.spec.ts
jahow: that shouldn't be impacted here, unless I'm missing something; in general we should strive to avoid reducing the test coverage :)

f-necas: Oh yeah indeed, I don't know why my IDE displayed it in red. Sry @Angi-Kinas for the misunderstood !

Angi-Kinas (approved):
Thanks for the rework of the button component. It reduces the logic and complexity :-) :+1: 
I'll approve the PR, but it should not be merged yet because of the new release of the 2.1.0 version.

PR #748, opened 2024-01-04
feature(DH): Add more contact information

Angi-Kinas:
This PR adds more contact information in the metadata contact section if it exists.

It will look like this:
image

image

image


libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.spec.ts
```diff
@@ -80,4 +106,14 @@ describe('MetadataContactComponent', () => {
       expect(a.attributes.target).toBe('_blank')
     })
   })
+
+  describe('#parseAddress', () => {
+    test.each(addressCases)(
+      'adressString: $address',
+      ({ address, expectedResult }) => {
+        console.log(component)
```
cmoinier, line 114: Forgot to delete this console.log :slightly_smiling_face:

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.ts
```diff
@@ -34,6 +34,37 @@ export class MetadataContactComponent {
     )
   }
 
+  parseAddress(inputAddress) {
+    const addressParts = inputAddress.split(',').map((part) => part.trim())
```
cmoinier, line 38: The only thing that worries me is if one address doesn't have any commas (as we spoke about). Is this a case that can happen? @fgravin

cmoinier (changes requested):
Thanks for the work :slightly_smiling_face: 
I tested and it works well with all of my datasets.

Only UI thing to fix IMO is the spacing between the address and the email in case there is no website, i.e. without website (too narrow) :

image

and with website : 

image

I also would like to wait for a second opinion about the risk of having addresses without commas (if there is no risk then I will approve the PR) :slightly_smiling_face:

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -29,4 +29,44 @@
...
+          {{ contacts[0].firstName }} {{ contacts[0].lastName }}
+        </p>
+        <p class="text-gray-700 text-sm">{{ contacts[0].phone }}</p>
+      </div>
+    </div>
+  </div>
+  <div *ngIf="contacts[0].address" class="mb-2">
+    <div class="flex">
+      <mat-icon
+        class="material-symbols-outlined !w-[20px] !h-[20px] !text-[20px] opacity-75"
```
fgravin, line 49: Please use built-in classes like `w-5` instead, when possible.

fgravin (commented):
A similar work has been started in https://github.com/geonetwork/geonetwork-ui/pull/575
Dunno if it helps.

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
+        class="contact-website text-primary text-sm cursor-pointer hover:underline transition-all"
+        >{{ shownOrganization.website }}
+        <mat-icon
+          class="material-symbols-outlined !w-[12px] !h-[12px] !text-[12px] opacity-75 shrink-0"
+          >open_in_new</mat-icon
+        >
+      </a>
+    </div>
+  </div>
+  <div *ngIf="contacts[0].phone" class="mb-5">
```
jahow, line 36:
```suggestion
  <div *ngIf="contacts[0]?.phone" class="mb-5">
```

Otherwise JS errors pop in case of a missing contact

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
-    target="_blank"
-    >{{ contacts[0].email }}</a
-  >
-  <div *ngIf="shownOrganization.website" class="mb-2">
-    <a
-      [href]="shownOrganization.website"
-      target="_blank"
-      class="contact-website text-primary text-sm cursor-pointer hover:underline transition-all"
-      >{{ shownOrganization.website }}
+  <div *ngIf="contacts[0].email" class="mb-5">
```
jahow, line 47:
```suggestion
  <div *ngIf="contacts[0]?.email" class="mb-5">
```

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
+        *ngIf="contacts.length"
+        [href]="'mailto:' + contacts[0].email"
+        class="text-sm hover:underline ml-2"
+        target="_blank"
+        data-cy="contact-email"
+        >{{ contacts[0].email }}</a
+      >
+    </div>
+  </div>
+  <div *ngIf="contacts[0].firstName || contacts[0].lastName" class="mb-5">
```
jahow, line 64:
```suggestion
  <div *ngIf="contacts[0]?.firstName || contacts[0]?.lastName" class="mb-5">
```

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
+    <div class="flex">
+      <mat-icon
+        class="material-symbols-outlined !w-5 !h-5 !text-[20px] opacity-75 shrink-0"
+        >person_outline</mat-icon
+      >
+      <div class="flex flex-col ml-2">
+        <p class="text-sm">
+          {{ contacts[0].firstName }} {{ contacts[0].lastName }}
```
jahow, line 72: This might render weirdly if only one of the two is defined

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
+        </p>
+      </div>
+    </div>
+  </div>
+  <div *ngIf="contacts[0].address" class="mb-2">
```
jahow, line 77:
```suggestion
  <div *ngIf="contacts[0]?.address" class="mb-2">
```

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
-  </p>
-  <div class="flex mb-1">
+<div class="py-4 px-5 rounded bg-gray-100 text-black">
+  <div *ngIf="shownOrganization" class="mb-6 sm:mb-12">
+    <p class="text-sm mb-3 font-medium" translate>record.metadata.contact</p>
     <div
-      class="text-primary font-title text-21 mr-2 cursor-pointer hover:underline"
-      (click)="onOrganizationClick()"
+      *ngIf="contacts[0].organization?.logoUrl?.href"
+      class="mb-3 flex items-center justify-center border-solid border border-gray-300 rounded-md bg-white h-32"
```
jahow, line 6: this needs an `overflow-hidden` class otherwise there is a "cut edges" effect on logos, see:
image

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -1,32 +1,94 @@
...
+      >
+        location_on</mat-icon
+      >
+      <div class="flex flex-col ml-2">
+        <p class="text-sm">
+          {{ address[0] }}
+        </p>
+        <p *ngFor="let addressPart of address.slice(1)" class="text-sm">
+          {{ addressPart }}
+        </p>
```
jahow, line 90: why would we need to split the logic here? having `ngFor` loop on all parts would be enough I guess

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.ts
jahow: yes definitely, there's no rule for addresses as far as I know; but then it'll simply end up being a long line, so we just need to make sure that the text overflow works well :) (looks to be the case in my tests)

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.ts
```diff
@@ -34,6 +34,37 @@ export class MetadataContactComponent {
...
+        const postcodeCity = `${part} ${addressParts[i - 1]}`
+        // delete duplicate city
+        if (postcodeCity.includes(addressParts[i - 1])) {
+          addressArray.pop()
+        }
+        addressArray.push(postcodeCity)
+      } else {
+        // Treat as a separate line
+        addressArray.push(part)
+      }
```
jahow, line 58: I'm not sure about this logic; it makes a lot of assumptions about the structure of an address, and honestly I don't think we can assume that much as address fields can often be filled completely freely by metadata authors; e.g. assuming that "CS" will be present before a postcode (which is specific to France I guess), assuming that the city name comes before the postal code...

Relying on the "," separator is enough IMO, and putting each part in a separate line.

jahow (commented):
Added some more comments, thanks for the contribution!

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.ts
jahow: On the other hand it looks like the GN editor reliably produces this kind of output... but that might change anytime in the future

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.ts
Angi-Kinas: Okay in this case I will keep it simple as you were proposing @jahow and only split by ',' and if that's not there, it will be a long line of text which overflows, as you said. I implemented it to fit the design, but I guess there are too many different variations of the user input.

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
Angi-Kinas: I think now it should be handled correctly.

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
Angi-Kinas: Good catch! Thanks :)

jahow (approved):
Thanks for the improvements! As discussed, the layout can probably be slightly improved using `grid` and `gap` instead of `mb` classes. I'm approving this, please merge when you feel this is ready!

PR #752, opened 2024-01-08
[Datahub] Fix : add missing tooltip

cmoinier:
This PR adds a tooltip for the `gn-ui-link-cards` and `gn-ui-api-cards`, to display titles that are too long for the cards.

image


jahow (approved):
Thanks!! :pray:

PR #753, opened 2024-01-08
Datahub: Fix organisation search to work with pagination

tkohr:
Currently, organisation search results are not displayed once the user navigates to a page > 1. This PR ensures displaying the search results by going back to the first page. The pagination is not impacted when changing sort order.

Bug before the fix:
org_filter_paginate


jahow (approved):
Very good catch, thanks!!

PR #754, opened 2024-01-08
[Datahub] Fix : Improve linkify REGEX 

cmoinier:
This PR improves the Linkify directive's REGEX to avoid including end parenthesis.

--> End parenthesis used to be detected as part of the URLS, ex : 

image-2024-01-08-12-04-25-004

--> Now they're not : 

image


jahow (approved):
Very well done, regexes are tricky :)

PR #758, opened 2024-01-10
Automatically publish NPM package on every commit on main branch

jahow:
This should also publish the package on every release, as well as attempt a build of the npm package on every PR to make sure that it doesn't break.

PR #760, opened 2024-01-11
Update declared dependencies of NPM package

jahow:
Follow-up of #758 

* Update OL and add a few missing dependencies according to the root package.json
* Remove nguniversal from peer dependencies as we don't do SSR in custom apps
* List new non-peer deps as allowed in ng-package.json


fgravin (commented):
What does it mean exactly ?
We need to maintain those version manually ?
What happens if the versions are not the same ?

jahow:
yes currently it's a manual process; the dependencies in `package/package.json` will be the ones installed in the custom application when doing `npm install`, so if they're out of date then gn-ui features might not work correctly.

fgravin (approved):

PR #761, opened 2024-01-11
feat(DH): Add basiclightbox for thumbnail preview

Angi-Kinas:
This PR adds a thumbnail preview to the dataset details page:

image

image


apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -33,13 +33,38 @@
...
+          *ngIf="lightbox$ | async as lightbox"
+          data-cy="record-thumbnail"
+          class="flex-shrink-0 bg-gray-100 rounded-lg overflow-hidden w-full border border-gray-300 h-36 group-hover:shadow-xl group-hover:border-0 mb-3"
         >
-        </gn-ui-metadata-contact>
-        <gn-ui-metadata-catalog [sourceLabel]="sourceLabel$ | async">
-        </gn-ui-metadata-catalog>
+          <gn-ui-thumbnail
+            class="relative h-full w-full"
+            [thumbnailUrl]="(facade.metadata$ | async).overviews?.[0]?.url"
```
tkohr, line 43: Looks like the same logic as for `lightbox$`. Maybe both could be resolved with one variable (eg. `thumbnailUrl` or `overviewUrl`) on this level, which can then be passed to different UI components?

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -33,13 +33,38 @@
...
+              [extraClass]="'!py-2 !px-0'"
+              (buttonClick)="openLightbox(lightbox)"
+            >
+              <mat-icon class="material-symbols-outlined font-extralight"
+                >zoom_out_map</mat-icon
+              >
+            </gn-ui-button>
+          </div>
+        </div>
+        <div>
```
tkohr, line 59: I think it would be nice to have a separate ui component for this (which would also encapsulate the lightbox feature).

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -77,4 +83,9 @@ export class RecordMetadataComponent {
       .getFiltersForOrgs([org])
       .subscribe((filters) => this.searchService.updateFilters(filters))
   }
+
+  openLightbox(src: string) {
+    this.lightboxInstance = basicLightbox.create(`<img src="${src}"/>`)
+    this.lightboxInstance.show()
```
tkohr, line 89: Wondering if we could just do `basicLightbox.create(``).show()` and get rid of the `lightboxInstance: any`?

tkohr (commented):
Thanks for this neat feature @Angi-Kinas. I left some suggestions inline. I think it would also be nice to have a little test for this (maybe e2e?).

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
Angi-Kinas: Yes this works :+1:

libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
```diff
@@ -85,6 +88,7 @@ export class ThumbnailComponent implements OnInit, OnChanges {
   private setNewSrcImage(image: ThumbnailImageObject) {
     this.imgFit = image.fit
     this.imgUrl = image.url
+    this.placeholderShown.emit(this.isPlaceholder)
```
tkohr, line 91: Just wondering why this is not inside `setPlaceholder()` always returning `true`.

tkohr (commented):
Thanks for the refactoring and error handling @Angi-Kinas !

I'm just noticing the following problem:

lightbox

My impression is that it occurs on records with CORS errors in the console (for the map for instance), not sure if this can be related.

Do you think you could also add a e2e test that checks if the overlay-preview opens and displays the image?

libs/ui/elements/src/lib/thumbnail/thumbnail.component.ts
jahow: I can answer for @Angi-Kinas since we've done this together: we wanted this output to emit either `true` or `false` according to the state of the thumbnail, and not just emit `true` when the placeholder shows up

jahow (approved):
Thanks for the added tests!

PR #762, opened 2024-01-11
NPM package: fix exports

fgravin:
Add `gn4` folder from `api/repository` lib.


jahow (approved):
That makes sense, thanks!

PR #763, opened 2024-01-11
Button type directly in class

f-necas:
Fix buttons not using their tailwind classes.


fgravin (commented):
I don't really see the meaning of this change, it complexifies the button api.
Couldn't you adopt another approach ?

f-necas:
@fgravin It's not really a big change, instead of `primary` we would have to write `btn-primary`. 

It's just a small step to fix the display of not secondary buttons.

It's intended to get rid of this gn-ui-button either which will result in a `<button type="button" class="btn-primary>" `

jahow (commented):
Wait, how come this fixes the issue? edit: sorry, just saw the above messages. I think if we're going to modify every gn-ui-button usage then we might as well use the opportunity to get rid of this component altogether? But I would be very interested to know the underlying reason for this problem.

jahow (approved):
Thanks!

PR #764, opened 2024-01-12
Improve search in organizations & advanced filters with common fuzzy logic

jahow:
This PR introduces a `createFuzzyFilter` util that is used both in organizations & advanced filters search. This filter is only client side and will split compared strings by whitespace and compare each part while ignoring special chars etc.

In the future we might want to use a more advanced fuzzy filtering library.


libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.ts
```diff
@@ -74,9 +77,10 @@ export class DropdownMultiselectComponent {
   }
 
   get filteredChoicesByText() {
-    return this.choices.filter((choice) =>
-      choice.label.toLowerCase().includes(this.searchInputValue?.toLowerCase())
-    )
+    const filterPattern = this.searchInputValue?.toLowerCase()
```
tkohr, line 80: `?.toLowerCase()` might not be necessary anymore here, as the `FuzzyFilter` seems to handle this.

libs/util/shared/src/lib/utils/fuzzy-filter.ts
```diff
@@ -0,0 +1,32 @@
+type FuzzyFilter = (input: string) => boolean
+
+function normalize(input: string) {
+  return input
+    .toLowerCase()
+    .normalize('NFD') // explode composite chars (e.g. é) into multiple chars
+    .replace(/[\u0300-\u036f]/g, '') // remove accents
+    .replace(/œ/g, 'oe') // remove accents
```
tkohr, line 8: Looking at french specificities, `æ` is not relevant in modern french anymore, is it?

tkohr (approved):
Nice @jahow, very elegant solution to make this feature generic and reusable, thanks!

libs/util/shared/src/lib/utils/fuzzy-filter.ts
jahow: not ae but oe yes, like oeuf :)

libs/ui/inputs/src/lib/dropdown-multiselect/dropdown-multiselect.component.ts
jahow: right!

libs/util/shared/src/lib/utils/fuzzy-filter.ts
tkohr: yep, I just saw that æ also ([used to](https://fr.wikipedia.org/wiki/%C3%86)) exist, but seems to have been replaced mostly by "accent aigu" nowadays, if I understand right :-)

jahow: To be honest I would have never thought of it, but I encountered a metadata record with it!

PR #765, opened 2024-01-12
[Datahub] [Resdesign] Header section

cmoinier:
This PR applies the redesign of the header section, namely : 

- [x] add text to the star toggle
- [x] redesign and position the language switcher on the top right 
- [x] redesign the back button
- [x] remove sticky navigation bar
- [x] redesign the title layout
- [x] show the dataset type below the title on the left
- [x] show update date below the title
- [x] show status below the title

To do : 

- [x] Address review
- [x] Add condition to language switcher (if in conf)

This is NOT COVERED (will be in another PR):

- [x] thumbnail
- [x]  "share" link

### To test

Dataset `/dataset/01491630-78ce-49f3-b479-4b30dabc4c69` has all of the needed parameters of the design.

### Screenshot 

image


apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -1,36 +1,63 @@
...
     </div>
-
-    <div class="absolute" style="left: 16px; top: 16px">
-      <gn-ui-navigation-button
-        (click)="back()"
-        [label]="'datahub.search.back' | translate"
-        [icon]="'navigate_before'"
+    <div class="flex flex-row text-white gap-3 mb-3 ml-4">
+      <div
+        *ngIf="(displayMap$ | async) === true"
```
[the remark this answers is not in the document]
tkohr: I agree, or maybe `isGeodata$` if it's a boolean.

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -84,6 +84,7 @@ export interface BaseRecord {
   extras?: Record<string, unknown>
   landingPage?: URL
   updateFrequency?: UpdateFrequency
+  status?: RecordStatus
```
tkohr, line 87: FYI `status` is already present in `DatasetRecord`.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
```diff
@@ -27,6 +27,7 @@ import { PlatformServiceInterface } from '@geonetwork-ui/common/domain/platform.
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class FavoriteStarComponent implements AfterViewInit, OnDestroy {
+  @Input() displayCount: string
```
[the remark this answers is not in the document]
tkohr: Watch out, the type should be a `boolean`.

tkohr (changes requested):
Thanks for the clean adaptions @cmoinier! There are still some details that don't exactly match the mockups. Could you have a second check on that? Some I see at first glance are:
- the margin/padding below the "Geographical dataset" etc. line should be slightly larger
- background color of language switcher should be primary
- letter spacing of "BACK" should be wider
- should the time of last update be displayed? (It is not included in mockups)
- the date is not formatted as in mockups (which is not french btw) and not localized. We should eventually pass the current locale to toLocaleString().

I'm also wondering if there is a specific reason why we need to [reload the page](https://github.com/geonetwork/geonetwork-ui/blob/main/libs/ui/catalog/src/lib/language-switcher/language-switcher.component.ts#L45) when switching the language (cc @fgravin)? The line does not come from this PR, but its effect is visible here.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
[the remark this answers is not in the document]
cmoinier: Thx for the suggestions, both are done :slightly_smiling_face:

libs/common/domain/src/lib/model/record/metadata.model.ts
cmoinier: Yes but for some reason, `@Input() metadata: CatalogRecord` in `home-record` which is of type `ServiceRecord | DatasetRecord`, does not recognize `status`.

cmoinier:
Thanks @Angi-Kinas & @tkohr for your helpful reviews :slightly_smiling_face: 

I made the following changes : 

- [ ] Changed `displayMap$` name to `isGeodata$`
- [ ] Changed `displayCount` type to `Boolean` and made it optionnal
- [ ] Changed the margin below the "Geographical dataset" etc. line
- [ ] Added the primary color to the language switcher
- [ ] Spaced out the "BACK" letters
- [ ] Formatted the date as in the mockups, but yes, I agree that the date should be in current locale.

- [ ] I didn't add the time of last update, since it's not in the mockups. wdyt @fgravin @jahow ?

libs/common/domain/src/lib/model/record/metadata.model.ts
fgravin: There might be a reason !
The status only exists for Dataset records, so the solution is not to update the Parent type, but probably to cast the Record in the component.
If the record page is designed for dataset record, then its type should be `DatasetRecord` instead of `BaseRecord`

fgravin:
> I'm also wondering if there is a specific reason why we need to [reload the page](https://github.com/geonetwork/geonetwork-ui/blob/main/libs/ui/catalog/src/lib/language-switcher/language-switcher.component.ts#L45) when switching the language (cc @fgravin)? The line does not come from this PR, but its effect is visible here.

It is because changing the language has many consequences <hich are not handled internally:
- reload the search as the search is driven by the language (the index language we apply the search on)
- reload advanced filter values etc...

fgravin:
> * [x]  Changed `displayMap$` name to `isGeodata$`
> *

to me this is 2 different things, cf my comments in Slack

apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -1,36 +1,63 @@
...
+        <gn-ui-favorite-star
+          *ngIf="metadata?.uniqueIdentifier"
+          [record]="metadata"
+          [displayCount]="false"
+          class="flex text-background content-center"
+          [style.color]="foregroundColor"
+          [style.--star-toggle-enabled-color]="'var(--color-primary)'"
+          [style.--star-toggle-disabled-color]="foregroundColor"
+        ></gn-ui-favorite-star>
+        <gn-ui-language-switcher
```
fgravin, line 24: The language switcher should be displayed under a condition, isn't it ?
Is it the case or not ?

apps/datahub/src/app/record/header-record/header-record.component.ts
```diff
@@ -34,7 +34,14 @@ export class HeaderRecordComponent {
...
   get lastUpdate() {
-    return this.metadata.recordUpdated.toLocaleString().split(',')[0]
+    const date = this.metadata.recordUpdated
+    return (
+      ('0' + date.getDate()).slice(-2) +
+      '.' +
+      ('0' + (date.getMonth() + 1)).slice(-2) +
+      '.' +
+      date.getFullYear()
+    )
```
tkohr, line 44: I'd try to prevent a custom function here (that will make localization much more complicated as well).
How about a `return this.metadata.recordUpdated.toLocaleDateString(this.translateService.currentLang)`?
This will not display the dots from the mockup, but slashes are French standard.

libs/ui/catalog/src/lib/language-switcher/language-switcher.component.html
```diff
@@ -5,6 +5,8 @@
   [selected]="currentLang"
   ariaName="languages"
   [showTitle]="false"
-  [extraBtnClass]="'flex justify-items-center border border-white !pl-2 !py-1'"
+  [extraBtnClass]="
+    'flex justify-items-center border border-white bg-primary text-white !pl-2 !py-1'
+  "
```
tkohr, line 10: Sorry, I think it should actually be transparent, as it picks up the header background color in the mockups.

cmoinier: Okay, that's what I thought too, primary doesnt look very good

apps/datahub/src/app/record/header-record/header-record.component.ts
cmoinier: Okay! I knew it wasn't the best, but I didn't find anything better to have the dots. If we ditch the dots, then yes, your suggestion is much better!

apps/datahub/src/app/record/header-record/header-record.component.html
cmoinier: In `home-header`, yes, there is a condition for its display. But in the mockups (nor in the ticket), this wasn't specified. I assumed that the switcher had to be there at all times.

libs/common/domain/src/lib/model/record/metadata.model.ts
cmoinier: Thanks @fgravin , it's changed :slightly_smiling_face:

tkohr:
Thanks for all the adaptions @cmoinier! The date formats adapts well when switching the language on my side :+1: 

When using the geo2france config (`header_background = "#1f4490"` from mockups or the similar value from primary color), I still encounter a couple of problems:

dh_header

- The favorite star disappears when favoring a record due to used colors
- "Donnée géographique" is not distinct anymore => should perhaps be `bg-primary-darker` if this is ok for generic design
- I'd guess the [] around the status were just to represent a placeholder and can actually be removed

Maybe it'd be nice to add some e2e tests as well.

apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -1,36 +1,62 @@
...
+        </gn-ui-navigation-button>
+      </div>
+      <div class="flex flex-row mr-6 content-center gap-3">
+        <gn-ui-favorite-star
+          *ngIf="metadata?.uniqueIdentifier"
+          [record]="metadata"
+          [displayCount]="false"
+          class="flex text-background content-center"
+          [style.color]="foregroundColor"
+          [style.--star-toggle-enabled-color]="'var(--color-primary)'"
```
tkohr, line 21: I think one last adaption we need here to handle themes where the header background will be the same as the primary color (like the new mockups) is:
```suggestion
          [style.--star-toggle-enabled-color]="foregroundColor"
```
Otherwise an active star won't be visible.

tkohr (approved):
Thanks for the continued effort on this @cmoinier ! I let you apply the last suggested change and then it LGTM.

PR #769, opened 2024-01-15
[Datahub][Redesign] Details exapandable panel

cmoinier:
This PR applies the redesign of the header section, namely : 

- [x] Change the title of the "details" panel to "technical information"
- [x] Show the last metadata update time
- [x] Show the name of the source catalog
- [x] Show the unique dataset identifier
- [x] Show categories/themes
- [x] Show keywords

To do : 

- [x] Use gn-ui-copy-button
- [x] Revert linkify button
- [x] Make themes not clickable
- [x] Do not split the themes

This is NOT COVERED :

- [ ] Downloads count: this information is not yet available and might require a modification of the backend

### To test

Dataset `/dataset/01491630-78ce-49f3-b479-4b30dabc4c69` has all of the needed parameters of the design.

### Screenshot 

image


fgravin (commented):
I've just looked at the screen shot, it looks great :ok_hand: 
I see some icons on the mockups you might have forgotten them :upside_down_face: 
Thanks

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -49,6 +49,18 @@ export class MetadataInfoComponent {
     }
   }
 
+  get getThemes() {
```
tkohr, line 52: I don't think these themes (like "cadastre, aménagement" in the example) actually need to be split (cc @jahow @fgravin ). I'm noticing that clicking on (unsplit) themes, but also "constructions et ouvrages" for example does not return any results in the search, however. Do we need to add them to the search request or index?

tkohr (commented):
Thanks for the proper layout @cmoinier ! I just left a comment about an uncertainty regarding the themes.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
cmoinier: I did split them because indeed, clicking on "unsplit" themes didn't return results while clicking on "split" ones did. But I can undo that if it's not the wanted behavior.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -106,9 +106,23 @@
     </div>
     <div *ngIf="metadata.uniqueIdentifier">
       <p class="text-sm" translate>record.metadata.uniqueId</p>
-      <p class="text-primary font-medium">
-        {{ metadata.uniqueIdentifier }}
-      </p>
+      <div class="flex flex-row content-align gap-2">
+        <button
```
tkohr, line 110: Could we reuse the `gn-ui-copy-text-button` here?

libs/ui/elements/src/lib/metadata-info/linkify.directive.ts
```diff
@@ -75,6 +75,6 @@ export class GnUiLinkifyDirective implements OnInit {
   }
 
   private createLink(displayValue: string, url: string): string {
-    return `<a href="${url}" target="_blank" class="text-primary cursor-pointer hover:underline">${displayValue} <mat-icon class="material-symbols-outlined !w-[12px] !h-[14px] !text-[14px] opacity-75">open_in_new</mat-icon></a>`
+    return `<a href="${url}" target="_blank" class="text-primary cursor-pointer hover:underline"><mat-icon class="material-symbols-outlined mr-1 !w-[12px] !h-[14px] !text-[14px] opacity-75">open_in_new</mat-icon> ${displayValue}</a>`
```
tkohr, line 78: Just to make sure, is this change desired everywhere the directive is used in the new design?

cmoinier: The directive isn't visibly used anywhere else in the mockups, so it's just me assuming here.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
cmoinier: I will look into doing that :+1:

tkohr (approved):
Thanks for the adaptions @cmoinier, looks really clean now :slightly_smiling_face:

PR #773, opened 2024-01-18
[Datafeeder] Add CSV support

f-necas:
# CSV in DF

Add CSV support in Datafeeder.
Allow to add geospatial and non geospatial data with datafeeder.

Also add conditionnal link to geoserver (if it's not geospatial) and a new one to an OGC API Features.

For the backend part : https://github.com/georchestra/georchestra/pull/4137

## Screenshot
image

# Helpers
Mock API (backend) with [Mockoon](https://mockoon.com/) : 
[datafeeder.json](https://github.com/geonetwork/geonetwork-ui/files/13974165/datafeeder.json)


apps/datafeeder/src/app/presentation/pages/dataset-validation-csv-page/dataset-validation-csv-page.html
```diff
@@ -0,0 +1,91 @@
...
+        >
+          <tbody>
+            <tr
+              *ngFor="let row of csvData; let isFirstRow = first"
+              class="{{isFirstRow ? 'uppercase font-bold' : ''}} rounded-2xl bg-white border-b bg-secondary-white"
+            >
+              <td
+                *ngFor="let value of row; let isOdd = odd; let columnIndex = index;"
+                class="{{isOdd? 'bg-slate-100' : ''}} {{isFirstRow ? 'border-b' : ''}} text-sm px-2 py-2 border-spacing-0 border-separate"
+              >
```
jahow, line 59: this can probably be done using a built-in gn-ui component?

jahow (commented):
I think this looks quite good so far, haven't tested yet though.

I feel like the "csv validation page" is doing a lot of duplicate logic from other places though, and in principle I would say that we shouldn't have to add a complete new page for each format. Keep in mind that the datafeeder only handles 2 formats now, but there are dozens out there. Will we have to do a geopackage page, a geojson page, a KML page, a geotiff page, etc.? That seems cumbersome to me.

Also maybe some gn-ui components could be reused here? I'm thinking at least of the table, but that might not be easy considering it should contain other components.

What do you think?

jahow:
> > Will we have to do a geopackage page, a geojson page, a KML page, a geotiff page, etc.?
> 
> How can it be if there's no specific page for each format ? Each of them have their specifics ...

I agree, but also some things are common, e.g. showing a preview of the data as a map and/or a table. Basically we could consider that each format has its "parameters" that can be reviewed/tweaked by the user. Encoding for instance is something that can be relevant for other format as well.

fgravin:
I just had a demo from @f-necas and it looks great in term of UX.
Honestly, what @jahow raises makes sense, but I think that a better integration and refactoring should come when the work on the editor will be more precise. We should have more funding for that.
I am not even sure about the future of the datafeeder, so I wouldn't invest more on this PR and this budget for now.

support-services/docker-compose.yml
```diff
@@ -153,3 +153,18 @@ services:
...
+        [
+          'CMD-SHELL',
+          'curl -s -f http://localhost:80/datahub/ >/dev/null || exit 1',
+        ]
+      interval: 30s
+      timeout: 10s
+      retries: 10
+    restart: always
+    ports:
+      - '8082:80'
```
jahow, line 170: I don't think this should be kept?

jahow (approved):
I think this can be merged once all comments have been addressed; thanks!

PR #775, opened 2024-01-18
Suppress CSS warnings on build related to /deep/

jahow:
Replaced by `::ng-deep`, also made sure that we always have `:host` in front to avoid leaking styles.

See https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep

We still rely on applying styles further down an element children so we can't really get rid of these. I'm not sure if an alternative will show up in the future, but at least for now we won't have as many warnings on builds.

### DIFF
--- a/libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
+++ b/libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
@@ -1,5 +1,5 @@
 /** Body **/
-:host /deep/ .markdown-body {
+:host ::ng-deep .markdown-body {
   -ms-text-size-adjust: 100%;
   -webkit-text-size-adjust: 100%;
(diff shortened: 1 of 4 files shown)


libs/ui/elements/src/lib/markdown-parser/markdown-parser.component.css
```diff
@@ -1,5 +1,5 @@
 /** Body **/
-:host /deep/ .markdown-body {
+:host ::ng-deep .markdown-body {
```
fgravin, line 2: btw why do we need `ng-deep` if it's in the same component ?

jahow: because the content generated through innerHtml will not be affected by the component stylesheet.

fgravin (approved):

Issue #756, opened 2024-01-09
Expand Map Viewer functionalities


jahow:
The following features will be added to the Map Viewer app:

* Add layers from third party sources: WMS, WFS, OGC API Features, GeoJSON file
* Manipulate layers: order, opacity, delete
* Zoom on a layer extent
* Show the map legend
* Import and export a map using a JSON-based file format
* Print the map in PNG or PDF
* Add a geocoding widget
* Show current user location
* Show cursor coordinates (both in map projection and geographic coordinates)
* When using the map viewer web component:
  * Ability to specify a map JSON and make it read-only

jahow:
Since funding is lacking I am moving this to the next milestone

PR #776, opened 2024-01-18
[Map-Viewer] Add layer from WFS service

ronitjadhav:
This pull request adds the feature that allows users to add a layer from a Web Feature Service (WFS) to their map.

- Added a new `add-layer-from-wfs` component in `src/app/components/add-layer-from-wfs/add-layer-from-wfs.component.ts`
- Updated the `map-context.service.ts` and `map-context.model.ts` to add an option which limits WFS features count

**Task List:**
- [x] Ensure the initial display includes a text input for the WFS URL entry.
- [x] Implement a loading indicator upon URL submission.
- [x] Display a list of layers post-load, with each layer featuring a title and an "Add" button.
- [x] Enable layer addition to the map while keeping the panel open.
- [x] Apply default OL styling for added layers.
- [x] Enforce a maximum of 10,000 features per layer to prevent browser overload.

**Screenshots:**

image

image

#756

### DIFF
--- a/libs/feature/map/src/lib/add-layer-from-wfs/add-layer-from-wfs.component.ts
+++ b/libs/feature/map/src/lib/add-layer-from-wfs/add-layer-from-wfs.component.ts
@@ -1,0 +1,65 @@
+import { ChangeDetectorRef, Component, OnInit } from '@angular/core'
+import { WfsEndpoint, WfsFeatureTypeBrief } from '@camptocamp/ogc-client'
+import { Subject } from 'rxjs'
+import {
+  MapContextLayerModel,
+  MapContextLayerTypeEnum,
+} from '../map-context/map-context.model'
+import { MapFacade } from '../+state/map.facade'
+import { debounceTime } from 'rxjs/operators'
+
+@Component({
+  selector: 'gn-ui-add-layer-from-wfs',
+  templateUrl: './add-layer-from-wfs.component.html',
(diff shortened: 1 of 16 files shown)


libs/feature/map/src/lib/map-context/map-context.model.ts
```diff
@@ -32,6 +32,7 @@ interface MapContextLayerWfsModel {
   type: 'wfs'
   url: string
   name: string
+  featureCount?: number
```
jahow, line 35: See other comment, this shouldn't be necessary

libs/feature/map/src/lib/add-layer-from-wfs/add-layer-from-wfs.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      this.loading = true
+
+      if (this.wfsUrl.trim() === '') {
+        this.layers = []
+        return
+      }
+
+      this.wfsEndpoint = await new WfsEndpoint(this.wfsUrl).isReady()
+      this.layers = this.wfsEndpoint.getFeatureTypes()
+      console.log(this.layers)
```
jahow, line 45: This line can be removed I think :slightly_smiling_face:  did we forget a console.log in WMS as well?

jahow (approved):
Thanks, simple and clean, good job!

I made a comment about always applying the features count limit to WFS layers, feel free to merge once this is addressed.

PR #779, opened 2024-01-24
[Datahub] [Redesign] Thumbail preview

cmoinier:
### Description

This PR introduces changes for the redesign of the thumbnail image. It includes the following changes : 

- [x] Move the contact & lightbox section up in the about section
- [x] Increase thumbnail size
- [x] e2e tests

### Architectural changes

Added `basiclightbox` library.

### Screenshots

image

---

**This work is sponsored by [GeoCat](https://www.geocat.admin.ch/fr/home/evenements.html)**.

### DIFF
--- a/apps/datahub/src/app/record/record-metadata/record-metadata.component.html
+++ b/apps/datahub/src/app/record/record-metadata/record-metadata.component.html
@@ -30,15 +30,22 @@
           *ngIf="showOverlay"
         >
         </gn-ui-image-overlay-preview>
-        <div>
-          <gn-ui-metadata-contact
-            (organizationClick)="onOrganizationClick($event)"
+        <gn-ui-metadata-contact
+          (organizationClick)="onOrganizationClick($event)"
+          [metadata]="facade.metadata$ | async"
+        >
+        </gn-ui-metadata-contact>
+        <div *ngIf="metadataQualityDisplay">
+          <p class="text text-gray-700 text-xs mb-3 uppercase" translate>
+            record.metadata.quality
+          </p>
+          <gn-ui-metadata-quality
             [metadata]="facade.metadata$ | async"
-          >
-          </gn-ui-metadata-contact>
-          <gn-ui-metadata-catalog [sourceLabel]="sourceLabel$ | async">
-          </gn-ui-metadata-catalog>
+            [metadataQualityDisplay]="metadataQualityDisplay"
+          ></gn-ui-metadata-quality>
         </div>
+        <gn-ui-metadata-catalog [sourceLabel]="sourceLabel$ | async">
+        </gn-ui-metadata-catalog>
       </div>
     </div>
   </div>
--- a/apps/datahub/src/app/record/record-metadata/record-metadata.component.html
+++ b/apps/datahub/src/app/record/record-metadata/record-metadata.component.html
@@ -23,16 +23,7 @@
         >
         </gn-ui-metadata-info>
       </div>
-      <div>
-        <div *ngIf="metadataQualityDisplay">
-          <p class="text text-gray-700 text-xs mb-3 uppercase" translate>
-            record.metadata.quality
-          </p>
-          <gn-ui-metadata-quality
-            [metadata]="facade.metadata$ | async"
-            [metadataQualityDisplay]="metadataQualityDisplay"
-          ></gn-ui-metadata-quality>
-        </div>
+      <div class="-mt-44">
         <gn-ui-image-overlay-preview
           [imageUrl]="thumbnailUrl$ | async"
           (isPlaceholderShown)="showOverlay = !$event"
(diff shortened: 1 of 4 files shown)


apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -23,16 +23,7 @@
...
-      <div>
-        <div *ngIf="metadataQualityDisplay">
-          <p class="text text-gray-700 text-xs mb-3 uppercase" translate>
-            record.metadata.quality
-          </p>
-          <gn-ui-metadata-quality
-            [metadata]="facade.metadata$ | async"
-            [metadataQualityDisplay]="metadataQualityDisplay"
-          ></gn-ui-metadata-quality>
-        </div>
```
jahow, line 35: hmm, was there a reason to remove that? This indicator should stay somewhere on the page, although I agree that now, above the thumbnail is probably not the best.

Also our E2E tests are lacking because we're not testing that this indicator indeed show up :grimacing:

libs/ui/elements/src/lib/image-overlay-preview/image-overlay-preview.component.html
```diff
@@ -1,29 +1,25 @@
-<gn-ui-content-ghost
+<div
+  *ngIf="imageUrl"
```
jahow, line 2: Looks like the "content-ghost" for the thumbnail is gone, could you please bring it back and see if your changes still work? thanks :)

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
```diff
@@ -19,6 +19,7 @@
           class="text-primary font-title text-21 mr-2 cursor-pointer hover:underline"
           (click)="onOrganizationClick()"
           data-cy="organization-name"
+          *ngIf="shownOrganization?.name"
```
jahow, line 22: was that change intentional? I'm not sure effect it has but it looks not intended :wink:

jahow (commented):
Thanks for the changes, I think some unintentional changes slipped through, could you try and see if you can narrow down what was really intended? thanks!

libs/ui/elements/src/lib/image-overlay-preview/image-overlay-preview.component.html
cmoinier: I removed that because it was creating a weird display of the thumbnail, but if it's important I can look into having it back.

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
cmoinier: It was intentional! I had a case where the organization had no name (or an error on it perhaps), so I added it as a fallback. I know it's not included in the PR, but figured it was shorter a fix than creating a ticket for it...

libs/ui/elements/src/lib/image-overlay-preview/image-overlay-preview.component.html
jahow: hmm, maybe the content ghost doesn't work well on a colored background? what do you mean by "weird display"? The content ghost is important to avoid layout shifts when the data loading is finished

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
jahow: Yes ok, I see what you mean. These kind of "hidden" components are tricky. This component is important for platforms which use it, but enabling it is not necessarily straightforward so we tend to forget them. See 

@lucilecollignon do you have a suggestion on what do to with this component? Currently it looks like this:
image

libs/ui/elements/src/lib/metadata-contact/metadata-contact.component.html
jahow: ok, then I would suggest trying to identify what would cause an organization to have no name. This can have other implications so I think it'd be best if you could go ahead and create a bug ticket? thanks!

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
cmoinier: Interesting, I never saw this component before :slightly_frowning_face: 

Waiting for Lucile's input, but I feel like it should go under the lightbox and contact information.

libs/ui/elements/src/lib/image-overlay-preview/image-overlay-preview.component.html
cmoinier: Basically, it's like the image is hidden by a mask and there is a "loading animation"... Also, the console shows no error, so I'm a bit at lost on what is wrong here.

But maybe I'm missing something, like a parameter ? Here I have added back the "content-ghost" and its class `ghostClass="h-36 mb-3"`. 

image

jahow:
I amended this PR to address the issues, we changed the position of the md quality indicator like so:
image

jahow (approved):

PR #782, opened 2024-01-26
Datahub: Add generic "keyword" filter with translated values

jahow:
### Description

This PR introduces a generic `keyword` filter. Unlike the INSPIRE keyword filter which fetches translation dynamically from the GeoNetwork API, the `keyword` filter simply looks at either `tag.default` or `tag.langxyz` fields depending on whether a hardcoded metadata language is defined in the configuration.

This PR also simplifies the implementation of both topics and INSPIRE keyword filters by introducing a `TranslatedSearchField` class which relies on the platform service to show a label in the correct language.

### Architectural changes

No significant changes.

### Screenshots

image

---

**This work is sponsored by Swisstopo / [Geocat.ch](geocat.ch)**.

### DIFF
--- a/apps/datahub-e2e/src/e2e/datasets.cy.ts
+++ b/apps/datahub-e2e/src/e2e/datasets.cy.ts
@@ -157,6 +157,12 @@ describe('datasets', () => {
     }
 
     beforeEach(() => {
+      // this will enable all available filters
+      cy.intercept('GET', '/assets/configuration/default.toml', {
+        fixture: 'config-with-all-filters.toml',
+      })
+      cy.visit('/search')
+
       // expand filters
       cy.get('datahub-search-filters')
         .find('[data-cy=filters-expand]')
@@ -164,7 +170,7 @@ describe('datasets', () => {
         .click()
     })
     it('should display all filters', () => {
-      cy.get('@filters').filter(':visible').should('have.length', 6)
+      cy.get('@filters').filter(':visible').should('have.length', 10)
       cy.get('@filters')
         .children()
         .then(($dropdowns) =>
@@ -179,6 +185,10 @@ describe('datasets', () => {
           'topic',
           'isSpatial',
           'license',
+          'inspireKeyword',
+          'keyword',
+          'resourceType',
+          'representationType',
         ])
     })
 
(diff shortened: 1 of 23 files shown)


fgravin (commented):
Not looked at the code yet but just tested a bit, some issues on geocat dev for instance:
- the limit of 1000 for the `tag` facet is not enough as there are more than 1000 keywords
- the limit of 1000 for thesaurus keys is not enough for `http://www.eionet.europa.eu/gemet/concept` thesaurus
- the runtime_mapping slows a lot the request
- the runtime_mapping is present in all state search request payloads, which makes the search very slow, even when the runtime is not needed
- could be some issues, for instance `opendata.swiss` which have 2 entries, one for the keywords from the thesaurus, one for the same keyword which is not in the thesaurus
- `opendata.swiss` entries sum count is not the same as the same facet in gn legacy ui `900 + 1149 != 2566`
- might be better to sort alphabetically

I think our method is accurate, but leads to some inconsistency and is slower than just targeting an index entry (eg `tag.langfre`)
Might need to talk with Benoit about what is the best tradeoff.

jahow:
You're right, the performance on the keyword fields with runtime mappings is abysmal on large catalogs, I didn't even realize that. We could use the pipeline system to register a scripted field specifically for keywords, and use it if available to improve the results. If not available, we can just target a `lang<xxx>` field according to the UI language.

As for the GEMET concepts, the thesaurus contains more than 1000 entries so currently some won't be fetched; I thought about doing pagination on these queries, not sure yet. Maybe this is going too far in terms of trying to improve the search results.

I'm going to leave this PR for now and will revisit it later; I'll create another PR to extract the changes to the docker composition though, because this will improve e2e tests.

jahow:
I have reworked the PR to not rely on pipelines or runtime fields, and simply look for either `tag.default` or `tag.langxyz` depending on whether a fixed metadata language is defined in the config.

This is ready for review, although I still have to add documentation on the topic.

jahow:
> The code looks good. I've tested the `keyword` and works well, but the `inspireKeyword` list is empty, on geocat dev.

Thanks, there was indeed a regression on the inspire keywords. I've added E2E tests for all the new filters to make sure this does not happen again.

Angi-Kinas (approved):
Thanks @jahow ! I manually tested it and it LGTM! :rocket:

PR #790, opened 2024-02-02
Add editable label directive to UI module

LHBruneton-C2C:
### Description

This PR introduces a new directive, to edit in place fields like titles and names.

### Architectural changes

Strangely enough, the `RecordFormComponent` was declared as standalone, but not importing anything more than `CommonModule`. Had to at least import `UiInputsModule` for it to work properly.

Added the new `EditableLabelDirective` to the `UiInputsModule` too.

### DIFF
--- a/libs/ui/inputs/src/lib/editable-label/editable-label.directive.ts
+++ b/libs/ui/inputs/src/lib/editable-label/editable-label.directive.ts
@@ -1,0 +1,47 @@
+import {
+  Directive,
+  ElementRef,
+  Renderer2,
+  AfterViewInit,
+  EventEmitter,
+  Output,
+  Input,
+} from '@angular/core'
+
+@Directive({
+  selector: '[gnUiEditableLabel]',
+})
+export class EditableLabelDirective implements AfterViewInit {
+  @Output() editableLabelChanged = new EventEmitter<string>()
+  @Input() gnUiEditableLabel?: boolean
+
+  constructor(private el: ElementRef, private renderer: Renderer2) {}
+
+  ngAfterViewInit() {
+    if (this.gnUiEditableLabel !== false) {
+      const appendedInput = this.renderer.createElement('input')
+
+      this.renderer.setStyle(appendedInput, 'background', 'inherit')
+      this.renderer.setStyle(appendedInput, 'color', 'inherit')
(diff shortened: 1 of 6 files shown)


jahow:
> @jahow can you confirm this is the technical approach you had in mind ? If the technical approach is OK, what would be the proper way to deliver this feature? With a story into StoryBook only? This is only a draft, I will complete with further tests and some refactoring / renaming.

Yes I also like the approach of a directive, it gives more freedom for styling the affected label. I would suggest trying to use the element inner text as `initalValue` input, this would also make it easier to apply this directive conditionnally.

Having a complete story on storybook is enough to validate and merge this. Thanks!

libs/ui/inputs/src/lib/editable-label/editable-label.directive.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.renderer.setStyle(appendedInput, 'border', 'inherit')
+      this.renderer.setStyle(appendedInput, 'width', '100%')
+      this.renderer.setStyle(appendedInput, 'padding', 'inherit')
+      this.renderer.setStyle(appendedInput, 'margin', '0')
+      this.renderer.setStyle(appendedInput, 'height', 'inherit')
+      this.renderer.setStyle(appendedInput, 'line-height', 'inherit')
+      this.renderer.setStyle(appendedInput, 'text-decoration', 'inherit')
+
+      const hostContent = this.el.nativeElement.innerHTML
+      this.renderer.setProperty(appendedInput, 'value', hostContent)
```
jahow, line 36: `hostContent` needs to be trimmed, otherwise there will be blanks in the first and last positions

jahow: (or innerText could be used instead, I'm not sure if that would change anything)

jahow (approved):
Thanks, simple and efficient! I just noticed an issue with the original content of the lael being prepended and appended with spaces. Feel free to merge when this is sorted out :)

libs/ui/inputs/src/lib/editable-label/editable-label.directive.ts
LHBruneton-C2C: Well spotted! As we decided this should be specific to text elements, I'll go with `textContents`, and `trim()`.

PR #797, opened 2024-02-16
Metadata Editor: rework pages and introduce an interactive table UI component

jahow:
### Description

This PR does:
* introduce a `gn-ui-interactive-table` UI component in ui/layout which is used both for records and users lists
* introduce a `gn-ui-results-table` smart component in feature/search which is bound to the search facade, shows a list of records and offers sorting options; also responsible for the selection/deselection of records
* simplify the components in the metadata-editor app to rely on these components
* slightly change the layout of the tables in the editor to match the mockups

### Architectural changes

Made several components standalone, removed the `gn-ui-record-table` component which was too specific to the editor and isn't used anymore.

### Screenshots

image

---


libs/ui/layout/src/lib/interactive-table/interactive-table-column/interactive-table-column.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  selector: 'gn-ui-interactive-table-column',
+  standalone: true,
+  imports: [CommonModule],
+  templateUrl: './interactive-table-column.component.html',
+  styleUrls: ['./interactive-table-column.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class InteractiveTableColumnComponent {
+  @ContentChild('header') header: TemplateRef<unknown>
+  @ContentChild('cell') cell: TemplateRef<unknown>
```
LHBruneton-C2C, line 22: I couldn't find in the review where those objects header and cell are used, what's their purpose?

jahow: yes exactly! I think it was a good way to give the possibility to render anything in each column using templates

libs/api/repository/src/lib/gn4/selection/selection.service.ts
```diff
@@ -33,48 +33,43 @@ export class SelectionService {
...
-      map(() => undefined)
-    )
+  selectRecords(records: CatalogRecord[]) {
+    const newIds = records.map((record) => record.uniqueIdentifier)
+    this.selectionsApi
+      .add(BUCKET_ID, newIds)
+      .pipe(
+        tap(() => {
+          this.addIdsToSelected(newIds)
+        })
```
LHBruneton-C2C, line 43: I'm not a huge fan of tap, and as I'm only discovering the code, I wonder:
- if addIdsToSelected is storing selected ids locally, why not read from the state instead
- if it's a facade above the state, why not use an effect instead?

libs/feature/search/src/lib/results-table/results-table.component.html
```diff
@@ -0,0 +1,135 @@
...
+        hits: (recordCount$ | async)
+      }"
+    >
+      results.records.hits.displayedOn
+    </div>
+    <div
+      *ngIf="(selectedRecords$ | async).length as selectedRecordCount"
+      class="my-8 ml-6 block text-gray-400 selected-records"
+      translate
+      [translateParams]="{ amount: selectedRecordCount }"
```
LHBruneton-C2C, line 17: To discuss: how does translation work on this project? I will need it for the rich text editor component.

libs/feature/search/src/lib/results-table/results-table.component.ts
```diff
@@ -0,0 +1,165 @@
...
+  imports: [
+    CommonModule,
+    UiInputsModule,
+    InteractiveTableComponent,
+    InteractiveTableColumnComponent,
+    MatIconModule,
+    TranslateModule,
+  ],
+})
+export class ResultsTableComponent {
```
LHBruneton-C2C, line 38: So many methods, and the rendering part is huge too... Should we split the logical part to a service provided by the component, and keep only the rendering part here? That means splitting the unit tests too.

libs/feature/search/src/lib/results-table/results-table.component.html
```diff
@@ -0,0 +1,135 @@
+<div>
+  <div class="flex records-information">
```
LHBruneton-C2C, line 2: Out of curiosity, the css file is empty, but the `records-information` looks like a project specific class, where does it come from? I couldn't find it as a selector in the unit tests either, although I'd much rather use the `data-test` attribute selector.

libs/feature/search/src/lib/results-table/results-table.component.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  setSortBy = jest.fn()
+}
+class SelectionServiceMock {
+  selectRecords = jest.fn()
+  deselectRecords = jest.fn()
+  clearSelection = jest.fn()
+  selectedRecordsIdentifiers$ = new BehaviorSubject([])
+}
+
+describe('ResultsTableComponent', () => {
```
LHBruneton-C2C, line 30: I' don't see any css selector in these tests, even though there is rendering logic to tests (several `*ngIf`s). This would be easier to test if the component was split into a rendering part and a logical part (see my comment about a provided service). This is just a suggestion, I don't know the project well enough, might be tested in another way.

libs/feature/search/src/lib/results-table/results-table.component.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      it('should return false when the record is not in the selectedRecords array', async () => {
+        selectionService.selectedRecordsIdentifiers$.next(['1', '2', '3'])
+        const record = { uniqueIdentifier: '4' } as CatalogRecord
+        expect(await firstValueFrom(component.isChecked(record))).toBe(false)
+      })
+    })
+
+    describe('#handleRecordSelectedChange', () => {
+      it('should call selectRecords when checkbox is clicked', () => {
```
LHBruneton-C2C, line 147: Possible misleading comment, as the click is not part of the test, if at some point the event is changed to another, the comment might not be changed.

libs/feature/search/src/lib/results-table/results-table.component.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        selectionService.selectedRecordsIdentifiers$.next([
+          '1',
+          '2',
+          '3',
+          '4',
+          '5',
+        ])
+        expect(await firstValueFrom(component.isSomeSelected())).toBe(false)
+      })
+      it('returns true if more than one record selected', async () => {
```
LHBruneton-C2C, line 182: `more than one`, or `one or more`?

libs/feature/search/src/lib/results-table/results-table.component.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        selectionService.selectedRecordsIdentifiers$.next(['2', '3'])
+        expect(await firstValueFrom(component.isSomeSelected())).toBe(true)
+      })
+      it('returns false if no record selected', async () => {
+        selectionService.selectedRecordsIdentifiers$.next(['4', '5'])
+        expect(await firstValueFrom(component.isSomeSelected())).toBe(false)
+      })
```
LHBruneton-C2C, line 189: I don't understand this test compared to the previous one? Are there no selected records because 4 and 5 are not part of the possible selection? If so, this should be explicit in the description.

libs/feature/search/src/lib/results-table/results-table.component.ts
```diff
@@ -0,0 +1,165 @@
...
+    return formats
+  }
+
+  formatUserInfo(userInfo: string | unknown): string {
+    const infos = (typeof userInfo === 'string' ? userInfo : '').split('|')
+    if (infos && infos.length === 4) {
+      return `${infos[2]} ${infos[1]}`
+    }
+    return undefined
+  }
```
LHBruneton-C2C, line 83: Is this tested somewhere I couldn't find? Quite a bit of logic there.

libs/feature/search/src/lib/results-table/results-table.component.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  handleRecordSelectedChange(selected: boolean, record: CatalogRecord) {
+    if (!selected) {
+      this.selectionService.deselectRecords([record])
+    } else {
+      this.selectionService.selectRecords([record])
+    }
+  }
+
+  async selectAll() {
```
LHBruneton-C2C, line 131: Maybe rename to `toggleCompleteSelection` or `toggleSelectAll`, as this can deselect too.

apps/metadata-editor/src/app/my-org-users/my-org-users.component.ts
```diff
@@ -1,48 +1,27 @@
...
 })
-export class MyOrgUsersComponent implements OnDestroy {
-  orgData: {
-    orgName: string
-    logoUrl: string
-    recordCount: number
-    userCount: number
-    userList: UserModel[]
-  }
+export class MyOrgUsersComponent {
```
LHBruneton-C2C, line 23: This component is already not doing much in terms of logic, would it cost much to make it dumb?

apps/metadata-editor/src/app/my-org-users/my-org-users.component.spec.ts
```diff
@@ -59,43 +47,32 @@ describe('MyOrgUsersComponent', () => {
...
+    let orgData
+
+    beforeEach(() => {
+      orgData = null
+      component.orgData$.subscribe((data) => (orgData = data))
+    })
+
     it('should get the org name', () => {
-      expect(component.orgData.orgName).toEqual('wizard-org')
+      expect(orgData.orgName).toEqual('wizard-org')
```
LHBruneton-C2C, line 67: This feels like you are testing the data type, and not the rendering. Why not select the html element bearing the dynamic content and test that this is correct?

apps/metadata-editor/src/app/records/records-list.component.html
```diff
@@ -29,6 +29,16 @@ <h1 class="text-[56px] font-title grow">{{ title }}</h1>
...
+      <gn-ui-button
+        type="light"
+        *ngIf="userCount > 0"
+        (buttonClick)="showUsers()"
+      >
+        {{ userCount }}&nbsp;
+        <span translate [translateParams]="{ count: userCount }">
+          dashboard.records.users
+        </span>
+      </gn-ui-button>
```
LHBruneton-C2C: New logic without new test?

LHBruneton-C2C (commented):
Very interesting! Sorry for all the comments outside the review, most of them will be subjects of discussions I guess.

libs/api/repository/src/lib/gn4/selection/selection.service.ts
jahow: Right, I simplified the selection service to hopefully make it more readable in https://github.com/geonetwork/geonetwork-ui/pull/797/commits/b0eeb67aa92e2aece5f63009949bc0b6041cf2ae

This service provides a local state as well as handles interaction with the backend API, there is no NgRx store for this part because repository interfaces can have various implementations.

libs/feature/search/src/lib/results-table/results-table.component.html
jahow: Discussed separately; for the record, here is the doc about i18n in the project: https://geonetwork.github.io/geonetwork-ui/main/docs/reference/i18n.html

libs/feature/search/src/lib/results-table/results-table.component.html
jahow: it's used in the Cypress E2E tests; I've changed it to use a `data-cy` attribute which is a standard practice and less prone to confusion

libs/feature/search/src/lib/results-table/results-table.component.ts
jahow: Ideally yes, but this would go too far for this PR, especially since the focus should not be on the dashboard view any more. So I would be in favor of not pushing the refactoring further for now.

libs/feature/search/src/lib/results-table/results-table.component.spec.ts
jahow: You're right, several logic branches are not tested yet. This should be handled later when/if this component can be split up into logic/presentation further.

apps/metadata-editor/src/app/records/records-list.component.html
jahow: This should actually be a link instead of a button, but it was made like that initially and I am not going to invest more time in this. Also this will need to be reworked to match the new mockups anyway

apps/metadata-editor/src/app/my-org-users/my-org-users.component.spec.ts
jahow: the users will be fed to another component; but really even this test is a bit superfluous since there's barely any logic in this component.

jahow:
@LHBruneton-C2C I addressed all your comments I think, the fact is that this PR is only a partial refactoring of the dashboard view and there will probably be other ones later on to match the latest mockups, which is why I am not pushing this too far. Let me know if you feel that makes sense!

LHBruneton-C2C (approved):
@jahow Thanks for the explanations! And great work, this is quite a beast.

jahow:
Thanks for the extensive review :slightly_smiling_face:

PR #798, opened 2024-02-19
Datahub: Remove warning from console if filter geometry undefined

tkohr:
### Description

This PR prevents displaying a warning in the console if no `filter_geometry` could be found (which is already logged in the console). This might occur when a request to the geometry aborts without success (due to access rights or a wrong url for example).

image-2024-02-15-08-47-08-973

Follow-up on #635 

### DIFF
--- a/libs/feature/search/src/lib/state/effects.spec.ts
+++ b/libs/feature/search/src/lib/state/effects.spec.ts
@@ -492,8 +492,8 @@ describe('Effects', () => {
           it('skips the geometry in the search', async () => {
             await firstValueFrom(effects.loadResults$)
             expect(repository.search).toHaveBeenCalledWith(
-              expect.not.objectContaining({
-                filterGeometry: { type: 'Polygon', coordinates: [[]] },
+              expect.objectContaining({
+                filterGeometry: undefined,
               })
             )
           })
--- a/libs/feature/search/src/lib/state/effects.spec.ts
+++ b/libs/feature/search/src/lib/state/effects.spec.ts
@@ -483,6 +483,21 @@ describe('Effects', () => {
             )
           })
         })
+        describe('when geometry is undefined', () => {
+          beforeEach(() => {
+            effects['filterGeometry$'] = of(undefined) as any
+            effects = TestBed.inject(SearchEffects)
+            actions$ = of(new RequestMoreResults('main'))
+          })
+          it('skips the geometry in the search', async () => {
+            await firstValueFrom(effects.loadResults$)
+            expect(repository.search).toHaveBeenCalledWith(
+              expect.not.objectContaining({
+                filterGeometry: { type: 'Polygon', coordinates: [[]] },
+              })
+            )
+          })
+        })
       })
       describe('when useSpatialFilter is disabled', () => {
         beforeEach(() => {
--- a/libs/feature/search/src/lib/state/effects.ts
+++ b/libs/feature/search/src/lib/state/effects.ts
@@ -130,17 +130,19 @@ export class SearchEffects {
             }
             return this.filterGeometry$.pipe(
               tap((geom) => {
-                try {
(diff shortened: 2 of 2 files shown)


libs/feature/search/src/lib/state/effects.ts
```diff
@@ -130,17 +130,19 @@ export class SearchEffects {
...
                   }
-                } catch (error) {
-                  console.warn(
-                    'Error while parsing the geometry filter\n',
-                    error
-                  )
-                  throw new Error()
                 }
               }),
               map((geom) => [state, favorites, geom]),
```
tkohr, line 148: Was just wondering if it was better to return `null` here (like in the `catchError` below) if the `geom` is `undefined`...
```suggestion
              map((geom) => [state, favorites, geom ?? null]),
```

libs/feature/search/src/lib/state/effects.ts
```diff
@@ -130,17 +130,19 @@ export class SearchEffects {
             }
             return this.filterGeometry$.pipe(
               tap((geom) => {
-                try {
-                  const trace = validGeoJson(geom, true) as string[]
-                  if (trace?.length > 0) {
-                    throw new Error(trace.join('\n'))
+                if (geom) {
```
jahow, line 133: could you maybe invert this statement? e.g. `if (!geom) { return; }`
That would save one level of depth :)

libs/feature/search/src/lib/state/effects.ts
jahow: yes that sounds good, it would mean we would reliably get `null` for a geometry in all those cases

libs/feature/search/src/lib/state/effects.spec.ts
```diff
@@ -483,6 +483,21 @@ describe('Effects', () => {
...
+        describe('when geometry is undefined', () => {
+          beforeEach(() => {
+            effects['filterGeometry$'] = of(undefined) as any
+            effects = TestBed.inject(SearchEffects)
+            actions$ = of(new RequestMoreResults('main'))
+          })
+          it('skips the geometry in the search', async () => {
+            await firstValueFrom(effects.loadResults$)
+            expect(repository.search).toHaveBeenCalledWith(
+              expect.not.objectContaining({
```
jahow, line 495: could you be more specific in your assertion here and make sure that `geom` was null?

jahow (approved):
Thanks! I made a few comments, also I think your idea about always returning null is good. In general the fix looks good so I'm approving, I hope the tests will eventually turn green...

libs/feature/search/src/lib/state/effects.ts
tkohr: Actually, I noticed that we are converting nullish geometries to `undefined` [further down](https://github.com/geonetwork/geonetwork-ui/pull/798/files#diff-3d941fbb88225a34e9370b338aaf79c8c0fb0e4f4cdb5f060932f5d3e995ee36L173), so I'm hesitating a little where to use `null` and `undefined`. Returning `undefined` in line [173](https://github.com/geonetwork/geonetwork-ui/pull/798/files#diff-3d941fbb88225a34e9370b338aaf79c8c0fb0e4f4cdb5f060932f5d3e995ee36L173) seems reasonable, as it's a new object and we do the same for `filterIds`. So I kept `null` in the modified `switchMap` and `undefined` in the following. What do you think?

libs/feature/search/src/lib/state/effects.spec.ts
tkohr: As commented above, I made sure it's `undefined` in the test.

libs/feature/search/src/lib/state/effects.ts
tkohr: So as discussed, I consistently put `undefined` in the end, which should make the code more readable.

PR #804, opened 2024-02-23
Search: Do not change route path when clearing fuzzy search

tkohr:
### Description

Currently, the router navigates to the search when the fuzzy search entry is cleared (with the input's cross) on a different tab/page than the search. This PR intends to maintain the same path when clearing the search. The `fuzzy-search.component` now only calls `updateFilters` if `searchFilters.any` is not empty or undefined.

Also updates `@geospatial-sdk/geocoding` for the npm package and attempts to fix a flaky e2e test.

### DIFF
--- a/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
+++ b/libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
@@ -11,8 +11,8 @@ import {
   AutocompleteComponent,
   AutocompleteItem,
 } from '@geonetwork-ui/ui/inputs'
-import { Observable } from 'rxjs'
-import { map, tap } from 'rxjs/operators'
+import { Observable, firstValueFrom } from 'rxjs'
+import { map } from 'rxjs/operators'
 import { SearchFacade } from '../state/search.facade'
 import { SearchService } from '../utils/service/search.service'
 import { CatalogRecord } from '@geonetwork-ui/common/domain/model/record'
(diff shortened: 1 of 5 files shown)


libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -76,6 +79,8 @@ export class FuzzySearchComponent implements OnInit {
   }
 
   handleInputCleared() {
-    this.searchService.updateFilters({ any: '' })
+    if (this.currentSearchFilters.any) {
```
jahow, line 82: Instead of storing an internal state here you can do something like so:
```ts
async handleInputCleared() {
  const currentSearchFilters = await firstValueFrom(this.searchFacade.searchFilters$);
  if (!currentSearchFilters.any) { return; }
  this.searchService.updateFilters({ any: '' })
}
```

jahow (commented):
Thanks for all the added tests! I did a suggestion to make the code more straightforward, what do you think?

jahow (approved):

Angi-Kinas (approved):
Thanks @tkohr for changing the behavior for clearing the search input! It looks good to me :+1:

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
tkohr: Nice idea, also works as expected, thanks! I just inverted the `if` to save another line.

PR #805, opened 2024-02-25
Metadata Editor: rework search results layout, add "create record" button

jahow:
### Description

This PR affects the layout of the search results in the dashboard views of the editor:
* Removed `/records/all`, only keep `/records/search` route (similar to the datahub)
* Adapt the layout of the search results in the dashboard to make it closer to the mockups
* Create a new `RecordsCountComponent` class to extract the part that shows the amount of records and selected records in the table
* Add a "create record" button on the search results view
* Removed unused `CreatePageComponent`, now `EditComponent` is used both for editing and creating
* Document some best practices related to standalone components and testing

### Architectural changes

Add `ngx-translate-testing` dependency (instead of a local copy of the lib that was introduced while waiting for rxJS 7)

### Screenshots

image


apps/metadata-editor/src/app/records/search-records/search-records-list.component.html
```diff
@@ -1,9 +1,58 @@
...
+      </h1>
+      <div class="text-[12px]">
+        <md-editor-records-count></md-editor-records-count>
+      </div>
+    </ng-container>
+    <ng-template #allRecords>
+      <h1 class="text-[16px] text-main font-title font-bold" translate>
+        dashboard.records.all
+      </h1>
+      <div class="text-[12px]">
```
LHBruneton-C2C, line 19: I don't mind for gap, padding, margin, but should text size be hardcoded in pixels this way?

apps/metadata-editor/src/app/records/search-records/search-records-list.component.spec.ts
```diff
@@ -1,30 +1,85 @@
...
+import { By } from '@angular/platform-browser'
+import { DATASET_RECORDS } from '@geonetwork-ui/common/fixtures'
+import { Router } from '@angular/router'
+
+const results = [{ md: true }]
+const currentPage = 5
+const totalPages = 25
+
+@Component({
+  // eslint-disable-next-line
```
LHBruneton-C2C, line 26: Not even specifying which rule to disable?

docs/guide/best-practices.md
```diff
@@ -4,6 +4,156 @@ outline: deep
...
+```html
+<div>My component</div>
+<div *ngIf="data$ | async as currentData" data-test="show-data">
+  {{ currentData }}
+</div>
+```
+
+This makes the HTML template easier to read and avoid confusion with e.g. CSS classes.
+
+#### Mocking translations
```
LHBruneton-C2C, line 80: Is this valid for Storybook too?

LHBruneton-C2C (approved):
Thanks for this great work!

apps/metadata-editor/src/app/records/search-records/search-records-list.component.spec.ts
jahow: we've always been doing like that in tests when defining mock components because otherwise eslint complains about the component prefix; but I agree that specifying the rule to disable would make it more explicit

docs/guide/best-practices.md
jahow: Hmm, good question; I have the impression that translations just don't work in Storybook, which is often not very visible because UI components tend to have their labels provided to them already translated. We should make Storybook load the default EN translations I guess.

jahow: I took a look and actually translations work just fine in Storybook, I will document how to enable them.

libs/feature/search/src/lib/results-table/results-table.component.stories.ts
jahow, line 1: This was removed because it wasn't working and it basically needs a whole application to run, which doesn't really make sense in the context of storybook anymore

jahow:
Thanks for the review!

PR #808, opened 2024-03-01
Metadata Editor: rework E2E test to avoid random failures

jahow:


tkohr (approved):
Thanks for reworking this @jahow! Both changes make sense to me.

PR #811, opened 2024-03-05
GN4 mapper: Map keyword types/themes into a new Keyword type

tkohr:
### Description

This PR takes the keywords from ES `allKeywords` instead of `tags` in order to map the keyword type/theme like `place`, `theme` etc. into a new `Keyword` type.


libs/api/metadata-converter/src/lib/gn4/atomic-operations.ts
```diff
@@ -98,3 +100,19 @@ export const mapContact = (
...
+
+export const mapKeywords = (thesauri: Thesaurus[], language: string) => {
+  const keywords = []
+  for (const thesaurusId in thesauri) {
+    const thesaurus = thesauri[thesaurusId]
+    for (const keyword of thesaurus.keywords) {
+      keywords.push({
+        label: selectTranslatedValue<string>(keyword, language),
+        type: getKeywordTypeFromKeywordTypeCode(thesaurus.theme),
+        ...(thesaurus.id && { thesaurus: { id: thesaurus.id } }),
```
tkohr, line 112: Just wondering if we shouldn't try to recover the `name` and `url` here as well? I see optionally `title` and `link` in the ES response.

tkohr (commented):
Thanks for the huge work behind this @jahow ! As far as I can judge, all looks good. I just left one comment and added one small commit for the sake of renaming completeness.

libs/api/metadata-converter/src/lib/gn4/atomic-operations.ts
jahow: yes good point!

jahow: done

jahow (approved):
Looking good :) thanks for starting this work, now we have proper keywords in the pivot format!

tkohr:
Thanks for the latest adaption @jahow ! Looking at some other code a last doubt came to my mind, but I guess we can still continue using `tag` for the keywords filter [here](https://github.com/geonetwork/geonetwork-ui/blob/place-keywords/libs/feature/search/src/lib/utils/service/fields.service.ts#L63), right?

jahow:
> Thanks for the latest adaption @jahow ! Looking at some other code a last doubt came to my mind, but I guess we can still continue using `tag` for the keywords filter [here](https://github.com/geonetwork/geonetwork-ui/blob/place-keywords/libs/feature/search/src/lib/utils/service/fields.service.ts#L63), right?

Yes that should keep working identically

PR #815, opened 2024-03-06
[Datahub][API form] Empty string not read as negative limit

cmoinier:
### Description

This PR fixes that empty strings are interpreted as -1 for the limit input.
It also updates the docker compo with a new dump, which changes a deprecated OGC API URL to a working one.

Please run `docker-compose down -v` and `docker-compose up -d` to have the right dump.


libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
```diff
@@ -28,7 +28,7 @@
             <input
               class="mr-2 cursor-pointer"
               type="checkbox"
-              [checked]="(limit$ | async) === '-1'"
+              [checked]="(limit$ | async) === '-1' || (limit$ | async) === ''"
```
tkohr, line 31: I think we could introduce two Observables for the inputs that use logic in this template. This would prevent multiple subscriptions for the same logic and multiplication of the same expression if we need to override the template.

tkohr (commented):
Thanks for the fix @cmoinier ! It works fine now. I left one comment regarding the template. Maybe we could also leverage this PR to resolve the typing issues we had in MEL and apply the same changes here.

tkohr (approved):
Thanks for the adaptions @cmoinier !

PR #826, opened 2024-03-19
[Datahub][Redesign] : Redesign the about panel

cmoinier:
### Description

This PR changes the styling and adds some new data to the "about the data panel". It introduces the variables "languages" and "publication date" in the mapper.

### Metadata to test : 

`/dataset/04bcec79-5b25-4b16-b635-73115f7456e4` --> for complete data
`/dataset/ee965118-2416-4d48-b07e-bbc696f002c2` --> dataset without an end date for its temporal range
`/dataset/8698bf0b-fceb-4f0f-989b-111e7c4af0a4` --> dataset without a start date for its temporal range

### Screenshots

image


libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -73,6 +73,37 @@ export class MetadataInfoComponent {
     }
   }
 
+  get temporalExtent(): string {
```
jahow, line 76: This should probably not be a string

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -55,29 +55,89 @@ <h5 translate class="font-medium text-black text-sm mb-[2px] mt-[16px]">
...
+    class="py-6 px-6 rounded bg-gray-100 grid grid-cols-2 gap-y-6 gap-x-[20px] text-gray-700"
   >
-    <div *ngIf="metadata.recordUpdated">
-      <p class="text-sm" translate>record.metadata.updatedOn</p>
+    <div *ngIf="metadata.recordCreated">
+      <p class="text-sm" translate>record.metadata.creation</p>
       <p class="text-primary font-medium mt-1">
-        {{ metadata.recordUpdated && metadata.recordUpdated.toLocaleString() }}
+        {{
+          metadata.recordCreated && metadata.recordCreated.toLocaleDateString()
```
jahow, line 132:
```suggestion
          metadata.recordCreated.toLocaleDateString()
```

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -55,29 +55,88 @@ <h5 translate class="font-medium text-black text-sm mb-[2px] mt-[16px]">
...
-  <p
+  <div
     *ngIf="metadata.lineage"
-    class="mb-5 pt-4 whitespace-pre-line break-words text-gray-900"
-    gnUiLinkify
+    class="text-gray-900 flex flex-col mb-5 pt-4 gap-2"
   >
-    {{ metadata.lineage }}
-  </p>
+    <p class="inline-block" translate>record.metadata.genealogy</p>
```
jahow, line 70: you can get rid of this header I think

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
jahow: Also two more cases need to be handled: when there is only a start date, and only an end date

jahow (commented):
Looking good!

jahow (approved):
Thanks for all the base work @cmoinier. It turns out modifying the "alipe convention" record in the GN editor will break all its keywords, which is why you had to adapt the datasets.cy test originally. I have omitted the changes on that record, and only kept the changes on INSEE and SCOT records.

Also I've made the history hopefully more clear :) there's still work to do on the ISO19139 converter but I'm working on it for the editor so for now it can stay like that.

I'll merge when it's green!! :rocket:

PR #827, opened 2024-03-19
Editor: rework save/publish logic, add top toolbar

jahow:
### Description

This PR brings the following changes to the `/edit/{uuid}` route:
* The placeholders from the previous layout were removed
* A top toolbar now shows with the save/publish button
* The publish button will show whether something changed in the form and the record can be saved again
* When saving a record, the `recordUpdated` field will be set to the current time

Note that errors when saving do not show up yet, they will be handled in another PR.

### Architectural changes

A new feature state was introduced for the `feature-edit` library, handling:
* opening a record
* saving a record
* showing a list of fields according to a configuration
* running processes (i.e. expressions) on save; this is the case for `recordUpdated` which has the expression `${dateNow()}`

Also buttons have received a significant update and can now be customized with css variables:
* `gn-ui-button-padding`
* `gn-ui-button-font-size`
* `gn-ui-button-background`
* `gn-ui-button-rounded`
* `gn-ui-button-width`
* `gn-ui-button-height`

This lets customize buttons based on the context without resorting to hacky "extra classes". The button storybook was updated to showcase more possibilites.

### Screenshots

image


libs/feature/editor/src/lib/+state/editor.reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+    changedSinceSave: false,
+  })),
+  on(EditorActions.saveRecordFailure, (state, { error }) => ({
+    ...state,
+    saving: false,
+    saveError: error,
+  })),
+  on(EditorActions.updateRecordField, (state, { field, value }) => ({
+    ...state,
+    record: {
```
LHBruneton-C2C, line 55: Why not set `changedSinceSaved` here?

libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -65,43 +31,19 @@ export class EditorService {
...
-  setCurrentRecord(record: CatalogRecord) {
-    this.record$.next(record)
-  }
-
-  updateRecordField(fieldName: string, value: unknown) {
-    this.record$
-      .pipe(
-        take(1),
-        map((record) => ({ ...record, [fieldName]: value }))
+  // TODO: use the catalog repository instead
```
LHBruneton-C2C: By this TODO, do you mean "use the OpenAPI generated client"?

libs/ui/inputs/src/lib/form-field/form-field.component.ts
```diff
@@ -17,7 +17,9 @@ import { Observable } from 'rxjs'
 export class FormFieldComponent {
   @Input() config: FormFieldConfig
   @Input() set value(v: unknown) {
-    this.formControl.setValue(v)
+    this.formControl.setValue(v, {
+      emitEvent: false,
```
LHBruneton-C2C, line 21: I'm curious, what is this for?

libs/feature/editor/src/lib/expressions.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('expressions', () => {
+  let evaluator: ExpressionEvaluator
+  describe('operators', () => {
+    describe('dateNow', () => {
+      beforeEach(() => {
+        evaluator = evaluate('${dateNow()}')
+      })
+      it('returns the current time at evaluation', () => {
+        expect(evaluator({ config: SAMPLE_CONFIG, value: 'bla' })).toEqual(
+          new Date()
```
LHBruneton-C2C, line 26: You don't need to mock the Date API for the test to "always" work?

Edit: fixed in last commit

libs/feature/editor/src/lib/+state/editor.effects.ts
```diff
@@ -16,10 +20,18 @@ export class EditorEffects {
...
+      withLatestFrom(
+        this.store.select(selectRecord),
+        this.store.select(selectRecordFieldsConfig)
+      ),
+      switchMap(([, record, fieldsConfig]) =>
+        this.editorService.saveRecord(record, fieldsConfig).pipe(
+          switchMap((newRecord) =>
+            of(
+              EditorActions.saveRecordSuccess(),
+              EditorActions.openRecord({ record: newRecord })
```
LHBruneton-C2C, line 28: A bit overkill opening what you just saved, no? Or there might be some changes made by the backend?

libs/feature/editor/src/lib/+state/editor.selectors.spec.ts
```diff
@@ -36,5 +38,32 @@ describe('Editor Selectors', () => {
       const result = EditorSelectors.selectRecordChangedSinceSave(state)
       expect(result).toBe(true)
     })
+
+    it('selectRecordFieldsConfig() should return the current "fieldsConfig" state', () => {
+      const result = EditorSelectors.selectRecordFieldsConfig(state)
+      expect(result).toEqual(DEFAULT_FIELDS)
+    })
+
+    it('selectRecordFields() should return the current "fieldsConfig" state', () => {
```
LHBruneton-C2C, line 47: Same "should" as above?

libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -31,19 +33,36 @@ export class EditorService {
...
+    // run onSave processes
+    for (const field of fieldsConfig) {
+      if (field.onSaveProcess && field.model) {
+        const evaluator = evaluate(field.onSaveProcess)
+        savedRecord[field.model] = evaluator({
+          config: field,
+          value: record[field.model],
+        })
+      }
+    }
```
LHBruneton-C2C, line 52: New critical feature, not tested?

Edit: fixed in last commit

LHBruneton-C2C (approved):
Nice work! I'm approving already, as none of my remarks are critical.
I've learned a bit more about GN-UI with this PR.

libs/feature/editor/src/lib/+state/editor.effects.ts
```diff
@@ -0,0 +1,41 @@
...
+      )
+    )
+  )
+
+  markAsChanged$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(EditorActions.updateRecordField),
+      map(() => EditorActions.markRecordAsChanged())
+    )
+  )
```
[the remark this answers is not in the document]
jahow: So my approach here was to avoid putting what could be considered "business logic" inside the reducer. I have the impression that it's good practice to have a very simple reducer, and avoid handling side-effects like that there. This is why I went with an effect, although I agree that it's debatable.

I think I like the fact that an effect is more explicit, then there's no "magic" happening in the reducer

libs/feature/editor/src/lib/+state/editor.reducer.ts
jahow: See above

libs/ui/inputs/src/lib/form-field/form-field.component.ts
jahow: This is to avoid having an event emitted out of the component when we set its value; otherwise what happens is that the value of the field is changed once the record is loaded, and so the record is "marked as changed", which means that it can be saved although it hasn't received any change yet

libs/feature/editor/src/lib/+state/editor.selectors.spec.ts
jahow: I used copilot a bit too much :grimacing:

jahow:
Thanks for the review @LHBruneton-C2C :slightly_smiling_face:

PR #832, opened 2024-03-22
chore: make bg white in dropdowns, force to transparent for language switcher

jahow:
before:

image

after:

image


tkohr (approved):
Thanks @jahow !

jahow (commented):
Thanks @tkohr!!

PR #843, opened 2024-04-05
Metadata Editor / Implement iso19115-3 converter

jahow:
### Description

This PR introduces a new `Iso191153Converter` class deriving from the `Iso19139Converter` class.

Some elements are encoded differently in iso19115-3 so new reading/writing logic for introduced. Many other elements could be kept from iso19139 by just replacing the namespaces after the XML generation.

To do:
- [x] implement a way to automatically detect what iso schema is used in an XML doc

### Architectural changes


libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
```diff
@@ -1,6 +1,197 @@
...
+    extractCharacterString()
+  )
+  const getNameParts = pipe(
+    getIndividual,
+    findChildElement('cit:name'),
+    extractCharacterString(),
+    map((fullName) => {
+      if (!fullName) return []
+      const parts = fullName.split(/\s+/)
+      if (!parts.length) return [fullName, null]
```
LHBruneton-C2C, line 101: If `fullName` is truthy, I don't see how `parts.length` could be falsy here.

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
```diff
@@ -1,6 +1,197 @@
...
+      const first = parts.shift()
+      return [first, parts.join(' ')]
```
LHBruneton-C2C, line 103: Out of curiosity, and for further reference, is it the standard to split only on the first space?

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
```diff
@@ -1,6 +1,197 @@
...
+    mapArray(findNestedElement('cit:phone', 'cit:CI_Telephone', 'cit:number')),
+    mapArray(extractCharacterString())
+  )
+  const getEmails = pipe(
+    getAddressRoots,
+    mapArray(
+      pipe(
+        findChildElement('cit:electronicMailAddress', false),
+        extractCharacterString(),
+        map((email) => (email === null ? 'missing@missing.com' : email))
```
LHBruneton-C2C: Wouldn't testing for truthy email be more prudent? What about spaces?

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
```diff
@@ -1,6 +1,197 @@
...
+        organization,
+        emails,
+        addresses,
+        phones,
+      ]) =>
+        emails.map((email, i) => ({
+          email,
+          role,
+          organization,
+          ...(position && { position }),
```
LHBruneton-C2C: What's the purpose of this syntax? Only adding the property if it has a value?

LHBruneton-C2C (approved):
Thanks for this PR.
Nothing much to say, this didn't look like fun dev ^^'

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
jahow: exactly! setting a property as undefined and not setting it are two different things, and having properties that are present but undefined can be confusing

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
jahow: that would be up to the quality of the metadata; this is a text field so it can contain anything, even an empty string; in that case, I think it's better to keep the email as an empty string and have it show up like so, because that's what the metadata really contains

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
jahow: Right, good point... I was also wondering about that myself. It's not standard at all and depending on cultures it can give wrong results. On the other hand this is not critical, as the firstName and lastName fields are just there for convenience and are concatenated back when writing (actually I even wonder if it's a good idea to have those and not just a "fullName" field...)

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
jahow: good point! I'll see if I can also fix and factorize this with the iso19139 converter

libs/api/metadata-converter/src/lib/iso19139/write-parts.ts
```diff
@@ -238,6 +239,50 @@ export function appendResponsibleParty(contact: Individual) {
...
+      )
+    ),
+    contact.address
+      ? appendChildren(
+          pipe(
+            createElement('gmd:deliveryPoint'),
+            writeCharacterString(contact.address)
+          )
+        )
+      : noop
```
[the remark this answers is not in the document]
jahow: I had never heard of this!! will try :slightly_smiling_face:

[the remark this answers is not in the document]
jahow: no wait, `nop` is only in webassembly right? https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Control_flow/nop

jahow:
thanks for the review :) i won't lie, it wasn't the funnest stuff I've ever done

PR #847, opened 2024-04-09
e2e(datahub): wait more precisely for favorite to be registered

jahow:
Before that the filtering on favorites was sometimes done too early

### DIFF
--- a/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
+++ b/libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
@@ -1,6 +1,7 @@
 <div class="flex flex-row items-center">
   <span
-    class="align-text-top mr-1.5 favorite-count"
+    class="align-text-top mr-1.5"
+    data-test="favorite-count"
     *ngIf="hasFavoriteCount && displayCount"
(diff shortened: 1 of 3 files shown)


libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="flex flex-row items-center">
   <span
-    class="align-text-top mr-1.5 favorite-count"
+    class="align-text-top mr-1.5"
+    data-test="favorite-count"
```
tkohr, line 4: We are already using `data-cy` in other locations. Maybe it's worth to name it the same. Then again I see it's also used in the UTs, so I let you judge :-)

tkohr (approved):
Thanks, @jahow, looks like a promising solution!

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
jahow: exactly, it used in the UT so I gave priority to `data-test` :)

jahow:
Thanks for the quick review!!

PR #849, opened 2024-04-10
[Datahub]: add comment and discussion module to record page

rcaplier:
### Description

This PR introduces add the comment and discussion module to the record page.

### Changes description

I had to change some things in the lib ui to fit the mockups : 
- added the possibility to override css classes in the gn-ui-text-input and gn-ui-text-area component as it is already done for the gn-ui-button component


libs/ui/inputs/src/lib/text-area/text-area.component.ts
```diff
@@ -24,6 +31,30 @@ export class TextAreaComponent implements AfterViewInit {
...
+
+    this.disabledClasses = [
+      'cursor-not-allowed'
+    ].join(' ')
+  }
+
+  get classList() {
+    return `${this.baseClasses} ${this.extraClass} ${this.disabled ? this.disabledClasses	: ''}`
+  }
+
```
jahow, line 57: This can be done using a `disabled:` modified in TW: https://tailwindcss.com/docs/hover-focus-and-other-states#form-states

support-services/.env
```diff
@@ -1 +1 @@
-GEONETWORK_VERSION=4.2.2
+GEONETWORK_VERSION=4.2.8
```
jahow, line 1: This is important because we are supposed to maintain compatibility with the 4.2.2 version. If there is no way to have the system work with 4.2.2 then we have the option to disable feedbacks but it has to be justified.

libs/ui/elements/src/lib/user-feedback-item/time-since.pipe.ts
```diff
@@ -0,0 +1,48 @@
...
+    } else if (hours < 24) {
+      return this.translate.instant('timeSincePipe.hoursAgo', { value: hours, s: hours > 1 ? 's' : '' });
+    } else if (days < 30) {
+      return this.translate.instant('timeSincePipe.daysAgo', { value: days, s: days > 1 ? 's' : '' });
+    } else if (months < 12) {
+      return this.translate.instant('timeSincePipe.monthsAgo', { value: months });
+    } else {
+      return this.translate.instant('timeSincePipe.yearsAgo', { value: years, s: years > 1 ? 's' : '' });
+    }
+  }
```
jahow, line 47: This can probably be done by a dependency, although it can also make sense to reimplement it if no dependency is light/fitting enough :)

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  userFeedbacks$ = this.mdViewFacade.userFeedbacks$
+
+  addUserFeedback(userFeedback: UserFeedback) {
+    return this.mdViewFacade.addUserFeedback(userFeedback)
+  }
+
+  loadUserFeedbacks(metadataUuid: string) {
+    return this.mdViewFacade.loadUserFeedbacks(metadataUuid)
```
jahow, line 24: the component can use the facade directly for that

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  sortByDateFromNewestToOldest(userFeedbackA: UserFeedback, userFeedbackB: UserFeedback): number {
+    return new Date(userFeedbackB.date).getTime() - new Date(userFeedbackA.date).getTime()
+  }
```
jahow, line 29: using Date objects will make this logic simpler, and it can probably go back in the component

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  sortByDateFromOldestToNewest(userFeedbackA: UserFeedback, userFeedbackB: UserFeedback): number {
+    return new Date(userFeedbackA.date).getTime() - new Date(userFeedbackB.date).getTime()
+  }
+
+  createViewModel(baseUserFeedback: UserFeedback): UserFeedbackViewModel {
+    const userAvatarUrl = this.gravatarService.getProfileIcon(baseUserFeedback.authorUserId?.toString())
+
+    return {
+      ...baseUserFeedback,
+      avatarUrl: userAvatarUrl
```
jahow, line 40: In principle this should be provided by the platform alongside the rest of the feedback. This might require quite a bit of refactoring

jahow (commented):
Thanks you, it's looking really good so far!

cmoinier (changes requested):
First of all, very well done on your first PR @rcaplier :tada: 
It's a really good work. I have left some comments, most of them are about missing unit tests, languages and e2e improvement. Let me know if you have any questions :smiley:

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -1,24 +1,24 @@
-<div *ngIf="(facade.error$ | async) === null">
+<div *ngIf="(metadataViewFacade.error$ | async) === null">
```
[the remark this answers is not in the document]
rcaplier: I struggled a bit as a new commer because it wasn't explicit enough for me.
I don't need to make my brain work to find what I'm manipulating, an extrem opposite would have been to rename it `mdvf `.

It's my personnal point of view, I' down to follow the naming convention if there is any ?

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
```diff
@@ -688,3 +688,83 @@ describe('api form', () => {
...
+    })
+    it('should sort comments', () => {
+      cy.get('gn-ui-user-feedback-item')
+        .find('[data-cy="commentText"]')
+        .as('commentText')
+
+      cy.get('@commentText')
+        .first()
+        .then((div) => {
+          const premierCommentaireAvantTri = div.text().trim()
```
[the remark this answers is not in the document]
rcaplier: 1 year working on a french project hurts :/

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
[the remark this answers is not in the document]
cmoinier: I understand that feeling! But I would say that usually we try to not rename the existing things (especially if they are used multiple times), unless there is a specific reason (ie if there could be a confusion with something that you added). But maybe it's just my own feeling. WDYT @jahow @Angi-Kinas @tkohr ?

libs/ui/elements/src/lib/user-feedback-item/user-feedback-item.component.html
```diff
@@ -0,0 +1,73 @@
+<div
+  *ngIf="userFeedbackParent.published"
+  class="flex flex-col bg-white rounded w-full"
+  [ngClass]="[isAnAnswer ? 'ps-4 ' : 'p-4']"
```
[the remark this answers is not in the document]
cmoinier: Interesting! I didn't know it :thinking:

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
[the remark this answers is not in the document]
jahow: IMO this can stay like this, it's more explicit now.

cmoinier (approved):
Thanks for the rework @rcaplier, well done ! :smiley:

PR #857, opened 2024-04-16
Allow disabling cdn from datahub and datafeeder

f-necas:
### Description

This PR introduces removes CDN resources of Datahub, Datafeeder and Metadata-editor.

---


jahow (commented):
Why not make all these fonts local once and for all? this shouldn't change anything for users using the default fonts as well as for the ones using custom fonts.

jahow (changes requested):
With the current PR the Datahub and Editor apps still rely on google fonts, because of this:
https://github.com/geonetwork/geonetwork-ui/blob/6083baae75be83fec080c85bed3643e73c93ac48/apps/datahub/src/app/app.module.ts#L216-L217

The stylesheet used for fonts should be configurable and not hardcoded in the HTML (this is not the case for icons font, which is always loaded no matter what). When I remove the google fonts stylesheet, the default-fonts.css stylesheety doesn't work because the paths in the CSS are not right.

I'm also wondering about the format; I feel like using "woff2" fonts would make more sense as it is a format optimized for the web. When looking at https://fonts.googleapis.com/css2?family=Readex+Pro&family=Rubik&display=swap (the default font stylesheet), there is also much more to it than just the current content of default-fonts.css. We should maybe just copy all this into the assets of the project, just to make sure we don't have invalid characters somewhere down the road?

Thanks!

f-necas:
When we hit https://fonts.googleapis.com/css2?family=Readex+Pro&family=Rubik&display=swap there are some font face for arabic, cyrillic, hebrew and vietnamese chars. Should we implement them also ? As they are not implemented in those apps. 

I can implement the woff by retrieving from the url provided in this css file but, when we download it from the google font, I got only the ttf. 

Is it ok for you if I put only an empty string ? 
```
      getThemeConfig().FONTS_STYLESHEET_URL || ''
```

jahow:
> When we hit https://fonts.googleapis.com/css2?family=Readex+Pro&family=Rubik&display=swap there are some font face for arabic, cyrillic, hebrew and vietnamese chars. Should we implement them also ? As they are not implemented in those apps. Also I think this ttf implement all chars instead of google's woff2 which seems to be small parts of the font. Indeed they should be more optimized and a lot smaller.

Honestly I think you can just copy paste the whole CSS and not worry to much about it; we don't have such languages for now but there's no reason for it to stay that way (e.g. core-geonetwork now supports georgian, ukrainian etc.).

> I can implement the woff by retrieving from the url provided in this css file but, when we download it from the google font official link, I got only the ttf.

I would go for woff2 to replicate the behavior that we had before, just to make sure we don't break anything. thanks :)

> Is it ok for you if I put only an empty string ?
> 
> ```
>       getThemeConfig().FONTS_STYLESHEET_URL || ''
> ```

I'm not sure I see the logic; users can decide to use a different stylesheet with different fonts, why would we also include the default one then? I would keep the same behaviour as before, e.g. pointing on a default stylesheet.

jahow (approved):
I have added a commit with a few fixes and some doc, and rebased on main (to hopefully fix the e2e test failing)

Merge when you see fit :) PS: this should be squashed

PR #861, opened 2024-04-22
[Datahub]: Api form formats output

cmoinier:
### Description

This PR adds the output formats dynamically to the API form, with the use of the `ogc-client` lib.
It also adds more MIME formats to the formats parser.

### Screenshots

image


apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
```diff
@@ -632,7 +632,7 @@ describe('api form', () => {
     cy.get('@apiForm').find('input[type="checkbox"]').check()
     cy.get('@apiForm').find('gn-ui-text-input').first().should('have.value', '')
   })
-  it('should reset all 3 inputs and link on click', () => {
+  it.only('should reset all 3 inputs and link on click', () => {
```
Angi-Kinas, line 635: There's an `.only` here ;-)

libs/util/shared/src/lib/links/link-utils.ts
```diff
@@ -94,6 +94,33 @@ export const FORMATS = {
     color: '#de630b',
     mimeTypes: ['application/x-dxf', 'image/x-dxf'],
   },
+  html: {
+    extensions: ['html', 'htm'],
+    priority: 12,
+    color: '#f2bb3a',
+    mimeTypes: ['text/html'],
+  },
+  fgb: {
```
jahow, line 103: these keys are also used for display I think, so maybe going with "flatgeobuf" would make it clearer :)

libs/util/shared/src/lib/links/link-utils.ts
```diff
@@ -94,6 +94,33 @@ export const FORMATS = {
...
+      'application/vnd.ogc.fg+json',
+      'application/vnd.ogc.fg+json;compatibility=geojson',
+    ],
+  },
+  pap: {
+    extensions: ['pap'],
+    priority: 15,
+    color: '#f2bb3a',
+    mimeTypes: ['application/vnd.policy.attributes'],
+  },
```
jahow, line 123: not a relevant format, to remove!

libs/ui/elements/src/lib/record-api-form/record-api-form.component.ts
```diff
@@ -70,4 +71,34 @@ export class RecordApiFormComponent {
     this.limit$.next(DEFAULT_PARAMS.LIMIT)
     this.format$.next(DEFAULT_PARAMS.FORMAT)
   }
+
+  parseOutputFormats() {
+    const apiUrl =
+      this.apiBaseUrl.slice(-1) === '?'
+        ? this.apiBaseUrl.slice(0, -1)
+        : this.apiBaseUrl
```
jahow, line 79: does that cause an error down the line? having a trailing "?" should not break anything in principle

libs/ui/elements/src/lib/record-api-form/record-api-form.component.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { TestBed, ComponentFixture } from '@angular/core/testing'
 import { RecordApiFormComponent } from './record-api-form.component'
 import { DatasetServiceDistribution } from '@geonetwork-ui/common/domain/model/record'
-import { firstValueFrom } from 'rxjs'
+import { firstValueFrom, of } from 'rxjs'
```
Angi-Kinas, line 4: I think the of import is not used and can be removed.

libs/util/shared/src/lib/links/link-utils.ts
cmoinier: Okay ! :) But won't it put "flatgeobuf" as a format in the url? i tested and `https://demo.ldproxy.net/zoomstack/collections/airports/items?f=flatgeobuf` returns an error, while  `https://demo.ldproxy.net/zoomstack/collections/airports/items?f=fgb` is alright (its a download though, but it will be relevant for the other ticket).

But maybe my understanding of this function is wrong! I understand that the key is used as the format name.

Angi-Kinas (commented):
Thanks @cmoinier. Just found small things. Otherwise LGTM :+1:

libs/ui/elements/src/lib/record-api-form/record-api-form.component.ts
cmoinier: Yes. it creates an error when `ogc-client` tries to get to the conformance document and ends up here : `https://mel.integration.apps.gs-fr-prod.camptocamp.com/data/ogcapi?%2F=&f=json`

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
cmoinier: Good catch!

libs/ui/elements/src/lib/record-api-form/record-api-form.component.spec.ts
cmoinier: Indeed, thanks!

libs/ui/elements/src/lib/record-api-form/record-api-form.component.ts
jahow: then I guess this is an issue with ogc-client, I'll see if I can reproduce on its side

libs/util/shared/src/lib/links/link-utils.ts
jahow: hmm, maybe that approach is wrong then; ogc-client offers a "getCollectionItemsUrl" function, this function could be responsible for taking in a mime type and generting a correct url out of it, instead of the application being responsible for appending a "f=" format param itself

cmoinier: Yes, but in this case, the whole logic of the api-form would have to be changed :grimacing: The format, limit and offset being added dynamically when the input values change. Would it mean that the baseUrl would now come from getCollectionItemsUrl? Wouldn't it be easier to have a `displayName` for each mimeType, instead of using the key?

jahow: Then I guess we can stay with fgb for now; in principle I would say the url generation should be the responsibility of ogc-client here, but this can be done later

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
```diff
@@ -641,14 +641,16 @@ describe('api form', () => {
     cy.get('@secondInput').clear()
     cy.get('@secondInput').type('87')
 
-    cy.get('@apiForm').find('gn-ui-dropdown-selector').click()
-    cy.get('button[data-cy-value="csv"]').click()
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.wait(3000)
```
jahow, line 645: Instead of adding this wait time, it might be better to intercept & mock the results from the test API. But this can be done in a followup PR

jahow (commented):
Hey @cmoinier, I think this PR is good to go, thanks! There is a conflicting change on the dump, I hope it won't be too much trouble to reproduce the changes in the database. Please merge when this is rebased & green, thanks!

PR #872, opened 2024-05-07
Datahub: OGC API fixes

tkohr:
### Description

This PR applies a couple of fixes for the OGC API integration into the datahub:

- Prevent the download list from breaking when WFS and OGC API links are present and the latter do not succeed due to CORS => visible in the SCOT test dataset
- Fix a bug on app load in the OGC API form
- Recover the message displayed for download links generated by WFS 

- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label => I think the bug fixes are for features introduced after the latest release (besides the recovered message)


libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -156,20 +156,25 @@ export class DataService {
...
+            type: 'download',
+            url: new URL(collectionInfo.bulkDownloadLinks[downloadLink]),
+            mimeType: getMimeTypeForFormat(
+              getFileFormatFromServiceOutput(downloadLink)
+            ),
+          }
+        }
+      )
+    } catch (error) {
+      return Promise.resolve([])
```
jahow, line 176: I think the error can be caught much later, this is already the case for WFS for instance. The errors are caught here: https://github.com/geonetwork/geonetwork-ui/blob/3371b676033bae088ee0423b96863a4eed463e2a/apps/datahub/src/app/record/record-downloads/record-downloads.component.ts#L78

What's happening is maybe that
1/ even if an error is caught it is not shown (could be shown using an overlay for instance?)
2/ when an error is caught, it is assumed that the error comes from wfs; ut now, it can come from somewhere else, so the logic has to be adjusted

jahow: I think this looks better now, we could probably have a common `handleError` callback which assigns the error and returns `of([])` and use that both for WFS and OGC API. I'm not sure using a Promise is needed here anyway? But yes to have the equivalent of `of` you probably need to do `from(Promise.resolve(...))`, that makes sense to me.

tkohr: Hum, I was not able to refactor this further with a common `handleError` function. Any attempt I did to try to catch the error once the `Promise` has been transformed into en `Obervable` broke the download list, so I needed to keep catching it within the `Promise`.
I've added a separate translation key for the OGC API error. If I saw right, `ogc-client` does not provide more info on the error as for `WFS`, so there is just one generic message so far. It is displayed in the preview components like the map, but I decided to not also add it to the download list to not "pollute" the UI when there might be many links already displayed from working sources.

jahow (approved):
thank you! not tested but it looks really good :+1:

PR #874, opened 2024-05-13
Datahub - add info when metadata has no online ressource

rcaplier:
### Description

This PR introduces a new warning block for when a record dataset has neither API, download or other links to display.

### Screenshots

image

---

**This work is sponsored by Swisstopo geocat**.


libs/ui/elements/src/lib/search-results-error/search-results-error.component.html
```diff
@@ -13,14 +16,29 @@
...
     <div translate>search.error.couldNotReachApi</div>
   </div>
-  <div *ngIf="type === types.RECEIVED_ERROR">
+  <div *ngIf="type === types.NO_LINK" data-test="no-link-error">
+    <div class="relative opacity-40">
+      <mat-icon class="material-symbols-outlined face">computer</mat-icon>
+      <mat-icon class="material-symbols-outlined question-mark1"
+        >question_mark</mat-icon
+      >
+      <mat-icon class="material-symbols-outlined question-mark2"
```
cmoinier, line 25: What are these two question marks icons for ? They are not in the screenshot you provided.

translations/fr.json
```diff
@@ -332,6 +332,7 @@
   "search.error.couldNotReachApi": "Problème de connexion à l'API",
   "search.error.receivedError": "Erreur retournée",
   "search.error.recordNotFound": "Cette donnée n'a pu être trouvée.",
+  "search.error.nolink": "Ce dataset n'a pas encore de lien, réessayez plus tard s'il vous plaît.",
```
cmoinier, line 335: I would've made "liens" plural :slightly_smiling_face:

cmoinier (changes requested):
Thanks for this feature @rcaplier ! :smiley: It looks really good already. Two remarks/questions under here + would be nice to add e2e testing (difficult to do for other errors, but for the absence of links, you would just have to point to a dataset where you would have removed all the links, perhaps there is already an existing one that you used for this PR? LMK if you need help with that).

translations/fr.json
cmoinier: Mmmh seen like that, yes it makes sense for the singular, my bad!

libs/ui/elements/src/lib/search-results-error/search-results-error.component.html
rcaplier: Wrong copy/pasta. They don't even display 😆 
Fixed it. Good catch !

rcaplier:
> Thanks for this feature @rcaplier ! 😃 It looks really good already. Two remarks/questions under here + would be nice to add e2e testing (difficult to do for other errors, but for the absence of links, you would just have to point to a dataset where you would have removed all the links, perhaps there is already an existing one that you used for this PR? LMK if you need help with that).

Thank you. I've had to make a new dataset with no link in it to see the result. Do I have to make another dump so we have it and point to it in the e2e tests ? Or maybe I can cheat in the e2e by simulating a record with no link ? 😎

cmoinier:
> > Thanks for this feature @rcaplier ! 😃 It looks really good already. Two remarks/questions under here + would be nice to add e2e testing (difficult to do for other errors, but for the absence of links, you would just have to point to a dataset where you would have removed all the links, perhaps there is already an existing one that you used for this PR? LMK if you need help with that).
> 
> Thank you. I've had to make a new dataset with no link in it to see the result. Do I have to make another dump so we have it and point to it in the e2e tests ? Or maybe I can cheat in the e2e by simulating a record with no link ? 😎

Yes you can use that one! I think simulating a record without any links would be too time consuming, I would use the one you got :wink:

jahow (commented):
Just a general comment, could we maybe be more specific and instead of "noLink" use "recordHasNoLink"? E.g. `search.error.RecordHasNoLink`, `ErrorType.RECORD_HAS_NO_LINK`, etc.

cmoinier (approved):
Thanks for the last changes, LGTM :ok_hand:

PR #875, opened 2024-05-14
[Datahub] Allow building custom URL for WFS service

ronitjadhav:
### Description

This pull request introduces a new feature that allows users to generate a URL for downloading a dataset from a WFS endpoint with specific parameters. The changes include:

- [x] UI updates to include a form for WFS parameters, replacing the "copy link" button with a "..." button that opens this form.
- [x] Updates to the WfsEndpoint class in the ogc-client library to include a supportsStartIndex() method.
- [x] Updates to the method in ogc-client that generates the URL for downloading the dataset to include the parameters specified in the form.
- [x] Updates to the method that retrieves output formats from the WFS service to replace mime types with human-readable labels.
- [x] If the WFS version does not support the starting index parameter, it should appear as greyed out with an orange warning icon next to it and a tooltip telling the user that this is not supported on this service

For output format only "known" mime types (i.e. the ones that match a format already present in gn-ui) are shown.

### Screenshots

image

If the service does not support startIndex

image

---

**This work is sponsored by [Organization ABC](xx)**.


libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
```diff
@@ -37,15 +37,37 @@
...
+        <p class="text-sm" [class.text-gray-600]="!supportOffset" translate>
+          record.metadata.api.form.offset
+        </p>
+        <div class="flex items-center">
+          <gn-ui-text-input
+            class="w-20"
+            [value]="offset$ | async"
+            [extraClass]="
+              supportOffset ? '' : 'text-gray-600 bg-gray-200 opacity-70'
+            "
```
jahow, line 50: can't you simply make this component disabled instead of giving it css classes?

libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
```diff
@@ -37,15 +37,37 @@
...
+            (valueChange)="supportOffset ? setOffset($event) : null"
+            hint=""
+          >
+          </gn-ui-text-input>
+          <div
+            *ngIf="!supportOffset"
+            class="flex items-center gap-2 text-red-500 z-10 ml-3 cursor-pointer"
```
jahow, line 57: maybe an orange "warning" color would be more fitting here? also the pointer cursor isn't necessary IMO, since there's no possible interaction

libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
```diff
@@ -37,15 +37,37 @@
...
+              matTooltip="Not supported on this service"
+            >
+              warning
+            </span>
+          </div>
+        </div>
+        <div
+          *ngIf="!supportOffset"
+          class="absolute inset-0 bg-transparent cursor-not-allowed"
+        ></div>
```
jahow, line 70: I'm not sure this is really necessary; if the text input is disabled and the warning icon shows up with a mat-tooltip that should be enough in terms of feedback

jahow (approved):
Thanks! I've made a few suggestions regarding the DOM of the offset text input, adapt as you see fit. The rest is OK, I think the logic to handle both OGC API and WFS could be made clearer with less ifs involved, but that can be done later on (since there's another PR coming for the OGC API url generation)

libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
ronitjadhav: I don't think that's possible as the `gn-ui-text-input` component does not have a property to disable.

libs/ui/elements/src/lib/record-api-form/record-api-form.component.html
ronitjadhav: The text input can't be directly disabled, but this provides an alternative to prevent user interaction.

PR #878, opened 2024-05-17
ME: support temporal extents in iso19139

LHBruneton-C2C:
### Description

This PR introduces support of temporal extent in the iso19139 converter.

Useful resource : http://www.datypic.com/sc/niem21/e-gmd_EX_Extent.html

### DIFF
--- a/libs/common/domain/src/lib/model/record/metadata.model.ts
+++ b/libs/common/domain/src/lib/model/record/metadata.model.ts
@@ -169,12 +169,12 @@ export interface DatasetSpatialExtent {
 }
 
 /**
- * At least a start or an end date should be provided
+ * Period if both start and end are provided
+ * Instant if only start is provided
  */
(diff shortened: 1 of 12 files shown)


libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -174,7 +174,6 @@ export interface DatasetSpatialExtent {
 export interface DatasetTemporalExtent {
   start?: Date
```
jahow, line 175: it looks like this is actually a required prop, since a time instant will be written using that field

jahow (approved):
looking good! maybe the functions for reading/writing can be made simpler but that can be done later. Was there no spatial extent in the existing fixtures at all? Maybe we should at least add some in the generic records (which are not from the "real world" but made by hand) to at least see it happen on full records :) thanks!

PR #879, opened 2024-05-21
Datahub: use the complete symbols font locally

jahow:
This is a followup of #857 

The local font use to be otf and not contain every symbol in every shape; this is fixed with this PR.

Before:
image

After:
image


jahow:
@f-necas I can't remember why we let than one slip through; let me know if you see anything that I don't, thanks :)

Angi-Kinas (approved):
Thanks @jahow !

PR #887, opened 2024-05-28
MEL - datahub - tile WMS

rcaplier:
### Description

This PR introduces a new parameter in the default.toml config file: `do_not_tile_wms` which is false by default.

### Screenshots

- When the parameter is set to false (by default), we can see on this dataset in the network that 15 tiles are fetched :

image

- When the parameter is set to true, we can see that only one image is fetched :

image

---

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -141,6 +141,9 @@ background_color = "#fdfbff"
 # Optional; if true, opens external viewer in new tab (default false)
 # external_viewer_open_new_tab = false
 
+# Optional; Will not tile WMS. False by default.
+# do_not_tile_wms = false
+
 # Optional; if true, the default basemap will not be added to the map.
 # Use [[map_layer]] sections to define your own custom layers (see below)
 # do_not_use_default_basemap = false
--- a/docs/guide/configure.md
+++ b/docs/guide/configure.md
@@ -214,6 +214,10 @@ The map section lets you customize how maps appear and behave across GeoNetwork-
   max_extent = [-418263.418776, 5251529.591305, 961272.067714, 6706890.609855]
   ```
 
+- `do_not_tile_wms` (optional)
+
+  Will not use tiling when requesting WMS services. Defaults to `false` (WMS are tiled). Not using tiles for WMS might incur performance loss since the client will not benefit from an eventual tile cache anymore. On the other hand, visual quality might improve in case a map tile server does not handle neighbouring tiles correctly, e.g. symbols or text being cropped at tile boundaries. This can be set true to prevent visual conflicts on tile borders, if the WMS server does not add a gutter, for example. gn-ui does not add a gutter on the client side, in order to allow server-side caching.
+
 - `do_not_use_default_basemap` (optional)
 
   If set to `true`, the default basemap will not be added to the map. Defaults to `false` (base map is shown).
--- a/libs/feature/map/src/lib/map-context/map-context.service.ts
+++ b/libs/feature/map/src/lib/map-context/map-context.service.ts
@@ -29,6 +29,8 @@ import OGCVectorTile from 'ol/source/OGCVectorTile.js'
 import { MVT } from 'ol/format'
 import VectorTileLayer from 'ol/layer/VectorTile'
 import OGCMapTile from 'ol/source/OGCMapTile.js'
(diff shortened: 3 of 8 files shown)


libs/feature/map/src/lib/map-context/map-context.service.ts
```diff
@@ -112,13 +116,22 @@ export class MapContextService {
...
+            source: new ImageWMS({
+              url: layerModel.url,
+              params: { LAYERS: layerModel.name },
+            }),
+          })
+        } else {
+          return new TileLayer({
+            source: new TileWMS({
+              url: layerModel.url,
+              params: { LAYERS: layerModel.name },
```
jahow, line 130: to let backend components know that we're querying tiles we should also set the `TILED: true` parameter
See e.g. https://docs.geoserver.org/main/en/user/geowebcache/using.html

jahow (commented):
Thanks! Haven't tested but it looks good. We have to keep in mind to 1/ document the setting [here](https://github.com/geonetwork/geonetwork-ui/blob/main/docs/guide/configure.md) and 2/ make sure that the next version will mention this change (exp. the gutter being gone) and let people know that they might need to use this setting if they have artefacts close to tile borders.

libs/feature/map/src/lib/map-context/map-context.service.ts
```diff
@@ -112,13 +116,22 @@ export class MapContextService {
...
-            url: layerModel.url,
-            params: { LAYERS: layerModel.name },
-            gutter: 20,
-          }),
-        })
+        if (mapConfig?.DO_NOT_TILE_WMS) {
+          return new ImageLayer({
+            source: new ImageWMS({
+              url: layerModel.url,
+              params: { LAYERS: layerModel.name, TILED: false },
```
jahow, line 123:
```suggestion
              params: { LAYERS: layerModel.name },
```
This param can be omitted safely, we usually don't include it if it's false

docs/guide/configure.md
```diff
@@ -206,6 +206,10 @@ The map section lets you customize how maps appear and behave across GeoNetwork-
 
   Will limit the possibility to zoom in past a certain zoom level.
 
+- `do_not_tile_wms` (optional)
+
+  Will not tile WMS. Defaults to `false` (WMS are tiled).
```
jahow, line 211:
```suggestion
  Will not use tiling when requesting WMS services. Defaults to `false` (WMS are tiled). Not using tiles for WMS might incur performance loss since the client will not benefit from an eventual tile cache anymore. On the other hand, visual quality might improve in case a map tile server does not handle neighbouring tiles correctly, e.g. symbols or text being cropped at tile boundaries.
```

conf/default.toml
```diff
@@ -122,6 +122,9 @@ background_color = "#fdfbff"
 # Optional; Will limit the possibility to zoom in past a certain zoom level
 # max_zoom = 10
 
+# Optional; Will not tile WMS. False by default.
+# do_not_tile_wms = false
```
tkohr, line 126: Maybe it'd make sense to move this param a little further down (just above `do_not_use_default_basemap`) and add some more background info on this here and/or in the doc, like:
`... This can be set `true` to prevent visual conflicts on tile borders, if the WMS server does not add a gutter, for example. gn-ui does not add a gutter on the client side, in order to allow server-side caching.`
What do you think about the formulation @jahow ?

tkohr (approved):
Thanks @rcaplier, tested it and it works fine! Geowebcache now responds with a `Geowebcache-Cache-Result: HIT` when keeping the default behavior (WMS being tiled) :slightly_smiling_face:

conf/default.toml
jahow: > Maybe it'd make sense to move this param a little further down (just above do_not_use_default_basemap)

yes agreed, it should not be above the extent, layers etc.

As for formulation, it can be elaborated a little but the big explanation should be given in the doc website (see https://github.com/geonetwork/geonetwork-ui/pull/887/files#diff-9088b1e53d9bb428254eea7cc21785db455294f9bea3c903806cd8dfa06f7024)

tkohr:
Thanks for the adaptions @rcaplier! I think this one can be merged once the conflict is resolved. Watch out rebasing, there's a new `attributions` property now.

PR #893, opened 2024-05-29
Datahub - Map is not reset when navigating between datasets

rcaplier:
### Description

This PR makes the map reset when exiting the dataset.

---


jahow (changes requested):
I'm not sure this is the right way to approach this issue. I would say that instead we should try to reinitialize the map when a new dataset is opened, instead of assuming that the map will be empty initially (which is what we currently do).

cmoinier:
> I'm not sure this is the right way to approach this issue. I would say that instead we should try to reinitialize the map when a new dataset is opened, instead of assuming that the map will be empty initially (which is what we currently do).

I looked into this because my PR is kind of related. I think `dropdownChoices$` take some time to change/reload the new links when you switch records, creating this issue. I think that emptying this Observable when the user leaves the page should do the trick.

jahow:
> I looked into this because my PR is kind of related. I think `dropdownChoices$` or `compatibleMapLinks$` perhaps take some time to change/reload the new links when you switch records, creating this issue. I think that emptying this Observable when the user leaves the page should do the trick.

I agree that we should work with observables. Why not do a `startwith` somewhere to make sure that as soon as a record is opened an empty map is shown? Instead of clearing it on record close, which would solve the issue in a more indirect way.

rcaplier:
> > I looked into this because my PR is kind of related. I think `dropdownChoices$` or `compatibleMapLinks$` perhaps take some time to change/reload the new links when you switch records, creating this issue. I think that emptying this Observable when the user leaves the page should do the trick.
> 
> I agree that we should work with observables. Why not do a `startwith` somewhere to make sure that as soon as a record is opened an empty map is shown? Instead of clearing it on record close, which would solve the issue in a more indirect way.

I tried the startWith solution and it didn't work for me... And then I discovered that we actually have something to create an emptyMap in the mapUtilsManager.

jahow (approved):

PR #894, opened 2024-06-05
Metadata Editor / keep a draft of a record being edited

jahow:
### Description

This PR brings many changes to the Records Repository abstraction and the editor state:
* The Records Repository now allows for opening a record for edition
  * This will return a record in its most recent version (e.g. a draft if any), as well as its source in XML and whether this record is published or not
* It also allows for saving a record, saving a record as draft, and checking whether a record as a draft
* Currently this is implemented by storing the record as XML inside the client local storage; **this might not be the best solution but we're going with it for now because it's very simple**
* When modifying the active record in the form, a draft will be saved every few hunder milliseconds automatically
* The records list in the dashboard now shows if a record has an active draft
* Because we now also get the record as XML when opening it for edition, we can then save it back using the XML source, thus preserving most of the record and only changing the parts that really need changing (and also keeping the same format when converting)
* The record source and whether it's been published are also available in the editor state

Another change is that the badge UI component is now standalone and can be customized using CSS variables, kind of like the button component.

Opening a record:
image

After modifying it:
image

In the records list:
image

**_Caveat: creating a new record is now broken!! Will be fixed in upcoming PRs._**

### Architectural changes

Opening records for edition and saving them is now done through the RecordsRepository abstraction.


libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -1,50 +1,22 @@
...
 
-  constructor(
-    private http: HttpClient,
-    @Optional()
-    @Inject(Configuration)
-    private apiConfiguration: Configuration
-  ) {}
-
-  // TODO: use the catalog repository instead
-  loadRecordByUuid(uuid: string): Observable<CatalogRecord> {
```
LHBruneton-C2C, line 28: Found what I was looking for!

libs/ui/widgets/src/lib/ui-widgets.module.ts
```diff
@@ -41,7 +39,6 @@ import { CommonModule } from '@angular/common'
     StepBarComponent,
     LoadingMaskComponent,
     PopupAlertComponent,
-    BadgeComponent,
```
LHBruneton-C2C, line 44: Wondering about this, this is making a switch to standalone components mandatory, while still exporting from modules could allow both use.

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.ts
```diff
@@ -12,9 +16,37 @@ import { MatIconModule } from '@angular/material/icon'
...
+    ]).pipe(
+      map(([alreadySavedOnce, changedSinceSave]) => {
+        if (!alreadySavedOnce) {
+          return 'saved_not_published'
+        }
+        return changedSinceSave
+          ? 'saved_published_changes_pending'
+          : 'saved_published_up_to_date'
+      })
+    )
```
LHBruneton-C2C, line 49: I'm a bit confused about the save status and already saved once here. Are we talking about the same "save" (draft on localStorage vs record on server)?
As I understand it, the draft is always saved, hence all the status beginning with "saved_".
But the record might not be saved, hence the need to call "alreadySavedOnce".
Could we maybe switch to a word like "stored" instead of "saved" for the localStorage?

LHBruneton-C2C (approved):
Not tested yet but looking great already!
Thanks for the comments where needed and the really readable git history!
I'll test first thing tomorrow.

LHBruneton-C2C:
Tested OK! (except for this new record caveat)

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -165,4 +177,131 @@ export class Gn4Repository implements RecordsRepositoryInterface {
...
+  openRecordForEdition(
+    uniqueIdentifier: string
+  ): Observable<[CatalogRecord, string, boolean] | null> {
+    const draft$ = of(
+      window.localStorage.getItem(
+        this.getLocalStorageKeyForRecord(uniqueIdentifier)
+      )
+    )
+    const recordAsXml$ = this.loadRecordAsXml(uniqueIdentifier)
+    return combineLatest([draft$, recordAsXml$]).pipe(
```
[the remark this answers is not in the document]
jahow: good point, originally I made it return an observable because the draft was saved in the backend; I left it like that because I thought it would make sense to allow another implementation of the draft system which would be asynchronous. But of course YAGNI would dictate to get rid of it and make it synchronous... I'll rework it

libs/ui/widgets/src/lib/ui-widgets.module.ts
jahow: The way I understand it, standalone components when used by non-standalone ones can either be part of an "outgoing" module (here UI widgets) or an "incoming" module (for instance an app module).

I feel like a benefit of converting things to standalone is that the lib modules will become slimmer and eventually disappear. For that to happen we need to import the standalone components as close as possible to their consumer, which would mean in the app module.

I hope I'm clear enough :grimacing:

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.ts
jahow: This is also a good point, it can be confusing. I guess we could change the statuses like so:
* "saved as draft only" => when creating a record
* "record is up-to-date" => when the record was just published (ie saved on the server)
* "draft saved, changes pending" => when the record was modified and not yet published

libs/ui/widgets/src/lib/ui-widgets.module.ts
LHBruneton-C2C: OK, if its a conscious choice to get definitely rid of Angular modules, this is the way to go indeed.

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.ts
LHBruneton-C2C: I find those new labels much better! Will you change the technical statuses names too?

jahow: yep :)

jahow:
thank you for the review @LHBruneton-C2C!!

PR #898, opened 2024-06-12
feat(ui-inputs): Fix and extend autocomplete component

Angi-Kinas:
Add a new Input `minCharacterCount` to the autocomplete component to allow for a minimum character count before triggering the action. That means that if `minCharacterCount`  is 0, the dropdown should be filled initially with all available values. Additionally after clearing the input value, the dropdown is also populated with all available values instead of using the previous selected value.
The storybook was updated to reflect these changes. 

To Do:
- Make `action` be an input also in the storybook 

### Screenshots
image
image

### DIFF
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.spec.ts
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.spec.ts
@@ -183,7 +183,6 @@ describe('AutocompleteComponent', () => {
   })
 
   describe('@Input() value', () => {
-    let anyEmitted
     describe('when set', () => {
       beforeEach(() => {
         const simpleChanges: any = {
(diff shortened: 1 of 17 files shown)


jahow:
Thanks for this PR @Angi-Kinas! I've added some commits to hopefully improve the autocomplete component and also to make it standalone.

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.spec.ts
```diff
@@ -164,9 +184,6 @@ describe('AutocompleteComponent', () => {
       it('sends a submitted value', () => {
         expect(anyEmitted).toEqual(['bla'])
       })
-      it('closes the autocomplete panel', () => {
-        expect(component.triggerRef.closePanel).toHaveBeenCalled()
-      })
```
jahow, line 169: This test is not needed since when we click the button the input becomes out of focus and the overlay closes anyway.

PR #900, opened 2024-06-13
[Fix]: Reset map dropdown links

cmoinier:
### Description

This PR fixes the behavior of the map preview dropdown, which wrongfully included previous dataset's links in any newly opened dataset.

---

### DIFF
--- a/libs/common/fixtures/src/lib/records.fixtures.ts
+++ b/libs/common/fixtures/src/lib/records.fixtures.ts
@@ -233,6 +233,14 @@ Malgré l'attention portée à la création de ces données, il est rappelé que
         description: 'Téléchargement du fichier',
         mimeType: 'x-gis/x-shapefile',
       },
+      {
+        type: 'service',
+        url: new URL('https://my-org.net/ogc'),
+        accessServiceProtocol: 'ogcFeatures',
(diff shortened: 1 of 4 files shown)


jahow:
> Maybe adding a test (UT or e2e) for your changes would be nice if not too complicated.

I would suggest simply an UT test focusing on this observable, thanks @cmoinier :)

cmoinier:
@tkohr @jahow Thanks for the reviews :) 

In the last commit I have added a new UT, and used the "loading" variable to keep the map loading until the dropdown has its options. I was not able to test as the options always load pretty fast on my end, so let me know if it doesn't do the trick for you.

tkohr:
Thanks for the added test @cmoinier! Setting `this.loading` does not change anything on my side, so I think we can discard this modification. IMO, we can still improve this later (maybe along with the "record not found" message that I also see popping up during loading). But this is clearly out of scope of this PR.

tkohr:
It just came to my mind, that we'll also need a backport of this for the 2.3 release, I guess, but I don't see a `2.3.x` branch for now. Do we still need to create this branch and then the according backport label or has the workflow changed recently @jahow ?

jahow:
We need to first create the branch and then put the backport tag :)

PR #905, opened 2024-06-17
[FIX]: Export popup-alert component

cmoinier:
### Description

This PR adds back the export popup-alert component, which is needed in MEL.


jahow (commented):
I think this component is standalone now, so it should not appear there I think. Do you get an error on the MEL application side?

cmoinier:
This was fixed on MEL in the end.

PR #909, opened 2024-06-20
EDITOR - Record field: graphic overview

rcaplier:
### Description

This PR introduces the graphical overview component. Now editor allows to edit the overview of the metadata.

### Screenshots

image

---


libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -2,12 +2,27 @@ import {
...
+import { Subject, takeUntil } from 'rxjs'
+
+const extractFileNameFormUrl = (url: URL, metadataUuid: string): string => {
+  const pattern = new RegExp(
+    `records/${metadataUuid}/attachments/([^/?#]+)(?:[/?#]|$)`,
+    'i'
+  )
+  const match = url.href.match(pattern)
+  return match ? match[1] : ''
+}
```
LHBruneton-C2C, line 25: I don't understand why this needed when you already have the resource fileName at your disposal...

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -20,7 +20,7 @@ import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
 export class OverviewUploadComponent implements OnInit {
   @Input() metadataUuid: string
 
-  resourceFileName: string
+  imageAltText: string
```
LHBruneton-C2C: Is this really a good change? We are using the resource file name as the alt text, not the other way around.

LHBruneton-C2C (commented):
Looking good, not tested.

Although, there is some added complexity which I don't understand.

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
rcaplier: I don't know why this has been flagged as added code, maybe I moved it from another line. This not my code.

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
rcaplier: Yes... This is a point of view. My logic is to name the variable as it will be used in the code, not the source of it. If I have some number that I want to use as the the tax amout for exemple, then I will name my variable "taxAmount", not "someNumber".

Keeping the source as the name is disturbing for me.

libs/data-access/gn4/src/openapi/model/metadataResource.api.model.ts
```diff
@@ -19,7 +19,7 @@ export interface MetadataResourceApiModel {
   metadataResourceExternalManagementProperties?: MetadataResourceExternalManagementPropertiesApiModel
   lastModification?: string
   version?: string
-  url?: string
+  url?: URL
```
jahow, line 22: these files are auto-generated; this change will be list if we regenerate the API client, so I would avoid doing this

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
jahow: it was definitely added by this PR, which makes sense since this file makes use of `extractFileNameFormUrl` and it wasn't available in that context. Maybe try factorizing the original function or simply exporting it?

libs/feature/editor/src/lib/components/record-form/form-field/form-field.component.spec.ts
```diff
@@ -150,4 +171,17 @@ describe('FormFieldComponent', () => {
...
+  describe('overviews field', () => {
+    let formField
+    beforeEach(() => {
+      component.model = 'overviews'
+      fixture.detectChanges()
+      formField = fixture.debugElement.query(
+        By.directive(FormFieldOverviewsComponent)
+      ).componentInstance
+    })
+    it('creates an array form field', () => {
```
jahow, line 183:
```suggestion
    it('creates an overview upload form field', () => {
```

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -31,40 +50,98 @@ export class OverviewUploadComponent implements OnInit {
   ngOnInit(): void {
     this.recordsApiService
       .getAllResources(this.metadataUuid)
-      .subscribe((resources) => {
-        this.resourceFileName = resources[0]?.filename
-        this.resourceUrl = resources[0]?.url
-        this.cd.markForCheck()
+      .pipe(takeUntil(this.destroy$))
```
jahow, line 53: I don't think you need that; the `getAllResources` observable is an HTTP call, so it is short-lived: as soon as the result is emitted, it also completes.

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -31,40 +50,98 @@ export class OverviewUploadComponent implements OnInit {
...
   }
 
   handleFileChange(file: File) {
     this.recordsApiService
       .putResource(this.metadataUuid, file, 'public')
-      .subscribe((resource) => {
-        this.resourceFileName = resource.filename
-        this.resourceUrl = resource.url
-        this.cd.markForCheck()
+      .pipe(takeUntil(this.destroy$))
```
jahow, line 76: See above, probably not needed

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -31,40 +50,98 @@ export class OverviewUploadComponent implements OnInit {
...
   }
 
   handleUrlChange(url: string) {
     this.recordsApiService
       .putResourceFromURL(this.metadataUuid, url, 'public')
-      .subscribe((resource) => {
-        this.resourceFileName = resource.filename
-        this.resourceUrl = resource.url
-        this.cd.markForCheck()
+      .pipe(takeUntil(this.destroy$))
```
jahow, line 96: See above, probably not needed

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -31,40 +50,98 @@ export class OverviewUploadComponent implements OnInit {
...
+    const fileName = extractFileNameFormUrl(this.resourceUrl, this.metadataUuid)
+
     this.recordsApiService
-      .delResource(this.metadataUuid, this.resourceFileName)
-      .subscribe(() => {
-        this.resourceFileName = null
-        this.resourceUrl = null
-        this.cd.markForCheck()
+      .delResource(this.metadataUuid, fileName)
+      .pipe(takeUntil(this.destroy$))
```
jahow, line 118: See above, probably not needed

LHBruneton-C2C:
@rcaplier could you resolve the conflicts on this PR, so we can do a final review before merging? Thanks!

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
rcaplier: As seen with @jahow, this is needed finally.

rcaplier:
> @rcaplier could you resolve the conflicts on this PR, so we can do a final review before merging? Thanks!

The final review will come when I will finish it.

Like I said in dailies, I still have to adapt the component regarding the version of geonetwork. I still don't know if I have to hide it disable it. Does this need design @jahow ?

Plus, I have to readd the `extractFileNameFormUrl()` method since it is indeed needed after all.

But thanks for the reminder

libs/feature/editor/src/lib/fields.config.ts
```diff
@@ -83,14 +83,25 @@ export const RECORD_ABSTRACT_FIELD: EditorField = {
   },
 }
 
+export const RECORD_GRAPHICAL_OVERVIEW_FIELD: EditorField = {
+  model: 'overviews',
+  formFieldConfig: {
+    labelKey: marker('editor.record.form.field.overviews'),
```
LHBruneton-C2C, line 89: This will need a translation.

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -17,54 +29,115 @@ import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
...
-      .getAllResources(this.metadataUuid)
-      .subscribe((resources) => {
-        this.resourceFileName = resources[0]?.filename
-        this.resourceUrl = resources[0]?.url
+    this.recordsApiService.getAllResources(this.metadataUuid).subscribe({
+      next: (resources) => {
+        if (resources && resources.length > 0) {
+          this.resourceUrl = new URL(resources[0]?.url)
+          this.imageAltText = resources[0].filename
+          this.ressourceFileName = extractFileNameFormUrl(resources[0]?.url)
```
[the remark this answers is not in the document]
rcaplier: The `resources[0].filename` is a text field that is used to store the altTextImage initially. But the user can change it if he wants. So it's not even clear to me how the back is retreiving the image afterward but I suppose that it store the initial value at creation. And if so, then if we store the altDescription of the image, the only place I can still find the real name of the image is in the url.

libs/feature/editor/src/lib/components/overview-upload/overview-upload.component.ts
```diff
@@ -2,12 +2,24 @@ import {
...
+  Output,
 } from '@angular/core'
 import { CommonModule } from '@angular/common'
 import { RecordsApiService } from '@geonetwork-ui/data-access/gn4'
 import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
+import { FormControl } from '@angular/forms'
+import { GraphicOverview } from '@geonetwork-ui/common/domain/model/record'
+import { Subject } from 'rxjs'
+
+const extractFileNameFormUrl = (url: string): string => {
```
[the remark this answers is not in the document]
rcaplier: Good catch ! It's from

LHBruneton-C2C (changes requested):
I've tested it this time. The alternative text is not being saved, not on draft and not on record.
Tell me if you need to have a look on my machine.

libs/feature/editor/src/lib/fields.config.ts
rcaplier: This had a translation but should have been lost with a rebasing ;/

LHBruneton-C2C (approved):
Thanks for the fix, working as it should now!

PR #913, opened 2024-06-25
[Search page] : Add new filtering options

cmoinier:
### Description

This PR introduces two new filtering options : 

- Data quality score
- Territories

For the territory one, since the thesaurus name is specific to MEL (and to each instance, actually), it needs to be passed down to fields.service, where the filter is initialized with no thesaurus name. If someone has a better solution, it's very welcomed :slightly_smiling_face:

### DIFF
--- a/libs/feature/search/src/lib/utils/service/fields.service.ts
+++ b/libs/feature/search/src/lib/utils/service/fields.service.ts
@@ -32,7 +32,7 @@ marker('search.filters.standard')
 marker('search.filters.topic')
 marker('search.filters.contact')
 marker('search.filters.qualityScore')
-marker('search.filters.territories')
+marker('search.filters.customTranslatedSearchField')
 
 @Injectable({
   providedIn: 'root',
@@ -73,7 +73,11 @@ export class FieldsService {
     license: new LicenseSearchField(this.injector),
     owner: new OwnerSearchField(this.injector),
     qualityScore: new SimpleSearchField('qualityScore', this.injector, 'desc'),
-    territories: new TranslatedSearchField(undefined, this.injector, 'asc'),
+    customTranslatedSearchField: new TranslatedSearchField(
+      undefined,
+      this.injector,
+      'asc'
+    ),
   } as Record<string, AbstractSearchField>
 
   get supportedFields() {
@@ -86,7 +90,7 @@ export class FieldsService {
     if (this.supportedFields.indexOf(fieldName) === -1) {
       throw new Error(`Unsupported search field: ${fieldName}`)
     }
-    if (fieldName === 'territories' && thesaurusName) {
+    if (fieldName === 'customTranslatedSearchField' && thesaurusName) {
       this.fields[fieldName] = new TranslatedSearchField(
         thesaurusName,
         this.injector,
--- a/libs/feature/search/src/lib/utils/service/fields.service.ts
+++ b/libs/feature/search/src/lib/utils/service/fields.service.ts
@@ -31,6 +31,8 @@ marker('search.filters.resourceType')
 marker('search.filters.standard')
 marker('search.filters.topic')
 marker('search.filters.contact')
+marker('search.filters.qualityScore')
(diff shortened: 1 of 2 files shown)


libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -70,6 +72,8 @@ export class FieldsService {
     q: new FullTextSearchField(),
     license: new LicenseSearchField(this.injector),
     owner: new OwnerSearchField(this.injector),
+    qualityScore: new SimpleSearchField('qualityScore', this.injector, 'desc'),
+    territories: new TranslatedSearchField(undefined, this.injector, 'asc'),
```
tkohr, line 76: Hum, not sure if there is a cleaner solution for this (cc @jahow), but I'd rename `territories` to `customTranslatedSearchField` here as this is not limited to the territories use case.

jahow:
For quality score, what exactly is the intent here? shouldn't it be something like "qualityScoreAbove" and filter out values below a certain threshold?

About the "territories" filter, maybe it can be implemented on the MEL custom app side? here for instance: https://github.com/camptocamp/mel-dataplatform/blob/7e89f672046e91d047ba99122fe6547ea381ec5c/apps/datahub/src/app/search/service/fields.service.ts#L10-L23

cmoinier:
> For quality score, what exactly is the intent here? shouldn't it be something like "qualityScoreAbove" and filter out values below a certain threshold?
> 
> About the "territories" filter, maybe it can be implemented on the MEL custom app side? here for instance: https://github.com/camptocamp/mel-dataplatform/blob/7e89f672046e91d047ba99122fe6547ea381ec5c/apps/datahub/src/app/search/service/fields.service.ts#L10-L23

For the quality, I think the mockup shows a modal with two fields to request above and under values, but I was thinking of handling that part on MEL, and just creating the filter on gn-ui because I don't know if it's the same intent on gn-ui with this filter later on.

Well I had no idea we had an override for that on MEL, so that's on me :smile: It would make sense, but seeing https://camptocamp.atlassian.net/browse/GSCIRAD-84 about some new filters, and since one of them mentions the territories, so wouldn't it make sense to already have a "custom tag" field already? Or will it be in a fork and so it doesn't matter here ? 

I'm also happy to close this PR and do the whole thing on MEL

jahow:
One thing that we considered when doing the MEL filters was to allow some king of configuration of filters instead of just enablig them. Currently we can do:
```toml
advanced_filters = ['publisher', 'format', 'inspireTheme']
```

We could have something like
```toml
advanced_filters = ['publisher', 'format', 'fromThesaurus[inspire_theme_thesaurus_id]', 'fromThesaurus[territories_thesaurus_id]']
```
or
```toml
advanced_filters = ['publisher', 'format', 'fromThesaurus.inspire_theme_thesaurus_id', 'fromThesaurus.territories_thesaurus_id']
```
or even
```toml
advanced_filters = [
  'publisher',
  'format',
  { type: 'fromThesaurus', parameter: 'inspire_theme_thesaurus_id' },
  { type: 'fromThesaurus', parameter: 'territories_thesaurus_id' }
]
```

So each advanced filter would also support some kind of parameter. I think this would be a great feature, but in terms of implementation this would be a big change in GN-UI. Handling custom filters in the MEL custom app would probably be much cheaper, so in the end it depends on the resource we have available for this feature.

tkohr:
Ah right, I didn't think of the overrides in the `fields.service` yesterday. This might be the most rational solution here, as the MEL thesaurus is really custom, and we exploit the overrides of the service already. When I see right, the CIRAD would like to use a standard GN thesaurus, so it might be ok to implement their need with the specific thesaurus in gn-ui.
Of course, the configurable system would be pretty nice, but I don't know if we have the time for it within this feature.

cmoinier:
Okay, based on the last comments I will close this PR to do the implem on MEL's end :slightly_smiling_face:

PR #915, opened 2024-06-25
fix(platform-service): Exclude place thesarus keywords

Angi-Kinas:
### Description

This PR introduces changes to the `#searchKeywords` function in order to get all keywords except the ones of type `place`.


jahow (approved):
Very good, thanks!! It looks like the i18n task wasn't run for a while, I hope it's not going to break anything...

PR #924, opened 2024-07-02
UI component switch toggle

Angi-Kinas:
### Description

This PR introduces a new UI Input component "SwitchToggleComponent", which is able to toggle between options similar to radio buttons.

The component should be accessible with the keyboard. Please verify this in the Storybook.

### Screenshots

image


libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SwitchToggleComponent {
+  @Input() options: SwitchToggleOption[]
+  @Input() ariaLabel? = ''
+  @Input() extraClasses? = ''
+  @Output() selectedValue = new EventEmitter<SwitchToggleOption>()
+
+  onChange(selectedOption: SwitchToggleOption) {
+    this.options.forEach((option) => {
```
LHBruneton-C2C, line 28: Small possible improvement here (nothing critical): as you can't select several options at the same time, you don't need to loop over the whole array, you can stop as soon as you've found the selected element.
To do this, you can use the [array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) method.

libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  value: string
+  checked: boolean
+}
+
+@Component({
+  selector: 'gn-ui-switch-toggle',
+  templateUrl: './switch-toggle.component.html',
+  styleUrls: ['./switch-toggle.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
```
LHBruneton-C2C, line 20: This component should be created as standalone.

libs/ui/inputs/src/lib/ui-inputs.module.ts
```diff
@@ -46,6 +48,7 @@ import { ImageInputComponent } from './image-input/image-input.component'
     CopyTextButtonComponent,
     CheckboxComponent,
     SearchInputComponent,
+    SwitchToggleComponent,
```
LHBruneton-C2C, line 51: All the changes in this file won't be needed anymore when switching to standalone component.

LHBruneton-C2C (changes requested):
Looking good, working great (tested on the storybook github page).

You just need to change the component to standalone for this PR to be approved.

libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
Angi-Kinas: I tried this and it works! I'm just wondering how it sets the other options.checked to `false`? Maybe it does not need to because as you already said, it is not possible to select multiple at a time. But how does the component know which to style differently if all options.checked are `true`?

Angi-Kinas:
Thank you for the review @LHBruneton-C2C :-) 
I changed the component to be standalone and simplified the logic a bit as you suggested. Can you have another look?

libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
LHBruneton-C2C: I didn't read the Angular Material toggle implementation, but most probably they rely on some behavior (native radio button or custom one) to make sure that if one option is checked, the other ones are unchecked. It almost never makes sense to explicitly set an option to unchecked, unless you allow your toggle switch to be in a an "all unchecked" state.

LHBruneton-C2C (approved):
Looking great, thanks for the fixes!

PR #926, opened 2024-07-02
[Datahub]: Fix WFS error on QGIS

cmoinier:
### Description

This PR fixes an error that appears when copying the API form's links into QGIS.
It adds the root URL copy button to the API card to this end.

### This work has already been reviewed here : https://github.com/camptocamp/mel-dataplatform/pull/94

---

**This work is sponsored by  MEL**.


tkohr (approved):
Thanks @cmoinier !

PR #942, opened 2024-07-17
ME: duplicate local record

LHBruneton-C2C:
### Description

This PR introduces the local record duplication feature, from the action menu on each record, in all records tables.

The click interaction on a record to open it has been moved to only the title, to allow a click on the menu.

When clicking on the duplicate action, the app navigates to a new duplicate route, which take an UUID as parameter, to read a local record, save it as a copy draft, then enter edition mode.

### Architectural changes

The ui-search results-table component now depends on the MatMenuModule from Angular Material.

### Screenshots

image

### DIFF
--- a/libs/api/repository/src/lib/gn4/gn4-repository.ts
+++ b/libs/api/repository/src/lib/gn4/gn4-repository.ts
@@ -235,15 +235,12 @@ export class Gn4Repository implements RecordsRepositoryInterface {
     uniqueIdentifier: string
   ): Observable<[CatalogRecord, string, false] | null> {
     return this.loadRecordAsXml(uniqueIdentifier).pipe(
-      switchMap(async (xml) => {
-        const converter = findConverterForDocument(xml)
-        const record = await converter.readRecord(xml)
-        return [record, converter] as [CatalogRecord, BaseConverter<string>]
-      }),
-      switchMap(async ([record, converter]) => {
+      switchMap(async (recordAsXml) => {
+        const converter = findConverterForDocument(recordAsXml)
+        const record = await converter.readRecord(recordAsXml)
         record.uniqueIdentifier = `TEMP-ID-${Date.now()}`
         record.title = `${record.title} (Copy)`
-        const xml = await converter.writeRecord(record)
+        const xml = await converter.writeRecord(record, recordAsXml)
         window.localStorage.setItem(
           this.getLocalStorageKeyForRecord(record.uniqueIdentifier),
           xml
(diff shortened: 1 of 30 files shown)


libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -230,6 +231,28 @@ export class Gn4Repository implements RecordsRepositoryInterface {
...
+  openRecordForDuplication(
+    uniqueIdentifier: string
+  ): Observable<[CatalogRecord, string, false] | null> {
+    return this.loadRecordAsXml(uniqueIdentifier).pipe(
+      switchMap(async (xml) => {
+        const converter = findConverterForDocument(xml)
+        const record = await converter.readRecord(xml)
+        return [record, converter] as [CatalogRecord, BaseConverter<string>]
+      }),
+      switchMap(async ([record, converter]) => {
```
jahow, line 243: you can probably leave that aside and have everything in the same `switchMap` :)

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -230,6 +231,28 @@ export class Gn4Repository implements RecordsRepositoryInterface {
...
+        record.uniqueIdentifier = `TEMP-ID-${Date.now()}`
+        record.title = `${record.title} (Copy)`
+        const xml = await converter.writeRecord(record)
```
jahow, line 246: The original XML document should be provided to `writeRecord` so that only the affected parts of the XML are changed, otherwise an completely new XML doc will be generated from scratch.

```suggestion
        const xml = await converter.writeRecord(record, xml)
```

libs/ui/search/src/lib/results-table/results-table.component.html
```diff
@@ -120,4 +121,27 @@
       {{ dateToString(item.recordUpdated) }}
     </ng-template>
   </gn-ui-interactive-table-column>
+
+  <!-- ACTION MENU COLUMN -->
+  <gn-ui-interactive-table-column>
+    <ng-template #header> </ng-template>
+    <ng-template #cell let-item>
+      <button
```
jahow, line 129: do you think you could use a gn-ui-button there? to have some hover/focus effects (the stop propagation part will also be handled that way)

jahow (approved):
Looking and working great, thanks! I made a few suggestions in the code, especially when generating the XML of the duplicated record.

Also it looks like this PR is rebased on another one, but I'm not sure which. Should we wait for another PR to be merged before merging this one?

libs/ui/search/src/lib/results-table/results-table.component.html
LHBruneton-C2C: I had tried, but I'll try again.
Opening the menu didn't seem to work with a gn-ui-button, maybe because of the stop propagation indeed.

PR #944, opened 2024-07-22
ME: iso19139 spatial extents

LHBruneton-C2C:
### Description

This PR introduces support of the spatial extents in the ISO converters.
Both bbox and the first of multiple geometries are supported.
The description is supported only through the `MD_Identifier` tag.

### Architectural changes

The api-metadata-converter library now depends on OpenLayers.


libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -159,7 +159,8 @@ export interface GraphicOverview {
 }
 
 export interface DatasetSpatialExtent {
-  geometry: Geometry
+  geometries?: Geometry[]
+  bbox?: [number, number, number, number]
```
jahow, line 163: So now I understand what you meant by using a `bbox` property.

My original approach with this field was to describe simple bboxes as GeoJSON geometries as well (simple polygons). See for example the mapper for the GN4 index format:

https://github.com/geonetwork/geonetwork-ui/blob/785657302d93b46785c0eb143a60ef4186934208/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts#L292-L316

What do you think about getting rid of the `bbox` property and only use the `geometries` one?

libs/api/metadata-converter/src/lib/xml-utils.ts
```diff
@@ -312,6 +352,7 @@ export function appendChildren(
 ): ChainableFunction<XmlElement, XmlElement> {
   return (element) => {
     if (!element) return null
+    childrenFns = childrenFns.filter((fn) => fn)
```
jahow, line 355: yes that makes sense! Maybe documenting this behavior (i.e. not appending a child if it is falsy) would be good?

libs/api/metadata-converter/src/lib/xml-utils.ts
```diff
@@ -225,14 +239,40 @@ export function xmlToString(
...
 ${parentPadding}`
   }
+
   return `
-${padding}<${el.name}${attrs}>${children}</${el.name}>
+${padding}<${el.name}${attrs}${gmlNamespace}>${children}</${el.name}>
 ${parentPadding}`
 }
 
+function getGmlNamespaceAsNeeded(
```
jahow, line 251: I don't think you need that; the namespaces are added at the end when producing the XML document, see:
https://github.com/geonetwork/geonetwork-ui/blob/f92455971a26a1110d02acb31a142cc27e7cd6d8/libs/api/metadata-converter/src/lib/xml-utils.ts#L34

libs/api/metadata-converter/src/lib/xml-utils.ts
```diff
@@ -448,3 +489,29 @@ export function renameElements(
   doReplace(rootElement)
   return rootElement
 }
+
+export function readGeometry(el: XmlElement): Geometry {
```
jahow, line 493: these two functions (read/write geom) might be better of in the iso19139/utils directory in their own file (e.g. geometry.ts) because the xml utils are meant to be shared across any XML schema

jahow (commented):
Thank you!! most of this looks really good. My main question is about the "bbox" property, which I wish we could avoid adding; using only geometries would make things simpler overall I think.

Also the logic for adding the GML namespace should not be needed, the XML document generation already adds all necessary namespaces at the top in the end :)

libs/api/metadata-converter/src/lib/xml-utils.ts
LHBruneton-C2C: Do you mean in the code, or did you have somewhere specifically in mind?

libs/api/metadata-converter/src/lib/xml-utils.ts
LHBruneton-C2C: Actually, the way I parse/serialize the GML part with OpenLayers, I don't have the full access to the XmlDocument at that time, as OpenLayers only takes in native node objects, I have to use a string as a intermediate step at some point.
It's there that I'm missing the GML namespace.
If you know a way to have OL understand the Xml elements from the lib we use to parse the record, I'd be really interested.
I agree adding this namespace manually is ugly (and the serialization is ugly too, we change default namespace in the middle of the tree, but apparently it's allowed, so...).

LHBruneton-C2C:
> using only geometries would make things simpler overall I think.

I mostly agree, but as we are dealing with an array of spatial extents, I remove all the elements before writting the new ones, which in this case would mean changing the content of the original record, even when doing only partial edition. Is it OK for the original record to switch from a BBOX spatial extent to a Polygon spatial extent?

libs/api/metadata-converter/src/lib/xml-utils.ts
jahow: in the code :)

jahow:
> Is it OK for the original record to switch from a BBOX spatial extent to a Polygon spatial extent?

Hmm, I didn't think of that. Maybe there could be a way to figure out that a Polygon is in fact a bbox? something like so:
```ts
// this was generated with the help of chatgpt
function isBoundingBox(geom: Geometry) {
  if (geometry.type !== 'Polygon') {
    return false;
  }

  const coordinates = geometry.coordinates;

  // is it a square polygon with no holes?
  if (coordinates.length !== 1 || coordinates[0].length !== 5) {
    return false;
  }

  const [first, second, third, fourth, last] = coordinates[0];
  
  // is it a closed ring?
  if (first[0] !== last[0] || first[1] !== last[1]) {
    return false;
  }

  // is it an axis-aligned rectangle?
  return first[0] === second[0] && first[1] === fourth[1] &&
    second[1] === third[1] && third[0] === fourth[0] &&
    first[0] !== third[0] && first[1] !== third[1];
}
```

libs/api/metadata-converter/src/lib/xml-utils.ts
```diff
@@ -448,3 +489,29 @@ export function renameElements(
...
+  const gmlString = xmlToString(el)
```
jahow, line 494: So, to avoid the GML namespace stuff in `xmlToString` you can do the following here:

```suggestion
  const elWithNs = new XmlElement(
    el.name,
    {
      ...el.attributes,
      'xmlns:gml': 'http://www.opengis.net/gml/3.2',
      xmlns: 'http://www.opengis.net/gml/3.2',
    },
    el.children
  )
  const gmlString = xmlToString(elWithNs)
```

Essentially, we make a clone of the element to read and we set both the `gml` namespace and the default namespace to GML 3.2. This way, we're (almost?) sure that OL will be able to parse the geometry.

What do you think?

libs/api/metadata-converter/src/lib/xml-utils.ts
jahow: I've tried a few things, I think this can be handled inside the `readGeometry` function without any change to `xmlToString`, see my other comment

libs/api/metadata-converter/src/lib/xml-utils.ts
jahow: Oh, actually another even simpler solution:
```suggestion
  const gmlString = xmlToString(createDocument(el))
```

This way an actual document will be created out of the node, including all the namespaces.

LHBruneton-C2C:
> Hmm, I didn't think of that. Maybe there could be a way to figure out that a Polygon is in fact a bbox? something like so:

So, I'm reading the PR on how to add new spatial extents, and in fact we will only add Polygons that are bboxes. So we end up with a record with several geometries, we first clean those from the XML as this is an array, but then we don't know which one was the original bbox, as all our added geometries are bboxes... It feels like keeping the bbox field in the model is less troublesome :/

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -907,16 +908,12 @@ export function readTemporalExtents(rootEl: XmlElement) {
...
-      findChildrenElement('gmd:polygon', false),
-      mapArray((el) => {
-        const elements = el.children.filter(
-          (child) => child instanceof XmlElement
-        )
-        return readGeometry(elements[0] as XmlElement)
-      })
+      findChildElement('gmd:polygon', false),
+      firstChildElement,
+      map((el) => readGeometry(el) as Polygon | MultiPolygon)
```
LHBruneton-C2C, line 916: @jahow Can you explain why this cast to `Polygon | MultiPolygon`, when the GML can be of any kind, and OL is parsing it correctly?

jahow: sorry this can be removed, it is indeed useless!

LHBruneton-C2C:
Merging with e2e tests failure:

image

PR #946, opened 2024-07-22
Datahub: Make geo data badge more readable

tkohr:
### Description

PR removes the background color from geo data badge to make it more readable in a generic way.

**This work is sponsored by [geo2france](https://www.geo2france.fr/portail/)**.

### DIFF
--- a/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
+++ b/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
@@ -118,7 +118,7 @@ describe('dataset pages', () => {
           .find('.font-title')
           .next()
           .as('infoBar')
-        cy.get('@infoBar').children('div').should('have.length', 3)
+        cy.get('@infoBar').children().should('have.length', 3)
       })
       it('should return to the dataset list', () => {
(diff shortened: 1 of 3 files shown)


jahow:
This PR removes the background, right? Can't we simply use a normal badge, and not simply use the foreground color in that case?

tkohr:
> This PR removes the background, right? Can't we simply use a normal badge, and not simply use the foreground color in that case?

Indeed, it would be an alternative to use the same badge style as under the search input (with `badge-btn bg-primary-opacity-50` and some more modifications). Personally, I had a slight preference to the outcome with a style consistent to the update date (same foreground color and no badge layout anymore), but if we really want to keep a badge layout this is an option.

tkohr:
A second look using light text and opacity 1 for the background via the `gn-ui-badge` convinced me of its outcome :

image

I made the opacity configurable along the way. Hope this is what you had in mind @jahow.

jahow (approved):
Looking good, thanks!! The margins look a bit off compared to the mockup: 
image

I'll let you judge, merge when you want!

tkohr:
Hum, the figma mockup is a little different I have the impression

PR #949, opened 2024-07-24
Use geospatial-sdk for maps

jahow:
### Description

This PR refactors significantly the map-related logic of GeoNetwork-UI to rely mostly on the [geospatial-sdk](https://github.com/camptocamp/geospatial-sdk) library.

How things work now:
* The `gn-ui-map-container` component is used to display a map; it only takes a map context (`MapContext` type from geospatial-sdk) as an input, and offers the following outputs: `featuresClicked`, `featuresHovered`, `mapClick`
  * this component will add basemap layers to the received context, as well as apply view constraints (max zoom, max extent...)
  * the basemap layers and view constraints can be changed by using InjectionTokens
  * this components provides an `openlayersMap` promise that resolves to the OL map when it's ready, if necessary
* The `gn-ui-map-state-container` component is connected to the map state, accessible through the `MapFacade`; this facade lets us provide a new context to the map, and stores currently selected features

Depending on the use case, a map can be shown with or without using the state; both ways are valid.

A lot of code was removed and brought back into the geospatial-sdk: map and layer creation, extent computation, etc.

**Important**: these features are _not_ working yet and can be considered regressions:
* When selecting a vector feature on a map, it doesn't appear in a different color
* When zooming in on a feature, there is currently no animation

### Architectural changes

The following classes are gone:
* feature info service
* map instance
* map manager
* map container
* map context

Those features are now handled by the geospatial-sdk.

ogc-client was updated.

Several components were made standalone in the process. The UiMap module is gone entirely because all of its components are now standalone!


jahow:
Most of the features should work; still working on the missing tests

apps/datahub/src/app/app.module.ts
```diff
@@ -229,6 +238,30 @@ export const metaReducers: MetaReducer[] = !environment.production ? [] : []
...
       useValue: `${ROUTER_ROUTE_SEARCH}?${ROUTE_PARAMS.PUBLISHER}=\${name}`,
     },
+    {
+      provide: DO_NOT_USE_DEFAULT_BASEMAP,
+      useFactory: () => getOptionalMapConfig()?.DO_NOT_USE_DEFAULT_BASEMAP,
+    },
+    {
+      provide: BASEMAP_LAYERS,
+      useFactory: () =>
+        getOptionalMapConfig()?.MAP_LAYERS.map(getMapContextLayerFromConfig),
```
tkohr, line 248: I might be missing something, but shouldn't the token rather be called `MAP_LAYERS`?

libs/feature/map/src/lib/add-layer-from-catalog/add-layer-record-preview/add-layer-record-preview.component.ts
```diff
@@ -38,34 +34,35 @@ export class AddLayerRecordPreviewComponent extends RecordPreviewComponent {
...
+  getLayerFromLink(link: DatasetOnlineResource): Observable<MapContextLayer> {
     if (link.type !== 'service')
       return throwError(
         () => 'map layer could not be built for this distribution'
       )
     if (link.accessServiceProtocol === 'wms') {
       return of({
         url: link.url.toString(),
-        type: MapContextLayerTypeEnum.WMS,
+        type: 'wms',
```
tkohr, line 59: Shouldn't we keep some constants here?

tkohr (approved):
Waooo, thanks for the huge work @jahow ! It looks very clean and the geospatial-sdk seems handy to use. Maybe some words on the possibilities how to use map (with and without a map state) could be useful in the gn-ui doc?

When testing, most things worked smoothly, the only things I've noticed were:
- map-viewer:
  - an added layer does not show a title in the layer tree
  - when removing a layer via 'delete' button the app reloads
layers
  
- web components:
  - http://localhost:8001/webcomponents/gn-dataset-view-map.sample.html works fine
  - http://localhost:8001/webcomponents/gn-map-viewer.sample.html does not work, but not sure it did before;-)

- storybook:
  - selecting a row in the geotable zooms strangely (maybe to 0, 0), but not sure how it beghaved before
geotable

I didn't test the datafeeder. Not sure if the failing CI is currently expected.

libs/feature/map/src/lib/add-layer-from-catalog/add-layer-record-preview/add-layer-record-preview.component.ts
jahow: the SDK model uses string literals, it's simpler than constants and still efficient for typechecking :)

apps/datahub/src/app/app.module.ts
jahow: The token is used to add "base layers", layers that will be shown below any other layer that we ask for; I think the name "basemap layers" make sense. It is called "map layers" in the config but that name would be more confusing I think.

tkohr: Ah indeed, I ignored the fact that the config layers are base layers.

jahow:
Added a bit of doc :) thanks for the review @tkohr!

PR #950, opened 2024-07-25
ME: delete a record and its draft, delete a draft without record

LHBruneton-C2C:
### Description

This PR introduces deleting of a record and its draft, or a draft without, but not rollback of draft with a record.

### Architectural changes

The testing tool ng-mocks was added to the devDeps, and used in the files already concerned by this PR.

### Screenshots

image
image

- [X] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created


libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -323,6 +333,10 @@ export class Gn4Repository implements RecordsRepositoryInterface {
     )
   }
 
+  draftIsTemporary(uniqueIdentifier: string): boolean {
+    return uniqueIdentifier.startsWith('TEMP-ID-')
+  }
```
LHBruneton-C2C, line 338: @jahow I tried to do a proper check here, with loading the record XML, but there seems to be a retry mecanism, ending up with multiple requests, each taking more and more time.
Do you have any clue about this? Or an idea to make this check more robust?

LHBruneton-C2C:
@Angi-Kinas If you want, you can do a first review of this PR. I'll try to setup ng-mocks in a dedicated commit.

libs/ui/search/src/lib/results-table/results-table.component.ts
```diff
@@ -103,6 +99,10 @@ export class ResultsTableComponent {
     this.duplicateRecord.emit(item as CatalogRecord)
   }
 
+  handleDelete(item: unknown) {
```
Angi-Kinas, line 102: Can the type be set here already as `CatalogRecord`? I see it's also unknown in the other functions, but was wondering if that could be changed.

libs/ui/search/src/lib/results-table/results-table.component.ts
```diff
@@ -48,11 +41,14 @@ export class ResultsTableComponent {
   @Input() selectedRecordsIdentifiers: string[] = []
   @Input() sortOrder: SortByField = null
   @Input() recordHasDraft: (record: CatalogRecord) => boolean = () => false
+  @Input() recordIsDraft: (record: CatalogRecord) => boolean = () => false
```
Angi-Kinas, line 44: What is the difference between `recordHasDraft` and `recordIsDraft`?

libs/ui/search/src/lib/results-table/action-menu/action-menu.component.ts
```diff
@@ -1,22 +1,74 @@
...
+        cancelText: this.translateService.instant(
+          'editor.record.delete.confirmation.cancelText'
+        ),
+      },
+      restoreFocus: false,
+    })
+
+    // Manually restore focus to the menu trigger since the element that
+    // opens the dialog won't be in the DOM any more when the dialog closes.
+    dialogRef.afterClosed().subscribe((result) => {
```
Angi-Kinas, line 67: Is the "cancel of the deletion"-case also considered here?
Also do we need to unsubscribe from this Observable?

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    public dialogRef: MatDialogRef<ConfirmationDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public data: ConfirmationDialogData
+  ) {}
+
+  onConfirm() {
+    this.dialogRef.close(true)
+  }
+
+  onCancel() {
+    this.dialogRef.close(false)
```
Angi-Kinas, line 35: The dialog should close on cancel, right? Then I am wondering why there is `close(false)` instead of `close(true)`. Or do I understand something wrong?

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.stories.ts
```diff
@@ -0,0 +1,47 @@
+import { Component, Input } from '@angular/core'
+import { MatDialog } from '@angular/material/dialog'
+import { ButtonComponent } from '@geonetwork-ui/ui/inputs'
+import { Meta, moduleMetadata, StoryObj } from '@storybook/angular'
+import { ConfirmationDialogComponent } from './confirmation-dialog.component'
+
+@Component({
+  selector: 'gn-ui-launcher',
+  template: ` <gn-ui-button (buttonClick)="launch()">Open</gn-ui-button> `,
```
Angi-Kinas, line 9: Should this component be standalone maybe?

libs/feature/search/src/lib/results-table/results-table-container.component.spec.ts
```diff
@@ -1,71 +1,49 @@
...
-        },
-        {
-          provide: SelectionService,
-          useClass: SelectionServiceMock,
-        },
-        {
-          provide: RecordsRepositoryInterface,
-          useClass: RecordsRepositoryMock,
-        },
+        MockProviders(
```
Angi-Kinas, line 31: This is so much nicer! Thanks for introducing ng-mocks :tada:

jest.setup.ts
```diff
@@ -43,3 +54,31 @@ beforeEach(() => {
     value: new LocalStorageRefStub().getLocalStorage(),
   })
 })
+
+// ng-mocks global configuration
```
Angi-Kinas, line 58: :+1:

libs/ui/search/src/lib/results-table/results-table.component.ts
LHBruneton-C2C: `recordHasDraft` means a record with a draft (ie. changed locally but not published), `recordIsDraft` means a draft without record (ex: new record not yet published, duplicate record not yet published).

But you're right, this is not clear enough, I'll change the name.

libs/ui/search/src/lib/results-table/action-menu/action-menu.component.ts
LHBruneton-C2C: This is a short lived observable, only as long as the dialog is open, so no need to unsubscribe.
But good call!

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
LHBruneton-C2C: The value inside the `close()` call is the returned value of the dialog. So the dialog is closed either way, but we can distinguish between confirm and cancel.

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.stories.ts
LHBruneton-C2C: This is only a wrapper for the dialog mechanism to work properly in storybook, so no bother ;)

libs/ui/search/src/lib/results-table/results-table.component.ts
Angi-Kinas: I see! Maybe `draftWithoutRecord`?

libs/ui/search/src/lib/results-table/results-table.component.ts
LHBruneton-C2C: I just tried again, but the table component is generic (it can work with items of any types as long as the data matches the template), so the row is of unknown type in the event.

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
Angi-Kinas: Ok makes sense :+1:

libs/ui/search/src/lib/results-table/results-table.component.ts
LHBruneton-C2C: I went for `hasDraft` (can only be a record then), `isDraft` and `isDraftWithoutRecord`.

jahow:
@LHBruneton-C2C thank you for starting with ng-mocks, don't forget to include this in the best practices (to rename "code guide") in the docs!

apps/metadata-editor-e2e/src/e2e/delete.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+  describe('record with draft', () => {
+    it('should delete the record, delete its associated draft and refresh the interface', () => {
+      // First create a record and its draft
+      cy.get('[data-cy="create-record"]').click()
+      cy.get('gn-ui-form-field[ng-reflect-model=abstract] textarea').type(
+        'record abstract'
+      )
+      cy.get('md-editor-publish-button').click()
+      // eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(1000) // waiting for publication
```
jahow, line 16: Ok I understand better why you added the wait here; it makes sense, you might be able to replace it with a named request like in https://docs.cypress.io/api/commands/wait#Wait-for-a-specific-request-to-respond but it's not such a big deal IMO

libs/ui/search/src/lib/results-table/results-table.component.ts
```diff
@@ -47,12 +40,16 @@ export class ResultsTableComponent {
   @Input() records: CatalogRecord[] = []
   @Input() selectedRecordsIdentifiers: string[] = []
   @Input() sortOrder: SortByField = null
-  @Input() recordHasDraft: (record: CatalogRecord) => boolean = () => false
+  @Input() hasDraft: (record: CatalogRecord) => boolean = () => false
+  @Input() isDraft: (record: CatalogRecord) => boolean = () => false
```
jahow, line 44: After thinking about it, I feel like these two distinct functions might be confusing. When we're showing a list of records, we're showing _records_ and nothing else; drafts are not another type of records, they are only a bunch of pending changes that were not saved yet.

As such, a record would only have two different "flags":
* "hasDraft" means the record has pending changes that were not saved yet, and
* "notSavedYet" means the record was never saved (i.e. equivalent to "isDraftWithoutRecord")

I think it's important here to not give drafts an sort of existence on their own as a separate type of entity, otherwise it's going to be confusing in the long run. What do you think?

jahow: As for the fact that we currently don't allow deleting a draft from the draft menu if it has been saved already, I would go with a simple `allowDelete: (record) => bool` optional input that would always return true by default. I feel like this might be temporary anyway since we'll need to find a way to handle this case when rolling back will be implemented.

jahow (commented):
Thank you, very good work!! Also nice job on the E2E tests :slightly_smiling_face: 

I made a few comments on top of what we discussed. I'll give a last review once you make the modifications on the records repository. Please let me know if anything is unclear. Thanks!

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.html
```diff
@@ -0,0 +1,12 @@
+<h1 mat-dialog-title>{{ data.title }}</h1>
+<div mat-dialog-content>{{ data.message }}</div>
+<div mat-dialog-actions>
+  <gn-ui-button (buttonClick)="onCancel()">{{ data.cancelText }}</gn-ui-button>
+  <gn-ui-button
+    (buttonClick)="onConfirm()"
+    cdkFocusInitial
```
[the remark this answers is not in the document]
LHBruneton-C2C: Oh right, that was about accessibility, but the last focused element is the action menu button, which has been removed when the modal was opened... I guess we'll just let the focus go to the next element in focus order.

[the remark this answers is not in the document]
LHBruneton-C2C: Hm in fact it's how we preselect the confirm button when opening the dialog, otherwise it's the cancel button that is selected first. What's the most strange is that the directive actually works, even with the warning... Is it that bad to keep the warning?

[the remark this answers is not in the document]
jahow: I agree it's important to keep it; it's really not that bad to have a warning in the console, it can be considered technical debt :)

jahow:
oh also feel free to rebase to have the e2e tests actually working from time to time :)

libs/ui/search/src/lib/results-table/results-table.component.ts
LHBruneton-C2C: I'll try moving the `canDuplicate` and `canDelete` to the inputs of the `results-table`.

LHBruneton-C2C:
TODO: documentation!!!

libs/ui/search/src/lib/results-table/results-table.component.ts
jahow: yes that's what I meant, sorry that wasn't super clear

PR #951, opened 2024-07-29
EDITOR - Record field: Contacts for resource

rcaplier:
### Description

This PR introduces the contacts for resources field.

image

[...]


libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,188 @@
...
+      (organizations) => {
+        this.allOrganizations = new Map<string, Organization>(
+          organizations.map((organization) => [organization.name, organization])
+        )
+        this.changeDetectorRef.markForCheck()
+      }
+    )
+  }
+
+  async ngOnChanges(changes: SimpleChanges): Promise<void> {
```
Angi-Kinas, line 84: Does this function have to be async?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    TranslateModule,
+    ContactCardComponent,
+  ],
+})
+export class FormFieldContactsForResourceComponent
+  implements OnInit, OnChanges
+{
+  @Input() control: FormControl<Individual[]>
+
+  allUser$: Observable<UserModel[]>
```
Angi-Kinas, line 51: I would add a plural `s` to allUser, since it will be an array of users.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,188 @@
...
+
+      this.filterRolesToPick()
+
+      this.changeDetectorRef.markForCheck()
+    }
+  }
+
+  addRoleToDisplay(roleToAdd: string) {
+    this.rolesToDisplay.push(roleToAdd)
+    this.filterRolesToPick()
```
Angi-Kinas, line 109: I would rename this function to maybe `handleRolesToDisplay`, since it is adding and filtering. Just a suggestion.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,188 @@
...
+      ]),
+      debounceTime(300),
+      distinctUntilChanged()
+    )
+  }
+
+  /**
+   * gn-ui-autocomplete
+   */
+  async addContact(contact: UserModel, role: string) {
```
Angi-Kinas, line 151: Does this have to be async?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,188 @@
...
+    let newContactsforRessource = {
```
Angi-Kinas, line 152:
```suggestion
    let newContactsForRessource = {
```
really minor thing, sorry :D

libs/feature/editor/src/lib/components/contact-card/contact-card.component.html
```diff
@@ -0,0 +1,22 @@
...
+      <div class="flex flex-row justify-between">
+        <span class="flex flex-wrap font-bold w-full"
+          >{{ contact.firstName }} {{ contact.lastName }}</span
+        >
+        <gn-ui-button
+          data-test="removeContactButton"
+          type="light"
+          extraClass="w-[20px] h-[20px] flex items-center justify-center"
+          (buttonClick)="removeContact(contact)"
+          ><span class="text-[14px] leading-[0] font-bold">&#x2715;</span>
```
Angi-Kinas, line 16:
```suggestion
          ><span class="material-symbols-outlined">
close
</span>
```
Should we rather use the material symbol? it is more readable at least.

Angi-Kinas (commented):
@rcaplier thank you for this feature! I haven't tested anything yet, just had a look at the code. It's really readable, thanks for that!
I have some suggestions, most of them minor things :)

libs/feature/editor/src/lib/fields.config.ts
```diff
@@ -83,6 +83,14 @@ export const RECORD_ABSTRACT_FIELD: EditorField = {
   },
 }
 
+export const CONTACTS_FOR_RESOURCE_FIELD: EditorField = {
+  model: 'contactsForResource',
+  formFieldConfig: {
+    // labelKey: marker('editor.record.form.field.contactsForResource')
```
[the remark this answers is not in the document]
rcaplier: It doesn't have a label.

translations/en.json
```diff
@@ -177,9 +177,6 @@
   "editor.record.form.field.title": "Metadata title",
   "editor.record.form.field.uniqueIdentifier": "Unique identifier",
   "editor.record.form.field.updateFrequency": "Update frequency",
-  "editor.record.form.abstract": "Abstract",
```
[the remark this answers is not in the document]
rcaplier: Because I refactored it when I did the pages and sections ticket.

libs/common/domain/src/lib/model/record/contact.model.ts
```diff
@@ -24,6 +25,39 @@ export const RoleValues = [
   'user', // Party who uses the resource
 ]
 
+export const RoleLabels = new Map<string, string>([
+  ['unspecified', marker('metadata.contactForResource.role.unspecified')],
```
jahow, line 29:
```suggestion
  ['unspecified', marker('domain.contact.role.unspecified')],
```

This kind of translation key would make more sense here IMO

libs/feature/editor/src/lib/components/contact-card/contact-card.component.html
```diff
@@ -0,0 +1,28 @@
+<ng-container *ngIf="organization && contact">
```
jahow, line 1: I think it is fair to assume that a contact and an org will be provided for this component, so no need to have this check

libs/feature/editor/src/lib/components/contact-card/contact-card.component.ts
```diff
@@ -0,0 +1,33 @@
...
+  selector: 'gn-ui-contact-card',
+  templateUrl: './contact-card.component.html',
+  styleUrls: ['./contact-card.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  imports: [CommonModule, MatIconModule, ButtonComponent],
+})
+export class ContactCardComponent {
+  @Input() contact: Individual
+  @Input() organization: Organization
```
jahow, line 26: I'm surprised that this input is needed; the `Invididual` model already contains an organization, shouldn't that be enough?

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -61,6 +61,7 @@ export class AutocompleteComponent
   @Input() action: (value: string) => Observable<AutocompleteItem[]>
   @Input() value?: AutocompleteItem
   @Input() clearOnSelection = false
+  @Input() completeOnSelection = false
```
jahow, line 64: I think this new input makes sense; but it looks like the existing components (e.g. fuzzy search) which relied on "clearOnSelection" haven't been changed to use this new input instead.

Also would it be possible to add a bit of doc in a comment to clarify what both of these do? thanks!

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+    'owner',
+    'point_of_contact',
+    'author',
+  ]
+
+  rolesToDisplay = []
+
+  allOrganizations: Map<string, Organization> = new Map()
+
+  addOptions: Array<{ buttonLabel: string; eventName: string }> = []
```
jahow, line 71: I think this is not needed anymore

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+  ],
+})
+export class FormFieldContactsForResourceComponent
+  implements OnInit, OnChanges
+{
+  @Input() control: FormControl<Individual[]>
+
+  allUsers$: Observable<UserModel[]>
+
+  rolesToPick: string[] = [
```
jahow, line 59:
```suggestion
  rolesToPick: Role[] = [
```

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    if (
+      contactsForResource &&
+      contactsForResource.currentValue !== contactsForResource.previousValue
+    ) {
+      const rolesToAdd = contactsForResource.currentValue.value.map(
+        (contact: Individual) => contact.role
+      )
+
+      console.log(contactsForResource.currentValue)
```
jahow, line 103: to remove

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+  addRoleToDisplay(roleToAdd: string) {
+    this.rolesToDisplay.push(roleToAdd)
+    this.filterRolesToPick()
+  }
+
+  filterRolesToPick() {
+    this.rolesToPick = this.rolesToPick.filter(
+      (role) => !this.rolesToDisplay.includes(role)
+    )
+  }
```
jahow, line 126: To add to the comment from @Angi-Kinas , I think you can even do this in a more straightforward way with getters:

```ts
const availableRoles = [
    'resource_provider',
    'custodian',
    'owner',
    'point_of_contact',
    'author',
  ]

get rolesToPick() {
  return availableRoles.filter(
      (role) => !this.control.value.any(contact => contact.role === role)
    )
}

get rolesToDisplay() {
  return this.control.value.reduce((prev, curr) => prev.indexOf(curr.role) > -1 ? prev : [...prev, curr.role], [])
}
```

(untested!)

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  protected roleToLabel(role: string): string {
+    return RoleLabels.get(role)
+  }
+
+  /**
+   * gn-ui-autocomplete
+   */
+  displayWithFn: (user: UserModel) => string = (user) =>
+    `${user.name} ${user.surname}`
```
jahow, line 143: Maybe add the org name in parenthesis here, if there's one?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,220 @@
...
+      (organizations) => {
+        this.allOrganizations = new Map<string, Organization>(
+          organizations.map((organization) => [organization.name, organization])
+        )
+        this.changeDetectorRef.markForCheck()
+      }
+    )
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
```
jahow, line 92: See my comment below about using getters, that would mean that you wouldn't have to rely on ngOnChanges anymore

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
rcaplier: I think you meant some not any.
But anyway, I agree that this is a one liner right, but I don't think this is more readable and more importantly the list of role to display is not only the list of role coming from the input control, it should also contain the ones that the user just clicked !!  So I have to manage another array.

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
jahow: > So I have to assume that all the components that were using the clearOnSelection in fact wanted an completeOnSelection in the first place ? How can we be sure about that ? Maybe what they wanted was clearOnSelection since that what was asked in the first place ? I don't get the logic here.

The logic is that we don't want to break existing stuff :)

> A doc ? It seems to be straight forward to me but ok sure. In which form ? A jsdoc comment above the function that use it or a single comment over each input ?

just a comment above the function, at the very least

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,264 @@
...
+        await lastValueFrom(this.organizationsServiceInterface.organisations$)
+      ).map((organization) => [organization.name, organization])
+    )
+    this.updateContactsForRessource()
+    this.manageRoleSectionsToDisplay(this.control.value)
+    this.filterRolesToPick()
+
+    this.changeDetectorRef.markForCheck()
+
+    this.subscription.add(
```
jahow, line 100: is this subscription supposed to be tore down somewhere?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,264 @@
...
+      this.control.valueChanges.subscribe((contactsForResource) => {
+        console.log(contactsForResource)
```
jahow, line 102: to remove

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -0,0 +1,264 @@
...
+      email: contact.email ?? '',
+      role,
+      address: '',
+      phone: '',
+      position: '',
+    } as Individual
+
+    const newControlValue = [...this.control.value, newContactsForRessource]
+
+    console.log(newControlValue)
```
jahow, line 260: to remove

jahow (commented):
Some remaining comments on my side, I'll let @Angi-Kinas give the final review since she was the one looking in to this first :) thanks!

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
jahow: Set is indeed a better way to remove duplicates in an array :) I see you're not going with the getter-based approach; it works the way it is but please keep in mind that using getters for computing values in an immediate way instead of maintaining a local intermediary state (in that case the `contactsForRessourceByRole` and `contactsAsDynElemByRole` variables) gives better code readability and is less error-prone because the intermediary state does not need to be kept up-to-date manually

jahow (approved):
looking good! thanks!

PR #957, opened 2024-08-07
[Editor]: Add an open data switch

cmoinier:
### Description

This PR introduces an open-data switch that controls whether the license & constraints fields are shown. It is also used to set the license to a default open data one if the switch is checked.

### Notes

A new dump was added for the e2e tests.
Since the switch was not present in the mockups, I have added it in the section where it makes the most sense to me (also because the data needed is easier to get in this section), but I'm open to other suggestions !

### Screenshots

image


libs/feature/editor/src/lib/components/record-form/record-form.component.html
```diff
@@ -33,7 +33,15 @@
               trackBy: fieldTracker
             "
           >
-            <ng-container *ngIf="!field.config.hidden">
+            <gn-ui-check-toggle
```
Angi-Kinas, line 36: This render logic should be in a new component (e.g. `form-field-open-data.component`).
The new component can be rendered inside of /components/record-form/form-field/form-field.component.html and the model gets checked in the corresponding `form-field.component.ts` file. You can take the example from the title:
`  get isTitle() {
    return this.model === 'title'
  }`

conf/default.toml
```diff
@@ -37,6 +37,9 @@ proxy_path = ""
 # Enables displaying a "contact block" wherever relevant in applications
 # contact_email = "opendata@mycompany.com"
 
+# Allows to customize the list of licenses considered as "open data".
```
Angi-Kinas, line 40: Maybe we should mention the editor here as well?
```suggestion
# Allows to customize the list of licenses inside the metadata editor considered as "open data".
```
WDYT?

Angi-Kinas (changes requested):
Thanks @cmoinier for the open data switch component and you contribution to the metadata-editor!

I think it would be nice to split the logic a bit in "smart" and "dumb" components, to keep the current setup.
Let me know if you have any questions regarding my comments.

libs/ui/inputs/src/lib/check-toggle/check-toggle.component.ts
```diff
@@ -18,11 +19,59 @@ import { FormsModule } from '@angular/forms'
...
+  //   }
+  // }
   @Input() color: 'primary' | 'secondary' = 'primary'
+  @Input() model: string
   @Output() toggled = new EventEmitter()
+  checked = false
 
-  toggle(event: Event) {
-    this.toggled.emit(event)
+  ngOnChanges(changes: SimpleChanges) {
```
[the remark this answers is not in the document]
cmoinier: I wondered if it was worth creating a component just for this, but alright I will :) I'm just not 100% sure how to make it display in the right place (I don't really understand how the list of components in the "form-field" component knows where to display the components).

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -8,19 +8,38 @@ import {
...
+  isHidden: boolean
+
+  get config() {
+    console.log(getGlobalConfig().LICENSES)
+    return getGlobalConfig().LICENSES
+  }
   constructor(public facade: EditorFacade) {}
 
-  handleFieldValueChange(model: string, newValue: EditorFieldValue) {
+  onOpenDataToggled(args) {
```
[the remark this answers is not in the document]
cmoinier: Yes, true!

conf/default.toml
cmoinier: Yes, good idea :+1:

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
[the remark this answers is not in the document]
cmoinier: Actually, not sure I agree entirely. I think the logic of setting `isHidden` has to kept here right? because it's where we decide if the field is going to be displayed or not, so the information has to come here in the form of an Output function. WDYT?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-open-data/form-field-open-data.component.ts
```diff
@@ -0,0 +1,43 @@
...
+import { CheckToggleComponent } from '@geonetwork-ui/ui/inputs'
+import { getGlobalConfig } from '@geonetwork-ui/util/app-config'
+import { TranslateModule } from '@ngx-translate/core'
+
+@Component({
+  selector: 'gn-ui-form-field-open-data',
+  templateUrl: './form-field-open-data.component.html',
+  styleUrls: ['./form-field-open-data.component.css'],
+  standalone: true,
+  imports: [CheckToggleComponent, TranslateModule],
```
LHBruneton-C2C, line 19: We try to create all new components with the OnPush changeDetectionStrategy, as a best practice, but I agree here it doesn't make much of a difference on performance.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-open-data/form-field-open-data.component.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Input() control: FormControl
+  value = false
+  @Output() visibilityChange = new EventEmitter<boolean>()
+
+  get config() {
+    return getGlobalConfig().LICENSES
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    if ('control' in changes) {
```
LHBruneton-C2C, line 31: I'm working on this for the draft rollback, so not critical, but it's best to know, the formControl will never change, it's its value that's changing, and you detect that by subscribing on valueChanges.

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -8,19 +8,21 @@ import {
   EditorFieldWithValue,
   EditorSectionWithValues,
 } from '../../+state/editor.models'
+import { getGlobalConfig } from '@geonetwork-ui/util/app-config'
+import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
```
LHBruneton-C2C, line 12: What do you need these for?

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -8,19 +8,21 @@ import {
...
   changeDetection: ChangeDetectionStrategy.OnPush,
   standalone: true,
-  imports: [CommonModule, FormFieldComponent, TranslateModule],
+  imports: [CommonModule, FormFieldComponent, TranslateModule, UiInputsModule],
 })
 export class RecordFormComponent {
   constructor(public facade: EditorFacade) {}
 
-  handleFieldValueChange(model: string, newValue: EditorFieldValue) {
+  handleFieldValueChange(model: any, newValue: EditorFieldValue) {
```
LHBruneton-C2C, line 25: Why did you need this? The model is a key into a dictionary of values, the key being always a string.

LHBruneton-C2C (changes requested):
This is nice work, but I think you did split the components a bit too much. The logic is mostly one sub-component by record field (not true for spatial extents, but that's the only exception as of now). Here, the open data switch is not an information stored on the record, rather a shortcut for the license edition. As such, I would pull all your logic inside the license sub-component, and not create your own.

LHBruneton-C2C:
Also, this is not DH work anymore, for the branch name, but not sure you can change it after the PR has been created.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -57,4 +57,49 @@ describe('editor form', () => {
     })
     cy.get('md-editor-publish-button').click()
   })
+
+  describe('Access and contact', () => {
```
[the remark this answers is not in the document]
cmoinier: I think so yes, there was a rebase around this line, I think something weird happened? Good catch :)

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -57,4 +57,49 @@ describe('editor form', () => {
...
+    describe('Open data switch', () => {
+      describe('When the open data switch is checked', () => {
+        beforeEach(() => {
+          cy.visit('/edit/accroche_velos')
+          cy.get('md-editor-page-selector').find('gn-ui-button').last().click()
+        })
+        it('should not display the licence section', () => {
```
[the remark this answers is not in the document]
cmoinier: Oh okay! Makes sense, will change that.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -57,4 +57,49 @@ describe('editor form', () => {
...
+          cy.get('gn-ui-form-field-license').should('not.exist')
+        })
+        it('should display the license section when toggled', () => {
+          cy.get('gn-ui-check-toggle').find('span').first().click()
```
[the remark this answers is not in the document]
cmoinier: Fair enough, will add a data-cy selector ;)

I think I did try to click on the toggle directly, because it would be strange to complicate the test directly without doing so. With the selector it might work directly

libs/feature/editor/src/lib/components/record-form/form-field/form-field-open-data/form-field-open-data.component.ts
cmoinier: Oh I just didn't notice that you were using that strategy! I don't mind changing it, this was not on purpose, just default.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-open-data/form-field-open-data.component.ts
cmoinier: Good to know, will subscribe then :)

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
cmoinier: I need getGlobalConfig to get the list of open-data licenses, stored in the config file.
And I need UIInputsModule because the switch component is in that lib (I think? It could also be leftover from v1 of the PR, will check)

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
cmoinier: I think this is also leftover from v1, when I had changed that function a bit. Will rollback.

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
LHBruneton-C2C: Just so you know, all new components are implemented as standalone and imported directly. We try not to import Angular modules anymore in this project.

As for the licenses configuration, it could be part of the license field configuration in `libs/feature/editor/src/lib/fields.config.ts`, but I'll let the final word to @jahow.

jahow: Indeed it should go into fields.config.ts! thanks!

cmoinier: Noted, thanks :slightly_smiling_face:

cmoinier:
> This is nice work, but I think you did split the components a bit too much. The logic is mostly one sub-component by record field (not true for spatial extents, but that's the only exception as of now). Here, the open data switch is not an information stored on the record, rather a shortcut for the license edition. As such, I would pull all your logic inside the license sub-component, and not create your own.

So you would remove the `form-field-open-data` component altogether ? I'm a bit confused by this remark since @Angi-Kinas recommended the exact opposite last week (see [here](https://github.com/geonetwork/geonetwork-ui/pull/957#discussion_r1709006522), and other comments)  :slightly_frowning_face: 

I think a good middle-ground between your ideas would be to move only the config check logic (eg initialization & subscriptions) to the licenses component, but to keep the open-data component with the toggle logic, WDYT?

LHBruneton-C2C (approved):

PR #960, opened 2024-08-13
Fix E2E test failures

jahow:
### Description

This PR tweaks E2E tests further. Issues fixed are:
* sometimes the custom thesauri (GEMEt and INSPIRE themes) in the `init ` step were pushed to GN, but then the thesauri would be reset and the "regions" thesauri would be pushed instead; I added a delay before pushing the custom thesauri and it always works now!
* I also added a print of the indexation errors happening in the `init` step, so that we can make sure that indexation happens exactly like it does locally (it wasn't the case before the above change)
* I added a dependency from the metadata-editor E2E to datahub E2E because if the editor tests pass first they change the records so that the datahub tests will fail after that; obviously this is not ideal (it shows that the datahub tests are too sensitive) but we can live with that for now

### Architectural changes
none

### DIFF
--- a/apps/metadata-editor-e2e/project.json
+++ b/apps/metadata-editor-e2e/project.json
@@ -13,6 +13,12 @@
         "browser": "chrome",
         "port": "cypress-auto"
       },
+      "dependsOn": [
+        {
+          "projects": ["datahub-e2e"],
+          "target": "e2e"
+        }
+      ],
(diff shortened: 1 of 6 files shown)


LHBruneton-C2C (approved):
Thanks for taking the time to fix the e2e test!

PR #969, opened 2024-08-25
[Editor] : Dashboard columns

cmoinier:
## To visualize the right result, you have to switch to the "my-records" tab & switch back to the "main" dashboard (known bug, but out of scope to fix here)

This PR makes a few changes on the dashboard columns of the editor, and to the record's ownership.

- [x] In the "Etat" column, shows if a record was ever published
- [x] In the "Modified le" column, shows if a record was never published by showing "-" instead of a date.
- [x] In the "Modifié par" column, shows the record's owner.
- [x] When saving a record, it transfers the ownership of the record to the last person who saved it.
- [x] PR also fixes a bug that used to not show all of the records on /catalog/search
- [x] Renames "canDelete" to "isUnsavedDraft" because it is now used to multiple purposes. Exception made of the action menu component which still uses canDelete from the draft component

---

**This work is sponsored by [Organization ABC](xx)**.


libs/ui/search/src/lib/results-table/results-table.component.html
```diff
@@ -117,7 +122,7 @@
       <span translate>record.metadata.updatedOn</span>
     </ng-template>
     <ng-template #cell let-item>
-      {{ dateToString(item.recordUpdated) }}
+      {{ isUnsavedDraft(item) ? '-' : dateToString(item.recordUpdated) }}
```
Angi-Kinas, line 125: I wished this column would be centered horizontally. I think that it would look nicer:
image

Not sure if it is necessary though. Let me know what you think.
All the cells use the same styling, so it would be a bit more specific to change this one. Just an idea of how you could center it without changing the other cells styling could be:

```    
    <ng-template #cell let-item>
      <div
        *ngIf="isUnsavedDraft(item); else date"
        class="flex justify-center w-full"
      >
        -
      </div>
      <ng-template #date
        ><div class="flex justify-center w-full">
          {{ dateToString(item.recordUpdated) }}
        </div></ng-template
      >
    </ng-template>
```

Angi-Kinas (commented):
Thanks @cmoinier for changing the columns, this will help the users to really see the information they are looking for.

I didn't check the code too much yet, but I tried to verify the changes in the app. Here are some ideas and findings I had:
- I could only see the "Updated date" filled with information in the "my drafts" table but not on the main "metadata records" table
- For the "Updated date"-column I think it would look nicer if it would be centered (see comment), but this can be discussed
- For the "Edited by"-column I only see a "person"-icon, but no name, even when I hover over it. Is there something missing?
- The status column looks as expected if I understood correctly :+1: 

Let me know if something is unclear.

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
```diff
@@ -42,9 +48,48 @@ export class PublishButtonComponent {
...
+              groupId ? groupId.id : 0,
+              Number(userId.id)
+            )
+            .subscribe(
+              () => {
+                console.log('Ownership set successfully')
+              },
+              (error) => {
+                console.log('Error setting ownership :', error)
+              }
```
LHBruneton-C2C: I get it for dev purpose, but I don't think we should pollute the console log with this.

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
```diff
@@ -42,9 +48,48 @@ export class PublishButtonComponent {
...
+  record$ = this.facade.record$
+  activeUser$: Observable<UserModel>
+
+  constructor(
+    private facade: EditorFacade,
+    private recordsApiService: RecordsApiService,
+    private platformService: PlatformServiceInterface,
+    private groupApiService: GroupsApiService
+  ) {
+    this.activeUser$ = this.platformService.getMe()
```
LHBruneton-C2C, line 56: Can't we call directly `this.platformService.getMe()` where it's neded. the `activeUser$` variable doesn't bring any added value, does it?

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
```diff
@@ -42,9 +48,48 @@ export class PublishButtonComponent {
...
+            (grp) => grp.name === record.ownerOrganization.name
+          )
+          // setting to 0 by default because the API expects a number and not null & most records don't have a group
+          this.recordsApiService
+            .setRecordOwnership(
+              record.uniqueIdentifier,
+              groupId ? groupId.id : 0,
+              Number(userId.id)
+            )
+            .subscribe(
```
LHBruneton-C2C: Sooo many subscribe... Can't we avoid this through better pipes?

LHBruneton-C2C (approved):
Tested OK, thanks for this clean PR!
A few remarks, but nothing critical, so I'm approving already, and you can merge when you feel it's clean enough.

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
cmoinier: The groupApiService has been removed in my last commit. Can you tell me if the latest version is good with you or if it needs further rework? :)

LHBruneton-C2C: I would use a withLatestFrom to get the active user and record, as you only the current value for those, so no subscription needed there. And I would still try to put the set ownership api call inside the save pipe, with no added subscription.

Do you wanna try this? Or else I can take on.

PR #983, opened 2024-09-02
[ME]: Add coverage field (toggle)

Angi-Kinas:
### Description

This PR introduces an addition to the spatial extent field which is the coverage field. This field can be toggled between predefined values (as Keywords). For now we have "national" and "regional" as options.
When a value is selected, the corresponding Keyword will be added to the list of Keywords in the record.

## Note
**Update**: No need to add the thesauri. Having the keywords in the fields config should be enough.
I'm not sure if this is testable without me changing the dump, or manually adding the two new thesauri in geonetwork.
The urls for them are:
https://inspire.ec.europa.eu/metadata-codelist/SpatialScope/regional
https://inspire.ec.europa.eu/metadata-codelist/SpatialScope/national

And for adding them in geonetwork you need to provide the url to the RDF/XML:
national: [RDF/XML](http://inspire.ec.europa.eu/metadata-codelist/SpatialScope/national/national.en.rdf)
regional: [RDF/XML](http://inspire.ec.europa.eu/metadata-codelist/SpatialScope/regional/regional.en.rdf)

### Screenshots
image

### DIFF
--- a/libs/feature/editor/src/lib/components/record-form/form-field/form-field-spatial-extent/form-field-spatial-extent.component.spec.ts
+++ b/libs/feature/editor/src/lib/components/record-form/form-field/form-field-spatial-extent/form-field-spatial-extent.component.spec.ts
@@ -16,15 +16,7 @@ const NATIONAL_KEYWORD = {
   key: 'http://inspire.ec.europa.eu/metadata-codelist/SpatialScope/national',
   label: 'National',
   description: '',
-  type: 'place',
-  thesaurus: {
-    id: 'external.place.national.en',
-    name: 'National',
-    url: new URL(
-      'http://localhost:8080/geonetwork/srv/api/registries/vocabularies/external.place.national.en'
-    ),
-    type: 'place',
-  },
+  type: 'theme',
 }
 
 const SAMPLE_PLACE_KEYWORDS: Keyword[] = [
@@ -376,7 +368,6 @@ describe('FormFieldSpatialExtentComponent', () => {
         component = fixture.componentInstance
         fixture.detectChanges()
 
-        await component.ngOnInit()
         const results = await firstValueFrom(component.switchToggleOptions$)
         const nationalOption = results.filter(
           (result) => result.label === 'National'
@@ -394,7 +385,6 @@ describe('FormFieldSpatialExtentComponent', () => {
         component = fixture.componentInstance
         fixture.detectChanges()
 
-        await component.ngOnInit()
         const results = await firstValueFrom(component.switchToggleOptions$)
         const nationalOption = results.filter(
           (result) => result.label === 'National'
@@ -405,7 +395,7 @@ describe('FormFieldSpatialExtentComponent', () => {
     })
     describe('#onSpatialScopeChange', () => {
       it('removes all existing spatial scope keywords and add the selected one', async () => {
-        const spatialScopes = [{ label: 'national' }, { label: 'regional' }]
+        const spatialScopes = [{ label: 'National' }, { label: 'Regional' }]
 
         const allKeywords = await firstValueFrom(component.allKeywords$)
         const filteredKeywords = allKeywords.filter((keyword) => {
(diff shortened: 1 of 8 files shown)


libs/feature/editor/src/lib/components/record-form/form-field/form-field-spatial-extent/form-field-spatial-extent.component.spec.ts
```diff
@@ -11,6 +12,21 @@ import { EditorFacade } from '../../../../+state/editor.facade'
...
+  label: 'National',
+  description: '',
+  type: 'place',
+  thesaurus: {
+    id: 'external.place.national.en',
+    name: 'National',
+    url: new URL(
+      'http://localhost:8080/geonetwork/srv/api/registries/vocabularies/external.place.national.en'
+    ),
+    type: 'place',
```
tkohr, line 26: Shouldn't this be of type `theme`?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-spatial-extent/form-field-spatial-extent.component.spec.ts
```diff
@@ -345,4 +361,70 @@ describe('FormFieldSpatialExtentComponent', () => {
...
+        const nationalOption = results.filter(
+          (result) => result.label === 'National'
+        )[0]
+
+        expect(nationalOption.checked).toBe(false)
+      })
+    })
+    describe('#onSpatialScopeChange', () => {
+      it('removes all existing spatial scope keywords and add the selected one', async () => {
+        const spatialScopes = [{ label: 'national' }, { label: 'regional' }]
```
tkohr, line 408: Aren't the labels expected to be uppercase?

libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
```diff
@@ -10,7 +10,7 @@ import { MatButtonToggleModule } from '@angular/material/button-toggle'
 
 export type SwitchToggleOption = {
   label: string
-  value: string
+  value: any
```
tkohr, line 13: Is there a way we could still type this?

tkohr (commented):
Thanks @Angi-Kinas ! I've left some suggestions. The logic seems to work fine. I was not able to save a metadata record yet, but this is likely an issue on my side.

libs/ui/inputs/src/lib/switch-toggle/switch-toggle.component.ts
Angi-Kinas: I removed it completely since it will be of different value for each component that will use the SwitchToggleComponent. And it is also not used anymore inside the component. WDYT?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-spatial-extent/form-field-spatial-extent.component.ts
```diff
@@ -98,12 +108,32 @@ export class FormFieldSpatialExtentComponent {
     shareReplay(1)
   )
 
+  switchToggleOptions$: Observable<SwitchToggleOption[]>
```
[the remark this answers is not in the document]
Angi-Kinas: Yes that makes more sense! Thanks for the suggestion. It was not easy for me to assign the switchToggleOptions$ initially based on the keywords, but now I have a solution that works. Maybe you can have another look?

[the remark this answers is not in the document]
tkohr: Nice, it looks good to me. If you move the `switchToggleOptions$` behind the `allKeywords$` you could even reuse the latter and only need one map like:
```
switchToggleOptions$: Observable<SwitchToggleOption[]> =
  this.allKeywords$.pipe(
    map((keywords) =>
...
```

tkohr (approved):
Thanks for the adaption @Angi-Kinas. I was also able to retest it on my side and it works fine. I just see that there are some conflicts to be resolved before merging.

PR #985, opened 2024-09-05
[ME]: Create record by duplicating remote record.

rcaplier:
### Description

This PR introduces a new import button right next to the create button. For now the import button allow only to import with an external URL.

### Screenshots

image

image

**This work is sponsored by(xx)**.


libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -25,22 +26,32 @@ import {
...
-import { HttpErrorResponse } from '@angular/common/http'
+import {
+  HttpClient,
+  HttpErrorResponse,
+  HttpHeaders,
+} from '@angular/common/http'
 
 const TEMPORARY_ID_PREFIX = 'TEMP-ID-'
 
+export const LOCAL_STORAGE_RECORD_PREFIX = 'me-record-draft-'
```
jahow, line 44: This prefix should actually never be changed (unless a retrocompatibility system is done) because otherwise all existing drafts in users sessions will get lost and never be cleaned up

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -356,4 +348,99 @@ export class Gn4Repository implements RecordsRepositoryInterface {
       )
     )
   }
+
+  /**
+   * Retrieves a record in XML format from the API using a unique identifier.
+   *
+   * @param {string} uniqueIdentifier - The unique identifier of the record to be retrieved.
+   * @returns {Observable<string | null>} - An Observable that emits the XML content of the record as a string, or `null` if the record is not found.
+   */
```
jahow, line 357: Although I appreciate the fact that you're trying to add as much comment as possible, please keep in mind that comments need to be maintained just like code, and outdated comments can be more harmful than helpful. In the case of such a simple and straightforward function that is also private, I think it would be acceptable to leave it without a comment block

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -301,6 +274,34 @@ export class Gn4Repository implements RecordsRepositoryInterface {
     )
   }
 
+  importRecordFromExternalFileUrlAsDraft(
```
jahow, line 277: I think this should be called
```suggestion
  duplicateExternalRecord(
```

Here we're not actually importing (which would mean keeping the title and identifier intact), we're copying the record

apps/metadata-editor/src/app/records/search-records/search-records-list.component.ts
```diff
@@ -28,32 +43,88 @@ import { MatIconModule } from '@angular/material/icon'
...
+      .navigate(['/duplicate', record.uniqueIdentifier])
+      .catch((err) => console.error(err))
   }
 
   createRecord() {
-    this.router.navigate(['/create'])
+    this.router.navigate(['/create']).catch((err) => console.error(err))
+  }
+
+  importRecord() {
```
jahow, line 92:
```suggestion
  duplicateExternalRecord() {
```

libs/feature/editor/src/lib/components/import-record/import-record.component.html
```diff
@@ -0,0 +1,43 @@
...
+    <div
+      data-test="importMenuMainSection"
+      class="mt-2 border border-gray-100 p-2 flex items-center bg-white shadow-2xl rounded-2xl"
+    >
+      <ul class="flex flex-col gap-2 w-full">
+        <li *ngFor="let menuItem of importMenuItems">
+          <gn-ui-button
+            [attr.data-test]="menuItem.dataTest"
+            type="light"
+            extraClass="flex flex-row items-center gap-2 w-full justify-start test"
```
jahow, line 12:
```suggestion
            extraClass="flex flex-row items-center gap-2 w-full justify-start"
```

libs/feature/editor/src/lib/components/import-record/import-record.component.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+        { provide: Router, useValue: mockRouter },
+        { provide: TranslateService, useClass: MockTranslateService },
+        { provide: NotificationsService, useValue: mockNotificationsService },
+        {
+          provide: RecordsRepositoryInterface,
+          useValue: mockRecordsRepository,
+        },
+        { provide: ChangeDetectorRef, useValue: mockChangeDetectorRef },
+
+        ChangeDetectorRef,
```
jahow, line 58: Could you maybe try using ng-mocks here? This might simplify the test file a bit

jahow (commented):
Very good work! It all works quite well. I tried with a few records and one of them failed for a CORS reason:

https://sextant.ifremer.fr/geonetwork/srv/api/records/dcb73aea-b8d7-406b-9110-10f007bc69b6/formatters/xml

The error was: 
image

Do you think the error could be a little more helpful? Ideally mentioning CORS, or at least mentioning that the resource could not be read?

libs/api/repository/src/lib/gn4/gn4-repository.ts
rcaplier: But we are still in developpment we have no user right ?

So what should I do ? 

Remove this diff ? 

Add a comment saying that : 
> This prefix should actually never be changed (unless a retrocompatibility system is done) because otherwise all existing drafts in users sessions will get lost and never be cleaned up

????

jahow: 1. restore the previous prefix (which was also more explicit than "me-record-draft")
2. add a comment to emphasize the fact that changing this prefix is a breaking change

thanks!

jahow (approved):
Thank you! There's a test failure that I'm not familiar with, but it might just be a flaky test. The rest is OK now!

PR #988, opened 2024-09-10
[Editor] : Add missing e2e tests

cmoinier:
### Description

This PR reorganizes the e2e tests & adds missing ones, mostly in the editor form.

#### edit.cy

Note : this file now uses a duplicated record of "Stations d'épuration", to be able to change it at will without having to restore everything afterwards. The record is deleted at the end of the tests.
Note 2 : All of the tests are testing with saving + reload, thanks to a new custom command `publishAndReload`.

- [x] Title edition
- [x] Abstract edition
- [x] Adding & deleting graphic overviews
- [x] Resource updated edition
- [x] Update frequency & switcher
- [x] Add a temporal extent and delete them
- [x] Add a spatial extent keyword and delete it 
- [x] Attachment tests were split into 3 different ones for clarity
- [x] Adding a keyword & deleting one
- [x] License edition
- [ ] Page selector --> not directly tested but implicitly tested in other tests when switching pages. 

#### dashboard.cy

- [x] "My records" tab should show only records I own
- [x] Search input

#### record-actions.cy

- [x] Duplicate a record

### Architectural changes

For more clarity, the following test files were deleted and their tests moved to : 

- [x] `home` --> tests moved to `dashboard`
- [x] `create` / `undo` / `delete` --> tests moved to a new file `record-actions` because it felt like those files were too specific and used only for one test.


tools/e2e/commands.ts
```diff
@@ -152,6 +153,15 @@ Cypress.Commands.add('clearRecordDrafts', () => {
   cy.reload()
 })
 
+Cypress.Commands.add('publishAndReload', () => {
+  cy.wait(1200)
```
LHBruneton-C2C, line 157: What do we need this first wait for? We really try to wait as little as possible. Can it be replaced by checking for a specific state?

tools/e2e/commands.ts
```diff
@@ -152,6 +153,15 @@ Cypress.Commands.add('clearRecordDrafts', () => {
...
+  cy.get('md-editor-publish-button').click()
+  cy.wait(1200)
```
LHBruneton-C2C, line 159: Waiting for the return of the publish request can and should be explicit:
```
cy.intercept({
        method: 'PUT',
        pathname: '**/records',
      }).as('insertRecord')
      cy.get('md-editor-publish-button').click()
      cy.wait('@insertRecord')
```

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -1,12 +1,49 @@
...
     cy.login('admin', 'admin', false)
 
-    // Alpine convention record
-    cy.visit('/edit/8698bf0b-fceb-4f0f-989b-111e7c4af0a4')
-
     cy.clearRecordDrafts()
+    // Duplicate a record to manipulate it at will
+    cy.visit('/catalog/search')
+    cy.get('.table-header-cell').eq(1).click()
+    cy.get('.table-header-cell').eq(1).click()
```
LHBruneton-C2C, line 12: Visiting the URL with the search parameters feels more explicit than this double click, which might not work anymore if the order of columns changes.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -1,12 +1,49 @@
...
+    cy.get('[data-cy="table-row"]')
+      .first()
+      .find('[data-test="record-menu-button"]')
+      .click()
+    cy.get('[data-test="record-menu-duplicate-button"]').click()
+    cy.wait(1200)
```
LHBruneton-C2C, line 18: As duplication uses the draft system, can't we wait for the local storage to contain the wanted record?

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -1,12 +1,49 @@
...
+    cy.get('md-editor-publish-button').click()
+
+    cy.visit('/catalog/search')
+    cy.get('.table-header-cell').eq(1).click()
+    cy.get('.table-header-cell').eq(1).click()
```
LHBruneton-C2C, line 23: Same remark as before.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -1,12 +1,49 @@
...
+
+    cy.get('gn-ui-record-form')
+      .children('div')
+      .first()
+      .children('div')
+      .eq(1)
+      .find('gn-ui-form-field')
+      .eq(0)
+      .find('gn-ui-form-field-simple')
+      .find('input')
```
LHBruneton-C2C, line 34: 9 operators to find the correct element... Couldn't we had some data-cy attributes as shortcut selector?

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -18,191 +55,355 @@ describe('editor form', () => {
...
-    cy.get('gn-ui-record-form gn-ui-form-field').should('have.length.gt', 0)
-    cy.get('@abstractField')
-      .invoke('val')
-      .should('contain', 'Perimeter der Alpenkonvention in der Schweiz.')
-    cy.get('@saveStatus').should('eq', 'record_up_to_date')
-    cy.screenshot({ capture: 'fullPage' })
+  after(() => {
+    cy.visit('/catalog/search')
+    cy.get('.table-header-cell').eq(1).click()
+    cy.get('.table-header-cell').eq(1).click()
```
LHBruneton-C2C, line 72: Same remark as before.

LHBruneton-C2C (commented):
Really nice rework. Love the new publish and reload command!
It feels like we could go a step further, and properly wait for the draft to be saved in the local storage.

tools/e2e/commands.ts
cmoinier: It waits for the form to "realize" that a change has been made, quite similar to waiting for the draft to be saved in the local storage. Without that, the save command doesn't work.

tools/e2e/commands.ts
cmoinier: On the other hand, yes we can try to get rid of the second `wait`.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
cmoinier: Tried that in the first place, but I think the routing doesn't work properly for that because the columns were ordered in a random way when I did that.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
cmoinier: Well we would still have to wait for the draft saving to kick, right? And that would add many extra commands (which take time to execute) so I'm not sure it would be worth it...

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
cmoinier: True, I kind of forgot about this one

jahow:
This is green at last! Let's call it a day. Thanks @cmoinier and @LHBruneton-C2C

PR #996, opened 2024-09-20
Element ref wasn't working on ng-for wizard.component.html with multiple fields

f-necas:
### Description

This PR introduces a fix where fields weren't all verified because the ElementRef was checked only on the first element of the list.

On second step, we could hit "Next" button even if abstract wasn't filled.


jahow (approved):
Looking good, thanks! Could you please address the failing CI before merging?

PR #998, opened 2024-09-21
miscellaneous fixes to the French translation

fvanderbiest:
This PR replaces "Fiches de métadonnée" with "Jeux de données" (French term for "dataset", which is commonly used in the OpenData community, rather than "metadata sheet").

It also fixes several missing accentuated chars.

If accepted, all translation files should be updated accordingly. 
I can take care of it. Just tell me if I should go ahead or not.


translations/en.json
```diff
@@ -218,8 +218,8 @@
   "editor.record.form.field.onlineResource.fileSize": "{sizeMB}MB",
   "editor.record.form.field.onlineResource.modify": "Modify",
   "editor.record.form.field.overviews": "Overviews",
-  "editor.record.form.field.recordUpdated": "Record Updated",
-  "editor.record.form.field.resourceUpdated": "Resource Updated",
+  "editor.record.form.field.recordUpdated": "Dataset updated",
+  "editor.record.form.field.resourceUpdated": "Resource updated",
```
LHBruneton-C2C, line 222: I honestly don't get this diff. In technical terms, the record is the metadata and the resource is the data, but here you suggest the data for the record.

LHBruneton-C2C (commented):
@fvanderbiest I must say I am a bit confused about some of the changes you suggest. I agree that users are mostly looking at the metadata and seeing the data. But it sometimes feels misleading. Mostly when we talk about editing. The data will never be updated alongside the metadata, yet we ask a confirmation to the user if he wants to update the dataset. Same for the external import. No data is created then, only metadata.

translations/en.json
fvanderbiest: Indeed. 
The notion of dataset encompasses data and their metadata.

See eg https://demo.ckan.org/dataset/warandpeace - the word "metadata" or "record" is nowhere to see.
See also https://stable.demo.geonode.org/catalogue/#/datasets?d=11161%3Bdataset%3Bvector 

We make the difference between data and metadata because we're fully immersed in the technical context.
But for most users, this is too complex to grasp. We need to lower the barrier to entry.

LHBruneton-C2C: OK. As discussed with @xxiiss , these metadata fields about the update date should not even be shown. We will remove those, and then your PR will be valid.

LHBruneton-C2C:
Closing this PR, as it's been rebased there => https://github.com/geonetwork/geonetwork-ui/pull/1080

PR #999, opened 2024-09-22
[Backport 2.0.x] Element ref wasn't working on ng-for wizard.component.html with multiple fields

jahow:
Backport of #996


jahow:
Not needed

PR #1002, opened 2024-09-23
ME: edit record distributions

LHBruneton-C2C:
### Description

This PR introduces the section on associated resources, allowing to add both download and service as distributions.

TODO: split and share components between annexes and associated resources. 

### Screenshots

image
image


libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -381,6 +381,7 @@ export function extractDatasetOnlineResources(
     mapArray(
       ([isService, isDownload, protocol, url, name, description, mimeType]) => {
         if (isService) {
+          //FIXME: allow writing identifier only for wms and wfs?
```
LHBruneton-C2C, line 384: @jahow This is the only point on which I'm not sure how to fix the converter. As it is currently, for service protocols other than WMS and WFS, the identifier is not lost, but it's transferred to the name property.

This is particularly disturbing when reloading a record after it's been first saved as draft. The attributes aren't kept as they were first inputted.

jahow:
Did a quick test and found this:
* When trying to upload a file as a distribution, this error appears in the console and nothing happens:
image

* When clicking on the main block, the upload file dialog doesn't open

libs/ui/inputs/src/lib/url-input/url-input.component.ts
```diff
@@ -12,16 +19,23 @@ import { Subject } from 'rxjs'
...
   rawChange = new Subject<string>()
   @Output() valueChange = this.rawChange.pipe(filter((v) => !!v))
 
   constructor(private cd: ChangeDetectorRef) {}
 
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.value) {
+      console.log('changes.value', changes.value)
+    }
+  }
```
jahow, line 37: to delete

fvanderbiest:
Just chiming in - "OGC API" title is misleading because there are so many OGC APIs  (OGC API Maps, Features, Tiles...)
Have a look at https://ogcapi.ogc.org/#standards

jahow:
> Just chiming in - "OGC API" title is misleading because there are so many OGC APIs (OGC API Maps, Features, Tiles...) Have a look at https://ogcapi.ogc.org/#standards

On the other hand it makes sense because an OGC API endpoint can provide many different information at the same place I think

PR #1004, opened 2024-09-25
[ME]: Clean up attachments on publishing

rcaplier:
This PR add an effect cleanRecordAttachments on saveRecordSuccess action that delete all ressources that are not used in the record.

**This work is sponsored by**.

### DIFF
--- a/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
+++ b/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
@@ -1,14 +1,12 @@
 import { Injectable } from '@angular/core'
-import { combineLatest, Observable, of, switchMap, throwError } from 'rxjs'
 import {
-  combineLatest,
-  forkJoin,
+  catchError,
+  filter,
+  map,
   mergeMap,
-  Observable,
-  of,
-  switchMap,
-} from 'rxjs'
-import { catchError, filter, map, shareReplay, tap } from 'rxjs/operators'
+  shareReplay,
+  tap,
+} from 'rxjs/operators'
 import {
   MeApiService,
   RecordsApiService,
@@ -39,6 +37,14 @@ import {
 } from '@geonetwork-ui/api/metadata-converter'
 import { KeywordType } from '@geonetwork-ui/common/domain/model/thesaurus'
 import { noDuplicateFileName } from '@geonetwork-ui/util/shared'
+import {
+  combineLatest,
+  forkJoin,
+  Observable,
+  of,
+  switchMap,
+  throwError,
+} from 'rxjs'
 
 const minApiVersion = '4.2.2'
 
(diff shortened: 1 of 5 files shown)


libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
```diff
@@ -288,6 +296,44 @@ export class Gn4PlatformService implements PlatformServiceInterface {
...
+    ]).pipe(
+      map(([associatedResources, recordResources]) => {
+        const resourceIdsToKeep = [
+          ...((associatedResources as any).onlines ?? [])
+            .map((o) => o.title)
+            .map((o) => o['']),
+          ...((associatedResources as any).thumbnails ?? [])
+            .map((o) => o.title)
+            .map((o) => o['']),
+        ]
```
LHBruneton-C2C, line 312: What does this do? This is the kind of code that could benefit from some comments.

libs/feature/editor/src/lib/+state/editor.effects.spec.ts
```diff
@@ -167,6 +172,16 @@ describe('EditorEffects', () => {
...
   })
 
+  describe('cleanRecordAttachments$', () => {
+    it('dispatch markRecordAsChanged', () => {
+      actions = hot('-a-|', {
+        a: EditorActions.saveRecordSuccess(),
+      })
+      const expected = hot('---|')
+      expect(effects.cleanRecordAttachments$).toBeObservable(expected)
+    })
```
LHBruneton-C2C, line 182: I don't get this test... Is it the proper `it` description?
Also, the way I read the marble test, no emit is expected on the observable. Is it really what we want, or am I missing something?

LHBruneton-C2C (approved):
Looking good and tested OK, thanks!
I just have a few questions, but for my own understanding.

libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
rcaplier: Well the object returned from the API is not quite like expected so I have to cast it as any and go get the things I need. I can't put a comment above to explain yes.

libs/feature/editor/src/lib/+state/editor.effects.spec.ts
rcaplier: Forgot to change the it after the copy pasta. The observable returns void so I don't know if I can marble that differently ?

LHBruneton-C2C: The problem here is that no event is different from an event with no content. Searching on SO, most people expect an undefined tick: https://stackoverflow.com/questions/64258427/how-do-i-marble-test-an-observable-of-void-values

PR #1006, opened 2024-09-26
Metadata Editor: redirect to login URL if accessing the editor as anonymous user

jahow:
### Description

This PR adds a common route guard for all routes in the ME app.

Also removed an unused component.

### Architectural changes

none


apps/metadata-editor/src/app/app.routes.ts
```diff
@@ -10,100 +9,110 @@ import { NewRecordResolver } from './new-record.resolver'
...
-        path: 'my-org',
-        component: MyOrgUsersComponent,
-        pathMatch: 'prefix',
+        path: 'edit/:uuid',
+        component: EditPageComponent,
+        resolve: { record: EditRecordResolver },
       },
     ],
   },
-  { path: 'sign-in', component: SignInPageComponent },
```
jahow, line 93: this route was removed as it was obviously not needed anyway

apps/metadata-editor/src/app/app.routes.ts
```diff
@@ -10,100 +9,110 @@ import { NewRecordResolver } from './new-record.resolver'
...
 
 export const appRoutes: Route[] = [
-  { path: '', redirectTo: 'catalog/search', pathMatch: 'prefix' },
   {
-    path: 'catalog',
-    component: DashboardPageComponent,
-    outlet: 'primary',
+    path: '',
+    canActivate: [AuthGuardService],
     children: [
```
jahow, line 18: Adding a root route (:thinking:) makes it easier to apply the same guard to all routes

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
```diff
@@ -314,4 +314,16 @@ describe('dashboard', () => {
...
   })
+
+  describe('when the user is not logged in', () => {
+    beforeEach(() => {
+      cy.visit('/catalog/search')
+    })
+    it('redirects to the login page', () => {
+      cy.url().should(
+        'include',
+        '/catalog.signin?redirect=http://localhost:4200/catalog/search'
```
LHBruneton-C2C, line 325: Out of curiosity, where does this `redirect` value come from? The referer? It's often needed to be redirected after successful login, but I don't see it written in our code.

LHBruneton-C2C (approved):
Thanks for this improvement.
There is still one corner case when the authentication might fail, when the user is already wrongly connected (XSFR cookie set on the wrong path from the legacy UI), and the redirection is not activated then.
Otherwise, all good !

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
jahow: it's computed here : https://github.com/geonetwork/geonetwork-ui/blob/main/libs/api/repository/src/lib/gn4/auth/auth.service.ts#L13

PR #1009, opened 2024-09-27
[ME]: Trying to upload the same file as a resource or attachment twice fails

rcaplier:
### Description

This PR fixes the overview and online resources file upload by checking if the filename already exists and renaming it if necessary.


libs/feature/editor/src/lib/components/record-form/form-field/form-field-overviews/form-field-overviews.component.ts
```diff
@@ -53,7 +54,26 @@ export class FormFieldOverviewsComponent {
...
+            const newFileName = `${baseName}_${Date.now()}.${extension}`
+
+            fileToUpload = new File([file], newFileName, { type: file.type })
+          }
+          return this.platformService.attachFileToRecord(
+            this.metadataUuid,
+            fileToUpload
+          )
+        })
+      )
```
jahow, line 76: This should be abstracted away in `platformService.attachFileToRecord` IMO, to avoid repeating it and having too much responsibility in the components.

That means this method should probably return the full path of the file on upload success.

jahow: Also I would put the filename manipulation logic in a separate util, so has to be able to test it a bit more. For example this code will probably fail on a filename without dots, right?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-online-link-resources/form-field-online-link-resources.component.ts
```diff
@@ -94,7 +114,7 @@ export class FormFieldOnlineLinkResourcesComponent {
             this.valueChange.emit([...this.allResources, newResource])
           }
         },
-        error: (error: Error) => this.handleError(error.message),
+        error: (error: any) => this.handleError(error.error.message),
```
jahow, line 117: This is strange. We should handle an Error here, it seems that we get a nested error in an error instead. It might be worthwhile to figure out where that comes from.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-overviews/form-field-overviews.component.ts
```diff
@@ -68,7 +88,7 @@ export class FormFieldOverviewsComponent {
             })
           }
         },
-        error: this.errorHandle,
+        error: (error: any) => this.handleError(error.error.message),
```
jahow, line 91: Same remark as above: handleError should take an Error as input.

Also please keep in mind that every `any` in the code is technical debt, and we should strive to avoid & get rid of them as much as possible.

jahow (changes requested):
Well done, thank you! I think the logic is perfect. In terms of implementation, I think we can achieve less repetition and better test coverage, see my remarks! Let me know if anything is unclear.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-online-link-resources/form-field-online-link-resources.component.ts
rcaplier: It comes from the backend.

So either we change the error object sent by the backend, either we adapt it into the way and form we want it to be on the fly.

Just tell me how you want it to be solved.

jahow: I would like `handleError` to take in an `Error` object, and the observable to emit an `Error` object in case of exception :)

libs/feature/editor/src/lib/components/record-form/form-field/form-field-overviews/form-field-overviews.component.ts
rcaplier: Yes, but what do you want me to do for it ? Make sure that the code still work if the uploaded file has no extension ?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-online-link-resources/form-field-online-link-resources.component.ts
rcaplier: The Error object come from the backend, so how should we proceed ? 

Do we know if the backend always send its errors nested in another Error object ? If that's the case, do we intercept all error object from the backend and reshape them into a proper Error object before treating them ?  :)

jahow (approved):
thanks!

PR #1011, opened 2024-10-04
Metadata Editor / support multilingual values in certain fields

jahow:
### Description

This PR introduces support for reading and writing multilingual values in various fields of a catalog record:

These fields are currently:

- title and abstract of a record
- name of an organization
- name and description on an online resource
- description of a geographical extent
- text of a constraint or license
- label of a keyword

The converters (both ISO and DCAT-AP) are now able to read and write translations alongside a field value.

Note: in the model, **a record has a mandatory default language**. Multilingual fields have **a default value** (in the default language). They can also have **values in other languages**, these values are called **translations**.

A record also have a field called `otherLanguages` which lists all languages in which translations have been found.

The Metadata Editor currently only modifies the default values of multilingual fields; this means that translations should not be impacted at all by the app for now.

This PR has the important effect that multilingual fields will _not_ lose their translations if modified in the editor; the translations can not be edited but at least they are preserved.

### Architectural changes

none


apps/metadata-converter/src/app/components/record-form/record-form.component.ts
```diff
@@ -100,6 +98,9 @@ export class RecordFormComponent implements AfterViewInit {
       keywords: hasPrevious ? this.record.keywords : [],
       topics: hasPrevious ? this.record.topics : [],
       onlineResources: [],
+      otherLanguages: [],
+      defaultLanguage: 'en',
+      overviews: [],
```
jahow, line 103: I have adjusted these fixtures to make them correct; there were typing errors that were hidden because of the cast to CatalogRecord below

libs/api/metadata-converter/src/lib/dcat-ap/read-parts.ts
```diff
@@ -1,48 +1,78 @@
...
-  return findNodeLocalized(dataStore, dataset, DCTERMS('title'), null, 'en')
-    ?.value
+export function readTitle(
+  dataStore: Store,
+  recordNode: NamedNode,
+  translations: RecordTranslations,
+  defaultLanguage: LanguageCode
+): string {
+  const dataset = getDatasetNode(dataStore, recordNode)
+  const [title] = readLocalizedValue<RecordTranslations>(
```
jahow, line 47: `readLocalizedValue` is introduced in this PR and does two things:
* add any found translations to the `translations` object given to it
* return both the default value and the translations array as a tuple: `[value, translations]`

jahow:
I took the opportunity to better document the datahub & editor apps!

apps/datahub/README.md
```diff
@@ -1,147 +1,7 @@
...
-
-Notice how the `GN4_API_URL` and `PROXY_PATH` variables are used to override any values present in the app configuration file.
-**This override will happen everytime the docker container is started.**
-
-The application will be available on http://localhost:8080/datahub/.
-
-#### Specifying a custom configuration file
-
-If the `GN4_API_URL` and `PROXY_PATH` environment variables are not enough and you want to specify a full configuration file,
-you can do so like this:
```
tkohr, line 34: Maybe it's worth mentioning that this is currently still only available/used in the datahub (if I remember right)?

libs/api/metadata-converter/src/lib/xml-utils.ts
```diff
@@ -314,7 +314,19 @@ export function createElement(
   return () => new XmlElement(name, {}, [])
 }
 
-export function addAttribute(
+export function createNestedElement(
```
tkohr, line 317: Maybe some added tests for the two new functions here would make sense?

tkohr (approved):
Thanks, @jahow, for the huge work and for reworking the doc! Not sure if I got everything, but as far as I can see PR LGTM.

apps/datahub/README.md
jahow: it's actually available for any app using the default Dockerfile :)

libs/api/metadata-converter/src/lib/xml-utils.ts
jahow: I haven't tested them because I considered that they were sufficiently covered by all the other tests in this module. But I added some comments to clarify what they do and hopefully simplified a bit the utils, because it might have been hard to understand from outside.

jahow:
@tkohr thank you for the review!!

PR #1016, opened 2024-10-15
Datahub: fix untranslated text and wrong search filters in organizations page

jahow:
### Description

The search filters for the org's datasets were not correctly updated when an organization was opened a second time.


tkohr (approved):
Thanks for fixing/reworking this @jahow ! Code LGTM and organization page works fine. 

I'm just wondering if the case that an organization has no dataset (which shouldn't occur that much in reality, I guess) also works. When creating the ticket, I had the impression there was an issue with the `orgHasNoDataset` else, but maybe it was related to the change detection you just fixed.

jahow:
Hmm, actually that should never happen since organizations are taken from the search results. Maybe if we use groups instead, I'll give it a try.

jahow:
> Thanks for fixing/reworking this @jahow ! Code LGTM and organization page works fine.
> 
> I'm just wondering if the case that an organization has no dataset (which shouldn't occur that much in reality, I guess) also works. When creating the ticket, I had the impression there was an issue with the `orgHasNoDataset` else, but maybe it was related to the change detection you just fixed.

Confirmed, an organization should never have zero datasets (otherwise it won't show in the datahub). I simplified the code as a result :)

PR #1022, opened 2024-10-18
[DH/Editor] : Sort by changeDate by default

cmoinier:
### Description

This PR ensures that both editor and datahub search/landing pages sort their results by changeDate and not by relevancy.
It also makes sure to revert back to a sorting by relevancy when a full text search is done.


apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -47,32 +47,40 @@ describe('datasets', () => {
...
-        .should('equal', 'desc,_score')
+        .should('equal', 'desc,createDate')
     })
   })
 
   describe('display of dataset previews', () => {
-    it('should display a logo for first and a placeholder for second result', () => {
-      cy.get('@sortBy').selectDropdownOption('desc,createDate') // this makes the order reliable
-      cy.get('@sampleResult')
+    it('should display a placeholder for first result and a logo for second result', () => {
```
cmoinier, line 55: This test has been changed because "sampleResult" dataset (Metadata for testing purposes) had an erroneous image and was not displaying a logo anymore. Besides, the new sorting is also reliable because the favorites are cleared before the e2e are run.

jahow (changes requested):
Thanks! I think there might be a better way to do this. This PR will immediately  put a `_sort` parameter in the URL if none is present. The previous behavior (when using the router) was to fall back to relevancy because of this code: 
https://github.com/geonetwork/geonetwork-ui/blob/main/libs/feature/router/src/lib/default/state/router.effects.ts#L64

I think we can change that to CHANGE_DATE, so we can have the datahub or editor load without a `_sort` parameter showing up.

Also the sort-by component should offer CHANGE_DATE alongside CREATE_DATE I guess:
https://github.com/geonetwork/geonetwork-ui/blob/main/libs/feature/search/src/lib/sort-by/sort-by.component.ts

(the create date criteria can be kept under the name "Newest" and change date can be "Last modification")

libs/feature/router/src/lib/default/state/router.effects.ts
```diff
@@ -61,7 +61,7 @@ export class RouterEffects {
         let sortBy =
           ROUTE_PARAMS.SORT in newParams
             ? sortByFromString(newParams[ROUTE_PARAMS.SORT])
-            : SortByEnum.RELEVANCY
+            : SortByEnum.CHANGE_DATE
```
jahow, line 64: the sme change should be applied to line 73 below

apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -47,7 +47,7 @@ describe('datasets', () => {
       cy.get('@sortBy')
         .getActiveDropdownOption()
         .invoke('attr', 'data-cy-value')
-        .should('equal', 'desc,_score')
+        .should('equal', 'desc,changeDate')
```
jahow, line 50: the title of the `it` block should be changed accordingly

jahow (approved):
Thank you! two minor comments to address and it's good to merge :)

PR #1023, opened 2024-10-22
[Editor]: Fix cleanup attachments function 

cmoinier:
### Description

This PR fixes a wrong behavior in the editor's form, where the cleanup attachment function deletes the newly added attachments (eg : if an image is added, then the record saved, and then reloaded, the new image would be deleted).

This was because of a mapping that was not working : the names of the saved attachments were supposed to match the names of the new attachments, and the mapping was not taking into account that the name object had language keys in it, thus returning "undefined" for all attachment's names.

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
@@ -186,6 +186,13 @@ describe('editor form', () => {
             'src/fixtures/sample.png'
           )
           cy.editor_publishAndReload()
+          cy.intercept({
+            method: 'GET',
+            url: '**/attachments/sample.png',
+          }).as('importUrlRequest')
+          cy.get('@importUrlRequest')
+            .its('response.statusCode')
+            .should('eq', 200)
           cy.get('@saveStatus').should('eq', 'record_up_to_date')
           cy.get('gn-ui-image-input').find('img').should('have.length', 1)
         })
--- a/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.spec.ts
+++ b/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.spec.ts
@@ -791,6 +791,37 @@ describe('Gn4PlatformService', () => {
         record.uniqueIdentifier
       )
     })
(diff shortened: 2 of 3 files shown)


jahow (commented):
Thank you! I wonder if the original mapping wasn't done against another version of GN? It's possible that the API response differs (e.g. with localized names) between 4.2.5 and 4.4.0 for instance.

Maybe having a test of this in an e2e test would make sure this works against all version; essentially we just need to test that once an image has been uploaded and the record saved the image still shows up correctly (i.e. it has not been cleaned up).

cmoinier:
> Thank you! I wonder if the original mapping wasn't done against another version of GN? It's possible that the API response differs (e.g. with localized names) between 4.2.5 and 4.4.0 for instance.
> 
> Maybe having a test of this in an e2e test would make sure this works against all version; essentially we just need to test that once an image has been uploaded and the record saved the image still shows up correctly (i.e. it has not been cleaned up).

@jahow I think this e2e covers it already, and it passed :slightly_smiling_face: [[](url)](https://github.com/geonetwork/geonetwork-ui/blob/8665bf59f124494bd81f99a3249737bf70393e30/apps/metadata-editor-e2e/src/e2e/edit.cy.ts#L183)

jahow:
> > Thank you! I wonder if the original mapping wasn't done against another version of GN? It's possible that the API response differs (e.g. with localized names) between 4.2.5 and 4.4.0 for instance.
> > Maybe having a test of this in an e2e test would make sure this works against all version; essentially we just need to test that once an image has been uploaded and the record saved the image still shows up correctly (i.e. it has not been cleaned up).
> 
> @jahow I think this e2e covers it already, and it passed with all versions on the CI 🙂 [](
> 
> https://github.com/geonetwork/geonetwork-ui/blob/8665bf59f124494bd81f99a3249737bf70393e30/apps/metadata-editor-e2e/src/e2e/edit.cy.ts#L183
> 
> )

OK so there's already an e2e test, but the test would not fail if the underlying image file ends up being deleted. Could you try and expand a bit the test to make sure that the url of the image does not end up in a 404? thanks!

jahow (commented):
Thanks!

PR #1032, opened 2024-10-28
Metadata Editor: constraints fields

Angi-Kinas:
### Description

This PR introduces the possibility to add legal/security/other constraints to a record in the Metadata-Editor.

TODO:
- [x] toggling one of the conditions toggle and then untoggling the same one, will create a "glitch" (expected behaviour: see all constraint sections that contain data, actual: all constraint sections are hidden until you click somewhere outside)
- [x] the saving of the url is only being triggered if you press Enter (should be triggered on each key stroke) + saving the url is only possible if it is a valid url
- [x] testing

### Architectural changes
Changes have been made in the way we use the `EditorField`in order to be able to have two components, one that actually represents the data that will be saved and the other one to show and hide this component.
If the `EditorField` has no `model` but instead a `componentName` this component will be displayed. This is the component that toggles the visibility of the other component.

### Screenshots

image


jahow:
@Angi-Kinas I rebased the branch and added some commits to handle the missing cases. The constraints-shortcuts was reworked quite a bit, hopefully it still makes sense to you. I also introduced separate utilities to check whether a constraint matches the "no applicable constraint" or "no known constraint" type, but this is not yet finished.

What's also missing is proper styling of the buttons and maybe improve the layout a bit.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-constraints-shortcuts/form-field-constraints-shortcuts.component.ts
```diff
@@ -137,60 +145,42 @@ export class FormFieldConstraintsShortcutsComponent implements OnInit {
...
+        toggleName === 'noApplicableConstraint'
+          ? NOT_APPLICABLE_CONSTRAINT
+          : NOT_KNOWN_CONSTRAINT
+      this.editorFacade.updateRecordField('legalConstraints', [
+        presetConstraint,
+      ])
+      this.hideAllConstraintSections()
+    } else {
+      // if the toggle is turned off, remove the constraint
+      this.editorFacade.updateRecordField('legalConstraints', []) //remove all legal constraints
```
Angi-Kinas, line 163: Let's assume we edit a record where there are some legal constraints already present.
Then we toggle either of the toggle possibilites and untoggle the same one again. After that all legal constraints that were previously in the record will be gone, right? Is that ok for the user?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-constraints/form-field-constraints.component.ts
```diff
@@ -45,58 +43,25 @@ marker('editor.record.form.constraint.header.otherConstraints')
...
-    console.log('updatedConstraints', updatedConstraints)
-    this.valueChange.emit(updatedConstraints)
-  }
-
-  handleConstraintTextChange(text: string, index: number) {
-    const updatedConstraints = [...this.value]
-    updatedConstraints[index].text = text
-    this.valueChange.emit(updatedConstraints)
+  handleConstraintChange(constraint: Constraint, index: number) {
+    this.value = [...this.value]
```
Angi-Kinas, line 59: Shouldn't we avoid assigning something to `this.value` as it is an `@Input`? Maybe instead we could do it in a similar way as in `handleConstraintsOrderChange`?

libs/feature/editor/src/lib/components/record-form/form-field/form-field-constraints-shortcuts/constraints.utils.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+      expect(
+        matchesNoKnownConstraint({
+          text: ' bonjour monde ',
+          translations: {
+            text: { en: 'Conditions unknown', de: 'hallo welt' },
+          },
+        })
+      ).toBe(true)
+    })
+    it('returns false otherwide', () => {
```
Angi-Kinas, line 91: same typo

libs/feature/editor/src/lib/components/record-form/form-field/form-field-constraints/form-field-constraints.component.ts
jahow: good point, i'll try changing it and see if it works better :)

libs/feature/editor/src/lib/components/record-form/form-field/form-field-constraints-shortcuts/form-field-constraints-shortcuts.component.ts
jahow: you're right, that's probably too radical; we should only remove constraints that match the "no applicable constraints" or "no known constraints" criteria

tkohr (approved):
Thanks, @Angi-Kinas and @jahow ! I only tested the PR as you already paired on it. I had the feeling I saw an error on an angular hook in the console once, but I cannot reproduce it. Was maybe just due to switching branches. Everything works fine besides that. You just forgot to extract and translate the hint translation key `editor.record.form.constraint.markdown.placeholder` I think.

jahow:
Thank you for the review!

PR #1036, opened 2024-11-12
Editor: make use of minCharCount and clearOnSelection of autocomplete component

Angi-Kinas:
### Description

This PR uses the functionality of the autocomplete component to achieve the following behaviour for the keyword edition:

- When the user chooses a result, the typed characters should be cleaned
- When the user clears the input, the dropdown is closed


cmoinier (changes requested):
Thanks for the work :)

However, while the changes work for the second point (closing the dropdown on clear), it doesn't work for the first one. Here I typed "asia" & selected the keyword, but the typed text wasn't cleared.

image

Angi-Kinas:
@cmoinier thanks for the review and testing!
You are right, the clearing of the input field did not work after selection. I fixed this now.
While testing I encountered a different bug though. 
When selecting a keyword and clicking inside the input field again, the previous selected keyword is being saved as search query and therefore the dropdown suggestions only reflect the previous selected keyword.
This happens just for the keywords (not for the spatial extent keywords).
Until now I could not figure out why this is happening.

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -585,7 +585,9 @@ describe('editor form', () => {
         })
         it('should add a keyword', () => {
```
tkohr, line 586: Could it make sense to add a test, which checks that the input is cleared on selection?

libs/feature/editor/src/lib/components/generic-keywords/generic-keywords.component.html
```diff
@@ -4,9 +4,10 @@
     [displayWithFn]="displayWithFn"
     [action]="autoCompleteAction"
     (itemSelected)="handleItemSelection($event)"
-    [preventCompleteOnSelection]="true"
-    [minCharacterCount]="0"
+    [preventCompleteOnSelection]="false"
```
tkohr, line 7: I see `false` is the default value, but it might indeed be more explicit to set it here.

tkohr (approved):
Thanks for the fix @Angi-Kinas ! Works like a charm. Happy to see it needed less logic than expected :-)

tkohr:
Thanks for the added test! LGTM

PR #1037, opened 2024-11-13
Editor: Add a search filters summary component

tkohr:
### Description

This PR introduces components to display currently selected search filter values in a zone below the search filters.

To-dos:

- [x] format date
- [x] format user
- [x] finish UI
- [x] tests

### Screenshots

image


apps/metadata-editor/src/app/dashboard/search-filters-summary-item/search-filters-summary-item.component.html
```diff
@@ -0,0 +1,22 @@
+<div *ngIf="(fieldValues$ | async)?.length > 0">
+  <div *ngIf="fieldType === 'dateRange'; else valuesBadge">
+    <span translate>search.filters.summaryLabel.{{ fieldName }}</span>
+    <gn-ui-badge
+      #valuesBadge
+      *ngFor="let fieldValue of dateRange$ | async"
+      [removable]="true"
+      (badgeRemoveClicked)="removeFilterValue(fieldValue)"
+      >{{ fieldValue['start'] }} - >{{ fieldValue['end'] }}</gn-ui-badge
```
jahow, line 9: This can probably be encapsulated in a "getReadableValue" method to avoid having a ngIf in the template

apps/metadata-editor/src/app/dashboard/search-filters-summary-item/search-filters-summary-item.component.ts
```diff
@@ -0,0 +1,84 @@
...
+      this.fieldValues$
+    )
+    const updatedFieldValues = currentFieldValues.filter(
+      (value: string | DateRange) => value !== fieldValue
+    )
+    this.fieldsService
+      .buildFiltersFromFieldValues({
+        [this.fieldName]: updatedFieldValues as FieldValue[],
+      })
+      .subscribe((filters) => this.searchService.updateFilters(filters))
```
jahow, line 82: how about directly modifying the current filters instead of going from filters to values and then values to filters? simply deleting the corresponding key in the filters should be enough

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.html
```diff
@@ -0,0 +1,12 @@
+<div
+  class="flex flex-row py-3 px-4 gap-4 overflow-hidden grow border-[1px] border-gray-200"
+  *ngIf="searchFilterActive$ | async"
+>
+  <md-editor-search-filters-summary-item
+    *ngFor="let field of searchFields; let i = index"
+    [fieldName]="field"
+  ></md-editor-search-filters-summary-item>
```
jahow, line 8: I would suggest avoiding creating a separate "summary-item" component, and do everything in the search-filters-summary component. Do you think that would be doable?

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.ts
```diff
@@ -0,0 +1,39 @@
+import { Component, Input } from '@angular/core'
+import { CommonModule } from '@angular/common'
+import { SearchFacade, SearchService } from '@geonetwork-ui/feature/search'
+import { map } from 'rxjs'
+import { SearchFiltersSummaryItemComponent } from '../search-filters-summary-item/search-filters-summary-item.component'
+import { TranslateModule } from '@ngx-translate/core'
+
+@Component({
+  selector: 'md-editor-search-filters-summary',
```
jahow, line 9: this can maybe go in the feature-search lib ?

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.ts
```diff
@@ -0,0 +1,39 @@
...
+  templateUrl: './search-filters-summary.component.html',
+  styleUrls: ['./search-filters-summary.component.css'],
+  standalone: true,
+})
+export class SearchFiltersSummaryComponent {
+  @Input() searchFields: string[] = []
+
+  searchFilterActive$ = this.searchFacade.searchFilters$.pipe(
+    map((filters) => this.hasNonEmptyValues(filters))
+  )
```
jahow, line 20: I'm wondering if instead of taking a list of filters as input we can simply listen for the search filters and show everyone of them. This means that if another filter (which does not appear in the advanced filters above) is active then it will also show up. What do you think?

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -39,6 +39,9 @@ marker('search.filters.publisherOrg')
 marker('search.filters.user')
 marker('search.filters.changeDate')
 
+marker('search.filters.summaryLabel.user')
+marker('search.filters.summaryLabel.changeDate')
```
jahow, line 43: how about marking labels for all fields? It might make more sense taken from outside I think. Another approach would be to use the field normal label if a summaryLabel is not available.

jahow (commented):
Thanks! I did put some comments on what's already there, I'm not 100% sure what is the best way, let me know if you have better ideas!

tkohr:
Thanks for the suggestions @jahow, I'll have a look how to refactor things this way.

apps/metadata-editor/src/app/dashboard/search-filters-summary-item/search-filters-summary-item.component.html
tkohr: Good point, I've added f4d10b643e1aa1ea38f536670377926b949bbd18 for this.

apps/metadata-editor/src/app/dashboard/search-filters-summary-item/search-filters-summary-item.component.ts
tkohr: I must be thinking too complicated, but what would this look like? An additional method in the service or facade to delete one filter by its key? Also, date ranges don't use their value as a key in the object.

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.html
tkohr: Hum, I've started with one component in the beginning, but afterwards I felt it became more readable, especially for understanding which field is targeted with a separate component.

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.ts
tkohr: Also, good point! I've moved it. I would've liked to separate logic and UI more to make it more reusable, but could not get it done. Maybe something for the future.

apps/metadata-editor/src/app/dashboard/search-filters-summary/search-filters-summary.component.ts
tkohr: I gave it a try, but had some difficulties with the fact that the field names do not always correspond to the ES field names in the filters. I'm also wondering if it is not more intuitive from a user perspective to only display the summary badges for the filters visible in the filter component above. For example, on the `myRecords` page the filter `{ owner: user.id }` should always be present.

libs/feature/search/src/lib/utils/service/fields.service.ts
tkohr: Finally, I've added a fallback  in [6b01ea9](https://github.com/geonetwork/geonetwork-ui/pull/1037/commits/6b01ea922a2955741921e530aa92146664274d09) to use the filters label, as we don't have the exact wording for more filter summary labels yet.

jahow (commented):
Thank you, this looks really good! Just a few final comments but it works well in the app and I think the implementation makes a lot of sense now.

apps/metadata-editor/src/app/dashboard/search-filters-summary-item/search-filters-summary-item.component.ts
jahow: Ok I looked into it and honestly I couldn't find a better way, so I think your approach makes sense. It ends up being verbose because we need to alternate between the "filters" model and the "values" model, but I guess this is something we cannot avoid...

libs/feature/router/src/lib/default/state/query-params.utils.ts
```diff
@@ -42,8 +44,8 @@ export function expandQueryParams(
       } else if (isDateUrl(value)) {
         const [start, end] = value.split('..')
         expanded[key] = {
-          ...(start && { start: new Date(start) }),
-          ...(end && { end: new Date(end) }),
+          ...(start && { start: new Date(`${start}T00:00:00`) }),
```
Angi-Kinas, line 47: I guess `T00:00:00` stands for the time... Is the time necessary here?

Angi-Kinas (commented):
Thank you @tkohr, this will improve the usability of the filters quite a lot!

The code looks good to me.
While testing I had some behaviors that we might need to have a look at:
- When we use `user=` as a query param, we get an empty badge
image
- When we do the same with `changeDate`, we see an empty search filters summary component with the "reset" button
image

And as mentioned to you before, the filters are only working as expected on "Metadata records" but not on "My Records". I think this is more important than the issues mentioned above. When this is fixed, you can merge from my side :rocket:

libs/feature/router/src/lib/default/state/query-params.utils.ts
tkohr: Yes, good point, the time is necessary here to prevent timezone issues. Actually, the URL does indeed not include the time as it is not needed, but as a `Date` in the app has a time it's set to midnight when it's read from the URL. Then, within the app, the date is handled in the local time of the browser and when it's written to the browser it's "cut" again. If we don't handle the time within the app, there are conflicts between the date chosen in the datepicker and the one encoded in the URL and sent to elasticsearch. 

You can test it by switching the timezone of your browser. I discovered the issue since cypress is not running in the same timezone in the CI as my machine.

tkohr:
> Thank you @tkohr, this will improve the usability of the filters quite a lot!
> 
> The code looks good to me. While testing I had some behaviors that we might need to have a look at:
> 

Thanks for the review and the testing @Angi-Kinas  !

Good catches! I've added two commits to fix the errors you encountered. The first one ([8302b76](https://github.com/geonetwork/geonetwork-ui/pull/1037/commits/8302b76d2e5cf92427942b96fd46fa9c13b44498)) allows to define filters that should be ignored by the summary component (like the active `owner` filter on the `myRecords` page, for example).

Both PRs are rebased again. Could you have a final look and approve the PRs if all is good for you?

Angi-Kinas:
@tkohr thank you for addressing my comments. The issue with the empty value for a filter seems to be fixed :+1: 

While testing the feature on the "My Records" tab, I encountered two other issues:
- when filtering on the last updated date, no query param is added to the url
- when resetting the filters, it also resets the `owner` filter of the "My Records" page

I think addressing these issues will take some more time and could be worked on in a separate ticket. It would take additional complexity out of this PR. What do you think?

tkohr:
Thanks for your additional tests and pointing out the resetting bug @Angi-Kinas !

> While testing the feature on the "My Records" tab, I encountered two other issues:
> 
>     * when filtering on the last updated date, no query param is added to the url

This is actually expected, as the `myRecords` search state is not a router search state. Only one router search state is possible per app, which is bound to the `allRecords` within the editor.
> 
>     * when resetting the filters, it also resets the `owner` filter of the "My Records" page
> 
Good catch, I've added [01d9e22](https://github.com/geonetwork/geonetwork-ui/pull/1037/commits/01d9e22933f0cb1822bfc6172b1f9ae780fef98c) to also ignore the filters from the `FILTER_SUMMARY_IGNORE_LIST` token when clearing filters.

Crossing fingers that everything works fine now :crossed_fingers:

Angi-Kinas (approved):
Thanks @tkohr for the adaption and the hard work on this feature! 
Now the filters work as expected :) :rocket:

PR #1039, opened 2024-11-14
[Editor]: Allow logging out

cmoinier:
### Description

This PR adds the logging out logic to the sidebar button in the editor.
When the user clicks on the logout button, the session ends and the user is redirected to the logging page.

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
@@ -344,11 +344,23 @@ describe('dashboard (authenticated)', () => {
   })
 })
 
-describe('when the user is not logged in', () => {
-  beforeEach(() => {
-    cy.visit('/catalog/search')
+describe('Logging in and out', () => {
+  describe('when the user is not logged in', () => {
+    beforeEach(() => {
+      cy.visit('/catalog/search')
+    })
+    it('redirects to the login page', () => {
+      cy.url().should('include', '/catalog.signin?redirect=')
+    })
   })
-  it('redirects to the login page', () => {
-    cy.url().should('include', '/catalog.signin?redirect=')
+  describe('Logging out', () => {
+    beforeEach(() => {
+      cy.login('admin', 'admin', false)
+      cy.visit('/catalog/search')
+    })
+    it('logs out the user', () => {
+      cy.get('md-editor-sidebar').find('gn-ui-button').eq(1).click()
+      cy.url().should('include', '/catalog.signin?redirect=')
+    })
   })
 })
--- a/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
@@ -359,8 +359,10 @@ describe('Logging in and out', () => {
       cy.visit('/catalog/search')
     })
     it('logs out the user', () => {
+      cy.get('gn-ui-avatar').should('be.visible')
       cy.get('md-editor-sidebar').find('gn-ui-button').eq(1).click()
       cy.url().should('include', '/catalog.signin?redirect=')
+      cy.get('gn-ui-avatar').should('not.exist')
     })
   })
 })
--- a/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
+++ b/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
@@ -31,8 +31,14 @@
         <gn-ui-button
           type="default"
           class="w-10 h-10 flex justify-center items-center hover:cursor-pointer"
+          (click)="logOut()"
         >
-          <img src="assets/system-shut.svg" alt="Log out" class="w-4 h-4" />
+          <img
+            src="assets/system-shut.svg"
+            alt="Log out"
+            [title]="'editor.sidebar.logout' | translate"
+            class="w-4 h-4 hover:invert-[.5]"
+          />
         </gn-ui-button>
       </div>
     </ng-container>
--- a/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
+++ b/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
@@ -35,7 +35,7 @@
         >
           <img
             src="assets/system-shut.svg"
-            alt="Log out"
+            [alt]="'editor.sidebar.logout' | translate"
             [title]="'editor.sidebar.logout' | translate"
             class="w-4 h-4 hover:invert-[.5]"
           />
--- a/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.spec.ts
+++ b/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.spec.ts
@@ -53,18 +53,14 @@ describe('SidebarComponent', () => {
         ok: true,
       } as Response)
 
-      delete window.location
-      Object.defineProperty(window, 'location', {
-        value: { href: 'http://app-url.com' },
-        writable: true,
-      })
+      const originalUrl = window.location.href
 
       await component.logOut()
 
       expect(window.fetch).toHaveBeenCalledWith(service.logoutUrl, {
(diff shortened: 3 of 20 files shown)


libs/api/repository/src/lib/gn4/auth/auth.service.ts
```diff
@@ -5,11 +5,15 @@ import { TranslateService } from '@ngx-translate/core'
 export const DEFAULT_GN4_LOGIN_URL = `/geonetwork/srv/\${lang3}/catalog.signin?redirect=\${current_url}`
 export const LOGIN_URL = new InjectionToken<string>('loginUrl')
 
+export const DEFAULT_GN4_LOGOUT_URL = `/geonetwork/signout`
+export const LOGOUT_URL = new InjectionToken<string>('loginUrl')
```
Angi-Kinas, line 9: It should be `logoutUrl` instead of `loginUrl`

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
```diff
@@ -31,8 +31,14 @@
         <gn-ui-button
           type="default"
           class="w-10 h-10 flex justify-center items-center hover:cursor-pointer"
+          (click)="logOut()"
         >
-          <img src="assets/system-shut.svg" alt="Log out" class="w-4 h-4" />
+          <img
+            src="assets/system-shut.svg"
+            alt="Log out"
```
Angi-Kinas, line 38: I'm wondering if the alt text should also be translated?

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
```diff
@@ -41,4 +45,20 @@ export class SidebarComponent implements OnInit {
...
     )
   }
+
+  logOut() {
+    fetch(this.authService.logoutUrl, {
+      method: 'GET',
+    })
+      .then((response) => {
+        if (response.ok) {
+          window.location.href = 'http://localhost:4200'
```
Angi-Kinas, line 55: In production there won't be a localhost url. I'd suggest to save the host before logging out and using it here instead. WDYT?

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.spec.ts
```diff
@@ -34,4 +46,25 @@ describe('SidebarComponent', () => {
...
+        value: { href: 'http://app-url.com' },
+        writable: true,
+      })
+
+      await component.logOut()
+
+      expect(window.fetch).toHaveBeenCalledWith(service.logoutUrl, {
+        method: 'GET',
+      })
+      expect(window.location.href).toBe('http://localhost:4200')
```
Angi-Kinas, line 67: Same here, we should use the actual host

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
```diff
@@ -344,11 +344,23 @@ describe('dashboard (authenticated)', () => {
...
-  it('redirects to the login page', () => {
-    cy.url().should('include', '/catalog.signin?redirect=')
+  describe('Logging out', () => {
+    beforeEach(() => {
+      cy.login('admin', 'admin', false)
+      cy.visit('/catalog/search')
+    })
+    it('logs out the user', () => {
+      cy.get('md-editor-sidebar').find('gn-ui-button').eq(1).click()
+      cy.url().should('include', '/catalog.signin?redirect=')
```
Angi-Kinas, line 363: Should the url not include something like: `/geonetwork/signout?redirectUrl=`?  Or is this already one step ahead?

Angi-Kinas (changes requested):
Thanks Camille for implementing a log out functionality, it will be important for the users! :tada: 

I found some little things which I think need to be changed/ reviewed again. I haven't manually tested the functionality yet.
Let me know if something is not clear.

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
cmoinier: Yes indeed, didn't think of that though it's a bit obvious :smile:

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.html
cmoinier: It was not translated so I didn't do it either, but sure, I don't see why not :slightly_smiling_face:

libs/api/repository/src/lib/gn4/auth/auth.service.ts
cmoinier: Thanks for catching this!

apps/metadata-editor-e2e/src/e2e/dashboard.cy.ts
cmoinier: And also I think it's better to have less re-routing, by using a request to signout, right?

cmoinier: I can add it as well in the same test :)

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.spec.ts
cmoinier: Well I think in the e2e it's always going to be localhost right?

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
```diff
@@ -41,4 +45,21 @@ export class SidebarComponent implements OnInit {
       (orgs, me) => orgs.filter((org) => org.name === me?.organisation)
     )
   }
+
+  logOut() {
+    const current_url = window.location.toString()
```
Angi-Kinas, line 50: Would it make sense to use `window.location.hostname` here?
I'm wondering how this works, since `window.location` returns an object, but I tested it and it returns the href.
I would rather use the hostname here, since you don't want to redirect to e.g. /edit/id12345.

Angi-Kinas: Or maybe even better to use the login url?
Since when I'm logged out, I want to see the login page again, right?

Angi-Kinas (commented):
Thanks @cmoinier the PR looks good!

I still have some doubts about setting the url (after logout) to the url that was present during log in. But maybe I did not understand it correctly.
Can you have another look?

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
cmoinier: Not sure we can call `LOGIN_URL`, I think it becomes `loginUrl` when exported from the service & therefore uses the `window.location`. I think using the hostname is easier too, will do that.

jahow:
@cmoinier GeoNetwork offers a logout service similar to the login one, which a redirect option. For instance this is what it can look like:

https://geonetwork-ui.apps.gs-ch-prod.camptocamp.com/geonetwork/signout?redirectUrl=https://geonetwork-ui.apps.gs-ch-prod.camptocamp.com/geonetwork/srv

cmoinier:
> @cmoinier GeoNetwork offers a logout service similar to the login one, which a redirect option. For instance this is what it can look like:
> 
> https://geonetwork-ui.apps.gs-ch-prod.camptocamp.com/geonetwork/signout?redirectUrl=https://geonetwork-ui.apps.gs-ch-prod.camptocamp.com/geonetwork/srv

I precisely tried something like that in the first place :slightly_frowning_face: But it kept bringing me back to the login page with an error saying the credentials are wrong...

jahow:
Oh right, actually it makes sense that it puts you again on the login form once you log out, because trying to access the editor while not logged in will redirect you to the login form... Your solution is the best I think!

Angi-Kinas (approved):
@cmoinier sorry for the back and forth. I was wrong with not keeping the previous URL.

Good job! :rocket:

PR #1041, opened 2024-11-15
[Editor]: Bug - Update frequency not kept

cmoinier:
### Description

This PR fixes the behavior of the updated frequency dropdown, which did not keep the value "2 times per day" upon save & refresh.

The bug only concerned the value "2 times per day", and was due to the fact that in the ISO reader, the mapper for `getUpdateFrequencyFromCustomPeriod` was always going to the case `days <= 7`, even for values that had 0 days, such as the value "2 times per day", which has the value 0 days and 12 hours. 

Also, in the `hours` case, the `udpatedTimes` value was calculated based on the days count, which is also wrong in this case and should divide the 24 hours with the hours count.


Angi-Kinas (approved):
Thanks @cmoinier for fixing this bug! :taco: 

I quickly tested it and it works. The code looks also good to me. :rocket:

jahow (commented):
well done!! :tada:

PR #1042, opened 2024-11-15
feat: adds opengraph to records

f-necas:
### Description

This PR introduces OpenGraph protocol to datahub's records. A nice implmentation would be to have a default image and base title for og:title.

But it will add more default.toml conf. I tried here to keep it simple.

🔗 ⚠️  Also add external viewer button to exports to be able to use it in other apps which use gn-ui

### Screenshots

image

**This work is sponsored by [MEL](https://lillemetropole.fr/)**.


apps/datahub/src/index.html
```diff
@@ -5,6 +5,9 @@
     <title>Datahub</title>
     <base href="/" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta property="og:title" content="Datahub" />
+    <meta property="og:type" content="website" />
+    <meta property="og:url" content="" />
```
jahow, line 10: Is this really required or will the angular code be enough to create those elements?

jahow (commented):
Thank you, very interesting feature! Could you maybe put all the meta-related features in a component in the feature/record lib? something like `gn-ui-record-meta` for example. Thanks!

libs/feature/record/src/lib/record-meta/record-meta.component.ts
```diff
@@ -0,0 +1,48 @@
...
+} from '@angular/core'
+import { Meta } from '@angular/platform-browser'
+import {
+  DatasetRecord,
+  ServiceRecord,
+} from '@geonetwork-ui/common/domain/model/record'
+
+@Component({
+  selector: 'gn-ui-record-meta',
+  template: '<>',
```
jahow, line 17: it would be great if you could make this component standalone, thanks!

jahow (approved):
thank you! I just added a comment to remind you to make the new component standalone, thanks :)

libs/feature/record/src/lib/record-meta/record-meta.component.ts
f-necas: Set as standalone and removed <> as it was adding this to html page :p

PR #1048, opened 2024-11-21
[Editor] : Access account settings

cmoinier:
### Description

This PR adds a link from the user avatar of the editor to the account settings.
It also removes the border of the avatar and adds a margin, as per the recettes file.

### DIFF
--- a/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
+++ b/apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
@@ -32,10 +32,7 @@ export class SidebarComponent implements OnInit {
   organisations$: Observable<Organization[]>
 
   get settingsUrl() {
-    return (
-      this.authService.loginUrl.split('/catalog')[0] +
-      '/admin.console#/organization/users?userOrGroup='
-    )
+    return this.authService.settingsUrl
   }
 
   constructor(
--- a/apps/metadata-editor/src/app/app.module.ts
+++ b/apps/metadata-editor/src/app/app.module.ts
@@ -27,6 +27,7 @@ import { EditorRouterService } from './router.service'
 import {
   LOGIN_URL,
   LOGOUT_URL,
+  SETTINGS_URL,
   provideGn4,
   provideRepositoryUrl,
 } from '@geonetwork-ui/api/repository'
@@ -74,6 +75,10 @@ import { FeatureEditorModule } from '@geonetwork-ui/feature/editor'
     {
       provide: LOGOUT_URL,
       useFactory: () => getGlobalConfig().LOGOUT_URL,
+    },
+    {
+      provide: SETTINGS_URL,
+      useFactory: () => getGlobalConfig().SETTINGS_URL,
     },
   ],
   bootstrap: [AppComponent],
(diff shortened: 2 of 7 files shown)


apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
```diff
@@ -28,10 +31,18 @@ export class SidebarComponent implements OnInit {
   public placeholder$ = this.avatarService.getPlaceholder()
   organisations$: Observable<Organization[]>
 
+  get settingsUrl() {
+    return (
+      this.authService.loginUrl.split('/catalog')[0] +
+      '/admin.console#/organization/users?userOrGroup='
```
Angi-Kinas, line 37: Would it make sense to create a constant (e.g. `SETTINGS_URL`) for this in the auth service like  `LOGIN_URL`?

I'm also wondering where exactly the url should point at. This looks very specific with the users. It is also not really clear in the ticket. I would suggest a more general path like `/geonetwork/srv/eng/admin.console#/settings`.
Another question that arises to me: Is it always `admin.console#` in the url even if the user is not an admin?

Angi-Kinas (commented):
Thanks @cmoinier. This will be really useful for all users :+1: 

As the ticket lacks some information I have a few questions.

apps/metadata-editor/src/app/dashboard/sidebar/sidebar.component.ts
cmoinier: The ticket said to either create a constant or forge from the login url, and I took the second option because I figured that it needed less manipulation that login/logout urls. But that's just personal opinion, it can be changed :slightly_smiling_face: 

For the other two questions, I don't know honestly. I tried to look into gn for a while and that's the URL that made more sense, so I used it. 

@LHBruneton-C2C I had more or less the same interrogations during the dev. I did what made most sense to me but I knew it would need some changes & some more infos for that.

Angi-Kinas: Ah yes you are right! You don't have to change it :-)

LHBruneton-C2C: Looking at how you need to split from the login URL, and not just complete it, I'd rather go for a specific SETTINGS_URL config key. But that's just an opinion here. This would also solve the second question, as the whole URL would be configurable.

cmoinier: Ok good :) Well I think as a user, I would be expecting to land on a page where I'm directly accessing my own parameters. Instead here, the user ends up on a page where they have to select their username to edit their account parameters. 
But I guess I was expecting something that doesn't exist yet :smile:

jahow: +1 for a SETTINGS_URL token, keep in mind that this url will be completely different on georchestra for instance 🙂

cmoinier: Okay, will do :slightly_smiling_face: 
It would be nice to have that decided in the ticket beforehand next time though! :wink:

jahow: 100% agreed, that was an oversight

Angi-Kinas (approved):
Thanks @cmoinier for the adaption. This makes the url much more configurable :+1:

PR #1051, opened 2024-12-02
[Editor]: Warn user if draft has been updated

cmoinier:
### Description

This PR introduces a warning when the user starts editing one of their drafts, if the record has been updated by another user since they created the draft.
Key features : 

- [x] Added detection for record change in backend when a draft is opened
- [x] Added the update of `recordUpdated` field when the draft is created, to be able to compare the draft's date with the record's date in the backend.
- [x] Added a banner at the top of the edit page when the record has been updated by someone else
- [x] Added an overlay warning message when the user tries to publish their version on top of someone else's, with cancel/proceed choices
- [x] E2E tests : to mock 2 users editing the same record in e2e tests, it's been decided to clear and re-populate the localStorage between 2 editions by the same user

### Screenshots

image

image


apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.html
```diff
@@ -11,12 +13,49 @@
...
+    <div
+      data-test="publish-warning"
+      class="w-96 p-5 flex flex-col gap-3 mt-2 border border-gray-100 bg-white shadow-2xl rounded-2xl mr-2"
+    >
+      <span
+        class="text-center"
+        translate
+        [translateParams]="{
+          date: publishWarning[0],
+          user: publishWarning[1]
```
tkohr, line 36: Not sure what has been discussed, just wondering if it could make sense to also display the given name and family name here?

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
```diff
@@ -58,12 +77,74 @@ export class PublishButtonComponent {
 
   record$ = this.facade.record$
 
+  @ViewChild(MatMenuTrigger) trigger: MatMenuTrigger
+
+  @ViewChild('actionMenuButton', { read: ElementRef })
+  actionMenuButton!: ElementRef
+  @ViewChild('template') template!: TemplateRef<any>
```
tkohr, line 84: Could this be typed?

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -151,6 +152,10 @@ export class EditPageComponent implements OnInit, OnDestroy {
       .subscribe((savedRecord) => {
         this.router.navigate(['edit', savedRecord.uniqueIdentifier])
       })
+
+    this.facade.record$.subscribe((record) => {
+      this.facade.hasRecordChangedSinceDraft(record)
+    })
```
tkohr, line 158: I think this could be wrapped into a `this.subscription.add()` to ensure unsubscribing.

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -365,6 +367,47 @@ export class Gn4Repository implements RecordsRepositoryInterface {
...
+        })
+      ),
+      this.getRecord(localRecord.uniqueIdentifier),
+    ]).pipe(
+      map(([draftRecordUpdated, recentRecord]) => {
+        if (recentRecord.recordUpdated > draftRecordUpdated) {
+          return [
+            this.formatDate(recentRecord.recordUpdated),
+            recentRecord.extras?.['ownerInfo'].toString().split('|')[0],
+          ]
```
tkohr, line 394: I'd rather return a typed object with a `date` and a `user` here than an array. That would make clearer what info is returned, IMO. Formatting might be done at component level, I think.

libs/feature/editor/src/lib/+state/editor.facade.ts
```diff
@@ -63,4 +66,8 @@ export class EditorFacade {
   setFieldVisibility(field: EditorFieldIdentification, visible: boolean) {
     this.store.dispatch(EditorActions.setFieldVisibility({ field, visible }))
   }
+
+  hasRecordChangedSinceDraft(record: CatalogRecord) {
+    this.store.dispatch(EditorActions.hasRecordChangedSinceDraft({ record }))
+  }
```
tkohr, line 72: I'm wondering if the naming could be clearer here between the observable above and the action, as the observable basically expresses the change resulting from the action. Would `checkHasRecordChanged` make sense, for example?

libs/feature/editor/src/lib/+state/editor.reducer.ts
```diff
@@ -24,6 +24,7 @@ export interface EditorState {
   changedSinceSave: boolean
   editorConfig: EditorConfig
   currentPage: number
+  hasRecordChanged: any[]
```
tkohr, line 27: Could we type this? Maybe as an object `{user: string, date: Date}` as proposed in my other comment?

libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -70,4 +71,8 @@ export class EditorService {
     this.recordsRepository.clearRecordDraft(record.uniqueIdentifier)
     return this.recordsRepository.openRecordForEdition(record.uniqueIdentifier)
   }
+
+  hasRecordChangedSinceDraft(localRecord: CatalogRecord): Observable<string[]> {
+    return this.recordsRepository.hasRecordChangedSinceDraft(localRecord)
+  }
```
tkohr, line 77: Not sure if this wrapper is really necessary, but as most methods are called via the service in the effects it might make sense to keep it. I let you judge.

tkohr (commented):
Thanks, @cmoinier ! Congrats for the clean PR on this essential feature ! I left a couple of comments with ideas in-line.

Almost everything works really well. I just discovered two egde/minor issues:

- When editing the same record (that has not been saved as a draft yet) at the same and submitting it on both sides, the one submitting last can still overwrite the changes of the one submitting first without being warned. 
- The popup does not close when accepting to overwrite changes.

Also, we should remember to clarify why we get a 400 error for non-admin users when submitting changes.

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.html
```diff
@@ -11,12 +13,49 @@
...
+  <ng-template #template>
+    <div
+      data-test="publish-warning"
+      class="w-96 p-5 flex flex-col gap-3 mt-2 border border-gray-100 bg-white shadow-2xl rounded-2xl mr-2"
+    >
+      <span
+        class="text-center"
+        translate
+        [translateParams]="{
+          date: formatDate(publishWarning.date),
```
tkohr, line 35: I think you could use angular's date pipe.

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.ts
```diff
@@ -41,6 +59,7 @@ export type RecordSaveStatus = 'saving' | 'upToDate' | 'hasChanges'
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class PublishButtonComponent {
```
tkohr, line 61: `...implements OnDestroy` ;-)

apps/metadata-editor/src/app/edit/edit-page.component.html
```diff
@@ -5,12 +5,26 @@
...
     </div>
     <div class="grow overflow-auto p-[32px]" #scrollContainer>
       <div class="container-lg mx-auto flex flex-col gap-[32px]">
+        <div *ngIf="hasRecordChanged$ | async as hasChanged">
+          <div
+            *ngIf="hasChanged.user && hasChanged.date"
+            class="p-4 border border-primary bg-primary-lightest rounded-lg text-center"
+            translate
+            [translateParams]="{
+              date: formatDate(hasChanged.date),
```
tkohr, line 20: Same as above.

tkohr (approved):
Thanks for the adaptions @cmoinier !

The issues mentioned above are fixed. There is a small glitch when aborting changes after the warning and reverting them, they are not applied in the open form. This may be out of scope for this PR.

warning

Clicking cancel and back arrow to revert changes =>

change

There is still an issue with the date formatting in the tests, potentially because the CI runs in a different timezone. I think you could rely on the angular pipe for formatting here and get rid of the test. I let you fix this and approve.

tkohr:
Just had a look at the e2e error log and I think the following errors are actually related to this PR, the issue on reverting changes I mentioned earlier:

- https://github.com/geonetwork/geonetwork-ui/actions/runs/12429155088/job/34703957757?pr=1051#step:8:276
- https://github.com/geonetwork/geonetwork-ui/actions/runs/12429155088/job/34703957757?pr=1051#step:8:285

apps/metadata-editor/src/app/edit/components/publish-button/publish-button.component.html
cmoinier: The issue with Angular's date pipe is that we cannot format it as "date at time"/ "jour à heure" (with the joining term) in the current langugage. Also, the test was not really useful since I was just using `toLocaleString`, which does not really need testing anyway.

PR #1061, opened 2024-12-11
Editor: Fix error when creating new record

tkohr:
### Description

The goal of this PR is to fix the error when creating a new record. The same exception sometimes occurs when editing fields of a draft. A first commit ensures errors displayed via the `notifications.service` still to be logged to the console.

The actual error is caused by the fact that the `EditRecordResolver` is called even when routing to `/create`. This happens due to the `navigate['edit']` calls within the component. Both navigations have been removed:
- 07dcd58b154bf47f0bf0d5e76f001e703123cbef does not seem essential, but maybe I've missed something. Creating a new record now stays on `/create`. => This prevents the exception when creating a new record.
- a97b566ab19a4a98b0ef3831aa8da821642a4a49 does not seem to be used anymore, since we've replaced the `uniqueIdentifier` with the `resourceIdentifier ` in the form. => This prevents the exception when editing fields of a draft.

Note to reproduce: The issue occurs on the first access to `/create` after app reload. It starts with the second access to `/create` or accessing `/create` after `/edit`. 

### Screenshots

image

---

### DIFF
--- a/apps/metadata-editor/src/app/edit/edit-page.component.spec.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.spec.ts
@@ -121,24 +121,32 @@ describe('EditPageComponent', () => {
     describe('publish version error', () => {
       it('shows notification', () => {
         ;(facade.saveError$ as any).next(new PublicationVersionError('1.0.0'))
-        expect(notificationsService.showNotification).toHaveBeenCalledWith({
-          type: 'error',
-          title: 'editor.record.publishVersionError.title',
-          text: 'editor.record.publishVersionError.body',
-          closeMessage: 'editor.record.publishVersionError.closeMessage',
-        })
+        expect(notificationsService.showNotification).toHaveBeenCalledWith(
+          {
+            type: 'error',
+            title: 'editor.record.publishVersionError.title',
+            text: 'editor.record.publishVersionError.body',
+            closeMessage: 'editor.record.publishVersionError.closeMessage',
+          },
+          undefined,
+          expect.any(PublicationVersionError)
+        )
       })
     })
 
     describe('publish error', () => {
       it('shows notification', () => {
         ;(facade.saveError$ as any).next(new Error('oopsie'))
-        expect(notificationsService.showNotification).toHaveBeenCalledWith({
-          type: 'error',
-          title: 'editor.record.publishError.title',
-          text: 'editor.record.publishError.body oopsie',
-          closeMessage: 'editor.record.publishError.closeMessage',
-        })
+        expect(notificationsService.showNotification).toHaveBeenCalledWith(
+          {
+            type: 'error',
+            title: 'editor.record.publishError.title',
+            text: 'editor.record.publishError.body oopsie',
+            closeMessage: 'editor.record.publishError.closeMessage',
+          },
+          undefined,
+          expect.any(Error)
+        )
       })
     })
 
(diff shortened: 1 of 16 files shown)


jahow:
I'm not sure that this fix is the right way forward; the automatic redirection from /create to /edit/TEMP-ID-XXX was made so that refreshing the page was not going to create a new record but instead leave the user on the same draft.

Maybe if it's just a timing issue a different fix could be found? I'm taking a look right now

apps/metadata-editor/src/app/edit/edit-page.component.spec.ts
```diff
@@ -165,22 +165,6 @@ describe('EditPageComponent', () => {
...
-    })
-    it('navigate from /create to /edit/uuid on first change', () => {
-      const router = TestBed.inject(Router)
-      const navigateSpy = jest.spyOn(router, 'navigate')
-      ;(facade.draftSaveSuccess$ as any).next()
-      expect(navigateSpy).toHaveBeenCalledWith(['edit', 'my-dataset-001'], {
-        replaceUrl: true,
-      })
-    })
-  })
```
jahow, line 174: This is actually needed because when a record is first saved its ID is replaced from a temporary one (TEMP-ID-xxxx) to a final UUID

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -137,18 +137,6 @@ export class EditPageComponent implements OnInit, OnDestroy {
...
-      .pipe(
-        filter(
-          (record) =>
-            record?.uniqueIdentifier !== currentRecord.uniqueIdentifier
-        ),
-        take(1)
-      )
-      .subscribe((savedRecord) => {
-        this.router.navigate(['edit', savedRecord.uniqueIdentifier])
-      })
```
jahow, line 153: This is actually needed because when a record is first saved its ID is replaced from a temporary one (TEMP-ID-xxxx) to a final UUID

tkohr:
Thanks for the feedback @jahow! I didn't think of reloading the app on create, good catch!

I've noticed that the reverted `subscribe` that is not yet added to the component's subscriptions still causes the error when editing draft records, so I also wrapped it into a `subscription.add()`. Also added some UTs for the subscriptions. I'll let you have a final look before merging.

libs/feature/editor/src/lib/components/import-record/import-record.component.spec.ts
```diff
@@ -7,6 +7,7 @@ import { NotificationsService } from '@geonetwork-ui/feature/notifications'
 import { RecordsRepositoryInterface } from '@geonetwork-ui/common/domain/repository/records-repository.interface'
 import { of, throwError } from 'rxjs'
 import { MockBuilder, MockComponent, MockModule, MockProviders } from 'ng-mocks'
+import exp from 'constants'
```
jahow, line 10: to remove

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -132,25 +138,29 @@ export class EditPageComponent implements OnInit, OnDestroy {
...
+    this.subscription.add(
```
jahow, line 151: This isn't needed because it's a short-live observable (there is a `take(1)` in the pipe); maybe getting rid of the subscription.add() and mentioning this fact in a comment will avoid confusion in the future?

jahow (approved):
Thank you! I just saw two new things that should be addressed, sorry :grimacing: I'll approve the PR and let you merge when you think it's done :)

libs/feature/editor/src/lib/components/import-record/import-record.component.spec.ts
tkohr: thanks, I've removed it

apps/metadata-editor/src/app/edit/edit-page.component.ts
tkohr: As discussed, I've kept this due to the error when editing a draft record.

PR #1063, opened 2024-12-12
Editor: Extract and remove national/regional toggle

Angi-Kinas:
### Description

This PR extracts and removes/hides the national/regional toggle.

In order to easily configure if the edit form should contain a national/regional toggle, the toggle was extracted out of the spatial extent field and became its own field.
Most of the logic was kept and just moved.
The approach for the form-field is similar to the `form-field-constraints-shortcuts`, meaning it does not have a `model` but a `componentName` in the config.

### Screenshots
This is the current state if the component is hidden:
image

This is how it looks if the component is visible:
image

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
@@ -441,27 +441,6 @@ describe('editor form', () => {
             .should('have.length', 1)
         })
       })
-      describe('spatial scope', () => {
-        it('toggle between national and regional spatial coverage', () => {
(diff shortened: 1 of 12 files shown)


tkohr (approved):
Thanks @Angi-Kinas ! Code LGTM. Toggle is hidden as expected and still works when activating it in the config :-) If I remember right, it's correct that the toggle value does not show up in the keyword list.

I noticed that the place keyword input is not cleared when selecting a value, but I think this is just since the PR has not been rebased. It is not in the scope of this PR anyway.

PR #1064, opened 2024-12-12
[Editor]: Search in the whole catalog from any dashboard

cmoinier:
### Description

This PR makes sure the search through the search header is done throughout the whole catalog, from any dashboard. For instance, the user can find records that they don't own if they start a search from "My records" dashboard.
It also clarifies the search state process, to make sure the states for "myRecords" and "allRecords" stay distinct and are not reinitialized at every render.

### DIFF
--- a/apps/metadata-editor/src/app/records/all-records/all-records.component.ts
+++ b/apps/metadata-editor/src/app/records/all-records/all-records.component.ts
@@ -3,32 +3,24 @@ import {
   ChangeDetectorRef,
   Component,
   ElementRef,
+  OnDestroy,
+  OnInit,
   TemplateRef,
   ViewChild,
   ViewContainerRef,
 } from '@angular/core'
-import {
-  ResultsTableContainerComponent,
-  SearchFacade,
-  SearchService,
-} from '@geonetwork-ui/feature/search'
+import { SearchFacade, SearchService } from '@geonetwork-ui/feature/search'
 import { TranslateModule } from '@ngx-translate/core'
 import { Router } from '@angular/router'
 import { RecordsCountComponent } from '../records-count/records-count.component'
-import { Observable } from 'rxjs'
+import { Observable, Subscription } from 'rxjs'
 import { UiElementsModule } from '@geonetwork-ui/ui/elements'
 import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
-import {
-  CdkConnectedOverlay,
-  CdkOverlayOrigin,
-  Overlay,
-  OverlayRef,
-} from '@angular/cdk/overlay'
+import { CdkOverlayOrigin, Overlay, OverlayRef } from '@angular/cdk/overlay'
 import { TemplatePortal } from '@angular/cdk/portal'
 import { ImportRecordComponent } from '@geonetwork-ui/feature/editor'
 import { RecordsListComponent } from '../records-list.component'
-import { map } from 'rxjs/operators'
-import { SearchHeaderComponent } from '../../dashboard/search-header/search-header.component'
+import { map, take } from 'rxjs/operators'
 import { SearchFiltersComponent } from '../../dashboard/search-filters/search-filters.component'
 import {
   NgIconComponent,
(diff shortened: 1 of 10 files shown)


apps/metadata-editor/src/app/records/all-records/all-records.component.ts
```diff
@@ -94,6 +92,27 @@ export class AllRecordsComponent {
     private cdr: ChangeDetectorRef
   ) {}
 
+  ngOnInit() {
+    this.searchFacade.init('editor')
```
tkohr, line 96: Do we really need to re-init the `searchFacade` each time we create the component? Isn't resetting the search sufficient?

apps/metadata-editor/src/app/records/all-records/all-records.component.ts
```diff
@@ -94,6 +92,27 @@ export class AllRecordsComponent {
...
+    this.searchFacade.searchFilters$
+      .pipe(
+        map((filters) => {
+          if ('owner' in filters) {
+            const { owner, ...rest } = filters
```
tkohr, line 101:
```suggestion
            const { ...rest } = filters
```

apps/metadata-editor/src/app/records/my-records/my-records.component.ts
```diff
@@ -91,7 +91,7 @@ export class MyRecordsComponent implements OnInit {
   ) {}
 
   ngOnInit() {
-    this.searchFacade.resetSearch()
+    this.searchFacade.init('myRecords')
```
tkohr, line 94: Same as above.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -72,7 +72,9 @@ export class FuzzySearchComponent implements OnInit {
     if (this.inputSubmitted.observers.length > 0) {
       this.inputSubmitted.emit(any)
     } else {
-      this.searchService.updateFilters({ any })
+      this.searchFacade.searchId === 'myRecords'
+        ? this.searchService.setFilters({ any })
+        : this.searchService.updateFilters({ any })
```
tkohr, line 77: This is a very specific change on a pretty generic component. Maybe you could listen to `inputSubmitted` in the `SearchHeaderComponent` and remove the owner filter from there if you're on myRecords (detecting this via a `ViewChild` or checking the current URL path for example) ?

libs/feature/search/src/lib/state/search.facade.ts
```diff
@@ -87,39 +88,53 @@ export class SearchFacade {
...
+    if (searchId !== this.searchId) {
+      this.searchId = searchId
+      if (!this.searchExists(searchId)) {
+        this.store.dispatch(new AddSearch(searchId))
+      }
+
+      this.results$ = this.store.pipe(select(getSearchResults, searchId))
+      this.layout$ = this.store.pipe(select(getSearchResultsLayout, searchId))
+      this.isLoading$ = this.store.pipe(
+        select(getSearchResultsLoading, searchId)
```
tkohr, line 112: Hum, I don't really like the changes on the facade and don't think this should be necessary. The facades are already initialized via the `SearchRouterContainerDirective` in the `AppComponent` and the `SearchStateContainerDirective` in the `MyRecordsComponent` and a distinct instance should be available for each state.

tkohr (commented):
Thanks, @cmoinier ! It works really well. I just saw some issues related to the search filters I've merged. I'll take a look at this. => Should now be fixed in https://github.com/geonetwork/geonetwork-ui/pull/1065

Regarding the code, I left some comments for things I think we should try to improve.

Thanks also for adding back the search to the draft page. We might still get rid of one of the gray separator bars under the search here and maybe move the title up? I can't find the mockups for it.

image

apps/metadata-editor/src/app/records/all-records/all-records.component.ts
tkohr: Sorry, the `owner` is actually needed I think

tkohr (approved):
@cmoinier I let you have a final look on this. From my side PR is good to merge (besides the e2e tests that keep staying red).

PR #1030, opened 2024-10-25
Editor: Add a date range search filter

tkohr:
### Description

This PR is still WIP. It prototypically allows setting a date range filter via the calendar for now.

### Quality Assurance Checklist

- [ ] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [ ] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](docs) 📚 has received the love it deserves

---

PR #1065, opened 2024-12-13
Editor: Ignore search value in filter reset button

tkohr:
### Description

Small follow-up PR to #1030 and #1037 to prevent 
- displaying the filter reset button, when doing a textual search
- resetting the textual search when clicking the button

---


tkohr:
>  I was only wondering about one thing. Would it maybe make sense to add `any` to the `FILTER_SUMMARY_IGNORE_LIST`? Then you would not have to introduce any new logic if I understand correctly. I'm fine with how it is, just wanted to understand if this would be a solution as well.

Thanks @Angi-Kinas, really good point ! This is indeed much cleaner :slightly_smiling_face: , didn't even think of it. I've adapted accordingly.

Angi-Kinas (approved):
Thanks @tkohr !

It looks much simpler now :) :rocket:

PR #1075, opened 2025-01-08
[Datahub] Fix dates

AlitaBernachot:
### Description

This PR fixes the content of the "About the data" section regarding the dates: instead of displaying the metadata dates, we now display the resource related dates (`recordUpdated`, `resourceCreated`, and `resourcePublished` instead of `recordCreated` and `recordPublished`).

### Architectural changes

Breaking change:

- `publicationDateForResource` now becomes `resourcePublished` after conversion (instead of `recordPublished`)
- `publicationDate` becomes `recordPublished` after conversion (new)

### Screenshots

No change in the UI.

---

### DIFF
--- a/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
+++ b/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
@@ -136,12 +136,16 @@ export class Gn4FieldMapper {
       ...output,
       recordUpdated: toDate(selectField<string>(source, 'changeDate')),
     }),
-    publicationDateForResource: (output, source) => ({
-      ...output,
-      recordPublished: toDate(
+    publicationDateForResource: (output, source) => {
+      const datePublished = toDate(
         selectField<string>(source, 'publicationDateForResource')
-      ),
-    }),
+      )
+      return {
+        ...output,
+        recordPublished: datePublished,
+        resourcePublished: datePublished,
+      }
+    },
     resourceLanguage: (output, source) => {
       const langList = getAsArray(
         selectField<string>(source, 'resourceLanguage')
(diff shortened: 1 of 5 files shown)


libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -135,12 +135,16 @@ export class Gn4FieldMapper {
...
-      recordPublished: toDate(
+    publicationDateForResource: (output, source) => {
+      const datePublished = toDate(
         selectField<string>(source, 'publicationDateForResource')
-      ),
-    }),
+      )
+      return {
+        ...output,
+        recordPublished: datePublished,
```
jahow, line 144: Am I guessing right that this is kept for backwards compatibility? I think there can theoretically be a `publicationDate` field coming from the index so I would rely on this one instead of `publicationDateForResource`, but maybe you had a good reason to do so

jahow (approved):
This looks really good, thanks a lot! I've added a comment regarding the GN4 formatter, could you please clarify? and then feel free to merge :)

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
AlitaBernachot: Here is an update: https://github.com/geonetwork/geonetwork-ui/pull/1075/files#diff-3e4fae1b2969801149a83f9b728e055e203e1b65a4b0e51c83951edfc8104573R145
Is that what you meant?

jahow: perfect :ok_hand: thanks!

PR #1076, opened 2025-01-09
[DH][IGN] : Quality score widget stuck at 87%

cmoinier:
#### What the issue was

The issue was that a filtering was done during the field mapping to remove any legal constraint that was also present in the licenses, to avoid having duplicates displayed. Since some records only have a legal constraint that happens to be a license, the array ends up empty.

#### Fix

The filtering was removed from the mapping and placed in a display component.

#### What issues remain

_**This PR fixes the behavior of the quality score widget specifically for IGN.**_

The backend still returns a wrong value for the quality score, which is taken instead of the one calculated by the front-end.
The IGN platform doesn't have a value returned for that field, which means that this fix will work on their platform (to test on our platform, we can simulate an empty value [here](https://github.com/geonetwork/geonetwork-ui/blob/17894ab49ccebf771ed6234174346f7360ce54e4/libs/ui/elements/src/lib/metadata-quality/metadata-quality.component.ts#L40)).

With this fix, all of our records that should be displaying a score of 100% will still show 87% with all indicators checked (like on the geo2france platform). 

### Screenshots

Screenshot from 2025-01-09 12-27-40

-----

Thanks @AlitaBernachot for the help on this :slightly_smiling_face:

### DIFF
--- a/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
+++ b/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
@@ -241,6 +241,21 @@ describe('dataset pages', () => {
             .children('div')
             .should('have.length', 5)
         })
+        it('should not display the same text twice in the constraints', () => {
+          // this dataset has the same text for the license and the legal constraints
+          cy.visit('/dataset/9e1ea778-d0ce-4b49-90b7-37bc0e448300')
+          cy.get('datahub-record-metadata')
+            .find('gn-ui-expandable-panel')
+            .first()
+            .click()
+          cy.get('datahub-record-metadata')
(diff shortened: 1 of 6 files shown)


cmoinier:
> Thanks, this makes a lot of sense! do you think you could check that the same text does not appear twice in the metadata-info block in the e2e tests? Maybe modify a record to have the same text in licenses and legalConstraints, and then make sure it only appears once on the record page

Noted and will do, I'm currently writing the tests :)

jahow (approved):
Thank you for all the tests! This fixes one part of the issue yes. The other part (wrong quality scores coming from the index) will be done in a separate PR.

PR #1070, opened 2024-12-19
Migration to Angular 18

jahow:
### Description

Migration to Angular 18 and Nx 20. Most of the work was:
* upgrading dependencies to have them all work together again
* running automatic and manual migrations from NX and Storybook
* fix individual issues appearing in different parts of the code

Note: migration to Angular 19 was attempted and then left out because ng-mocks is not compatible with it yet.

To do:
- [x] fix build for the various apps
- [x] fix failing tests
- [x] fix storybook

### Architectural changes

* Angular updated from 16 to 18
* NX updated from 16 to 20
* Required Node version upgraded to 20+
* Storybook upgraded from 7 to 8
* Many other runtime dependencies upgraded to their most recent versions
* Typescript upgraded from 5.1 to 5.5

### Screenshots

no visual changes

### Quality Assurance Checklist

- [x] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [ ] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [x] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](docs) 📚 has received the love it deserves

---

 **This work is sponsored by IGN**.

PR #1078, opened 2025-01-17
[Datahub] Declutter the download section

ronitjadhav:
### Description

This PR implements the `removeDuplicateFormats` method to prioritize OGC API Features and avoid displaying redundant download links for the same format when available through multiple protocols. Only the OGC API Features version will be shown when both OGC API Features and another protocol provide the same format, ensuring a cleaner interface. 

#### Additional Updates:

- "**Generated by an API**" label is displayed for both WFS and OGC API downloads.
- Added a new GeoNetwork record: "Zones de collecte de déchets en porte à porte - par flux de collecte, jour et horaire de tournée" from MEL.
- Updated the database dump (docker-entrypoint-initdb.d/dump) to reflect the newly added record.

### Screenshots

downloadDeclutter

---


libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -33,12 +33,37 @@ export class DownloadsListComponent {
 
   activeFilterFormats: FilterFormat[] = ['all']
 
+  private filterByProtocol(
+    links: DatasetOnlineResource[]
+  ): DatasetOnlineResource[] {
+    const preferredLinks = new Map<string, DatasetOnlineResource>()
+    const OGC_FEATURES_PROTOCOL = 'ogcFeatures'
```
jahow, line 40: I don't think we need a const here, using the string literal value should be enough :)

libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -33,12 +33,37 @@ export class DownloadsListComponent {
 
   activeFilterFormats: FilterFormat[] = ['all']
 
+  private filterByProtocol(
```
jahow, line 36: this could be called "removeDuplicateFormats" to clarify the intent

libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -33,12 +33,37 @@ export class DownloadsListComponent {
...
+      const format = getFileFormat(link)
+      if (!preferredLinks.has(format)) {
+        preferredLinks.set(format, link)
+      } else {
+        const existingLink = preferredLinks.get(format)
+        if (
+          link.accessServiceProtocol === OGC_FEATURES_PROTOCOL &&
+          existingLink?.accessServiceProtocol !== OGC_FEATURES_PROTOCOL
+        ) {
+          preferredLinks.set(format, link)
```
jahow, line 52: this logic will override a download link even if the `name` of the link is different; this might break datasets which have multiple layers offering the same format

jahow (approved):
Thank you! We can ignore the failing e2e test safely, it'll be fixed by #1070

PR #1082, opened 2025-01-21
[Datahub]: Change related datasets suggestions to be more relevant

Angi-Kinas:
### Description

This PR changes the elastic search service to base its search for related dataset suggestions on the title, abstract and keywords of the current dataset.
This makes sure that the suggestions will be more relevant and more similar to the current dataset.

[For Review]: please make sure you visit different datasets and verify that the suggestions are relevant to the dataset.


libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -148,9 +148,14 @@ export class ElasticsearchService {
                 fields: [
                   'resourceTitleObject.default',
                   'resourceAbstractObject.default',
-                  'tag.raw',
+                  'allKeywords',
+                ],
+                like: [
+                  {
+                    _index: 'gn-records',
```
Angi-Kinas, line 155: Not sure if this should stay hardcoded here. I found it in the docker file and somewhere else but not as a environment variable.

jahow: this can vary across deployments and should not be hardcoded; unfortunately the gn-ui apps don't really have access to this information :/

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -148,9 +148,23 @@ export class ElasticsearchService {
...
+                  {
+                    doc: {
+                      resourceTitleObject: {
+                        default: record.title,
+                      },
+                      resourceAbstractObject: {
+                        default: record.abstract,
+                      },
+                      allKeywords: record.keywords.map(
+                        (keyword) => keyword.label
```
cmoinier, line 163: Does the order here matter? I tested with the MEL instance (only around 50 records, I don't know if it's enough though), and some records have relevant suggestions IMO, but some others are too biased by the title. For instance, a record about wind turbines in Roubaix had suggestions like "women history in Roubaix", "trees in Roubaix" and "citizen crowdfunding in Roubaix" --> their only common point is Roubaix, I checked their keywords & abstract and they do not have anything in common. What if a datahub only has datasets with city names in their title? Will it suggest random records? 
I don't know much about elastic search, but could it check that the title match AND something else match, not just the title?

PR #1086, opened 2025-01-22
[Datahub] Add reuse metadata support

AlitaBernachot:
### Description

This PR adds the support of the new type of metadata: `Reuse metadata`, for:
-  `gn4` (read), 
- `iso19139` (read/write) 
- and `iso19115`-3 (read-write).

### Architectural changes

The new `ReuseRecord` Metadata Model has been added
```ts
export interface ReuseRecord extends BaseRecord {
  kind: 'reuse'
  onlineResources: Array<OnlineLinkResource>
  reuseType: ReuseType
  spatialExtents: Array<DatasetSpatialExtent>
  temporalExtents: Array<DatasetTemporalExtent>
}
```
along with with `ReuseType`
```ts
export type ReuseType = 'application' | 'map' | 'other'
```


libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -288,10 +289,12 @@ export class Gn4FieldMapper {
       ),
     resourceType: (output, source) => {
       const resourceType = getFirstValue(selectField(source, 'resourceType'))
-      const kind = resourceType === 'service' ? 'service' : 'dataset'
+      const kind = getResourceType(resourceType)
+      const reuseType = getReuseType(resourceType)
       return {
         ...output,
         kind,
+        reuseType,
```
jahow, line 297:
```suggestion
        ...(reuseType || { reuseType }),
```
We should not write this key if the reuseType is `undefined`

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
```diff
@@ -339,6 +340,7 @@ export class Iso19139Converter extends BaseConverter<string> {
 
     fieldChanged('uniqueIdentifier') &&
       this.writers['uniqueIdentifier'](record, rootEl)
+    fieldChanged('reuseType') && this.writers['kind'](record, rootEl)
```
jahow, line 343: This looks potentially confusing; maybe we can still have a "reuseType" writer and simply call the "kind" writer inside? At least we would have a homogeneous logic everywhere, even if it's an unnecessary step

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
AlitaBernachot: Right! I updated the gn4 field mapper and added a ut for this specific use case.

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
AlitaBernachot: Ok I have added a new writer for `reuseType` that is calling the `kind` writer.

Issue #971, opened 2024-08-26
Change the database dump to use SQL instead of binary


jahow:
The development database dump used in the E2E tests is currently expressed in "compressed" format (440KB).

Because it is a binary file, git cannot do diff and there is no way to prevent someone overwriting the dump with an older version by mistake.

By using SQL format for the dump (which makes the dump 1.8MB large), git should be able to better detect merge conflicts. Git LFS could be used to avoid making the repo to big.

PR #1088, opened 2025-01-24
Convert database dump in `support-services` to SQL (plain text)

jahow:
`pg_restore -f dump.sql dump`

Also removed the init-db script which isn't necessary anymore (the sql dump is automatically loaded)

Addresses https://github.com/geonetwork/geonetwork-ui/issues/971


ronitjadhav:
@jahow  Works great! 
I tested it locally and also added a new record to the SQL dump successfully.

PR #1080, opened 2025-01-20
Various wording fixes

LHBruneton-C2C:
### Description

This PR is meant to improve default translation for the editor.

PR #1096, opened 2025-01-29
e2e(me): apply label changes to the e2e tests

jahow:
Followup of #1080 

Also fixes a Node error when running e2e tests on node 22


jahow:
happens to the best of us ;)

PR #1097, opened 2025-01-29
Editor: Rework update frequency dropdown

fgravin:
### Description

This PR changes the update frequency dropdown list to fit ISO codelist.

Technically:
- we removed custom frequency from the dropdown, only strings from the codelist are present
- if we detect a custom frequency, we still handle it and display it as the first item of the list.


translations/fr.json
```diff
@@ -171,15 +171,24 @@
   "domain.record.status.ongoing": "Mise à jour continue",
   "domain.record.status.removed": "Supprimé",
   "domain.record.status.under_development": "En cours de création",
+  "domain.record.updateFrequency.annually": "Annuelle",
   "domain.record.updateFrequency.asNeeded": "Lorsque nécessaire",
+  "domain.record.updateFrequency.biannually": "Bi-annuelle",
+  "domain.record.updateFrequency.biennially": "Tous les 2 ans",
```
AlitaBernachot, line 177: To me, it would be:
biannually => "Semestrielle"
biennially => "Bisannuelle"

translations/es.json
```diff
@@ -170,16 +170,37 @@
...
+  "domain.record.updateFrequency.asNeeded": "Según sea necesario",
+  "domain.record.updateFrequency.biannually": "Bianualmente",
+  "domain.record.updateFrequency.biennially": "Bienalmente",
+  "domain.record.updateFrequency.continual": "Continuamente",
+  "domain.record.updateFrequency.daily": "Diariamente",
+  "domain.record.updateFrequency.fortnightly": "Quincenalmente",
+  "domain.record.updateFrequency.monthly": "Mensualmente",
+  "domain.record.updateFrequency.quarterly": "Trimestralmente",
+  "domain.record.updateFrequency.semimonthly": "Semimensualmente",
+  "domain.record.updateFrequency.weekly": "Semanalmente",
```
AlitaBernachot, line 183: from l173 to 183: keys are added twice in this file, with empty values below

libs/api/metadata-converter/src/lib/iso19139/write-parts.ts
```diff
@@ -267,6 +267,24 @@ export function getMaintenanceFrequencyCode(
...
+      return 'monthly'
+    case 'quarterly':
+      return 'quarterly'
+    case 'biannually':
+      return 'biannually'
+    case 'annually':
+      return 'annually'
+    case 'semimonthly':
+      return 'semimonthly'
+    case 'biennially':
```
AlitaBernachot, line 286: maybe i am missing something but i don't really understand the purpose of this function, is it only to output the frequency as a string? + it is missing the default case returning null.

libs/feature/editor/src/lib/components/record-form/form-field/form-field-update-frequency/form-field-update-frequency.component.ts
fgravin, line 1: You should setup the prettier plugin in your IDE for automatic formatting.

libs/api/metadata-converter/src/lib/iso19139/write-parts.ts
fgravin: Yes it seems useless now, maybe because you changed the way frequency code is handled. Dunno

translations/fr.json
fgravin: Ask PO :smile:

AlitaBernachot (approved):
Thanks for the PR!
I have added "Continual" and "Periodic" in the dropdown list and updated related translations for each available language. Also fixed the e2e tests.

libs/api/metadata-converter/src/lib/iso19139/write-parts.ts
AlitaBernachot: I didn't change anything regarding the frequency code handling, i have just allowed 2 more options to be displayed in the dropdown list.

PR #1104, opened 2025-01-30
Me image input preview

LHBruneton-C2C:
### Description

This PR changes the image input component, to use the same image preview as in the datahub (thumbnail + fullscreen). This way, the user can view the full image, and gets an idea of the final display in the datahub.

Also, the alternative text input is disabled, as changing it breaks the overview link.

### Architectural changes

The image-input component has been moved to the ui-elements lib, to allow the use of the image-overlay-preview component.

### Screenshots

image
image


libs/ui/elements/src/lib/image-input/image-input.component.html
```diff
@@ -2,31 +2,16 @@
...
+    <gn-ui-image-overlay-preview class="h-48" [imageUrl]="previewUrl">
+    </gn-ui-image-overlay-preview>
     <gn-ui-text-input
       *ngIf="showAltTextInput"
       [placeholder]="'input.image.altTextPlaceholder' | translate"
       [value]="altText ?? ''"
       (valueChange)="handleAltTextChange($event)"
       extraClass="gn-ui-editor-textarea"
-      [disabled]="disabled"
+      [disabled]="true"
```
AlitaBernachot, line 14: I am wondering why we show the alt text and the button add alt text if we cannot modify it?

AlitaBernachot (approved):
Thanks you for the PR, nice rendering! I don't know if an e2e test is worth here adding here, maybe?

This is not related to your work but:
- I can select any type of file (it won't upload but there is no error nor warning)
- when entering an url i cant click on Select image anymore, i am stuck with the url (also there are no error nor warning when the url is not working)
- regarding accessibility: when clicking on the full size button, there is no close button on the overlay, also i cannot escape/close the overview with the keyboard
- "Overviews" title is plural but i cannot add multiple images?

libs/ui/elements/src/lib/image-input/image-input.component.html
AlitaBernachot: I have read the comments in the ticket, it looks like we are going to fix in another PR?

LHBruneton-C2C: Because we should be able to modify it, it was initially developed that way. But now it's broken, and we need to release a version really soon.

LHBruneton-C2C:
@AlitaBernachot Thanks for your thorough testing and your feedback. I'll put it in the [dedicated Confluence page](https://camptocamp.atlassian.net/wiki/spaces/GeoNetwork/pages/209061480/User+feedback), so we keep it in mind.

PR #1108, opened 2025-01-31
Lock records if user isn't allowed to edit it

f-necas:
### Description

This PR introduces a check that displays the dashboard's records based on if the user is allowed to edit them.
This is based on the `edit` property, that we are getting from a search request.
If the user tries to access a record from the URL directly, an error message with the option to go back to the catalog is displayed.

### Testing tip

Testing can be done with the user "barbie", which has a few records in their org already.

### Architectural changes

none

### Screenshots

image
image

---


fvanderbiest:
Would it make sense to have all records which are editable for the current user grouped at the beginning of the list ?

LHBruneton-C2C:
> Would it make sense to have all records which are editable for the current user grouped at the beginning of the list ?

Sorting with pagination is always tricky, as you might hide valuable information to the user. I'd rather go for a filter option "show only editable records" which would make it clear why the user can only see some records and not others.

libs/api/metadata-converter/src/lib/gn4/gn4.converter.spec.ts
```diff
@@ -3346,6 +3356,7 @@ describe('Gn4Converter', () => {
               id: '8705',
               favoriteCount: 0,
               catalogUuid: 'ce008f24-8e0d-45a8-97f8-9f10399f0190',
+              edit: false,
```
Angi-Kinas, line 3359: I'm just wondering why all the `edit` properties are `false` in this test. Is it expected?

Angi-Kinas (approved):
@cmoinier and @f-necas thank you for this PR!

The code looks good! I just had a question.
I haven't tested it manually yet. Maybe this will make sense to do after rebasing.

libs/api/metadata-converter/src/lib/gn4/gn4.converter.spec.ts
cmoinier: Mmh TBH I put "false" because those tests were returning an error with "edit: false" missing, I didn't go much beyond that because the point of the test was just to check that a value is returned, no matter which. I just had a look and it looks like the fixture `search-response-fixture` is returning that value, but it was added 4 years ago so I think it's outdated. I will change it

f-necas:
[Delete .nvmrc](https://github.com/geonetwork/geonetwork-ui/pull/1108/commits/6ea4bd4fec7f21ece39d2f2381ae10e803e8847f) :'(

PR #1117, opened 2025-02-12
[MEL]: Fix size of share links

cmoinier:
### Description

This PR adds an extraClass to the `gn-ui-data-view-share` component, which has no impact on the display of the component in gn-ui, but will allow the webcomponent on MEL to be stretched all the way across the preview square (see https://mel.integration.apps.gs-fr-prod.camptocamp.com/catalogue/dataset/ville-roubaix::activit%C3%A9-sportive-sport-pass-roubaix).
Removing the `container-lg` and `lg:mx-auto` for the MEL fixes that.

### Architectural changes

none

### Screenshots

none

### DIFF
--- a/apps/datahub/src/app/record/record-data-preview/record-data-preview.component.html
+++ b/apps/datahub/src/app/record/record-data-preview/record-data-preview.component.html
@@ -88,6 +88,7 @@
   </div>
 </div>
 <gn-ui-data-view-share
+  extraClass="container-lg lg:mx-auto"
   *ngIf="displayViewShare$ | async"
   [viewType]="selectedView$ | async"
 ></gn-ui-data-view-share>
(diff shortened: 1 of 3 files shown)


tkohr (approved):
Thanks @cmoinier, good find! The change makes sense to me :-)

apps/datahub/src/app/record/record-data-preview/record-data-preview.component.html
```diff
@@ -88,6 +88,7 @@
   </div>
 </div>
 <gn-ui-data-view-share
+  extraClass="container-lg lg:mx-auto"
```
jahow, line 91: can you maybe just move these classes to a container div outside of the component, instead of adding an input to the component?

jahow (commented):
I think this change makes sense as a component such as `gn-ui-data-view-share` should not have this kind of layout responsibility. Thanks! Just made a suggestion regarding where to put those layout classes

apps/datahub/src/app/record/record-data-preview/record-data-preview.component.html
cmoinier: Yes, good idea!

cmoinier: Yes, good suggestion, done :)

PR #1122, opened 2025-02-18
[Webcomponents] : Full text search

cmoinier:
### Description

This PR makes numerous fixes on the `gn-search-input` webcomponent.

- [ ] Placeholder is now an Input to avoid having an empty value
- [ ] White background for dropdown
- [ ] Fixed space between input and dropdown
- [ ] Fixed `results-list` implementation in `gn-search-input-and-results`
- [ ] Ported various fixes from geocat
- [ ] Added cache cleaning in build script because changes are not always taken into account

### Architectural changes

none

---

### DIFF
--- a/apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
+++ b/apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
@@ -19,33 +19,25 @@ import { CatalogRecord } from '@geonetwork-ui/common/domain/model/record'
   encapsulation: ViewEncapsulation.ShadowDom,
   providers: [SearchFacade, SearchService],
 })
-export class GnSearchInputComponent
-  extends BaseComponent
-  implements AfterViewChecked
-{
+export class GnSearchInputComponent extends BaseComponent {
   @Input() openOnSearch: string
   @Input() openOnSelect: string
-  @ViewChild('searchInput') searchInput: FuzzySearchComponent
-
-  ngAfterViewChecked() {
-    if (this.openOnSearch) {
-      this.searchInput.inputSubmitted.subscribe(this.search.bind(this))
-    }
-    if (this.openOnSelect) {
-      this.searchInput.itemSelected.subscribe(this.select.bind(this))
-    }
-  }
+  @Input() placeholder = 'Search'
 
   search(any: string) {
-    const landingPage = this.openOnSearch.replace(/\$\{search}/, any)
-    window.open(landingPage, '_self').focus()
+    if (this.openOnSearch) {
+      const landingPage = this.openOnSearch.replace(/\$\{search}/, any)
+      window.open(landingPage, '_self').focus()
+    }
   }
 
   select(record: CatalogRecord) {
-    const landingPage = this.openOnSelect.replace(
-      /\$\{uuid}/,
-      record.uniqueIdentifier
-    )
-    window.open(landingPage, '_self').focus()
+    if (this.openOnSelect) {
+      const landingPage = this.openOnSelect.replace(
+        /\$\{uuid}/,
+        record.uniqueIdentifier
+      )
+      window.open(landingPage, '_self').focus()
+    }
   }
 }
--- a/apps/webcomponents/src/styles.css
+++ b/apps/webcomponents/src/styles.css
@@ -1,3 +1,5 @@
+@import '../../../tailwind.base.css';
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
--- a/apps/webcomponents/src/styles.css
+++ b/apps/webcomponents/src/styles.css
@@ -43,6 +43,7 @@
   display: flex;
   max-width: 100%;
   max-height: 100%;
+  top: 108.594px;
 }
 .cdk-overlay-connected-position-bounding-box {
   position: absolute;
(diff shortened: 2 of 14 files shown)


apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
```diff
@@ -19,33 +19,25 @@ import { CatalogRecord } from '@geonetwork-ui/common/domain/model/record'
...
 })
-export class GnSearchInputComponent
-  extends BaseComponent
-  implements AfterViewChecked
-{
+export class GnSearchInputComponent extends BaseComponent {
   @Input() openOnSearch: string
   @Input() openOnSelect: string
-  @ViewChild('searchInput') searchInput: FuzzySearchComponent
+  @Input() placeholder = 'Search'
```
tkohr, line 25: This won't be translated. I don't think we need to make the value configurable via an input.

apps/webcomponents/src/styles.css
```diff
@@ -43,6 +43,7 @@
   display: flex;
   max-width: 100%;
   max-height: 100%;
+  top: 108.594px;
```
tkohr, line 46: That's an interesting value :-)

apps/webcomponents/src/styles.css
```diff
@@ -123,3 +117,12 @@
   .mat-mdc-option.mat-selected:not(.mat-mdc-option-multiple) {
   background: rgba(0, 0, 0, 0.12);
 }
+.mdc-menu-surface.mat-mdc-autocomplete-panel {
+  margin-top: 10px !important;
```
tkohr, line 121: Do we need this margin? I still see a gap between the input and the autocomplete, which we could probably get rid off:
image

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.html
```diff
@@ -1,5 +1,5 @@
 <gn-ui-autocomplete
-  [placeholder]="'search.field.any.placeholder' | translate"
+  [placeholder]="placeholder ?? ('search.field.any.placeholder' | translate)"
```
tkohr, line 2: As mentioned above, I think we should stick with the default value here, which will be translated.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.ts
```diff
@@ -28,6 +28,7 @@ import { SearchFilters } from '@geonetwork-ui/api/metadata-converter'
 export class FuzzySearchComponent implements OnInit {
   @ViewChild(AutocompleteComponent) autocomplete: AutocompleteComponent
   @Input() autoFocus = false
+  @Input() placeholder: string
```
tkohr, line 31: Not needed IMO.

tkohr (commented):
Thanks for the fixes @cmoinier ! This mainly looks good to me.

Regarding the e2e tests, I think the `webcomponents-e2e` app only displays the list as it misses the "preparation" steps. The `webcomponents` app actually behaves the same way. Maybe we can solve the issue by generating a `demo-e2e` app? Which command did you use to generate the `webcomponents-e2e` one?

apps/webcomponents/src/styles.css
cmoinier: I added it because it was like that in the datahub and I thought it was the way to go :) But you are right, maybe we don't need it.

apps/webcomponents/src/styles.css
cmoinier: My point is that if it's fixed on geo2france, then we don't need this value here anymore

cmoinier: Wouldn't surprise me that it's fixed if they have been redeployed recently, because the wc looked good already when I started this work.

libs/feature/search/src/lib/fuzzy-search/fuzzy-search.component.html
cmoinier: Yes but in the ticket, it was written : "fixes: placeholder" --> since it didn't look like it had any issues to me, I thought it was because we needed to make it customisable because for some reasons, translations are not working for some users? But I think I misinterpreted that

cmoinier:
> Thanks for the fixes @cmoinier ! This mainly looks good to me.
> 
> Regarding the e2e tests, I think the `webcomponents-e2e` app only displays the list as it misses the "preparation" steps. The `webcomponents` app actually behaves the same way. Maybe we can solve the issue by generating a `demo-e2e` app? Which command did you use to generate the `webcomponents-e2e` one?

Thanks for the review! I think you are right with the 'demo-e2e' app idea. I did copy/pasted the datahub-e2e app for this work, because the cypress commands were adding a general cypress app and not a dedicated webcomponents one. I think I can easily switch the config from wc app to the demo app, I will try that now :)

apps/webcomponents/src/styles.css
cmoinier: image

Without the margin doesnt look too good actually, I think I prefer it with the margin & it matches what we have in the implems of the fuzzy searches : 

image

tkohr (approved):
Thanks for the adaptions @cmoinier ! I'll approve partially apart from the e2e tests that I've not gotten to work yet neither. Maybe we can push them to a separate branch for now.

PR #1123, opened 2025-02-18
DateService for consistent locale-based date formatting

ronitjadhav:
### Description
 
This PR introduces a centralized `DateService` that standardizes date formatting throughout the application based on the current user locale. The new service wraps the native `toLocaleDateString()` and `toLocaleString()` functions to ensure consistency and maintainability. The changes include:

- **New DateService Implementation:**  
  - A new `DateService` located at `libs/util/shared/src/lib/services/date.service.ts` that provides two methods:
    - `formatDate(date: Date | string, options?: Intl.DateTimeFormatOptions): string`  
      Wraps `toLocaleDateString()`
    - `formatDateTime(date: Date | string, options?: Intl.DateTimeFormatOptions): string`  
      Wraps `toLocaleString()`
  - The service retrieves the current locale from the `TranslateService`, defaulting to `'en-US'` if not available.
  - Unit tests for the `DateService` have been added to ensure that it correctly formats both dates and date-times.

- **Component Updates:**  Direct calls to `toLocaleDateString()` and `toLocaleString()` have been replaced with helper methods that delegate to `DateService.formatDate()` and `DateService.formatDateTime()` for following component:

  - Header-Record Component
  - Metadata-Info Component
  - Edit-Page Component
  - Metadata-Info Component
  - Publish-Button Component
  - Result-Table Component
  - Temporal-Extent-Union Utility
  - Wizard-Summarize Component

---

### DIFF
--- a/libs/feature/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
+++ b/libs/feature/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
@@ -1,5 +1,4 @@
 import { ComponentFixture, TestBed } from '@angular/core/testing'
-
 import { WizardSummarizeComponent } from './wizard-summarize.component'
 import { TranslateModule } from '@ngx-translate/core'
 import { BrowserModule, By } from '@angular/platform-browser'
@@ -7,11 +6,12 @@ import { HttpClientTestingModule } from '@angular/common/http/testing'
 import { NO_ERRORS_SCHEMA } from '@angular/core'
 import { WizardService } from '../../services/wizard.service'
 import { UiInputsModule } from '@geonetwork-ui/ui/inputs'
+import { DateService } from '@geonetwork-ui/util/shared'
 
 const TIME = new Date().getTime()
 
 const wizardServiceMock = {
-  getWizardFieldData: jest.fn((id) => {
+  getWizardFieldData: jest.fn((id: string) => {
     switch (id) {
       case 'title': {
         return 'title'
@@ -46,6 +46,16 @@ const wizardServiceMock = {
   }),
 }
 
+const dateServiceStub = {
+  formatDate: jest.fn(
+    (date: Date, options?: Intl.DateTimeFormatOptions): string => {
+      // For testing, we force the locale to 'en'
+      return date.toLocaleDateString('en', options)
+    }
+  ),
+  formatDateTime: jest.fn(),
+}
+
 describe('WizardSummarizeComponent', () => {
   let component: WizardSummarizeComponent
   let fixture: ComponentFixture<WizardSummarizeComponent>
@@ -65,6 +75,10 @@ describe('WizardSummarizeComponent', () => {
           provide: WizardService,
           useValue: wizardServiceMock,
         },
+        {
+          provide: DateService,
+          useValue: dateServiceStub,
+        },
       ],
     }).compileComponents()
   })
@@ -82,7 +96,6 @@ describe('WizardSummarizeComponent', () => {
   it('should display title', () => {
     const title = fixture.debugElement.query(By.css('.title')).nativeElement
       .textContent
-
     expect(title.trim()).toEqual(
       wizardServiceMock.getWizardFieldData('title').toUpperCase().trim()
     )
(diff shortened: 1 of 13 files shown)


libs/feature/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
```diff
@@ -91,29 +104,31 @@ describe('WizardSummarizeComponent', () => {
...
-        month: 'long',
-        day: 'numeric',
-      })
+    const options: Intl.DateTimeFormatOptions = {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+    }
+    // Use the dateServiceStub to compute the expected formatted date.
+    const expectedDate = dateServiceStub.formatDate(new Date(TIME), options)
```
AlitaBernachot, line 121: Maybe the expected date should simply be the date string?
eg. 
```js
const expectedDate = '01/01/2025'
expect(dateEl.trim()).toEqual(expectedDate)
```

libs/util/shared/src/lib/services/date.service.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      const result = service.formatDateTime(date, options)
+      expect(result).toEqual(expected)
+    })
+
+    it('should format a valid date string with default options', () => {
+      const dateString = '2020-01-01T12:34:56Z'
+      const date = new Date(dateString)
+      const expected = date.toLocaleString('en-US')
+      const result = service.formatDateTime(dateString)
+      expect(result).toEqual(expected)
```
AlitaBernachot, line 86: same comment as above, maybe I am missing something but I don't see why we should use new Date while we can directly set the expected string? (just asking, no need to change :) )

libs/util/shared/src/lib/services/date.service.ts
```diff
@@ -0,0 +1,51 @@
...
+    } else {
+      dateObj = date
+    }
+
+    return options
+      ? dateObj.toLocaleDateString(currentLocale, options)
+      : dateObj.toLocaleDateString(currentLocale)
+  }
+
+  formatDateTime(
```
AlitaBernachot, line 31: Maybe we can refactor these two function? What do you think? because I d'ont see any difference

libs/util/shared/src/lib/utils/temporal-extent-union.spec.ts
```diff
@@ -1,7 +1,30 @@
...
+    const dateObj = typeof date === 'string' ? new Date(date) : date
+    return options
+      ? dateObj.toLocaleString('en', options)
+      : dateObj.toLocaleString('en')
+  },
+} as unknown as DateService
 
-// lock locale to en
 beforeAll(() => {
+  // Override the native toLocaleDateString to use 'en' locale if called as a fallback.
```
AlitaBernachot, line 27: Can't we set the local in a different way? this would prevent mocking and redefining DateService each time we need to test dates maybe?

libs/util/shared/src/lib/utils/temporal-extent-union.ts
```diff
@@ -24,11 +27,11 @@ export function getTemporalRangeUnion(
     return {
       start:
         earliestStartDate !== -Infinity
-          ? new Date(earliestStartDate).toLocaleDateString()
+          ? dateService.formatDate(new Date(earliestStartDate))
```
AlitaBernachot, line 30: I think this utility should return unformatted Date, the formatting should be done in ui component, what do you thing?

ronitjadhav:
Thanks, @AlitaBernachot  for the review!

- I'll look into refactoring the DateService to remove the duplication between `formatDate` and `formatTime`.
- Regarding the formatting guidelines, I couldn't find them explicitly documented either—I followed the description in the ticket.
- I wasn’t aware of Angular’s DatePipe. I guess we should discuss whether switching to it is worth the effort and if we have the budget for it.

libs/util/shared/src/lib/services/date.service.spec.ts
ronitjadhav: I think in this case hardcoding the expected date string can lead to issues because the formatted string might differ across time zones, causing tests to fail unexpectedly.

libs/feature/editor/src/lib/components/wizard-summarize/wizard-summarize.component.spec.ts
ronitjadhav: Agree 👍🏼

libs/util/shared/src/lib/services/date.service.ts
ronitjadhav: Done 👍🏼

libs/util/shared/src/lib/utils/temporal-extent-union.ts
ronitjadhav: Do you think it is necessary ?
Because earlier we also use to format it directly here and later use it in metadata component.

libs/util/shared/src/lib/utils/temporal-extent-union.spec.ts
```diff
@@ -1,7 +1,30 @@
...
+
+// Create a minimal DateService mock by casting
+const dateServiceMock = {
+  formatDate: (
+    date: Date | string,
+    options?: Intl.DateTimeFormatOptions
+  ): string => {
+    const dateObj = typeof date === 'string' ? new Date(date) : date
+    return options
+      ? dateObj.toLocaleDateString('en', options)
```
AlitaBernachot, line 12: Should we be more precise and use `'en-US'`? i don't know if it changes something.

AlitaBernachot (approved):
Thanks for the rework!
`formatDate` and `formatDateTime` still have the same code, maybe this means one is useless? I think it can be refactored a bit more before merging.

PR #1128, opened 2025-02-20
Make e2e more robust

AlitaBernachot:
### Description

This PR introduces [...]

### Architectural changes

### Screenshots

[...]

---


apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
```diff
@@ -594,11 +594,12 @@ describe('dataset pages', () => {
...
-          cy.readFile(path.join('cypress/downloads', 'wfs.csv')).as(
-            'downloadedFile'
-          )
+          cy.readFile(
+            path.join('cypress/downloads', 'rectangles_200m_menage_erbm.csv')
+          ).as('downloadedFile')
           cy.get('@downloadedFile').should('exist')
-          cy.get('@downloadedFile').its('length').should('equal', 3579)
+          // FIXME: This spec always fails with Cypress v13
+          // cy.get('@downloadedFile').its('length').should('equal', 3579)
```
cmoinier, line 602: I saw the explanation you gave about this. I agree that it's the server's responsibility to produce the right file, and the datahub's to use the right path. So if this method cannot work anymore, then let's just check that the right path is called. Could you make that change? Thanks!

cmoinier (approved):
A big thanks for doing this work! It looks good to me. I'm sure it will improve our CI experience :smiley:

apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
AlitaBernachot: `cy.readFile()` is checking if the file exists so it is implied the url is the right one, but just in case, I have added one more check on the href attribute as you have suggested. (did not find a way to intercept window opening _blank successfully)

cmoinier: Thanks! Looks good :)

PR #1134, opened 2025-02-26
fix duplicated links OGC API

Guillaume-d-o:
### Description

This PR is about a bug encontred at MEL. The OGC and WFS links where duplicated because the OGC api does not have a name property linkd to the dataset. So to avoid this, in both OGCApi and WFS services, i reset the collection name. Now both OGC and WFS collection will have the same name, and the duplication remover works.

### Architectural changes

### Screenshots

[...]

---

### DIFF
--- a/apps/datahub/src/app/record/record-downloads/record-downloads.component.spec.ts
+++ b/apps/datahub/src/app/record/record-downloads/record-downloads.component.spec.ts
@@ -73,6 +73,18 @@ class DataServiceMock {
           },
         ])
   )
+  getDownloadUrlsFromOgcApi = jest.fn((url: string) =>
+    url.indexOf('error') > -1
+      ? Promise.reject(new Error('ogc.unreachable.unknown'))
+      : Promise.resolve({
+          name: 'collection1',
+          title: 'Collection 1',
+          bulkDownloadLinks: {
+            'application/geo+json': 'http://example.com/collection1.geojson',
+            'application/json': 'http://example.com/collection1.json',
+          },
+        })
+  )
 }
 
 describe('RecordDownloadsComponent', () => {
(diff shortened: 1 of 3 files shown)


libs/feature/dataviz/src/lib/service/data.service.ts
```diff
@@ -187,6 +188,7 @@ export class DataService {
     return Object.keys(collectionInfo.bulkDownloadLinks).map((downloadLink) => {
       return {
         ...ogcApiLink,
+        name: collectionInfo.title,
```
jahow, line 191: wouldn't you need to use `id` here? I think `title` is the human readable label, so it won't match with the WFS feature type

apps/datahub/src/app/record/record-downloads/record-downloads.component.spec.ts
```diff
@@ -63,14 +63,28 @@ class DataServiceMock {
...
+      ? Promise.reject(new Error('ogc.unreachable.unknown'))
+      : Promise.resolve({
+          name: 'collection1',
+          title: 'Collection 1',
+          bulkDownloadLinks: {
+            'application/geo+json': 'http://example.com/collection1.geojson',
+            'application/json': 'http://example.com/collection1.json',
+          },
+        })
+  )
```
jahow, line 87: Why did you need to add this change to the test? Is this an improvement to the existing test or was thee a failure?

jahow (commented):
Thank you! I added a few comments, I think we're using the wrong field from the collection info. Let me know what you think

apps/datahub/src/app/record/record-downloads/record-downloads.component.spec.ts
Guillaume-d-o: i make the mock in the wrong test and i forgot to delete I think. The goal is to see the forcing of the collection name... in the data service spec not this one.

libs/feature/dataviz/src/lib/service/data.service.ts
Guillaume-d-o: changed to id, both works

jahow (approved):
Thank you! it works :)

PR #1142, opened 2025-03-03
[MEL]: Application banner

cmoinier:
### Description

This PR introduces an application banner, shown on the datahub header when the translation key `appplication-banner` is set in the backend.
It includes different display & styling options as described in https://www.figma.com/design/gA3wp0RCnrZtvEBhtdVkpO/GN4-datahub-camptocamp?node-id=3902-13462&m=dev. Some adaptations will be done on MEL & that's why the component has been made configurable.

### To test

Login & go to `/geonetwork/srv/eng/admin.console#/settings/languages`.
Add the translation key "application-banner" and add the english/french translation (depending on your browser language). Don't forget to save!

### Architectural changes

none

### Screenshots

Screenshot from 2025-03-03 10-05-44


apps/datahub-e2e/src/e2e/header.cy.ts
```diff
@@ -154,4 +154,32 @@ describe('header', () => {
...
+      cy.login()
+      cy.visit('/geonetwork/srv/eng/admin.console#/settings/languages')
+      cy.get('.input-group').find('input').first().type('application-banner')
+      cy.get('.input-group').find('button').click()
+      cy.get('input')
+        .first()
+        .type(
+          'This is a warning message that should be shown when the key is set'
+        )
+      cy.get('.btn-group').eq(1).find('button').first().click()
```
AlitaBernachot, line 168: I don't know if there is an enpoint api to add a new translation key, maybe it will save time to do so as a POST instead of login and navigate in geonetwork admin.
But it works fine as is too, this is a thought about improving e2e runing time.

cmoinier: I think there is, and it would also fail a bit less. Will change that, thanks for the suggestion :)

apps/datahub/src/app/home/home-header/home-header.component.html
```diff
@@ -2,6 +2,11 @@
   <div
     class="container-lg relative h-full mx-auto flex flex-col-reverse justify-between sm:flex-col sm:justify-end"
   >
+    <gn-ui-application-banner
+      [message]="platformService.translateKey(bannerKey) | async"
+      class="container-sm"
+      type="secondary"
+    ></gn-ui-application-banner>
```
AlitaBernachot, line 9: It is a little broken with a long text 😅
image

libs/ui/elements/src/lib/application-banner/application-banner.component.html
```diff
@@ -0,0 +1,20 @@
+<div *ngIf="message" style="top: 0" class="absolute text-wrap bg-white mt-4">
```
AlitaBernachot, line 1: maybe use tailwind `top-0` instead of style atribute?

libs/ui/elements/src/lib/application-banner/application-banner.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    }),
+    provideNgIconsConfig({ size: '1.5em' }),
+  ],
+})
+export class ApplicationBannerComponent {
+  @Input() message: string
+  @Input() title: string
+  @Input() closeEnabled = false
+  @Input() extraClass = ''
+  @Input() icon = 'matWarningAmberOutline'
```
AlitaBernachot, line 33: what if i want to ause a different icon? i think it woud not find it because it could not be provided on the fly. Maybe if we want to customize the icon we should use slots.

libs/ui/elements/src/lib/application-banner/application-banner.component.ts
```diff
@@ -0,0 +1,61 @@
...
+        break
+      case 'secondary':
+        this.msgClass = 'bg-primary-opacity-50 border-primary-darker text-black'
+        break
+      case 'light':
+        this.msgClass =
+          'bg-primary-opacity-10 border-primary-lightest text-black'
+        break
+      default:
+        this.msgClass = 'bg-primary-opacity-50 border-primary-darker text-black'
```
AlitaBernachot, line 49: secondary and default are the same right? maybe regroup them.

```js
case 'secondary':
default:
....
```

libs/ui/elements/src/lib/application-banner/application-banner.component.ts
cmoinier: Yeah we did have this issue in the past with MEL tickets, but I think we solved that yet. The thing is, for MEL, the icons will be different than the ones on gn-ui, so this input will only be used on MEL the there will only be 2 possible icons (so not a customizable icon, actually).

libs/ui/elements/src/lib/application-banner/application-banner.component.ts
```diff
@@ -0,0 +1,61 @@
...
+        break
+    }
+  }
+
+  get classList() {
+    return `${this.msgClass} ${this.extraClass}`
+  }
+
+  closeMessage() {
+    this.message = ''
```
AlitaBernachot, line 59: I don't know if overriding an input value directly in Angular? maybe instead i would have triggered a closing event or something like that?

cmoinier: Actually the close functionality is not asked by the ticket (the banner should not be closable), but since I had time and it was in the design, I added that. Will try to change it if it doesn't take too much time

AlitaBernachot (approved):
Apart from the small display issue with long text, it is working great!
Thanks for the work!

libs/ui/elements/src/lib/application-banner/application-banner.component.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+}
+
+export const LightFull: StoryObj<ApplicationBannerComponent> = {
+  args: {
+    message: 'This is a warning message',
+    title: 'Warning',
+    closeEnabled: true,
+    type: 'light',
+  },
```
AlitaBernachot, line 60: image

I don't see colored background on the story book, i don't if this was made on purpose?

cmoinier: It wasn't, I don't know why it's happening like that in the storybook, maybe it's a matter of colors imports? Because in the datahub the right colors are shown when switching to secondary or light :/ 
Since it works in the app & it's only on storybook, I will look into it if there is time left

cmoinier: Spent 2h looking for the reason why it doesn't work properly, but couldn't find it. What is weird is that the tailwind classes for all borders & the background of `primary` work fine, so it's not a matter of styles import.
I think I will spend ~30mins more on that and then let it go

apps/datahub/src/app/home/home-header/home-header.component.html
cmoinier: So it has been decided that if the message is over a certain length (I decided of the length, so feel free to let me know what you think of it), the message will be scrollable.

Screenshot from 2025-03-06 11-36-07

AlitaBernachot: Ok working great in this case, thx!
Out of curiosity, why not keep the full length though?

cmoinier: Because the header of the datahub has a fixed height (it cannot be "not fixed" without really messing up the layout, it keeps the title in place for instance), so it cannot be growing depending on the banner's height. 

One thing we could've done is setting another fixed height depending on if there is a banner or not, but without knowing the precise size of the banner, the risk if to have a huge gap at the top (if the banner is too small), or still have a banner slipping under the title (in case the size is still too small for the text).

Also, we don't really want a huge header that might take the whole space on the page, if the text is super long :)

AlitaBernachot: Thx for the explanations :)

PR #1143, opened 2025-03-04
[Datahub]: Use EPSG:4326 as a projection for JSON/GeoJSON

cmoinier:
### Description

This PR makes sure the download and API links for json and geojson types are followed by a projection parameter in 4326.
This is done by using the geojson link for both json & geojson in the downloads (since a JSON with a geospatial extent is indeed, a geojson). It also displays the geojson download link if it exists, which was not done before.
In the api-form, the parameter is added to every link that has either the type geojson or json.

### Architectural changes

none

### Screenshots

no UI changes

---

### DIFF
--- a/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
+++ b/apps/datahub-e2e/src/e2e/datasetDetailPage.cy.ts
@@ -543,6 +543,7 @@ describe('dataset pages', () => {
               'csv',
               'excel',
               'json',
+              'geojson',
               'shp',
               'gml',
               'kml',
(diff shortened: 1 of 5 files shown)


LHBruneton-C2C:
Thanks @cmoinier for this PR.
The new SRSNAME parameter is working great!
Concerning the new geojson download link, I'm not so sure we want that, as it will now display twice exactly the same link. I think this will not bring any value, and make the UI less readable.
Can you just remove this (keep only the json link), and then we can merge.

cmoinier:
> Thanks @cmoinier for this PR. The new SRSNAME parameter is working great! Concerning the new geojson download link, I'm not so sure we want that, as it will now display twice exactly the same link. I think this will not bring any value, and make the UI less readable. Can you just remove this (keep only the json link), and then we can merge.

Yep, will do. I know they are the same links, but other users might not and I figured they might ask for it at some point.

PR #1145, opened 2025-03-05
Fix preview webcomponents

LHBruneton-C2C:
When switching to standalone components, the preview components had been removed from the export of their module, but not added directly to the imports of the webcomponents module.

### How to test
- build the webcomponents `npm run build:demo`
- launch the demo app `npm run demo`
- go to `http://localhost:8001/webcomponents/gn-dataset-view-map.sample.html`
- confirm that the webcomponent is working as it should


AlitaBernachot (approved):
Working good, thx!
The "How to test" section saved me a lot of time (i did not spend time to look how/where to test).

image

PR #1146, opened 2025-03-05
Force open panel when results in autocomplete

LHBruneton-C2C:
Sometimes, Angular Material autocomplete doesn't open the suggestion panel even when suggestions are provided (due to some weird async behavior most probably). Force opening the panel when we know suggestions are not empty fixes that.

### How to test
- build the webcomponents `npm run build:demo`
- launch the demo app `npm run demo`
- go to `http://localhost:8001/webcomponents/gn-search-input.sample.html`
- type various inputs (at least 10 different)

The same test on the current [demo](https://geonetwork.github.io/geonetwork-ui/main/demo/webcomponents/gn-search-input.sample.html) page would show the wrong behaviour of sometimes not opening the suggestions panel.

### DIFF
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
@@ -147,6 +147,12 @@ export class AutocompleteComponent
         this.error = null
       }),
       switchMap((value) => this.action(value)),
+      tap((suggestions) => {
+        // forcing the panel to open if there are suggestions
+        if (suggestions.length > 0) {
+          this.triggerRef?.openPanel()
+        }
+      }),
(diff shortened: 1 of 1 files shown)


cmoinier (approved):
Thank you for fixing this :)

AlitaBernachot:
I don't if we should open the dropdown and show "No result" when we type something and there is no result, because it feels it is not working when nothing happens. But this is more of an UX issue i would say.

LHBruneton-C2C:
> I don't know if we should open the dropdown and show "No result" when we type something and there is no result, because it feels it is not working when nothing happens. But this is more of an UX issue i would say.

I agree on both points. I'll talk with Pierre about it, and then this can be part of our gardening/suggestions.

PR #1147, opened 2025-03-05
[Datahub] Allow disabled ogc-client cache per request

AlitaBernachot:
### Description

This PR allows disabling the ogc cache per request.

### How to tests

This PR is a preparation work for the next PR where components will disable/enable cache according to dataset's update frequency when fetching data (eg. table and chart). So no manual testing.

### Architectural changes

A new BaseCacheReader has been introduced from which all other readers (Wfs, Csv, Gml, Geojson, Excel) inherits the `cacheActive` private property.

### Screenshots

No UI

---

### DIFF
--- a/libs/util/data-fetcher/src/lib/data-fetcher.ts
+++ b/libs/util/data-fetcher/src/lib/data-fetcher.ts
@@ -17,29 +17,39 @@ export async function openDataset(
     namespace?: string
     wfsVersion?: WfsVersion
     wfsFeatureType?: string
-  }
+  },
+  cacheActive?: boolean
 ): Promise<BaseReader> {
   const fileType = await inferDatasetType(url, typeHint)
   let reader: BaseReader
   try {
     switch (fileType) {
       case 'csv':
-        reader = new CsvReader(url)
+        reader = new CsvReader(url, cacheActive)
         break
       case 'json':
-        reader = new JsonReader(url)
+        reader = new JsonReader(url, cacheActive)
         break
       case 'geojson':
-        reader = new GeojsonReader(url)
+        reader = new GeojsonReader(url, cacheActive)
         break
       case 'excel':
-        reader = new ExcelReader(url)
+        reader = new ExcelReader(url, cacheActive)
         break
       case 'gml':
-        reader = new GmlReader(url, options.namespace, options.wfsVersion)
+        reader = new GmlReader(
+          url,
+          options.namespace,
+          options.wfsVersion,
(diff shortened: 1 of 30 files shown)


libs/util/data-fetcher/src/lib/data-fetcher.ts
```diff
@@ -17,29 +17,39 @@ export async function openDataset(
...
       case 'wfs':
-        reader = await WfsReader.createReader(url, options.wfsFeatureType)
+        reader = await WfsReader.createReader(
+          url,
+          options.wfsFeatureType,
+          cacheActive
+        )
         break
     }
     reader.load()
```
cmoinier, line 55: How about passing the cache param to `load` directly, instead of adding another layer of reader (with `base-cache`) ? I find that the Readers are already hard to read...

I did it in my PR and it worked too : https://github.com/geonetwork/geonetwork-ui/pull/1149/files#diff-9880751f5b34fe81d60fbb328d4f38d6f5fd1ff1f599057781c0a4e85ebaeefa

But that's just a suggestion, your way works fine too :)

cmoinier (approved):
Thanks for the changes! It looks good, just one suggestion but feel free to merge if you prefer to keep as is :)

libs/util/data-fetcher/src/lib/data-fetcher.ts
AlitaBernachot: This was to avoid changing functions signature. Ok I will simplify and add cache param like you suggest.

AlitaBernachot: Ah I remember why I did this, it is because also i didn't want to pass the activeCache params through all the sub component, cause now i need to pass `activeCache` to data-view > table-view > data-table, whereas if the cacheActive option is in the Reader I would not need all this modifications. What do you think? maybe we can ask a third opinion? Which one is best:
- let all the components and sub components to decide whether to cache or not (and then, we should add `@Input activeCache` to each components/child), simpler, but need edit everywhere when a request is done
- or let the reader handle this one, less modifications, but sub components won't be allowed to deactivate cache (can only be done in parent that creates the reader)

cmoinier: Mmmh okay, explained like that your way makes actually more sense! I think that no matter which component (data/table etc), the link will always have the same update frequency anyway, so it will be cached in all cases --> and that's what this ticket is all about.

PR #1150, opened 2025-03-06
Datahub: Display restricted access info for dataset previews 

tkohr:
### Description

This PR displays a warning to the user if a dataset is not available for preview due to restricted access (401 or 403 response by the service).

- ogc-client has been updated to obtain the http status for WFS
- geospatial-sdk has been updated to include 
  - https://github.com/camptocamp/geospatial-sdk/pull/22
  - https://github.com/camptocamp/geospatial-sdk/pull/23
  - https://github.com/camptocamp/geospatial-sdk/pull/24

### Screenshots

image

### How to test

- Add restricted online resources to a record, such as 
  - https://data.geopf.fr/private/wfs/ows
  - https://data.geopf.fr/private/wms-r
- Go to record in datahub
- Scroll to preview section and click on Map, Table and Chart tabs
- Select a service with restricted access in the dropdown
- The message from the screenshot above should appear

---


cmoinier (changes requested):
Thanks for this big work! It displays fine and the code looks good, but I am noticing the same glitch you had when testing my maxFeatureCount PR & the hiding behavior doesn't seem very consistent to me : 

When switching in the map view from a source with no error to the source with an error, the legend of the previous display stays on. 

image

Also, the table is fully hidden when the message is displayed, but not the map and chart, which I find weird...

image

To be more consistent, I think we should hide the map preview & chart preview as well, like we did for the maxFeatureCount error, WDYT? 
Or we could display the empty table instead to match the map/chart behavior, but choosing the "full hiding" strategy would also hide the map glitch.

tkohr:
Thanks for the review and pointing out the glitch @cmoinier ! Instead of hiding all preview components, I'd rather propose to display them and fix the legend glitch. The corresponding ticket states "make sure the error doesn’t prevent the display of the basemap layer" and I feel it has less visual impact when switching sources. I've added two commits for this. Maybe we can align the max_feature_count behavior with this in the future. Does this sound good to you?

cmoinier (approved):
@tkohr Tested again and it works well, I think this behavior is more logical IMO, thank you for the changes ! :)

PR #1157, opened 2025-03-12
Load multiple feature types from feature catalog

Guillaume-d-o:
### Description

This PR introduce a change on the feature catalog mapping. Previously we retrieved the feature catalog from an embedded extras / featureCatalog id, and we extracted as a POC the first attribute for each dataset. In this PR, the mapping was completed to map all feature types and all feature attributes. 

As it is a continuation of a previous ticket, this is a technical task without any front so the behaviour is only warranty by unit tests at the moment. 

### Architectural changes

No changes 

### How to test

As the developpement is a technical part there is no way to test it directly, only the unit tests can confirm the behaviour of the feature at the moment


libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
```diff
@@ -355,9 +395,7 @@ describe('Gn4Repository', () => {
     let results: CatalogRecord[]
     beforeEach(async () => {
       results = await lastValueFrom(
-        repository.getSimilarRecords(
-          datasetRecordsFixture()[0] as CatalogRecord
-        )
+        repository.getSimilarRecords(datasetRecordsFixture()[0])
```
cmoinier, line 398: why remove the typing? did it change?

cmoinier (commented):
Looks good to me on the code quality side, but a second opinion from someone who worked on the feature catalog already would be good for the functional side :) Thanks!

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -226,7 +226,16 @@ export interface DatasetRecord extends BaseRecord {
   temporalExtents: Array<DatasetTemporalExtent>
   spatialRepresentation?: SpatialRepresentationType
 }
+
 export type DatasetFeatureCatalog = {
+  featureTypes: Array<{
+    name: string
+    definition: string
+    featureAttributes: DatasetFeatureAttributes
```
jahow, line 234: I think this property should be called `attributes` and only contain an array, instead of another object with a single property

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -141,24 +141,33 @@ export class Gn4Repository implements RecordsRepositoryInterface {
       )
   }
 
+  private mapEmbeddedFeatureCatalog(
+    featureTypes: Array<any>
```
jahow, line 145: Using `any` should really be avoided; you can use an LLM if you need a model of the backend responses (by giving it a few samples)

jahow (commented):
Looking good, thanks! The model can be simplified a bit I think :)

libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
Guillaume-d-o: According to Sonar the typing was not necessary and i agree : 
image

libs/api/repository/src/lib/gn4/gn4-repository.ts
Guillaume-d-o: I hesitate between any and take the copilot model will args, so i will know for the next time :)

jahow: The choice is easy: _never any_ ;)

PR #1158, opened 2025-03-12
[Datahub]: Make analysis tab of preview work with WFS

cmoinier:
### Description

This PR makes the analysis tab of the preview section work with WFS, which was not allowed before.

### Architectural changes

none

### Screenshots

image

### DIFF
--- a/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
+++ b/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
@@ -2,7 +2,9 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
+  Inject,
   Input,
+  Optional,
   Output,
 } from '@angular/core'
 import { marker } from '@biesbjerg/ngx-translate-extract-marker'
--- a/libs/util/data-fetcher/src/lib/readers/wfs.ts
+++ b/libs/util/data-fetcher/src/lib/readers/wfs.ts
@@ -96,7 +96,7 @@ export class WfsReader extends BaseCacheReader {
     }
   }
 
-  protected async getData(aggregation?, groupedBy?) {
+  public async getData(aggregation?, groupedBy?) {
     if (aggregation || groupedBy) {
       return { items: await this.getQueryData() }
     }
@@ -129,7 +129,7 @@ export class WfsReader extends BaseCacheReader {
     )
   }
 
-  private async getQueryData() {
+  public async getQueryData() {
     const items = (await this.getData()).items
     const jsonItems = getJsonDataItemsProxy(items)
     const query = generateSqlQuery(
--- a/libs/util/data-fetcher/src/lib/readers/wfs.ts
+++ b/libs/util/data-fetcher/src/lib/readers/wfs.ts
@@ -3,8 +3,9 @@ import { DataItem, DatasetInfo, PropertyInfo } from '../model'
 import { fetchDataAsText } from '../utils'
 import { GmlReader, parseGml } from './gml'
 import { GeojsonReader, parseGeojson } from './geojson'
-import { marker } from '@biesbjerg/ngx-translate-extract-marker'
 import { BaseCacheReader } from './base-cache'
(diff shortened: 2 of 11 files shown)


libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -197,6 +202,12 @@ export class ChartViewComponent {
         .groupBy(['distinct', xProp])
         .aggregate(fieldAgg)
         .read()
+        .then((data) => {
+          if (this.maxFeatureCount && data.length > this.maxFeatureCount) {
+            this.handleError('maxFeature')
+          }
+          return data
+        })
```
jahow, line 210: This check doesn't make sense when we're aggregating, because the resulting amount of items after aggregation is most likely lower than the total count; instead, this check should happen earlier when determining whether a WFS link is appropriate for the chart view

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -49,6 +52,8 @@ marker('chart.aggregation.min')
 marker('chart.aggregation.average')
 marker('chart.aggregation.count')
 
+export const MAX_FEATURE_COUNT = new InjectionToken<string>('maxFeatureCount')
```
jahow, line 55: Wasn't that already defined somewhere else? If this is a duplicated of another injection token, this one will never get set

libs/util/data-fetcher/src/lib/readers/wfs.ts
```diff
@@ -129,6 +130,33 @@ export class WfsReader extends BaseReader {
...
+        this.filter,
+        this.sort,
+        this.startIndex,
+        this.count,
+        this.groupedBy,
+        this.aggregations
+      )
+      const result = await import('alasql').then((module) =>
+        module.default(query, [jsonItems])
+      )
```
jahow, line 158: I don't think this is the right place for this logic. Instead, this is what I would suggest:
1. make the `getData` function take in the following parameters as argument: `getData(selected, filter, sort, startIndex, count)`
2. modify the `getData` function with the following logic:

```
function getData(selected?, filter?, sort?, startIndex?, count?) {

  // if an aggregation or groupBy is set, load all items and use alasql to apply the query
  if (this.aggregations || this.groupBy) {
    const items = await (this.getData()).items // this will read ALL the items
      const jsonItems = getJsonDataItemsProxy(items)
      const query = generateSqlQuery(
        this.selected,
        this.filter,
        this.sort,
        this.startIndex,
        this.count,
        this.groupedBy,
        this.aggregations
      )
      const result = await import('alasql').then((module) =>
        module.default(query, [jsonItems])
      )
     return result.map(jsonToGeojsonFeature)
  }

  // ... the rest of the function using the input parameters
```

The `read()` function shouldn't need any change.

jahow (commented):
Hey, thanks for this proposition, I think it's very close to doing the job perfectly; let me know if my suggestions make sense!

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
cmoinier: Yes, and it's not needed here anymore though, will remove

libs/util/data-fetcher/src/lib/readers/wfs.ts
cmoinier: Okay, makes sense, will implement your suggestion, thanks for correcting! :)

jahow (approved):
Looking good! Not tested but the code looks good.

Could you please add a test for this specific case in the tests for the WFS reader? Feel free to merge once this is done.

cmoinier:
> Looking good! Not tested but the code looks good.
> 
> 
> 
> Could you please add a test for this specific case in the tests for the WFS reader? Feel free to merge once this is done.

I tried to, but I needed to expose a bunch of things from base to do that & I didn't want to change too many things that were common to all types... If that's okay to do then yes, will do

PR #1159, opened 2025-03-12
[Editor] : add unknown constraint when creating a new record

cmoinier:
### Description

This PR adds a constraint "unknown constraint" to the newly created records.
As such, the "Conditions are unknown" switch is already on, and the 3 buttons for the constraints types are hidden.

### Architectural changes

none

### Screenshots

image

### How to test

Create a new record :)


apps/metadata-editor/src/app/new-record.resolver.ts
```diff
@@ -38,7 +38,14 @@ export class NewRecordResolver {
           topics: [],
           keywords: [],
           licenses: [],
-          legalConstraints: [],
+          legalConstraints: [
+            {
+              text: 'Conditions unknown',
```
tkohr, line 43: Wondering if we could/should translate this string...
image

apps/metadata-editor/src/app/new-record.resolver.ts
```diff
@@ -38,7 +38,14 @@ export class NewRecordResolver {
...
+              url: new URL(
+                'http://inspire.ec.europa.eu/metadata-codelist/ConditionsApplyingToAccessAndUse/conditionsUnknown'
+              ),
+            },
```
tkohr, line 47: Couldn't we use the `NOT_KNOWN_CONSTRAINT` const here?

libs/feature/editor/src/lib/components/record-form/form-field/index.ts
```diff
@@ -9,3 +9,4 @@ export * from './form-field-object/form-field-object.component'
 export * from './form-field-array/form-field-array.component'
 export * from './form-field-spatial-extent/form-field-spatial-extent.component'
 export * from './form-field.component'
+export * from './form-field-constraints-shortcuts/form-field-constraints-shortcuts.component'
```
tkohr, line 12: Shouldn't it be
```suggestion
export * from './form-field-constraints-shortcuts/constraints.utils'
```
to use the const mentioned above?

tkohr (approved):
Thanks, @cmoinier, works as expected! I just left some comments inline.

apps/metadata-editor/src/app/new-record.resolver.ts
cmoinier: I tried, but lint wasn't too happy about it : 
image

apps/metadata-editor/src/app/new-record.resolver.ts
cmoinier: Yes that would be best :+1:

libs/feature/editor/src/lib/components/record-form/form-field/index.ts
cmoinier: Oh okay, will try

apps/metadata-editor/src/app/new-record.resolver.ts
tkohr: Indeed, it works, but now I'm wondering if it's actually a good idea to store a translation key in the metadata record. Didn't think about that. We can discuss it at the daily.

cmoinier: Yes, something similar is done [here](https://github.com/geonetwork/geonetwork-ui/blob/29c38455334bb94d951914f00011313a4e0465d4/apps/metadata-editor/src/app/new-record.resolver.ts#L28) though, but it's translated, so I'm unsure too. I won't be at the daily, so let me know.

tkohr: Ok so we discussed at the daily that we should stay with the English term for now.

PR #1151, opened 2025-03-06
[Datahub] Activate cache according to update frequency

AlitaBernachot:
### Description

This PR improves the caching mechanism for data fetching in the table and chart view components. The cache is now dynamically disabled when the record's update frequency is high (defined as more than one update per day or continual). In all other cases, the default behavior of fetching data with cache remains active.

This PR is based on https://github.com/geonetwork/geonetwork-ui/pull/1147

### Architectural changes

Added a new `isHighUpdateFrequency$` in `mdview.facade`

### Screenshots

None

### Quality Assurance Checklist

- [x] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [x] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](docs) 📚 has received the love it deserves

### How to test

Follow these instructions to confirm than the cache is still used when fetching data for a record that does not have a high frequency update rate, and the cache is disabled for record with a high frequency update rate (more than one / day).
- Record with high frequency rate: http://localhost:4200/dataset/ed34db28-5dd4-480f-bf29-dc08f0086131 => should not use the cache (see network devtool tab and check that an xhr call is always done for fetching data)
- Record with low frequency: => should use cache  (see network devtool tab and check that an xhr call is made once when going back to previous page on table view component for example)

---

PR #1171, opened 2025-03-19
Update OGC client package and avoid response error

Guillaume-d-o:
### Description

This PR is a continuation of the https://github.com/geonetwork/geonetwork-ui/pull/1151. An issue was detected on the plateform and data properties was not loaded because an error occurs when the body was readen. The cause is that the body of the GN server answer, was shared using sharedFetch with multiple code occurence while the machine can only read the body once. So the response was clone on shareFetch and dataFetcher. 

### Architectural changes

### Screenshots

[...]

### How to test

copy the dataset https://mel.integration.apps.gs-fr-prod.camptocamp.com/geonetwork/srv/fre/catalog.search#/metadata/d2f92a75-d03a-4048-a4b0-52ac95f0fd44 

change the "Maintenance and update frequency" to "Continual"

On datahub main, the table view still loading because the response body was not readeable 
On this branch, the table view is visible. 

---

### Architectural changes

### Screenshots

[...]

### How to test

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@angular/router": "18.2.13",
     "@bartholomej/ngx-translate-extract": "^8.0.2",
     "@biesbjerg/ngx-translate-extract-marker": "^1.0.0",
-    "@camptocamp/ogc-client": "1.1.1-dev.3e2d3cc",
+    "@camptocamp/ogc-client": "^1.2.1-dev.d666aef",
     "@geospatial-sdk/core": "0.0.5-dev.33",
     "@geospatial-sdk/geocoding": "0.0.5-dev.33",
     "@geospatial-sdk/legend": "0.0.5-dev.33",
--- a/libs/util/data-fetcher/src/lib/utils.ts
+++ b/libs/util/data-fetcher/src/lib/utils.ts
@@ -62,9 +62,11 @@ export function fetchDataAsText(
       })
       .then(async (response) => {
         if (!response.ok) {
-          throw FetchError.http(response.status, await response.text())
+          const clonedResponse = response.clone()
+          throw FetchError.http(response.status, await clonedResponse.text())
         }
-        return response.text()
+        const clonedResponse = response.clone()
+        return clonedResponse.text()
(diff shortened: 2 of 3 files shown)


Guillaume-d-o:
There is no unit testing created because the problem is between GNUI and ogc client. 
Ogc client gives a once readeable response, while we can need to read it multiple times. So we need to clone it. as we use a mock of sharefetch we can't really simulate the bug by unit test and give a clear test.

package.json
```diff
@@ -58,7 +58,7 @@
     "@angular/router": "18.2.13",
     "@bartholomej/ngx-translate-extract": "^8.0.2",
     "@biesbjerg/ngx-translate-extract-marker": "^1.0.0",
-    "@camptocamp/ogc-client": "1.1.1-dev.3e2d3cc",
+    "@camptocamp/ogc-client": "^1.2.1-dev.d666aef",
```
tkohr, line 61: I'd rather stay on a fixed tag here, especially since there is a hash in the version, which can confuse npm.

tkohr (approved):
Thanks @Guillaume-d-o, good catch! Tested and works as expected :-) Do you think the same issue can appear in other places where we use `sharedFetch`?

Guillaume-d-o:
> Thanks @Guillaume-d-o, good catch! Tested and works as expected :-) Do you think the same issue can appear in other places where we use `sharedFetch`?

I think it might appears, but in specific case and in a specific place. Because the dataset must be without cache, and the data requested must be read multiple times. 
Also there is few function that use the sharedfetch, and i think only one is not copying the answer before returning : fetchHeaders. that's why I think it might arrives again

jahow:
I still hope to fix this on ogc-client side, otherwise sharedFetch is very treacherous to use

Guillaume-d-o:
> I still hope to fix this on ogc-client side, otherwise sharedFetch is very treacherous to use

i think it's possible only if we can make a unlimited readable Response. But Response and Response.clone() gave a once readeable objects.

PR #1172, opened 2025-03-20
Change layout for cards depending on content and screen size

Guillaume-d-o:
### Description

This PR introduces a card list layout to make components use a harmonized grid layout. This Layout wrapper is in charge of the layout itself, the pagination of the cards ( the subcomponents) and also the size of its components ( to permit to show more information when there is few elements, and less when there is more elements). The layout is usable with mobile but use caroussel instead is preferable ( as the exemple in otherlink) 
The layout is also responsible of the number of elements per pages. 

### Architectural changes

### Screenshots
Desktop view 
image
image
image

Mobile view : 
image
image
image

### How to test

2 parts of testing. 

About Download part: 
- The component show download with a grid adapted
- in case of filtering, the grid and the pagination is adapted ( grid size can changes and pagination also) ,
- each filter action cause a go back to page 1 

About the link part : 
- in desktop and tablet, there is the layout with blocks elements 
- in mobile, there is a caroussel
- The destop layout has pagination if concerned, and the layout is adapt with the number of elements

I don't have a precise dataset to advise, but it is preferable to test this cases with 4 elements, between 4 and 12, 13 and 18 ( i don't have one) and more than 19. 

All of this cases are covered theorically by e2e and unit tests 

---

### DIFF
--- a/apps/datahub/src/app/record/record-apis/record-apis.component.html
+++ b/apps/datahub/src/app/record/record-apis/record-apis.component.html
@@ -1,16 +1,26 @@
 <div class="flex flex-row gap-x-4 items-center justify-center md:justify-start">
-  <p
-    class="font-title text-xl font-medium text-title text-center sm:text-left"
-    translate
-  >
-    record.metadata.api
-  </p>
+  <div>
+    <span
+      class="font-title text-xl font-medium text-title text-center sm:text-left"
+      translate
+    >
+      record.metadata.api
+    </span>
+    <span class="px-3">({{ apiLinksCount$ | async }})</span>
+  </div>
   <gn-ui-previous-next-buttons
-    *ngIf="carousel.pagesCount > 1"
+    class="block md:hidden"
+    *ngIf="carousel?.pagesCount > 1"
     [listComponent]="carousel"
+  ></gn-ui-previous-next-buttons>
+  <gn-ui-previous-next-buttons
+    class="hidden md:block"
+    *ngIf="list?.pagesCount > 1"
+    [listComponent]="list"
   ></gn-ui-previous-next-buttons>
 </div>
 <gn-ui-carousel
+  class="block md:hidden"
   (currentStepChange)="updateView()"
   containerClass="gap-4 pt-5 pb-7"
   #carousel
@@ -21,14 +31,25 @@
     [link]="link"
     [currentLink]="selectedApiLink"
     size="S"
-    [ngClass]="{
-      'card-shadow': link !== selectedApiLink || !selectedApiLink,
-      'bg-neutral-100': link === selectedApiLink,
-    }"
     (openRecordApiForm)="openRecordApiForm($event)"
   >
   </gn-ui-api-card>
 </gn-ui-carousel>
+<gn-ui-block-list
+  #blockList
+  class="hidden md:block"
+  containerClass="gap-4 pt-5 pb-7"
+>
+  <gn-ui-api-card
+    #block
+    *ngFor="let link of apiLinks$ | async"
+    [title]="link.name"
+    [link]="link"
+    [currentLink]="selectedApiLink"
+    [size]="blockList.subComponentSize"
+    (openRecordApiForm)="openRecordApiForm($event)"
+  ></gn-ui-api-card>
+</gn-ui-block-list>
 <div
   class="content transition-all duration-300 w-screen mx-auto -ml-[calc(50vw-50%)] overflow-hidden"
   [ngClass]="selectedApiLink ? 'ease-in' : 'ease-out'"
--- a/apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.html
+++ b/apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.html
@@ -1,37 +1,47 @@
 <div class="flex flex-row gap-x-4 items-center justify-center md:justify-start">
-  <p
-    class="font-title text-xl font-medium text-title text-center sm:text-left"
-    translate
-  >
-    record.metadata.links
-  </p>
+  <div>
+    <span
+      class="font-title text-xl font-medium text-title text-center sm:text-left"
+      translate
+    >
+      record.metadata.links
+    </span>
+    <span class="px-3">({{ linksCount$ | async }})</span>
+  </div>
   <gn-ui-previous-next-buttons
-    *ngIf="paginableElement?.pagesCount > 1"
-    [listComponent]="paginableElement"
+    class="block md:hidden"
+    *ngIf="carousel?.pagesCount > 1"
+    [listComponent]="carousel"
+  ></gn-ui-previous-next-buttons>
+  <gn-ui-previous-next-buttons
+    class="hidden md:block"
+    *ngIf="list?.pagesCount > 1"
+    [listComponent]="list"
   ></gn-ui-previous-next-buttons>
 </div>
 <ng-container *ngrxLet="otherLinks$ as otherLinks">
+  <gn-ui-carousel
+    class="block md:hidden"
+    containerClass="gap-4 pt-5 pb-7"
+    (currentStepChange)="updateView()"
+  >
+    <gn-ui-link-card
+      *ngFor="let otherLink of otherLinks"
+      [link]="otherLink"
+      class="w-80"
+      size="S"
+    ></gn-ui-link-card>
+  </gn-ui-carousel>
   <gn-ui-block-list
-    containerClass="gap-4 pt-5"
-    *ngIf="otherLinks.length >= 10; else carouselTemplate"
+    #blockList
+    class="hidden md:block"
+    containerClass="gap-4 pt-5 pb-7"
   >
     <gn-ui-link-card
       #block
       *ngFor="let otherLink of otherLinks"
       [link]="otherLink"
-      size="S"
+      [size]="blockList.subComponentSize"
     ></gn-ui-link-card>
   </gn-ui-block-list>
-  <ng-template #carouselTemplate>
-    <gn-ui-carousel
-      containerClass="gap-4 pt-5 pb-7"
-      (currentStepChange)="updateView()"
-    >
-      <gn-ui-link-card
-        *ngFor="let otherLink of otherLinks; let first = first; let last = last"
-        [link]="otherLink"
-        size="S"
-      ></gn-ui-link-card>
-    </gn-ui-carousel>
-  </ng-template>
 </ng-container>
--- a/apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.ts
+++ b/apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.ts
@@ -9,13 +9,13 @@ import { MdViewFacade } from '@geonetwork-ui/feature/record'
 import {
   BlockListComponent,
   CarouselComponent,
-  Paginable,
   PreviousNextButtonsComponent,
 } from '@geonetwork-ui/ui/layout'
 import { CommonModule } from '@angular/common'
 import { LinkCardComponent } from '@geonetwork-ui/ui/elements'
 import { LetDirective } from '@ngrx/component'
 import { TranslateModule } from '@ngx-translate/core'
+import { map } from 'rxjs/operators'
 
 @Component({
   selector: 'datahub-record-otherlinks',
@@ -36,11 +36,12 @@ import { TranslateModule } from '@ngx-translate/core'
 export class RecordOtherlinksComponent implements AfterViewInit {
   otherLinks$ = this.facade.otherLinks$
 
+  get linksCount$() {
+    return this.otherLinks$?.pipe(map((links) => links.length))
+  }
+
   @ViewChild(CarouselComponent) carousel: CarouselComponent
   @ViewChild(BlockListComponent) list: BlockListComponent
-  get paginableElement(): Paginable {
-    return this.carousel || this.list
-  }
 
   constructor(
     public facade: MdViewFacade,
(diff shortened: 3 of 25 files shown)


apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.html
```diff
@@ -11,9 +11,20 @@
   ></gn-ui-previous-next-buttons>
 </div>
 <ng-container *ngrxLet="otherLinks$ as otherLinks">
+  <gn-ui-carousel
+    *ngIf="isMobile$ | async"
+    containerClass="gap-4 pt-[26px] pb-[38px]"
+    (currentStepChange)="updateView()"
+  >
+    <gn-ui-link-card
+      *ngFor="let otherLink of otherLinks; let first = first; let last = last"
```
AlitaBernachot, line 20: `let first = first; let last = last`
are they used here? (looks like they are not)

apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.html
```diff
@@ -11,9 +11,20 @@
...
+      [link]="otherLink"
+      class="w-80"
+    ></gn-ui-link-card>
+  </gn-ui-carousel>
   <gn-ui-block-list
     containerClass="gap-4 py-10"
-    *ngIf="otherLinks.length >= 10; else carouselTemplate"
+    *ngIf="(isMobile$ | async) === false"
```
AlitaBernachot, line 27: can't we use if else here? it would be more understandable i think

libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -75,10 +99,16 @@ export class DownloadsListComponent {
       this.links.some((link) => this.isLinkOfFormat(link, format))
     )
   }
-
+  forceRefreshPagination() {
+    setTimeout(() => {
```
AlitaBernachot, line 103: this looks like a hack (using timeout), is there another way to trigger changes?

libs/ui/layout/src/lib/block-list/block-list.component.ts
```diff
@@ -71,14 +78,50 @@ export class BlockListComponent implements AfterViewInit, Paginable {
...
+
+  protected computePageSize(): number {
+    switch (this.subComponentSize) {
+      case 'M':
+        return 4
+      case 'S':
+        return 6
+      case 'XS':
+        return 8
+      default:
```
AlitaBernachot, line 113: case M and default are the same, maybe regroup them?

libs/ui/layout/src/lib/block-list/block-list.component.ts
```diff
@@ -71,14 +78,50 @@ export class BlockListComponent implements AfterViewInit, Paginable {
     })
   }
 
-  // pageIndex is 1-based
+  protected refreshSubComponentsVisibility = () => {
+    this.blocks.forEach((subComponent, index) => {
+      const element = subComponent.nativeElement
+      element.style.display =
```
AlitaBernachot, line 84: Not sure if this is a good practice to override css style  in js from the great parent. Can't this be done in Angular, with a ngif directive instead?

libs/ui/layout/src/lib/block-list/block-list.component.ts
```diff
@@ -22,14 +23,14 @@ import { PaginationDotsComponent } from '../pagination-dots/pagination-dots.comp
   imports: [CommonModule, PaginationDotsComponent],
 })
 export class BlockListComponent implements AfterViewInit, Paginable {
-  @Input() pageSize = 5
+  pageSize = 10
```
AlitaBernachot, line 26: for my own understanding, what is the difference between computePageSize and this pageSize property?

libs/ui/layout/src/lib/block-list/block-list.component.ts
```diff
@@ -13,6 +13,7 @@ import { CommonModule } from '@angular/common'
 import { Paginable } from '../paginable.interface'
 import { PaginationDotsComponent } from '../pagination-dots/pagination-dots.component'
 
+type ComponentSize = 'L' | 'M' | 'S' | 'XS'
```
AlitaBernachot, line 16: Is there something to do with 'L' size?

libs/ui/layout/src/lib/block-list/block-list.component.css
```diff
@@ -5,3 +5,31 @@
 :host {
   position: relative;
 }
+
```
AlitaBernachot, line 8: Any reason why you didn't use tailwind predefined classes in this file?

apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.ts
```diff
@@ -44,15 +46,23 @@ export class RecordOtherlinksComponent implements AfterViewInit {
...
   updateView() {
     this.changeDetector.detectChanges()
   }
 
   ngAfterViewInit() {
-    // this is required to show the pagination correctly
     this.changeDetector.detectChanges()
   }
+
+  isMobile$ = this.breakpointObserver.observe([Breakpoints.XSmall])?.pipe(
```
AlitaBernachot, line 61: I don't know why you need to detect screen size in js here, you can handle responsiveness only by using tailwind css classes (sm:, md:, ...) this will keep homogeneity regarding media queries breakpoints. (This may need a refacto between the carousel and the block component maybe)

AlitaBernachot:
Thx for the work.
There are some huge space appearing between sections here 
image

AlitaBernachot:
Also tested small iphone screen, links appear as if i had a big screen:

image

Guillaume-d-o:
demand from other task : make the count of each download , API, link, ... here by @LHBruneton-C2C

libs/ui/elements/src/lib/downloads-list/downloads-list.component.ts
```diff
@@ -94,6 +124,7 @@ export class DownloadsListComponent {
     if (this.activeFilterFormats.length === 0) {
       this.activeFilterFormats = ['all']
     }
+    this.forceRefreshPagination()
```
jahow, line 127: This absolutely shouldn't be here; pagination is definitely not a concern of this component!

libs/ui/layout/src/lib/block-list/block-list.component.ts
Guillaume-d-o: It was mandatory to define a page size for the page count ( cannot divide by 0 or null). To avoid error, i let the pageSize defined at 10. then at each update of the list, we execute : 
    this.pageSize = this.computePageSize()

so the pageSize property is the result of the function, after the init and for each rendering

libs/ui/layout/src/lib/block-list/block-list.component.ts
Guillaume-d-o: yep we can, i do it !

libs/ui/layout/src/lib/block-list/block-list.component.ts
```diff
@@ -47,18 +51,25 @@ export class BlockListComponent implements AfterViewInit, Paginable {
...
+      this.updateSizes()
+      this.refreshBlocksVisibility()
+      this.goToPage(1)
+      this.changeDetector.detectChanges()
+      this.listChanges.emit(this)
+    })
+    this.updateSizes()
     this.refreshBlocksVisibility()
 
-    // we store the first height as the min-height of the list container
```
jahow, line 59: is there a reason for deleting all these comments? :thinking:

Guillaume-d-o: A mistake i think it was mine

apps/datahub/src/app/record/record-otherlinks/record-otherlinks.component.html
Guillaume-d-o: done on the 3 layouts

AlitaBernachot:
Can we have screenshots in the dedicated PR section please? (desktop and smaller screens) 😬

AlitaBernachot:
Some spaces added on smaller screens for downloads
image

AlitaBernachot:
Different spacing between sub sections
image

AlitaBernachot:
Arrows are now squares with rounded border on the mockups

image

AlitaBernachot:
This could probably be homogenize
image

AlitaBernachot:
Looks like the link is align with the bottom in the mockups (instead of centered vertically)

image

LHBruneton-C2C:
> According to the ticket, we should have a fixed width
> 
> image

@AlitaBernachot How did you get this screenshot? A carousel for this many elements will only appear in small screen mode at the end of the dev (don't know where this PR stands, not yet the end I guess).

The width is fixed for grid and carousel, but adaptable for list elements (search and links if I remember correctly, not API and downloads).

LHBruneton-C2C:
> Maybe the translation value should be updated? (looks ok in EN, but KO in FR)
> 
> image

The second "Ressources & liens" shouldn't have changed. It was and must remain "Liens".

LHBruneton-C2C:
Thanks @AlitaBernachot for this detailed review!

Guillaume-d-o:
> Looks like the link is align with the bottom in the mockups (instead of centered vertically)
> 
> image

The mockup for the M size (here 2 elements = M ) => action are right

AlitaBernachot (commented):
I still don't get why we have different sizes here
image

AlitaBernachot (commented):
~ ~Thx for the work but i still find some inconsistancies: on smaller size, sometimes it's the Download that are size L, and the the API links are S, and when the size screen diminished, the sizes are inverted.~ ~

EDIT => OK this looks like the intended behavior

image

image

apps/datahub/src/app/record/record-apis/record-apis.component.html
```diff
@@ -1,16 +1,26 @@
...
+      class="font-title text-xl font-medium text-title text-center sm:text-left"
+      translate
+    >
+      record.metadata.api
+    </span>
+    <span class="px-3">({{ apiLinksCount$ | async }})</span>
+  </div>
   <gn-ui-previous-next-buttons
-    *ngIf="carousel.pagesCount > 1"
+    class="block md:hidden"
```
AlitaBernachot, line 12: any reason why we should hide prev/next buttons in some cases?

apps/datahub/src/app/record/record-apis/record-apis.component.html
```diff
@@ -21,14 +31,25 @@
...
     (openRecordApiForm)="openRecordApiForm($event)"
   >
   </gn-ui-api-card>
 </gn-ui-carousel>
+<gn-ui-block-list
+  #blockList
+  class="hidden md:block"
+  containerClass="gap-4 pt-5 pb-7"
+>
+  <gn-ui-api-card
```
AlitaBernachot, line 43: so you are duplicating elements to handle card sized? couldn't this have been avoided? this probably would add some more work, so, ok to keep like this.

tailwind.base.config.js
```diff
@@ -1,6 +1,9 @@
 module.exports = {
   theme: {
     extend: {
+      screens: {
+        xs: '320px',
```
AlitaBernachot, line 5: are you really using this one or was this added jsut for testings?

tailwind.base.css
```diff
@@ -134,20 +134,8 @@
...
-    @apply min-h-[135px] md:w-80;
-  }
-
-  .gn-ui-card-m {
-    @apply min-h-[68px] md:w-[487px];
-  }
-
-  .gn-ui-card-l {
-    @apply min-h-[68px] md:w-[992px];
+    @apply min-h-[64px] w-full xs:min-w-[260px] xs:max-w-[280px] sm:min-w-[300px] sm:max-w-[500px] md:min-w-[230px];
```
AlitaBernachot, line 138: any reason why you kept the ending "-s" here?
since all other css classes have been removed, there should be some cleaning to do elsewhere in the code.

AlitaBernachot:
It could be great to add some spacing here (like before)

image

apps/datahub/src/app/record/record-apis/record-apis.component.html
AlitaBernachot: oh i get it, it is when the component is displayed as a carousel

Guillaume-d-o:
> Also noticed: before filtering, cards are M, but after filtering when there is only one card, it is displayed as L, the ticket mention: "The size is based on the total of elements, not on the filtered ones ?"

That's true, but the ticket is only done for the desktop size, not for the mobile that keep the old behaviour as possible. That's why all caroussel thumbnail had a fixed size at the moment.

AlitaBernachot:
> > Also noticed: before filtering, cards are M, but after filtering when there is only one card, it is displayed as L, the ticket mention: "The size is based on the total of elements, not on the filtered ones ?"
> 
> That's true, but the ticket is only done for the desktop size, not for the mobile that keep the old behaviour as possible. That's why all caroussel thumbnail had a fixed size at the moment.

My test has nothing to do with mobile/ or desktop size, it happens on download links even on desktop.

AlitaBernachot:
> > > Also noticed: before filtering, cards are M, but after filtering when there is only one card, it is displayed as L, the ticket mention: "The size is based on the total of elements, not on the filtered ones ?"
> > 
> > 
> > That's true, but the ticket is only done for the desktop size, not for the mobile that keep the old behaviour as possible. That's why all caroussel thumbnail had a fixed size at the moment.
> 
> My test has nothing to do with mobile/ or desktop size, it happens on download links even on desktop.

EDIT: all right, it was not clear for me that this has been put aside because of the complexity, thanks for the clarification. It's working well, and as expected.

PR #1173, opened 2025-03-20
[Datahub]: Rework external links cards

cmoinier:
### Description

This PR reworks the cards for API links, Download links and other links.
Those cards are now available in L/M/S/XS sizes. They are set by default as L for the downloads, and S for the links and API to keep consistency with the current datahub layout.

### Architectural changes

none

### Screenshots

image

### How to test

Checkout the cards on Storybook where all sizes are available.


Guillaume-d-o (changes requested):

libs/ui/elements/src/lib/download-item/download-item.component.html
```diff
@@ -1,36 +1,36 @@
...
+  [ngClass]="cardClass"
 >
-  <div class="grow-1 w-full overflow-hidden">
-    <div
-      class="text-21 text-black truncate font-title w-11/12"
-      [title]="link.description || link.name"
-    >
+  <div class="flex flex-col justify-between">
+    <div class="gn-ui-card-title" [title]="link.description || link.name">
       {{ link.description || link.name }}
```
Guillaume-d-o, line 10: On the card, your titles, details and "subdetails" are on the same font size, that's not the case before and in the design

Guillaume-d-o: this comment can be applied to API and link card

Guillaume-d-o (changes requested):

libs/ui/elements/src/lib/download-item/download-item.component.html
```diff
@@ -1,36 +1,36 @@
...
       >
     </div>
   </div>
-  <div class="shrink-1 w-14 flex flex-col justify-center items-center">
-    <ng-icon
-      class="!w-8 !h-8 card-icon text-3xl"
-      name="matCloudDownloadOutline"
-    >
-    </ng-icon>
+  <div class="flex" [ngClass]="size === 'S' ? 'items-end' : 'items-center'">
```
Guillaume-d-o, line 27: i don't know if it was necessary but in L /M /S /XS elements are not adapted in their content. I  think XS card at least must be with less details. As it is a design part not necessary i think we will discuss about it tomorrow.

Guillaume-d-o:
Thanks for this very good PR. I think a quick update will be necessary regarding the card contents and card sizes (both of our programs limit their sizes).

A few code updates needed, but thank you very much!

libs/ui/elements/src/lib/download-item/download-item.component.html
cmoinier: It is necessary for the S card for the API because the title was being shrinked too much on the left as the two icons were taking up needless space. So this adaptation is needed
As for the détails and content of the cards, Laure-Helene and I did a lot of back and forth yesterday & the current content has been validated then :) (We decided that the mockups were not complete)

libs/ui/elements/src/lib/download-item/download-item.component.html
cmoinier: Good call!

Guillaume-d-o (approved):
all is OK !

PR #1176, opened 2025-03-20
[Editor]: Bug first letter eaten

cmoinier:
### Description

This PR fixes the behavior in the "Link to a service" input where the first letter was eaten and where http://example.com turned into  https://e/xample.com. 

### Architectural changes

none

### Screenshots

no UI changes

### How to test

---

### DIFF
--- a/libs/ui/inputs/src/lib/url-input/url-input.component.ts
+++ b/libs/ui/inputs/src/lib/url-input/url-input.component.ts
@@ -31,7 +31,7 @@ import { iconoirArrowUp, iconoirLink } from '@ng-icons/iconoir'
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UrlInputComponent implements OnChanges {
+export class UrlInputComponent {
   @Input() set value(v: string) {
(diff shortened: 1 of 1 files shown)


Guillaume-d-o (approved):
The field is visible from the first letter. There is no regression ( test and unit test & end to end test). All is fine Good job !

jahow:
Wait, are you sure that this fix does not break an existing behavior? It looks like it was there to update the field when its input value changes

jahow:
Oh yes indeed, it looks like updating `this.inputValue` was happening twice before this change? is that the root of the issue?

Guillaume-d-o:
> Wait, are you sure that this fix does not break an existing behavior? It looks like it was there to update the field when its input value changes

I don't see any break. on main ( so on this branch) i can' t submit a dataset. For the read / write / update / read again a dataset in draft there is no issue

cmoinier:
> Oh yes indeed, it looks like updating `this.inputValue` was happening twice before this change? is that the root of the issue?

Yes I think it was the issue, when logging the different updates it looked like they were clashing, hence why some characters were added at the wrong places.

PR #1178, opened 2025-03-21
[metadata editor] fix download url ( don't fetch the format) and fix drag and drop when add an url is clicked

Guillaume-d-o:
### Description

This PR introduces two fixes : 
- the first one, when an url was given, nothing was downloaded. the http wants header that are not given each time. So it create an empty file that doesn't exit
- The second ones is about url. Previously if we clicked on the url part, it was impossible to drag and drop or open a file. Now it is possible to change. 

### Architectural changes

### Screenshots

image

### How to test

---

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/edit.cy.ts
@@ -174,62 +174,10 @@ describe('editor form', () => {
             .should('eq', 'modified abstract')
         })
       })
-      describe('graphic overview', () => {
-        beforeEach(() => {
-          // Delete existing image if it exists
-          cy.get('gn-ui-image-input').then(($input) => {
-            if ($input.find('img').length > 0) {
-              cy.wrap($input).find('[data-cy="delete-image"]').click()
-              cy.editor_wrapPreviousDraft()
-              cy.editor_publishAndReload()
-              cy.get('gn-ui-image-input img').should('not.exist')
-            }
-          })
-        })
 
+      describe('graphic overview', () => {
         it('shows the graphic overview', () => {
           cy.get('gn-ui-image-input').should('be.visible')
-        })
-        it('allows switching between URL input and file upload', () => {
-          // First add by URL
-          cy.get('gn-ui-image-input')
-            .find('gn-ui-button')
-            .find('button')
-            .click()
-          cy.get('gn-ui-url-input').should('be.visible')
-          cy.get('gn-ui-url-input input').type('http://example.com/image.jpg')
-
-          // Then try to upload a file - URL input should disappear
-          cy.get('gn-ui-form-field-overviews label').selectFile(
-            'src/fixtures/sample.png',
-            { force: true }
-          )
-          cy.get('gn-ui-url-input').should('not.exist')
-
-          // Try URL input again - should clear previous file
-          cy.get('gn-ui-image-input')
-            .find('gn-ui-button')
-            .find('button')
-            .click()
-          cy.get('gn-ui-url-input').should('be.visible')
-        })
-
-        it('handles drag and drop file upload', () => {
-          // First add by URL
-          cy.get('gn-ui-image-input')
-            .find('gn-ui-button')
-            .find('button')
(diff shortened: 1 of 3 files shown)


apps/metadata-editor-e2e/src/e2e/edit.cy.ts
```diff
@@ -215,7 +218,59 @@ describe('editor form', () => {
             .should('be.visible')
         })
       })
+
+      describe('image field', () => {
+        beforeEach(() => {
+          // Delete existing image if it exists
+          cy.get('gn-ui-image-input').then(($input) => {
```
AlitaBernachot, line 225: I think this one could be simpler? (no `.then()` needed)

libs/ui/elements/src/lib/image-input/image-input.component.html
```diff
@@ -126,6 +126,7 @@
         type="gray"
         [disabled]="disabled"
         [extraClass]="disabled && 'cursor-not-allowed'"
+        data-test="display-url-input"
```
AlitaBernachot, line 129: `data-test` or `data-cy` maybe?

AlitaBernachot (approved):
Thx for the Pr, it is working well! Thx for the e2e, i guess i will have some rebase work to do inside https://github.com/geonetwork/geonetwork-ui/pull/1187.

Also it would have been great to add some more info in the "How to test" section in the PR description because I did not understand the first point (how to reproduce?).

Guillaume-d-o:
> Thx for the Pr, it working well! Thx for the e2e, i guess i will have some rebase work to do inside #1187.
> 
> Also it would have been great to add some more info in the "How to test" section in the PR description because I did not understand the first point (how to reproduce?).

Oups mistake from my part i totally forgot. Initially, when we click on the button add URL, the image box was disabled, so the user can't drop an image or select an image if url bar is shown

libs/ui/elements/src/lib/image-input/image-input.component.html
Guillaume-d-o: no one, i used it because my test failed but i removed it after => i won't keep it in production code

apps/metadata-editor-e2e/src/e2e/edit.cy.ts
Guillaume-d-o: I don't see how delete an optional image without a ".then." at this point we are not sure the image is on the gn-ui-image-input, so make a delete directly make the beforeEach fails.

cmoinier: +1 with @Guillaume-d-o, I don't think it is possible to remove `then` because cypress will throw an error when the image is gone before the second test. It's not an issue to keep it IMO.
If we want to simplify a bit it could be : 

```suggestion
  cy.get('gn-ui-image-input img').then(($img) => {
    if ($img.length > 0) {
      cy.get('[data-cy="delete-image"]').click();
      cy.editor_wrapPreviousDraft();
      cy.editor_publishAndReload();
      cy.get('gn-ui-image-input img').should('not.exist');
    }
  });
```

LHBruneton-C2C:
@Guillaume-d-o I think @AlitaBernachot was talking about this

> the first one, when an url was given, nothing was downloaded. the http wants header that are not given each time. So it create an empty file that doesn't exit

which we discussed and decided to remove from your work, as it was specific to you local tests. So maybe you could remove it from the PR description ?

Or I'm completly wrong, but then we'll still need some more explanations.

Guillaume-d-o:
> @Guillaume-d-o I think @AlitaBernachot was talking about this
> 
> > the first one, when an url was given, nothing was downloaded. the http wants header that are not given each time. So it create an empty file that doesn't exit
> 
> which we discussed and decided to remove from your work, as it was specific to you local tests. So maybe you could remove it from the PR description ?
> 
> Or I'm completly wrong, but then we'll still need some more explanations.

You're right i forgot to update the PR context after the discussion. In fact yes, this PR is only about the button to add an url that disable the drag and drop / image selector section.

PR #1179, opened 2025-03-21
Fix converter on records with empty feature catalog

LHBruneton-C2C:
### Description

This PR fixes the converter when a record has an empty related feature catalog array.

### How to test

Import this record:

https://dev.geo2france.fr/datahub/dataset/fr-120066022-jdd-0b0138ad-46f1-41d5-9725-67f917de260d

The record page should not display this error message:

image

### DIFF
--- a/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
+++ b/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
@@ -291,7 +291,7 @@ export class Gn4FieldMapper {
             <SourceWithUnknownProps>selectField(source, 'related'),
             'fcats'
           )
-        ),
+        ) ?? {},
         '_source'
       )
       const featureCatalogIdentifier = selectField(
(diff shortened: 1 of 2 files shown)


AlitaBernachot (approved):
Thx for the fix!

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -291,7 +291,7 @@ export class Gn4FieldMapper {
             <SourceWithUnknownProps>selectField(source, 'related'),
             'fcats'
           )
-        ),
+        ) ?? {},
```
AlitaBernachot, line 294: Just wondering, shouldn't this be the atomic operation's responsbility?

LHBruneton-C2C: I was wondering about that too, but then that would mean we can't check on the value being defined anymore. `getFirstValue` does what it says, without any changes to the value.

PR #1180, opened 2025-03-21
[Editor] : Save XML imports

cmoinier:
### Description

This PR introduces automatic saving (but not published) for imported XML records. The behavior is the same as created and duplicated ones : 

- [x] New record is NOT saved as draft
- [x] Status is "saved - not published"
- [x] Option to publish is directly available
- [x] No temporary UUID is set, it's already the definitive one

### Architectural changes

none

### Screenshots

no UI change

### How to test

Import an XML file and check that the above list is true.

### DIFF
--- a/apps/metadata-editor-e2e/src/e2e/record-actions.cy.ts
+++ b/apps/metadata-editor-e2e/src/e2e/record-actions.cy.ts
@@ -1,5 +1,8 @@
 // eslint-disable-next-line @nx/enforce-module-boundaries
-import { simpleDatasetRecordAsXmlFixture } from '@geonetwork-ui/common/fixtures'
+import {
+  duplicateDatasetRecordAsXmlFixture,
+  simpleDatasetRecordAsXmlFixture,
+} from '@geonetwork-ui/common/fixtures'
 
 describe('record-actions', () => {
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


AlitaBernachot (changes requested):
Thx for the work, code ok but I noticed some strange behavior: 
1. sometimes the new record won't appear in the dataset list (i can't  tell how to reproced, it happened only twice)
2. "Copy" is always added to newly added record
here, i have added two new records via the import button
image

cmoinier:
> Thx for the work, code ok but I noticed some strange behavior:
> 
>     1. sometimes the new record won't appear in the dataset list (i can't  tell how to reproced, it happened only twice)
> 
>     2. "Copy" is always added to newly added record
>        here, i have added two new records via the import button
>        image

For the first point, were you particularly quick to go back to the dashboard? I noticed this happening to new/duplicate links when we switch back too quickly, but after reloading the page or doing something else it ends up appearing. However I don't think it's related to this work, maybe we could create a bug ticket for that?

@LHBruneton-C2C The second point is something that existed before this PR too, it's in the function `duplicateExternalRecord` --> do we want to keep it ?

AlitaBernachot:
> The second point is something that existed before this PR too, it's in the function 

OK! i thought there was something missing in this PR. Since it is not related to yours approving the PR.

AlitaBernachot (approved):

AlitaBernachot:
> were you particularly quick to go back to the dashboard?

I may have been quick to go back to the dashboard :). The record appeared in the geonetwork dashboard but not in the editor dashboard, even after refresh. But I can't reproduce the issue anymore.

cmoinier:
> > were you particularly quick to go back to the dashboard?
> 
> I may have been quick to go back to the dashboard :). The record appeared in the geonetwork dashboard but not in the editor dashboard, even after refresh. But I can't reproduce the issue anymore.

Mmh that's strange that it didn't appear after refresh. Let's keep that in mind and if it comes up again we will create a bug ticket for that.

LHBruneton-C2C:
> @LHBruneton-C2C The second point is something that existed before this PR too, it's in the function `duplicateExternalRecord` --> do we want to keep it ?

Didn't have time to answer ^^'
This isn't critical so let's keep it like this. We already have more important issues to deal with.

PR #1187, opened 2025-03-25
[Editor] Image upload component UI error state

AlitaBernachot:
### Description

This PR introduces a better error display in the input image component when an image upload fails (CORS error, invalid url, invalid type file).

The input file has been forced to accept only image file type: in the prompt window, the user will only see image files (however, this does not prevent the user to drag and drop invalid file type).

### Architectural changes

This was not part of the jira ticket, but e2e tests for metadata-editor have been reorganized into folders and sub folders to make e2e writing and testing easier when ran locally.

image

NB. latest Cypress version allows the developer to trigger all specs  available in a folder by simply clicking a play button located after the folder name (in the spec tree view), this functionality is sadly not available in Cypress v13.17 we currently use.

### Screenshots

image

### How to test

Follow these instructions to confirm the component shows the correct display on image error:
- [ ] try a drag and drop of a non image file => should display message error
- [ ] try to type a wrong image url => should display message error
- [ ] try to type an image url with CORS error (eg. https://podaac.jpl.nasa.gov/Podaac/thumbnails/OISST_HR_NRT-GOS-L4-BLK-v2.0.jpg) => should display message error
- [ ] finally test the default behavior, upload a valid image => should upload the image

---


libs/ui/elements/src/lib/image-input/image-input.component.html
```diff
@@ -39,19 +43,21 @@
         'border-primary-lighter bg-primary-white': dragFilesOver,
         'border-gray-300': !dragFilesOver,
         'cursor-pointer hover:border-gray-500':
-          !isUploadInProgress && !uploadError && !showUrlInput && !disabled,
+          !isUploadInProgress && !imageFileError && !showUrlInput && !disabled,
         'cursor-not-allowed': disabled,
       }"
       [attr.tabindex]="
-        isUploadInProgress || uploadError || showUrlInput || disabled ? null : 0
+        isUploadInProgress || imageFileError || showUrlInput || disabled
```
cmoinier, line 50: Since this line and line 46 above check for the same variables, maybe that could be put in a getter to avoid two long lines in the template?

cmoinier (changes requested):
When testing I did find an issue though. All of the tests you suggested work fine separately but when switching from one case to the other, it doesn't work : 

1) Directly try with a faulty URL (CORS error or just a wrong one) --> error message displayed OK
2) Try to click on "retry" --> shortly displays the "Select an image" message, and then switches back to the error message, leaving no time to upload.
3) Try to drag and drop on top of error message --> KO, not allowed to. Although for this point, not sure we should allow it, WDYT? When I do the drag and drop first and get the error message, I can still try again to drag and drop, so we should pick a common behavior for both.

cmoinier:
> > 1. Directly try with a faulty URL (CORS error or just a wrong one) --> error message displayed OK
> > 2. Try to click on "retry" --> shortly displays the "Select an image" message, and then switches back to the error message, leaving no time to upload.
> 
> The retry button will try to upload the given url again.
> 
> > 3. Try to drag and drop on top of error message --> KO, not allowed to. Although for this point, not sure we should allow it, WDYT? But when I do the drag and drop first and get the error message, I can still try again to drag and drop, so we should pick a common behavior for both.
> 
> This is the purpose of #1178.

Okay I see. But what if the user decides that they want to not try again with the url, but to upload the file themselves instead? (Because maybe they noticed their file isn't at the link they want to use anymore & they have it locally) They would be stuck :/  

Also, while checking again, I noticed that if after a fail to upload a link, i remove the link totally, the "retry" stays on even though there is no URL left --> same thing, what if the user wants to move on? I think there should be a reset of some kind at some point.

LHBruneton-C2C:
Quickly reading you exchanges, here is how it was first designed:

- the dashed zone always remains a drag&drop zone, and clicking on it will open the file selector, even in an error state
- the error state replaces the initial state, but it's only to give the user feedback, all the same input interactions (d&d, select, url) remain valid

If too complicated, the retry option can be left out, or only displayed for URL input.

AlitaBernachot:
@cmoinier, @LHBruneton-C2C i have updated the PR (rebase and rework Retry), could you please check again?

Clicking on retry now resets the component to its default ui and opens the file prompt. I find this behavior a bit weird after trying to upload via an invalid url (but maybe this is acceptable?). Tell me what you think.

libs/ui/elements/src/lib/image-input/image-input.component.html
AlitaBernachot: ok done

LHBruneton-C2C:
For further review purpose, it's been decided to remove the "retry" logic for now, as it's not as essential as the error feedback, and not clear enough it the current state.

The focus remains on informing the user that something went wrong, and allowing all further input interactions.

AlitaBernachot:
Updated: no more Retry button.

Guillaume-d-o (approved):
It's seems good without the retry button. it tried several senarios and the behaviour is good. All is ok for me thanks for this PR !

PR #1190, opened 2025-03-26
[Editor]: Clear input on protocol switch but don't eat up letters

cmoinier:
### Description

This PR makes sure the `url-input` of the "Link to a service" section allows to : 

- [x] Type an url without it being eaten up (cf https://github.com/geonetwork/geonetwork-ui/pull/1176)
- [x] URL being reset when switching protocols when a valid URL was entered (was a regression from aforementioned PR)
- [x] URL being reset when switching protocols even when a non-valid URL was entered (eg google.com instead of https://google.com) --> was not the case before

### Architectural changes

none

### Screenshots

no UI changes

### How to test

- [ ] Try typing (no copy pasting because it was never an issue) a URL starting with https:// --> should not add a random slash or eat up any letter
- [ ] Try switching protocols --> should clear the input
- [ ] Try the same but starting with any text that's not https:// --> should not eat anything nor clear the input

### DIFF
--- a/libs/feature/editor/src/lib/components/online-service-resource-input/online-service-resource-input.component.html
+++ b/libs/feature/editor/src/lib/components/online-service-resource-input/online-service-resource-input.component.html
@@ -36,6 +36,7 @@
     [disabled]="disabled"
     [value]="_service.url?.toString()"
     [showValidateButton]="activeLayerSuggestion"
+    [resetUrl]="resetUrl"
   >
     <ng-content *ngIf="activeLayerSuggestion">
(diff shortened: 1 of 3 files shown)


cmoinier:
> Thanks for the quick fix! This is a bit disturbing this downward communication to the component, but I couldn't find anything better in a short amount of time with this context (OnPush, no empty string for a valid URL). I just renamed and retyped the reset parameter, it was looking like a parameter that would allow or not a behavior, when in fact it's mostly used as an event.

Ok thanks, will merge like that.

PR #1193, opened 2025-03-27
Section subsection rework

Guillaume-d-o:
### Description

This PR introduces section and subsection rework to be able to navigate throw a top menu. 
First of all we decided to use the Anchor component to make the menu. The anchor component / Directive use the id of the HTML object to link the element.

In this PR i tried my best to unified the way of the section are build, there margin, padding title and style. Some feedback on mockup will be done to make it better. 

### Architectural changes

### Screenshots

[...]

### How to test

Open a dataset on the datahub view
see if it correspond to the mockup 

---


apps/datahub/src/app/record/record-metadata/record-metadata.component.spec.ts
```diff
@@ -417,4 +418,36 @@ describe('RecordMetadataComponent', () => {
...
+      const sections = fixture.debugElement.queryAll(
+        By.directive(RecordSectionComponent)
+      )
+      const sectionIds = sections.map((section) => section.componentInstance.id)
+      expect(sectionIds).toEqual([
+        'about',
+        'data-preview',
+        'resources',
+        'related',
+        'userFeedbacks',
```
tkohr, line 449: Maybe we could use the same case for all ids. WDYT?

apps/datahub/src/app/record/record-section/record-section.component.ts
```diff
@@ -0,0 +1,26 @@
...
+import { CommonModule } from '@angular/common'
+import { UiLayoutModule } from '@geonetwork-ui/ui/layout'
+
+@Component({
+  selector: 'datahub-record-section',
+  standalone: true,
+  imports: [CommonModule, UiLayoutModule],
+  template: `
+    <section [id]="id">
+      <h1
```
tkohr, line 11: If we use `<h>` tags we must be careful to not break the hierarchy (if the record title already was in h1 for example). So far we don't use <h> tags yet I think. Not sure what's the best approach here for SPAs, to stay with `<divs>` or to take care implementing `<h>` tags hierarchically.

apps/datahub/src/app/record/record-section/record-section.component.ts
```diff
@@ -0,0 +1,26 @@
...
+        *ngIf="title"
+        class="title-font text-title font-medium pt-9 text-section-title container-lg px-4 lg:mx-auto"
```
tkohr, line 13: Actually, I'm noticing that the component finally only exists because of this line. So we could combine these classes into a `gn-ui-section-title` class in the `tailwind.base.css` and then use this class instead of this component. In this case we won't need the `text-section-title` anymore neither as it will only be used once. This would be much simpler in my eyes. WDYT? Sorry if I didn't point it out earlier, but I thought there were still other reasons for this component.

tkohr (commented):
Thanks for the PR @Guillaume-d-o, looks mostly clean to me! I've left some comments inline.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -1,112 +1,130 @@
...
-          [metadata]="metadataViewFacade.metadata$ | async"
-        >
-        </gn-ui-metadata-contact>
-        <div *ngIf="metadataQualityDisplay">
-          <p class="text text-gray-700 text-xs mb-3 uppercase" translate>
-            record.metadata.quality
-          </p>
-          <gn-ui-metadata-quality
+  <datahub-record-section id="about">
+    <div class="container-lg px-4 mb-9 mt-10 mb-16 pb-2 lg:mx-auto">
```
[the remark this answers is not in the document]
Guillaume-d-o: yes, i forgot the mb-9 only mb-16 must be kept

apps/datahub/src/app/record/record-metadata/record-metadata.component.spec.ts
Guillaume-d-o: I agree, and i convert it into kebab case

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -1,112 +1,130 @@
...
-    *ngIf="displayDownload$ | async"
+  </datahub-record-section>
+
+  <datahub-record-section id="data-preview">
+    <datahub-record-data-preview></datahub-record-data-preview>
+  </datahub-record-section>
+
+  <datahub-record-section
+    id="resources"
+    title="Ressources et liens"
```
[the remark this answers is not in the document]
Guillaume-d-o: i make the trad into the component as i think we have to keep it ( could be change after)

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -1,112 +1,130 @@
...
     <gn-ui-error
       data-test="dataset-has-no-link-block"
       [type]="errorTypes.DATASET_HAS_NO_LINK"
     ></gn-ui-error>
   </ng-container>
-  <div id="related-records" *ngIf="displayRelated$ | async">
+
+  <datahub-record-section
+    id="related"
+    title="Contenus associés"
```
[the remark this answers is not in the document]
Guillaume-d-o: i make the trad into the component as i think we have to keep it ( could be change after)

apps/datahub/src/app/record/record-section/record-section.component.ts
Guillaume-d-o: changed !

apps/datahub/src/app/record/record-section/record-section.component.ts
tkohr: Ok, I think we were just not precise enough when grooming the ticket and a "section component" could also mean a dedicated CSS class here. I still feel this would be the lighter solution, and the component feels like some overhead, but I let you decide. We can also bring this up in tomorrow's daily to get a decisive opinion on it.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -59,7 +59,7 @@
 
   <datahub-record-section
     id="resources"
-    title="Ressources et liens"
+    title="ressources.and.links"
```
tkohr, line 62: I think keys similar to the ones already present in the component (like `record.metadata.resources` for example) would be better following our current structure of translations keys.

Also, they should be present in all translation files. You can run `npm run i18n:extract` for that. Just be careful, as sometimes there have been unrelated changes in the translation files when running this script in the recent pas, I think.

AlitaBernachot:
image

AlitaBernachot:
Thx for the work! i did not check the code as @tkohr already did the code review, i left some comments and images above regading the ui.
Also section titles are usually wrapped with h1, h2, h3, h4, ... or span for accessibility, seo purposes and semantic html meaning (instead of divs and p). But it looks like we don't really care about this in gnui 😅.

apps/datahub/src/app/record/record-section/record-section.component.ts
Guillaume-d-o: i've done it, i use a style on tailwind and there is no datahub section component anymore

Guillaume-d-o:
> Preview and tab alignement is broken. image
> 
> Looks like some i18n are missing image
> 
> Missing seperation (but maybe this will be fixed with the dots, with the M/S/XS PR) image

=> Good catch , i make a mistake by reducing the title translation. I make the translte like tobias advise me 

The lack of space is normal, the space is the size of the dot pagination instroduce on the other PR ! that' why the other are ok and not the download !

tkohr (approved):
Thanks for the adaptions @Guillaume-d-o ! I left two small final comments you may fix before merging.

translations/it.json
```diff
@@ -531,7 +533,6 @@
   "tooltip.url.open": "Aprire l'URL",
   "ui.readLess": "Ridurre",
   "ui.readMore": "Leggere di più",
-  "wfs.aggregations.notsupported": "Aggregazioni non supportate per i servizi WFS",
```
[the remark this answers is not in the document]
Guillaume-d-o: oupsi i18n / chatGPT mistake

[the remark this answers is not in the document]
tkohr: Ah, you are right https://github.com/geonetwork/geonetwork-ui/pull/1158 recently made the message unnecessary, all good then:-)

PR #1195, opened 2025-03-27
[Datahub]: Error message when restricted access in preview

cmoinier:
### Description

This PR checks for every link used in preview if there is a restricted access link in the description object returned by elasticSearch. If there is, a restricted message is displayed & the layer is not loaded.

### Architectural changes

Added "restrictedAccess" to the mapper, which takes the link from the description object and checks if it constitutes a restriction.

### Screenshots

image

### How to test

Add the following tag to any existing record, within a CI_OnlineResource tag. Save and check on the record that the message is displayed for the online resource you modified.

`<gmd:description>
                <gmx:Anchor xlink:href="http://standards.iso.org/iso/19139/resources/gmxCodelists.xml#MD_RestrictionCode_restricted">Service à accès restreint</gmx:Anchor>
</gmd:description>`


libs/feature/dataviz/src/lib/table-view/table-view.component.ts
```diff
@@ -46,6 +46,10 @@ export class TableViewComponent {
     switchMap((link) => {
       this.error = null
       if (!link) return of(undefined)
+      if (link.applicationProfile) {
```
jahow, line 49: I feel like this should be checked earlier on. See other comment

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -485,6 +489,7 @@ export class Gn4FieldMapper {
           type,
           url: url,
           accessServiceProtocol,
+          applicationProfile: applicationProfile,
```
jahow, line 492: Instead of simply passing the `applicationProfile` value to the record, I think we can do more in the mapper. For instance, we could add an `accessRestricted` field in the model and set it to `true` if we find something in the applicationProfile telling us it's the case.

jahow (commented):
Hey! Took a look at the implementation and I left a comment regarding the way the application profile value is processed. Hope it's useful!

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
cmoinier: That's a good idea! That would be less checks. I'll keep that in mind and will make the changes once it's decided what needs to be parsed from applicationProfile :)

Guillaume-d-o:
Thanks for this PR. I tried to import gmx and gmd but i had error at the save ( there is a CI online resource). what is the exact way to add it ? 

The code seems good but i think it need a test to test that it is not diplayed.

cmoinier:
> Thanks for this PR. I tried to import gmx and gmd but i had error at the save ( there is a CI online resource). what is the exact way to add it ?
> 
> The code seems good but i think it need a test to test that it is not diplayed.

Which version of GN are you on? Also, can you provide a screenshot of the place were you added the tag?

Guillaume-d-o:
> > Thanks for this PR. I tried to import gmx and gmd but i had error at the save ( there is a CI online resource). what is the exact way to add it ?
> > The code seems good but i think it need a test to test that it is not displayed.
> 
> Which version of GN are you on? Also, can you provide a screenshot of the place were you added the tag?

I use the docker composition of the GN, and i edit in gn with XML : 
in the head gmd and gmx must be added i think ? 
image

So i try to add it : 
image

and after i put the XML you gave : 
image


and the data is visible

cmoinier:
> I use the docker composition of the GN, and i edit in gn with XML : in the head gmd and gmx must be added i think ? 

Nothing needs to be added to the head! Maybe my explanation was not clear enough, but it's only a tag inside of one of the CI_OnlineResource :)

>
> So i try to add it : 
> 
> and after i put the XML you gave : 

Not sure you added that in the correct section, because all links have a "name" tag and I don't see one.

> and the data is visible : 

Well first of all, the data displayed in your screenshot is from a link named "Service OGC API Feature", so it's not the link in which you added the XML. With the description tag we added, the name should be read from the free text part, that being "Service à accès restreint". So it's pretty normal that the data is displayed :smile: 

Look at this example, I'm using the SCOT record : 

So the existing link is here, as you can see, with a description and a name (all of the links should have both).
image

We are replacing the description tag with ours :

image

And now if we go on the dataset, and select the right link, the error message is being displayed : 

image

cmoinier:
@Guillaume-d-o Double reading your answer, I see in your screenshot that your record doesn't use gmd tags at all... All of my records do, I've never seen them with "mdb". why is that? did you use a specific record from a different format?

Guillaume-d-o:
behaviour confirmed, i used the wrong place and the wrong tag. Thanks @cmoinier 

image


As i ask in the first comment, Is it possible to add a unit test ? it is perfect for the production code for me, Thanks !

Guillaume-d-o (changes requested):
Unit test if possible

cmoinier:
> Unit test if possible

Sure, done for all 3 components.

Guillaume-d-o (approved):
unit test validated

PR #1198, opened 2025-03-31
Datahub: Add quick access menu

tkohr:
### Description

This PR introduces a new quick access "navigation bar" that scrolls down to the section of a dataset, service or reuse page. 

Note: So far the nav bar has been tested on the dataset page. Once the service and reuse page are present/merged we must ensure that the anchor ids in the page and in the nav bar "config" are the same.

### Screenshots

Desktop
image

Mobile
image

### How to test

Go to a dataset page.

To display the language switcher and test the outcome in the same colors as in the mockups, the following values can be set in the `default.toml`:

```
languages = ['en', 'fr', 'de']

primary_color = "#90C149"
secondary_color = "#F18345"
```

---


tailwind.base.config.js
```diff
@@ -69,6 +69,7 @@ module.exports = {
       },
       boxShadow: {
         'xl-hover': '0 20px 25px -5px rgb(0 0 0 / 0.2)',
+        navbar: '0px 0px 5.1px 0px var(--color-primary-darkest)',
```
jahow, line 72: This is not appropriate I think; why not specify the box shadow in the HTML?

tailwind.base.css
```diff
@@ -157,6 +161,11 @@
     @apply font-medium text-gray-500 text-ellipsis break-words text-sm;
   }
 
+  /* TODO: rework centering in gn-ui-card-icon and maybe merge both classes into one */
+  .gn-ui-nav-bar-icon {
+    @apply gn-ui-card-icon justify-center items-center h-8 min-w-10 hover:bg-primary;
+  }
```
jahow, line 167: this shouldn't be a new global class; global classes are for very generic elements like buttons, cards, input fields etc.

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
```diff
@@ -10,10 +10,4 @@
     (newValue)="toggleFavorite($event)"
     [disabled]="loading || (isAnonymous$ | async)"
   ></gn-ui-star-toggle>
-  <span
-    class="align-text-top ml-1.5"
-    data-test="favorite-count"
-    *ngIf="!displayCount"
-    translate="datahub.record.addToFavorites"
```
jahow, line 17: Maybe this translation key could be kept as a tooltip? It is a useful indication IMO (and probably required in terms of accessibility)

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
```diff
@@ -23,26 +42,42 @@
...
-    <button
-      type="button"
-      class="h-10 inline-flex items-center justify-center p-2 rounded-md shrink-0 sm:hidden"
-    >
-      <ng-icon class="align-middle" name="matExpandMoreOutline"></ng-icon>
-    </button>
+  </div>
+  <div
+    class="h-[60px] flex flex-row gap-3 min-w-40 justify-center items-center lg:min-w-60 lg:pl-4 lg:justify-start"
+    [class]="displayMobileMenu ? 'border-l-2 border-gray-300' : ''"
```
jahow, line 63: Does it work with `[class]`? I thought it needed ngClass:

```suggestion
    [ngClass]="displayMobileMenu ? 'border-l-2 border-gray-300' : ''"
```

apps/datahub/src/app/app.module.ts
```diff
@@ -181,6 +180,7 @@ export const metaReducers: MetaReducer[] = !environment.production ? [] : []
     }),
     OrganisationsComponent,
     LanguageSwitcherComponent,
+    NavigationBarComponent,
```
jahow, line 183: Is that really needed? this component is part of the same app

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
```diff
@@ -1,20 +1,39 @@
...
+    displayMobileMenu
+      ? 'items-start'
+      : 'items-center divide-x-2 divide-gray-300'
+  "
 >
+  <div
+    class="h-[60px] flex min-w-[60px] items-center justify-center lg:min-w-40 lg:justify-end lg:pr-4"
+    [class]="displayMobileMenu ? 'border-r-2 border-gray-300' : ''"
+  >
+    <gn-ui-navigation-button (click)="back()" [icon]="'matArrowBack'">
```
jahow, line 13: can't we try using a generic gn-ui-button here? the concept of a "navigation button" is a bit awkward IMO, it's just a button right? plus icons should be properly centered in gn-ui-buttons

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
```diff
@@ -5,12 +5,15 @@
...
+      disabled: !toggled,
       'transition hover:scale-125 will-change-transform': !disabled,
       'cursor-default': disabled,
     }"
   >
-    <ng-icon [name]="toggled ? 'matStar' : 'matStarBorder'"></ng-icon>
+    <ng-icon
+      [name]="toggled ? 'matStar' : 'matStarBorder'"
+      size="24px"
+    ></ng-icon>
```
jahow, line 16: isn't that going to change the appearance in the search results as well?

libs/ui/layout/src/lib/anchor-link/anchor-link.directive.ts
```diff
@@ -11,20 +12,29 @@ import {
...
   selector: '[gnUiAnchorLink]',
 })
-export class AnchorLinkDirective implements OnInit, OnDestroy {
+export class AnchorLinkDirective
+  implements OnInit, AfterViewChecked, OnDestroy
+{
   @Input('gnUiAnchorLink') targetId: string
   @Input('gnUiAnchorLinkDisabledClass') disabledClass: string
   @Input('gnUiAnchorLinkEnabledClass') enabledClass: string
+  @Input('gnUiAnchorLinkInViewClass') inViewClass: string
```
jahow, line 21: Remember to include this new input in the associated storybook entry :)

libs/ui/inputs/src/lib/navigation-button/navigation-button.component.ts
```diff
@@ -6,7 +7,7 @@ import { NgIcon } from '@ng-icons/core'
   templateUrl: './navigation-button.component.html',
   styleUrls: ['./navigation-button.component.css'],
   changeDetection: ChangeDetectionStrategy.OnPush,
-  imports: [NgIcon],
+  imports: [CommonModule, NgIcon],
   standalone: true,
 })
 export class NavigationButtonComponent {
```
jahow, line 13: as mentioned above, I think this component should go; everything should use the same buttons

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
```diff
@@ -5,12 +5,15 @@
     (click)="toggle($event)"
     [ngClass]="{
       enabled: toggled,
-      'disabled opacity-45': !toggled,
```
jahow, line 8: this is also going to have impacts on other places right?

jahow (commented):
Thank you, the new functionality of the anchor link directive looks good! I added several components because I feel that layout concerns in the Datahub are leaking all the way through the project, which shouldn't be the case IMO.

tailwind.base.config.js
tkohr: Ok, wasn't sure this was possible. What's the syntax for this?

tailwind.base.css
tkohr: Ok I'll try to use the `gn-ui-button` for this. I guess the `favorite-star` component can be wrapped into it.

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
tkohr: Yep, double-checked and it works:-)

apps/datahub/src/app/app.module.ts
tkohr: Ah right, it's not needed, this was a "make it standalone reflex".

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
tkohr: Right, I'll try to get rid of it.

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
tkohr: It's actually the size they are already displayed in, in the search, but for some reason they shrank in the nav bar.

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
tkohr: Yes, do you see another way to make the star black in the nav bar when untoggled to be consistent with the other buttons?

tkohr:
Thanks for the review @jahow, I'll see what I can do tomorrow.

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
jahow: maybe because the size of the star was proportional to the text size, which is really handy for UI components; fixing a size in px here means it's going to be impossible to have a star of a different size somewhere else

tailwind.base.config.js
jahow: not sure but if needed, you can just use CSS instead of tailwind classes

tkohr: Hum, ok I thought it would be preferable to have a custom tailwind class than CSS in our components. Does that mean that in the long run we should get rid of all the custom classes in the tailwind config?

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
tkohr: Ok I'll see if I get it sized via the text size

tailwind.base.config.js
jahow: In my opinion yes, global classes are technical debt unless they adress standard components like cards, buttons, icons, inputs...

jahow: Using tailwind classes or css in the component is honestly almost the same, sure we try to have as few CSS as possible in the project but it's preferable to have local CSS than global classes

tkohr: Ok I used tailwind interpolation

tailwind.base.css
tkohr: I refactored the toggle-star component to use the `gn-ui-button` and thus become "themable"

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.html
tkohr: It was actually not displayed anymore before this PR, so I've left it out for now. Displaying it as a simple tooltip could create a conflict with the "login message tooltip", displaying it in a tippy tooltip might be too intrusive and not necessarily improve accessibility.

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
tkohr: The component has now been removed.

libs/ui/layout/src/lib/anchor-link/anchor-link.directive.ts
tkohr: Good point, I've added it.

libs/ui/inputs/src/lib/navigation-button/navigation-button.component.ts
tkohr: It's gone now:-)

libs/ui/inputs/src/lib/star-toggle/star-toggle.component.html
tkohr: Yes it also impacts the display in the results list, which actually fits the new mockups. So I also changed the color from primary to secondary there. Will have to check if this is inline with the current work on the cards.

tkohr:
Thanks again for your feedback @jahow ! I've addressed your remarks and rebased the PR on main with the merged [section work](https://github.com/geonetwork/geonetwork-ui/pull/1193). It's now ready for a second review.

LHBruneton-C2C:
Thanks for this feature!

I noticed when preparing the demo that the menu is larger than the page content, unlike the mockups.

image

compared to

image

This should be discussed during the demo, because the default page content width is really smaller than in the mockups.

tkohr:
Ah, good point @LHBruneton-C2C, I completely missed that. I'm also noticing that the menu labels are not centered in the mockups.

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
```diff
@@ -1,48 +1,111 @@
...
     </div>
     <button
-      class="block px-3 py-3 rounded-md font-medium uppercase tracking-wider shrink truncate sm:hidden"
-      [class]="displayMobileMenu ? 'hidden' : 'block'"
+      type="button"
+      *ngFor="let l of anchorLinks"
+      [attr.data-cy]="l.anchor"
+      gnUiAnchorLink="{{ l.anchor }}"
+      gnUiAnchorLinkDisabledClass="sm:hidden"
+      gnUiAnchorLinkInViewClass="!border-b-primary border-b-4"
```
jahow, line 38: the bottom padding should be changed here as well, to avoid having the text "jump"

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
```diff
@@ -1,48 +1,111 @@
...
+        [style.--star-toggle-disabled-color]="black"
+        [style.--gn-ui-button-rounded]="'8px'"
+        [style.--gn-ui-button-height]="'32px'"
+        [style.--gn-ui-button-width]="'40px'"
+        [style.--gn-ui-button-color]="black"
+        [style.--gn-ui-button-background]="'var(--color-primary-lightest)'"
+        [style.--gn-ui-button-bg-hover]="'var(--color-primary)'"
+        [style.--gn-ui-button-border]="'var(--color-gray-300)'"
+        [style.--gn-ui-button-font-size]="'24px'"
+        [style.--gn-ui-button-padding]="'4px'"
```
jahow, line 99: FYI for later, css variables can be set on the host in the CSS file :) less verbose in HTML

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.html
tkohr: Yes, good point, it's fixed now.

tkohr:
PR is now ready from my side. There was still a mean bug hiding content in the top after using the nav-bar on mobile and scrolling back up. The nav-bar is now really `sticky` instead of `fixed` to solve this.

The latest comments from the PR and the demo have been addressed.

The background and hover colors have been adapted. There was actually a confusion between `bg-primary-lightest` and `bg-primary-white`. Hope it looks better now.

The menu title now updates on mobile when scrolling, as I finally found a solution with little evolution on the directive.

cmoinier (approved):
Thank you for this cool feature! Works great :)

PR #1201, opened 2025-03-31
Add the latest tag without publishing again on release

LHBruneton-C2C:
With the current workflows, NPM packages are published twice, on tag and on release, and the second time fails due to the version already used. The package is thus correctly published when a tag is created, but never set to latest.

For the fix, we keep the publishing on all github events that are not a release, and on a release we just add the latest tag to the correct version.


Guillaume-d-o (approved):

tkohr (approved):
Thanks @LHBruneton-C2C, the idea LGTM!

PR #1203, opened 2025-04-01
New truncate text component

Guillaume-d-o:
### Description

This PR introduces [...]

### Architectural changes

### Screenshots

[...]

### How to test

---


libs/ui/elements/src/lib/truncated-text/truncated-text.component.ts
```diff
@@ -0,0 +1,38 @@
...
+export class TruncatedTextComponent implements AfterViewInit {
+  @Input() text = ''
+  @ViewChild('textElement') textElement: ElementRef<HTMLElement>
+  isTextTruncated = false
+
+  constructor(private cd: ChangeDetectorRef) {}
+
+  ngAfterViewInit() {
+    this.checkTextTruncation()
+  }
```
jahow, line 31: could you make it so that this processing happens every time the text changes? so instead of using `afterViewInit`, use a setter?

libs/ui/elements/src/lib/truncated-text/truncated-text.component.html
```diff
@@ -0,0 +1,23 @@
+<div class="flex items-center justify-between w-full gap-2">
+  <span
+    #textElement
+    class="truncate"
+    [class.cursor-pointer]="isTextTruncated"
+    [matTooltip]="isTextTruncated ? text : ''"
```
jahow, line 6: I'm not sure about the use of `matTooltip` here, why not a simple `title`?

libs/ui/elements/src/lib/truncated-text/truncated-text.component.html
```diff
@@ -0,0 +1,23 @@
...
+  >
+    {{ text }}
+  </span>
+  <button
+    *ngIf="isTextTruncated"
+    [matTooltip]="text"
+    matTooltipPosition="above"
+    class="flex-shrink-0 border border-gray-300 px-2 py-1 rounded-lg w-11 h-9"
+  >
```
jahow, line 15: This button is strange: it does not do anything on click. Looking at the mockup, I feel like the popover with the text should appear when clicking the button. Also, the popover looks like an actual block of text, and not just a `matTooltip`.

jahow: And if the button shows the block of text on click, its tooltip should be something like "show all text".

jahow (commented):
Thank you :) This component shouldn't go in `ui/elements` but in `ui/layout` (see [this section of the docs](https://geonetwork.github.io/geonetwork-ui/main/docs/developers/architecture-overview.html#details)).

libs/ui/elements/src/lib/truncated-text/truncated-text.component.ts
```diff
@@ -0,0 +1,38 @@
...
+})
+export class TruncatedTextComponent implements AfterViewInit {
+  @Input() text = ''
+  @ViewChild('textElement') textElement: ElementRef<HTMLElement>
+  isTextTruncated = false
+
+  constructor(private cd: ChangeDetectorRef) {}
+
+  ngAfterViewInit() {
+    this.checkTextTruncation()
```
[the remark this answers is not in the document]
jahow: I agree with the robot :slightly_smiling_face:  :point_up:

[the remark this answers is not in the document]
Guillaume-d-o: I added a ngOnchange with the call to the function

libs/ui/elements/src/lib/truncated-text/truncated-text.component.ts
Guillaume-d-o: textElement was not define at the setter so it cause issues to display the button after

libs/ui/elements/src/lib/truncated-text/truncated-text.component.html
```diff
@@ -0,0 +1,23 @@
...
+    <ng-icon
+      class="align-middle text-secondary block"
+      name="iconoirExpand"
+      size="24"
+    >
+    </ng-icon>
+  </button>
```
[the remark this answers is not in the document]
Guillaume-d-o: done !

translations/en.json
```diff
@@ -547,6 +548,7 @@
   "service.metadata.spatialExtent": "Spatial extent",
   "share.tab.permalink": "Share",
   "share.tab.webComponent": "Integrate",
+  "show.all.text": "Show all text",
```
jahow, line 551: these keys are not very clear, how about:

* `ui.showAllText`
* `ui.hideFullText`

This is consistent with `ui.readMore` and `ui.readLess`

libs/ui/layout/src/lib/truncated-text/truncated-text.component.stories.ts
```diff
@@ -0,0 +1,66 @@
+import { Meta, StoryObj, applicationConfig } from '@storybook/angular'
+import { TruncatedTextComponent } from './truncated-text.component'
+import { TranslateModule } from '@ngx-translate/core'
+import { importProvidersFrom } from '@angular/core'
+
+const meta: Meta<TruncatedTextComponent> = {
+  component: TruncatedTextComponent,
+  title: 'Elements/TruncatedText',
```
jahow, line 8:
```suggestion
  title: 'Layout/TruncatedText',
```

libs/ui/layout/src/lib/truncated-text/truncated-text.component.html
```diff
@@ -0,0 +1,36 @@
+<div class="relative">
+  <div class="flex items-center justify-between w-full gap-2">
+    <span #textElement class="truncate">
+      {{ text }}
+    </span>
+    <gn-ui-button
+      *ngIf="isTextTruncated"
+      [matTooltip]="'show.all.text' | translate"
+      matTooltipPosition="above"
```
jahow, line 9: Minor but I wouldn't use matTooltip here; simple `title` attributes are fine and accessible

libs/ui/layout/src/lib/truncated-text/truncated-text.component.html
```diff
@@ -0,0 +1,36 @@
...
+      (buttonClick)="toggleExpand()"
+    >
+      <ng-icon name="iconoirExpand" size="24"> </ng-icon>
+    </gn-ui-button>
+  </div>
+
+  <!-- Expanded overlay -->
+  <div
+    *ngIf="isExpanded"
+    class="absolute left-0 top-0 bg-white shadow-lg border border-gray-300 z-10"
```
jahow, line 21: So I realize now that the positioning of the overlay is only done using `position: absolute`. This can give unexpected results and issues (scrolling, overlay not visible etc.) and typically when we want to show an overlay we use a lib like `tippy.js`  to make sure the overlay is attached to the body and not deep in the dom. Such a lib will also take care of positioning for you, which is a rabbit hole on its own.

Could you try getting it to work with the `gn-ui-popover` element? that would be ideal.

jahow (commented):
Hi! me again with my annoying comments :wink: 

I tried the component in Storybook and noticed a couple of differences with the mockup.

Mockup:

| closed | opened |
|----|----|
| image | image |

Storybook:

| closed | opened |
|----|----|
| image | image |

The issue is mainly margin/padding I believe, and also the fact that the overlay is not covering the parent element exactly.

Please refer to [this documentation](https://geonetwork.github.io/geonetwork-ui/main/docs/developers/styling-guide.html#css-variables) and the `gn-ui-button` Storybook entry to see how to customize gn-ui-button margins. Thanks!

libs/ui/layout/src/lib/truncated-text/truncated-text.component.html
fgravin: I realized while doing https://geocat-dev.dev.bgdi.ch/swissgeo-poc/search
That there is a glitch if the background is not `white`
Could this be handled easily ?

Guillaume-d-o: @jahow I agree it's not perfect but tippy is not corresponding to the mokup, I 'd rather rollback my last dev to use the tooltip on the div / button. So it could not work well with gn-ui-popover

@fgravin i take a look on the gn-ui-popover storybook and the background is always black also, so i think the component need an improvement to be updated / corrected

Guillaume-d-o: I use overlay on the advise of @LHBruneton-C2C and it will work !

cmoinier (changes requested):
Thank you for this useful and nice new component ! :slightly_smiling_face: 

In a more general way, I think that this feature could be only 1 component, if you follow what is done in the editor, for instance in `all-records` component, there is 1 component for whatever is opened when the button is clicked (in your case here that would be `truncated-text`, and what you do here in `overlay` would be handled directly wherever the button will be added, so no need for `overlay`.

cmoinier:
Also, from what I understand in the mockups, the overlay should be completly over the cell in which we want to see the text, whereas your component opens it underneath : 

image

Guillaume-d-o:
I completely agree with your comments, i merge the overlay into the truncate component. I also change the anchor of the overlay to the top left of the div of the truncate component to be as the mockup as possible

cmoinier (changes requested):
Thank you for the changes!! It looks much simpler & better already !

Just a couple of things & then it'll be perfect : 

- [ ] I still feel like this is different from the mockups. Looking at them (see screenshot under), it looks like the icon to close the overlay should be directly over the icon to open it, making the overlay's borders perfectly on top of the cell's border. Right now there is still some "décalage" between the cell and the overlay. Do you think you could improve that?

--> Mockups 

image

---> PR : 

image

- [ ] Also, maybe it's my screen, but the button inside of the overlay seems to have a background, and it shouldn't : 

image

cmoinier (approved):
Thank you for the adaptations, it looks really neat now!

PR #1204, opened 2025-04-01
Standalone Search

jahow:
### Description

This PR introduces the Standalone Search system, similar to Web Components:
* a `gn-standalone-search.js` file can be included in a web page
* this gives access to a `GNUI` global which lets us use the `recordsRepository` and `platformService` outside of an angular application

### Architectural changes

* Simplified the way the `GN_UI_VERSION` value was passed around; it's not an injection token anymore, it's simply read from the package json
* Moved some providers to the `gn4.provider` file; this is still lacking stuff, ideally it should provide everything needed to communicate with GN4 without having to import modules from libs/feature

### Screenshots

image

### How to test

Open the Web Components demo website and open the new sample to see the Standalone Search in action:

https://geonetwork.github.io/geonetwork-ui/standalone-search/demo/webcomponents/gn-standalone-search.sample.html

Locally:
1. run `npm run build:demo`
2. open your browser to `./dist/demo/webcomponents/gn-standalone-search.sample.html`

---

**This work is sponsored by Swisstopo for the Swiss Geoplatform project**.


apps/webcomponents/project.json
```diff
@@ -47,12 +47,6 @@
               "maximumError": "1mb"
             }
           ],
-          "fileReplacements": [
-            {
-              "replace": "apps/webcomponents/src/environments/environment.ts",
-              "with": "apps/webcomponents/src/environments/environment.prod.ts"
-            }
-          ],
```
tkohr, line 55: Indeed, as we do not have a real dev mode for the web components we can get rid of it entirely. Maybe something to set up in the future? Or do you think we won't be able to set up a different "dev env" than bootstrapping the web components as an app?

apps/webcomponents/src/app/standalone-search.module.ts
```diff
@@ -0,0 +1,60 @@
...
+    const recordsRepository = inject(RecordsRepositoryInterface)
+    const platformService = inject(PlatformServiceInterface)
+    window['GNUI'] = {
+      recordsRepository,
+      platformService,
+      init,
+    }
+  }
+
+  ngDoBootstrap() {
```
tkohr, line 55: Just a detail, but the class should implement `DoBootstrap`. Same for the `webcomponents.module`.

docs/guide/standalone-search.md
```diff
@@ -0,0 +1,42 @@
...
+GeoNetwork-UI offers a mechanism similar to Web Components but which only allows you to search in a GeoNetwork catalog using its search interface.
+
+The JS file to include is as a result much lighter, and this system is entirely appropriate if the search results coming from GeoNetwork should be used in a different UI framework.
+
+## Use
+
+To use the Standalone Search, you have to:
+
+- import the Standalone Search script exported by Angular (available via jsdelivr)
+- call `GNUI.init(...)` with the proper GeoNetwork API URL.
```
tkohr, line 16: Wondering if it's necessary to point out in the doc that it could potentially create a conflict with another global var?

tkohr (approved):
Thanks for this interesting feature and the rework of the version detection @jahow! Code LGTM and it works smoothly!

apps/webcomponents/project.json
jahow: I would like to do that in another PR yes

docs/guide/standalone-search.md
jahow: I don't think this is likely; on the other hand, having a conflict detection logic could make sense, I'll add that :)

jahow:
thank you for the review @tkohr :)

PR #1206, opened 2025-04-02
Display and filter feature catalogs

Guillaume-d-o:
### Description

This PR introduces the display of the Feature catalog section, the features Types nd the Feature Types attributes with a search and a count functionnality. 

### Architectural changes

Major change on the expandable panel to make the display conditionnal ( hidden before) to have good performance with large content. 

### Screenshots

image
image

### How to test

Before all : 2 Technical decision that can be challenged written in bold. 

use data mel and bd topo for example : 
https://ign-gpf-test.apps.gs-ch-prod.camptocamp.com/geonetwork/srv/fre/catalog.search#/metadata/50aa3037-ac17-4c0d-ad15-b43a5a0929ae

Use 2 differents dataset : one with only one feature type, one with multiple feature types. With one feature type, there is no search, no total number of attributes and total of object. the Feature type is displayed entirely. 

 
If there is 2 features types or more, there is the search bar which work by **using AND between each term**. 
the total number of attributes and object is synchronised with the number of elements displayed, they are shown only if the screen is **SM or bigger.**  The feature types panels are collapsed. 

---


cmoinier (changes requested):
First of all thank you very much for this big work! It's a really nice feature & it looks good already. Here are the UI points I think need fixing/improvements. I will look at the code afterwards : 

- [x] I feel like this should have a minimal height in case there are only a couple of tables : 

image

- [x] A border would be good at the bottom of the header to help distinct it from the table when scrolling down : 

image

- [x] Icon in search input should be `iconoirSearch` :

image

- [ ] The +/- should be vertically aligned & should be black, not grey : 

image

- [x] The gap between the bottom of the header and the top of the list should be bigger. I see that you applied the `py-2`, which represents the space below the purple line in the screenshot - but there is an additionnal `gap-3` in the mockups.

image

image

- [x] Abstract text size should be 14px, not 16px

- [x] Title text size should be 18px, not 20px

- [x] Missing gap of 10px between these two texts + right padding for the first number : 

image

cmoinier:
I think the border left should be removed for the left-end cells, since there is no border on right right-end cells : 

image

cmoinier:
image

Feature catalog tab is present in all records even though there is no feature catalog available. This behavior is similar to the fix in https://github.com/geonetwork/geonetwork-ui/pull/1227 --> this is because you probably didn't put the `ngIf` display condition of the feature catalog on the `section` tag, so the section technically still exists even if the child component is not displayed. Will highlight it in the code review.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
```diff
@@ -6,11 +6,23 @@ import { datasetRecordsFixture } from '@geonetwork-ui/common/fixtures'
 import { TranslateTestingModule } from 'ngx-translate-testing'
 import { TranslateMessageFormatCompiler } from 'ngx-translate-messageformat-compiler'
 import { DatasetRecord } from '@geonetwork-ui/common/domain/model/record'
+import { By } from '@angular/platform-browser'
 
 describe('MetadataInfoComponent', () => {
   let component: MetadataInfoComponent
   let fixture: ComponentFixture<MetadataInfoComponent>
 
+  const expandAllPanels = (fixture: ComponentFixture<any>) => {
```
cmoinier, line 15: I'm split on that. I understand that you need it for the your new component, but this is a component used elsewhere and having them all expanded it not the default behavior. Wasn't there any other way around?

libs/ui/elements/src/lib/simple-table/simple-table.component.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    ]
+    fixture.detectChanges()
+  })
+
+  it('should create', () => {
+    expect(component).toBeTruthy()
+  })
+
+  it('should display the correct number of columns', () => {
+    const headers = fixture.debugElement.queryAll(By.css('.font-bold'))
```
cmoinier, line 39: Same remark as above. Cf styling guide (and my opinion), you should avoid selecting DOM items by css and use `data-test`. Also, this should be an e2e test, not a UT. The same goes for the test just below.

libs/ui/elements/src/lib/simple-table/simple-table.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@Component({
+  selector: 'gn-ui-simple-table',
+  standalone: true,
+  imports: [CommonModule, TranslateModule],
+  templateUrl: './simple-table.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SimpleTableComponent implements OnInit, OnChanges {
+  @Input() data: unknown[]
```
cmoinier, line 20: Could this be typed more precisely?

libs/ui/elements/src/lib/simple-table/simple-table.component.ts
```diff
@@ -0,0 +1,39 @@
...
+  Component,
+  Input,
+  OnChanges,
+  OnInit,
+  SimpleChanges,
+} from '@angular/core'
+import { TranslateModule } from '@ngx-translate/core'
+
+@Component({
+  selector: 'gn-ui-simple-table',
```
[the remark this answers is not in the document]
cmoinier: Also - and it's just my opinion - this seems like a really simple & small component, and we have already 4 "table-like" components, now 5 adding this one. It seems like a lot, and `feature-catalog-list` which contains it is rather small too. Couldn't we move this component into it?

libs/ui/elements/src/lib/simple-table/simple-table.component.ts
```diff
@@ -0,0 +1,39 @@
...
+      .join(' ')
+  }
+
+  ngOnInit(): void {
+    this.updateGridTemplate()
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes['columns']) {
+      this.updateGridTemplate()
```
cmoinier, line 36: Why would the columns change, given that the input here is hard-coded ? 

image

cmoinier: Also, cf comments above, merging this component into its parent would make it more smooth, I think

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.ts
```diff
@@ -18,20 +22,58 @@ import { matAdd, matRemove } from '@ng-icons/material-icons/baseline'
   imports: [CommonModule, NgIcon],
   viewProviders: [provideIcons({ matAdd, matRemove })],
 })
-export class ExpandablePanelComponent {
-  @Input() title: string
-  @Input() collapsed = true
+export class ExpandablePanelComponent implements AfterViewChecked {
+  @Input() title?: string
+  @Input() iconColor? = 'var(--color-primary)' // Default color is 'inherit'
```
cmoinier, line 27: Well now it's not `inherit` anymore if it's set by default to `primary` ? Also, same remark as before, I think you can just use `text-primary` here, since this variable is already in the config, but I could be wrong, to test.

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.html
```diff
@@ -1,22 +1,30 @@
...
+    <ng-container *ngIf="titleTemplate; else simpleTitle">
+      <ng-container [ngTemplateOutlet]="titleTemplate"></ng-container>
+    </ng-container>
+    <ng-template #simpleTitle>
+      <div class="font-medium text-black text-sm">{{ title }}</div>
+    </ng-template>
   </div>
   <div
-    class="w-8 opacity-25 text-primary group-hover:opacity-100 transition-opacity"
+    class="w-8 opacity-25 text-primary group-hover:opacity-100 transition-opacity flex items-center shrink-0"
```
cmoinier, line 14: So, I don't know which of these changes had an impact on the expandable panels in the "About" section, but you have to be careful to other implementations (which are not always in the datahub, but might also be in custom apps like MEL, as they are exported) when making changes to `ui` components. Storybook should always be checked to make sure the components stay the exact same and only change in your implementation :wink: 

In this case, I noticed that the panels were true to the mockups before : 

image

But are not anymore : 

image

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.spec.ts
```diff
@@ -39,29 +71,36 @@ describe('ExpandablePanelComponent', () => {
     })
   })
 
+  describe('simple title rendering', () => {
+    it('shows simple title when no template is provided', () => {
+      component.title = 'Simple Title'
+      fixture.detectChanges()
+      const titleElement = fixture.debugElement.query(
+        By.css('.font-medium.text-sm')
```
cmoinier, line 79: Same remark about css selector. Actually you know, it can be used if you have a specifc class (like `icon-expand` for an icon, or `table-cell` if you wanna select multiple table cells at the same time), but a text size is not specific enough

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.ts
```diff
@@ -18,20 +22,58 @@ import { matAdd, matRemove } from '@ng-icons/material-icons/baseline'
...
+    if (value !== this._collapsed) {
+      if (value) {
+        this._collapsed = value
+        setTimeout(() => {
+          // Closing: wait for animation before hiding content
+          this.showContent = false
+          this.changeDetector.detectChanges()
+        }, this.ANIMATION_DURATION)
+      } else {
+        // Opening: show content immediately then animate
```
cmoinier, line 48: Reminder to remove comments :smiley:

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.ts
```diff
@@ -18,20 +22,58 @@ import { matAdd, matRemove } from '@ng-icons/material-icons/baseline'
...
-  @Input() collapsed = true
+export class ExpandablePanelComponent implements AfterViewChecked {
+  @Input() title?: string
+  @Input() iconColor? = 'var(--color-primary)' // Default color is 'inherit'
+  @ContentChild('titleTemplate') titleTemplate?: TemplateRef<HTMLElement>
   @ViewChild('contentDiv') contentDiv: ElementRef
-  maxHeight = this.setMaxHeight()
+  maxHeight = '0px'
+  showContent = false
+  private readonly ANIMATION_DURATION = 300
```
cmoinier, line 32: I don't think you need a variable for this if it's used only once :)

translations/en.json
```diff
@@ -358,6 +362,9 @@
   "record.action.view": "View",
   "record.card.metadata.contact": "Metadata Contact",
   "record.externalViewer.open": "Open in the external map viewer",
+  "record.metadata.feature.catalog": "Feature catalog",
+  "record.feature.catalog.number.total.attribute": "Total number of objects",
```
cmoinier, line 366:
```suggestion
  "record.feature.catalog.number.total.attribute": "Total amount of objects",
```

translations/en.json
```diff
@@ -358,6 +362,9 @@
...
+  "record.feature.catalog.number.total.object": "Total number of attributes",
```
cmoinier, line 367:
```suggestion
  "record.feature.catalog.number.total.object": "Total amount of attributes",
```

cmoinier (changes requested):
Code review done ! I have left some suggestions, let me know if something is not clear, I'm available :slightly_smiling_face: 

I think some e2e testing is needed, it would be great to add some!

Also, on the architecture level, I feel like `feature-catalog-list` and `search-feature-catalog` do not belong in `apps` (since they don't need any facade like the other components do. I think the first should be moved to `ui/elements` just like the `service-capabilities`, and the search to `ui/inputs`. This will also create the opportunity for some stories of `feature-catalog-list`, which I think could be used :smiley:

apps/datahub/src/app/record/record-feature-catalog/feature-catalog-list/feature-catalog-list.component.html
```diff
@@ -0,0 +1,44 @@
+<div class="grid gap-2 py-2 px-5 max-h-[550px] overflow-y-auto">
+  <div
+    *ngFor="let featureType of filteredFeatureCatalog?.featureTypes"
+    class="rounded shadow bg-white"
+  >
+    <gn-ui-expandable-panel
+      [collapsed]="filteredFeatureCatalog?.featureTypes?.length !== 1"
+      iconColor="var(--color-main)"
```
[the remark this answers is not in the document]
Guillaume-d-o: as it is black now do you think it is applicable ?

libs/ui/elements/src/lib/simple-table/simple-table.component.ts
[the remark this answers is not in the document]
Guillaume-d-o: In a way I agree because we have a lot of table but they are used in specific context. But i think we could simplify all this table. I will move th etable and see if i can reuse that to simply other components. But i'm forced to have this simple table in a way. I think it has sens to have this simple component that is very reusable. I i merge it on feature-catalog-list, it will be a component dropdown table that use a specific object, no reusable. IMO I think simple table must have it's own component, it's only my opinion ^^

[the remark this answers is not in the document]
cmoinier: I see what you mean, and it's fair enough - keep it then, hopefully we will reuse it somewhere else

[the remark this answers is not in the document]
jahow: Hi, I'd like to weigh in on this topic because I agree with @cmoinier, a simple table layout can be done using CSS and adding a new component seems like an unjustified complexity cost.

> I think it has sens to have this simple component that is very reusable.

Remember about [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it). We should _not_ assume that something will be reused in the future.

apps/datahub/src/app/record/record-feature-catalog/feature-catalog-list/feature-catalog-list.component.html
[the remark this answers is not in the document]
cmoinier: Yes I think it's acceptable now :)

apps/datahub/src/app/record/record-feature-catalog/record-feature-catalog.component.ts
```diff
@@ -0,0 +1,42 @@
...
+})
+export class RecordFeatureCatalogComponent implements OnInit {
+  @Input() metadata: CatalogRecord
+  featureCatalog$: Observable<DatasetFeatureCatalog>
+  filteredFeatureCatalog: DatasetFeatureCatalog
+
+  constructor(private readonly metadataViewFacade: MdViewFacade) {}
+
+  ngOnInit(): void {
+    this.metadataViewFacade.loadFeatureCatalog(this.metadata)
```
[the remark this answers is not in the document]
Guillaume-d-o: Good catch !

apps/datahub/src/app/record/record-feature-catalog/search-feature-catalog/search-feature-catalog.component.html
```diff
@@ -0,0 +1,34 @@
...
+      class="w-[220px] h-[32px] pl-3 pr-10 py-2 border rounded-md"
+      [(ngModel)]="searchTerm"
+      (ngModelChange)="onSearchChange($event)"
+    />
+    <ng-icon
+      class="absolute right-2.5 top-1 w-6 h-6 text-black"
+      name="matSearchOutline"
+    ></ng-icon>
+  </div>
+  <div class="text-right text-sm ui-med ml-auto hidden sm:inline">
```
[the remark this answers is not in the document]
Guillaume-d-o: wrong style report of the font ^^ => deleted

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -83,6 +85,16 @@ export class RecordMetadataComponent {
     map((records) => records?.length > 0)
   )
 
+  displayFeatureCatalog$ = combineLatest([
+    this.metadataViewFacade.metadata$,
+    this.metadataViewFacade.featureCatalog$,
+  ]).pipe(
+    map(
+      ([usedForRefresh, featureCatalog]) =>
```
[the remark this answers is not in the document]
Guillaume-d-o: We cannot load feature catalog without having the metadata loaded, becauses feature catalog can be embedded. This cause a synchronisation issue because if you load feature catalog alone, you don't have any metadata information, so you won't find a feature catalog. To make it clean as possible i decided to use a suscribe on metadata to refresh featureCatalog informations

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
Guillaume-d-o: I understand, i can expand the good panel before the test that need the content information

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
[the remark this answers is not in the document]
cmoinier: Okok, your explanation makes sense. Then I would rename to `usedForRefresh` to `metadata` for more clarity & add a comment like `subscribing to metadata to refresh featureCatalog information` :)

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
Guillaume-d-o: i make some modification, I propose a new way to test by openning the panel using the data test, and then we test the logic of display ( there is logic for the content inside the panel for some, and there is a logic of display or not the panel for other ) . I cannot do this by e2e test because i don't find a perfect dataset for this. But there is already e2e tes about the display or not for some datasets. 

So i let you decide if it is good for you or not :)

libs/ui/inputs/src/lib/search-feature-catalog/search-feature-catalog.component.ts
```diff
@@ -0,0 +1,76 @@
...
+          const searchedTerm = term.toLowerCase()
+          return (
+            featureType.name?.toLowerCase().includes(searchedTerm) ||
+            featureType.definition?.toLowerCase().includes(searchedTerm)
+          )
+        })
+      }),
+    }
+    this.filteredFeatureCatalog = filteredCatalog
+    this.filteredFeatureCatalogChange.emit(filteredCatalog)
```
jahow, line 74: There's a utility called FuzzyFilter that should be used for this kind of client-side search, see: https://github.com/geonetwork/geonetwork-ui/blob/main/libs/util/shared/src/lib/utils/fuzzy-filter.ts

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.ts
```diff
@@ -18,20 +22,55 @@ import { matAdd, matRemove } from '@ng-icons/material-icons/baseline'
...
+        this.showContent = true
+        this._collapsed = value
+      }
+      setTimeout(() => {
+        this.updateMaxHeight()
+        this.changeDetector.detectChanges()
+      })
+    }
+  }
+  get collapsed(): boolean {
```
jahow, line 55: I think you'd be better off using a ResizeObserver here, and recompute the maxheight whenever the content changes; this logic seems unreliable

support-services/docker-entrypoint-initdb.d/dump.sql
```diff
@@ -6409,6 +6395,12 @@ system/site/name	0	y	n	n	110	Test GeoNetwork-UI instance
 system/users/identicon	0	y	n	n	9110	gravatar:monsterid
 system/feedback/mailServer/host	0	y	n	y	630	this.will.not.work.smtp
 system/userFeedback/lastNotificationDate	0	n	n	y	1912	2024-06-14T06:26:33.6Z
+metadata/url/sitemapDoiFirst	2	y	n	y	9166	false
+metadata/url/dynamicAppLinkUrl	0	y	n	y	9167	\N
+metadata/publication/profilePublishMetadata	0	y	n	n	12021	Reviewer
+metadata/publication/profileUnpublishMetadata	0	y	n	n	12022	Reviewer
+system/platform/version	0	y	n	n	150	4.2.5
```
jahow, line 6402: this change shows that the database dump has been migrated to 4.2.5; it should stay in 4.2.2 because that is the first version the project supports.

To do that, make sure you're running GN in version 4.2.2 when updating the dump  :)

libs/ui/elements/src/lib/record-feature-catalog/feature-catalog-list/feature-catalog-list.component.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  })
+
+  it('should update grid template when columns change', () => {
+    component.columns = [
+      { key: 'type', label: 'Type', width: '50%' },
+      { key: 'name', label: 'Name', width: '50%' },
+    ]
+    component.ngOnInit()
+    fixture.detectChanges()
+    expect(component.gridTemplateColumns).toBe('50% 50%')
```
jahow, line 76: This is a comment from someone that hasn't followed everything: I find it hard to understand why the component supports updating the grid columns while `columns` is not declared as an input and does not seem to be used as such?

Guillaume-d-o:
I make all correction @cmoinier . the only one i can't is "The +/- should be vertically aligned & should be black " because it is already black , and we display the color without transparency only at the hoover. So to make the buton always black we must change the behaviour of the panel which is not conistent with the rest of the dropdown in the app

libs/ui/elements/src/lib/record-feature-catalog/feature-catalog-list/feature-catalog-list.component.spec.ts
Guillaume-d-o: You are right, It's because i merge the "simple table components " tests into the feature catalog list and i don't pay attention on this one, that has no sens in this context

support-services/docker-entrypoint-initdb.d/dump.sql
Guillaume-d-o: Ok ok, I don't pay attention that the GN version is 4.2.5 in the support service , did i pass the version of GN in .env to 4.2.2 ?

libs/ui/layout/src/lib/expandable-panel/expandable-panel.component.ts
Guillaume-d-o: Done without timeout

jahow: Thank you!

libs/ui/inputs/src/lib/search-feature-catalog/search-feature-catalog.component.ts
Guillaume-d-o: Done :)

AlitaBernachot:
- [ ] Here the title is not centered vertically

image
image

AlitaBernachot:
- [ ] Should this stays sticky like this?
image
image

AlitaBernachot (approved):
Thanks for the work! this is a big functionality!

Did not check the code as there are already some reviews, I just did some manual testings:
- the opened cells stay sticky, so ithey stay fixed even when scrolling down, i don't know if this is the right behavior
- the filtering is very long on the `exemple_iso19110_bduni_bdtopo_readyforGNimport.xml`, I don't know if this can be improved?

Approving the PR because other than these minor things, it works well.

libs/ui/elements/src/lib/record-feature-catalog/feature-catalog-list/feature-catalog-list.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    TruncatedTextComponent,
+  ],
+})
+export class FeatureCatalogListComponent implements OnInit {
+  @Input() filteredFeatureCatalog: DatasetFeatureCatalog
+
+  columns = [
+    {
+      key: 'type',
+      label: 'feature.catalog.attribute.type',
```
AlitaBernachot, line 27: @Guillaume-d-o Don't forget to use markers when translations are dynamic or not explicitly a translation key; like this case here, otherwise translations will disappear when running the i18n utility (there are missing translation on main currently).

AlitaBernachot:
Hello @Guillaume-d-o , we need to do the same overlay system for the new columns Values, but while testing, I still see fixed overlay on main, is it normal?
image

PR #1226, opened 2025-04-23
[Datahub] : Feature Catalog - User friendly labels

cmoinier:
### Description

This PR implements user friendly labels from feature catalog attributes, in the map feature details, the table columns and the charts dropdowns. It will attempt to match the feature's attribute "technical name" with the featureCatalog attribute name, and if a match exists, the featureCatalog "code" will be displayed instead of the feature's attribute "technical name". If there is no match (or no featureCatalog attributes), the technical name is kept.

### Architectural changes

none

### Screenshots

image

image

### How to test

I have created a custom record to test it - it's hacky so I won't add it to the dump. Contact me to get the XML :)

### DIFF
--- a/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
+++ b/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
@@ -2,9 +2,7 @@ import {
   ChangeDetectionStrategy,
   ChangeDetectorRef,
   Component,
-  Inject,
   Input,
-  Optional,
   Output,
 } from '@angular/core'
 import { marker } from '@biesbjerg/ngx-translate-extract-marker'
@@ -13,6 +11,7 @@ import {
   FetchError,
   FieldAggregation,
   getJsonDataItemsProxy,
+  PropertyInfo,
 } from '@geonetwork-ui/data-fetcher'
 import {
   DropdownChoice,
@@ -30,7 +29,10 @@ import {
 } from 'rxjs/operators'
 import { DataService } from '../service/data.service'
 import { InputChartType } from '@geonetwork-ui/common/domain/model/dataviz/dataviz-configuration.model'
-import { DatasetOnlineResource } from '@geonetwork-ui/common/domain/model/record'
+import {
+  DatasetFeatureCatalog,
+  DatasetOnlineResource,
+} from '@geonetwork-ui/common/domain/model/record'
 import { TranslateModule, TranslateService } from '@ngx-translate/core'
 import { CommonModule } from '@angular/common'
 import { ChartComponent } from '@geonetwork-ui/ui/dataviz'
@@ -67,6 +69,12 @@ marker('chart.aggregation.count')
   standalone: true,
 })
 export class ChartViewComponent {
+  public featureCatalog$ = new BehaviorSubject<DatasetFeatureCatalog | null>(
+    null
+  )
+  @Input() set featureCatalog(value: DatasetFeatureCatalog) {
+    this.featureCatalog$.next(value)
+  }
   @Input() cacheActive = true
   @Input() set link(value: DatasetOnlineResource) {
     this.currentLink$.next(value)
@@ -150,13 +158,8 @@ export class ChartViewComponent {
     }),
     shareReplay(1)
   )
-  properties$ = this.dataset$.pipe(
-    switchMap((dataset) =>
-      dataset.properties.catch((error) => {
-        this.handleError(error)
-        return []
-      })
-    ),
+  properties$ = combineLatest([this.dataset$, this.featureCatalog$]).pipe(
+    switchMap(([dataset, catalog]) => this.setProperties(dataset, catalog)),
     shareReplay(1)
   )
   yChoices$ = this.properties$.pipe(
@@ -235,6 +238,28 @@ export class ChartViewComponent {
     private translateService: TranslateService
   ) {}
 
+  setProperties(
+    dataset: BaseReader,
+    catalog: DatasetFeatureCatalog
+  ): Promise<PropertyInfo[]> {
+    return dataset.properties
+      .then((properties) => {
+        const featureAttributes = catalog?.featureTypes[0]?.attributes ?? []
+        return properties.map((p) => {
+          const matchingAttribute = featureAttributes.find(
+            (attr) => attr.name === p.label
+          )
+          if (matchingAttribute?.code) {
+            return { ...p, label: matchingAttribute.code }
+          }
+          return p
+        })
+      })
+      .catch((error) => {
+        this.handleError(error)
+        return []
+      })
+  }
   handleError(error: FetchError | Error | string) {
     if (error instanceof FetchError) {
       this.error = this.translateService.instant(
--- a/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
+++ b/libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
@@ -224,6 +224,18 @@ export class ChartViewComponent {
     shareReplay(1)
   )
 
+  prettyLabel$ = combineLatest([
+    this.aggregation$,
+    this.properties$,
+    this.yProperty$,
+  ]).pipe(
+    map(([aggregation, properties, yProperty]) => {
+      if (aggregation === 'count') return 'count()'
+      const prop = properties.find((p) => p.name === yProperty)
+      return prop ? `${aggregation}(${prop.label})` : ''
+    })
+  )
+
   get labelProperty() {
     if (!this.xProperty$.value) return ''
     return `distinct(${this.xProperty$.value})`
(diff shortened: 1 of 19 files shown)


libs/ui/dataviz/src/lib/data-table/data-table.component.ts
```diff
@@ -92,6 +93,7 @@ export class DataTableComponent implements OnInit, AfterViewInit, OnChanges {
 
   ngOnInit() {
     this.dataSource = new DataTableDataSource()
+    this.setProperties()
```
AlitaBernachot, line 96: won't this `setProperties()` happen twice? first one is triggered by the setter.

libs/ui/dataviz/src/lib/data-table/data-table.component.ts
```diff
@@ -61,14 +61,15 @@ export interface TableItemModel {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class DataTableComponent implements OnInit, AfterViewInit, OnChanges {
+  _featureAttributes = []
+  @Input() set featureAttributes(value: []) {
```
AlitaBernachot, line 65: could be nice to have a better typing

libs/ui/dataviz/src/lib/data-table/data-table.component.ts
```diff
@@ -115,6 +117,27 @@ export class DataTableComponent implements OnInit, AfterViewInit, OnChanges {
     this.readData()
   }
 
+  setProperties() {
+    this.dataset_.properties.then((properties) => {
```
AlitaBernachot, line 121: Not sure if the data-table component is really in charge of this label/code extracting process.

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -235,6 +238,28 @@ export class ChartViewComponent {
...
+          const matchingAttribute = featureAttributes.find(
+            (attr) => attr.name === p.label
+          )
+          if (matchingAttribute?.code) {
+            return { ...p, label: matchingAttribute.code }
+          }
+          return p
+        })
+      })
+      .catch((error) => {
```
AlitaBernachot, line 258: I don't know if this is due to these new changes but the chart is broken:

ok in int
image

ko in dev
image

AlitaBernachot (approved):
Ok working fine for map.

For table i can see the Code is selected and the label is set as a fallback when no code exists. But the table content is empty (there is a warning). When i switch back to main branch, the table is displayed correctly. => i don't know if it's due to the added feature catalog?

For the chart, the Y axis is missing but also on main, this is weird, i guess this is not due to your code modifications.

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
AlitaBernachot: Looks like I have the same result when switching to main.

libs/ui/map/src/lib/components/feature-detail/feature-detail.component.ts
```diff
@@ -14,11 +15,34 @@ const geometryKeys = ['geometry', 'the_geom']
...
+        if (this._featureAttributes.length) {
+          const matchingAttribute = this._featureAttributes.find(
+            (attr) => attr.name === p
+          )
+
+          if (matchingAttribute && matchingAttribute.code) {
+            key = matchingAttribute.code
+          }
+        }
+        acc[key] = this.feature.properties[p]
```
AlitaBernachot, line 42: Looks like you have done the code extraction many times, can't this be refactor (if the modification is simple)?

libs/ui/dataviz/src/lib/data-table/data-table.component.ts
cmoinier: Indeed... The one in the setter is from a commit to fix the tests, so I guess it should be the right one. Will remove this one and try it out, thanks for pointing this out!

libs/ui/dataviz/src/lib/data-table/data-table.component.ts
cmoinier: I can definitely move it to the `table-view` component :)

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
cmoinier: Oh right, I didn't even notice that this axis was down. Since this is also the case on main, it could be related to the recent layout changes (since we see it with the "old" layout on `int`. I will investigate because this should not be put on prod.

libs/ui/map/src/lib/components/feature-detail/feature-detail.component.ts
cmoinier: I would love to, but unfortunately the process is a bit different in all 3 cases, so I don't really see a way to harmonize it all...

cmoinier: In the chart, we need the output to be a pair of value/label, in the data-table, a simple string, and here, we need to do key/value matching. I could put in common the matching part, but that would just be one line... I don't know if that's enough

cmoinier:
@AlitaBernachot With the latest commits, I made the following changes : 

- [ ] Fixed the behavior of the data-table --> I removed the part where the aggregation was set to `count` if there was no `newProp` because it was not waiting long enough for the `featureCatalog` attribute to load, and therefore hiding the Y axis as if there were no values in it (which is wrong). It felt okay to removed that because we are actually already checking for values in `yProperty` through `aggregationChoices` and `isCountAggregation`, so the check I removed was kind of creating a loop anyway.

AlitaBernachot:
With dataset: n_tri_lill_inondable_s_059 (updated with the xml you gave to me)
i have an error in console and the legend does not display. The legend is now displayed only when clicking on a feature.
image

libs/feature/dataviz/src/lib/table-view/table-view.component.ts
```diff
@@ -105,4 +103,23 @@ export class TableViewComponent {
...
+        if (attributes.length) {
+          const matchingAttribute = attributes.find(
+            (attr) => attr.name === p.name
+          )
+          if (matchingAttribute && matchingAttribute.code) {
+            return { value: p.name, label: matchingAttribute.code }
+          }
+          return { value: p.name, label: p.name }
+        }
+        return { value: p.name, label: p.name }
```
AlitaBernachot, line 120: Minor: l. 118 and l.120 are the same? can't this be refacto?

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -78,6 +78,9 @@ export class ChartViewComponent {
   @Input() cacheActive = true
   @Input() set link(value: DatasetOnlineResource) {
     this.currentLink$.next(value)
+    if (value) {
+      this.aggregation$.next('sum')
```
AlitaBernachot, line 82: For my own understanding, why do we need to set aggr to 'sum' here?

cmoinier: That was the whole issue with the graph, even below this PR's scope (which was emphazising it) : whenever aggregation is "count", then the Y dropdown isn't shown. And the aggregation is set to count when there are no suitable choices in the Y dropdown... As you can tell, this is a loop. So when the first source link was not producing an error and not populating the dropdown options OR actually not having Y choices (which is valid!) --> when this happened, the Y dropdown was hidden, which is fine. The issue is that, without initializing the aggregation when switching source, then the YChoices observable was not being triggered/subscribed to because the aggregation variable, which wasn't re-initialized, was hiding it (ngIf in the template) .

libs/feature/dataviz/src/lib/chart-view/chart-view.component.ts
```diff
@@ -171,6 +174,9 @@ export class ChartViewComponent {
     tap((choices) => {
       if (!choices.find((choice) => choice.value === this.yProperty$.value)) {
         const newProp = choices[0]?.value || ''
+        if (!newProp && this.aggregation$.value !== 'count') {
+          this.aggregation$.next('count')
```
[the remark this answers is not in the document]
cmoinier: This is actually a revert of a previous commit, so this line is from the origin code :)
But I think this is because when there are no Y choices then you cannot really do any other aggregations than doing sums of the X axis data

AlitaBernachot:
I can see the table view but i cannot see the chart view: it looks like it does not refresh the chart with the WFS datasource ans stays on the csv source. When i test on main, the chart is displayed (with a broken displayed but it shows).
image

cmoinier:
> With dataset: n_tri_lill_inondable_s_059 (updated with the xml you gave to me) i have an error in console and the legend does not display. The legend is now displayed only when clicking on a feature. image
> 
> EDIT: I can reproduce on main, this is weird...

Yep, tested on the record I took this service from and I reproduce both this & the bug you reported on the graph, so I think this is the service's fault. I will create another record with a service & feature catalog that matches so you can properly test.

AlitaBernachot:
Tested with the new record you gave me => all good, thx for the work!

I don't know if it is also worth updating the label in the chart? Since it was not mentioned in the PR, I let you decide :)
image

PR #1228, opened 2025-04-24
[Editor]: Multilingual side panel

cmoinier:
### Description

This PR introduces a new multilingual side panel in the editor form.
This ticket only includes the creation of the side panel and its opening, not the listing of the languages which will be done in a follow-up ticket. At the moment, the language switch only passes a boolean but does nothing else - it will also be in the next ticket.

### Architectural changes

none

### Screenshots

image

### How to test

Open and close the panel in any record form.


apps/metadata-editor-e2e/src/e2e/edit/form-display.cy.ts
```diff
@@ -69,4 +69,10 @@ describe('editor form', () => {
         .should('contain', 'Cette couche de points reprend les informations')
     })
   })
+  describe('Multilingual panel', () => {
+    it('opens the multilingual panel', () => {
+      cy.get('md-editor-top-toolbar').find('gn-ui-button').eq(1).click()
```
AlitaBernachot, line 74: maybe test if its not visible before clicking?

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
```diff
@@ -61,8 +61,15 @@
...
   >
     <ng-icon name="matHelpOutlineOutline"></ng-icon>
-  </gn-ui-button>
+  </gn-ui-button> -->
   <gn-ui-button
+    [extraClass]="translatePanelOpen ? 'bg-neutral-200 border-0' : ''"
+    (buttonClick)="handleTranslatePanel()"
+    type="light"
+  >
+    <ng-icon name="iconoirTranslate"></ng-icon>
```
AlitaBernachot, line 70: Not very important: it would be great if the icon button has a label/title/tootltip saying this would open the translation panel for accessibility?

AlitaBernachot (approved):
Thx, working great!

I juste noticed two minor things (i don't know if this is related to your work though):

image

When the panel is open on small screen, there is a minimal width added to the viewport and we need to scroll right to go see the i18n panel. But maybe this one can be ignored, i don't know if we have rules for small screen sizes on the editor.

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
cmoinier: Yes, will add it :)

cmoinier:
@AlitaBernachot Changes done! Could you give me your opinion on the changes regarding the white space issue you mentionned ? The resolution is a bit different than what the mockups say (but they're not dynamic though), but I actually feel like this is an improvement because IMO it looked weird to scroll all the way down in the form and "loose" the left menu at some point. Now the scrolling is inside of the form making all menus (left, right, and top toolbar)  visible at all times, and I think it makes more sense. Let me know ! :)

AlitaBernachot:
> @AlitaBernachot Changes done! Could you give me your opinion on the changes regarding the white space issue you mentionned ? The resolution is a bit different than what the mockups say (but they're not showing how the scroll is supposed to work though), but I actually feel like this is an improvement because IMO it looked weird to scroll all the way down in the form and "loose" the left menu and top toolbar at some point. Now the scrolling is inside of the form making all menus (left, right, and top toolbar) visible at all times, and I think it makes more sense. Let me know ! :)

That is pretty good! thx for the improvement.

PR #1229, opened 2025-04-28
[Editor]: Add and remove languages to record

cmoinier:
### Description

This PR introduces several new features to handle multilingualism in the editor : 

- [x] Switch a record to multilingual / switch it back to monolingual
- [x] Select new languages for the record
- [x] Remove languages from the record
- [x] Add all GN-supported languages to the constants & translations
- [x] Delete languages from the three dots btn

#### What is visible but not implemented (followup tickets):

- [ ] Radio buttons for selected languages
- [ ] Switch default language in three dots btn menu

### Architectural changes

The library https://github.com/lipis/flag-icons has been added to display the languages flags.
Note that some flags have a different country code than the ISO codes, so a mapping is done for a few exceptions.

### Screenshots

image

image

### How to test

You will need to check in the `buckets` response directly since the changes are only done through the API, with nothing visible in the editor form yet (will be another ticket). 

In a mono-lingual record (you will know it's monolingual if you open the multilingual panel and the switch is not on) : 

- [ ] Open the translations panel & switch the record to multilingual --> should display a list of currently supported languages (which should be only 1 item, the language of the record you opened), the language should be set as "default"
- [ ] Click on the wheel icon --> should display a complete list of the supported languages, with the current language already selected (and disabled since it's not possible to delete a default language).
- [ ] Select a few more languages, click on "validate" and save the record -->  in the elasticsearch response, for the multilingual fields, you should have a key for each selected language & have the new languages in the `otherLanguages` list / the panel should also display a list of languages with radio buttons (those buttons don't do anything yet, will be another ticket).
- [ ] Remove languages, click on "validate" --> should open a validation menu --> if confirmed, should update the list of languages & the elasticsearch response should have removed the keys from the multilingual fields & the languages from the `otherLanguages` list // if not confirmed, should select again the languages (to match which languages are in the record)
- [ ] Switch off the multilingual feature --> should open a validation menu --> if confirmed, should not display anything in the panel except for the switch, elasticsearch should return an empty list for `otherLanguages` & no key besides "default" and the default language key should be present in the multilingual fields.


libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
```diff
@@ -1,14 +1,93 @@
...
+      <div class="flex flex-row justify-between border-t border-gray-300 p-3">
+        <span class="mt-2 text-sm text-gray-600" translate
+          >editor.record.form.multilingual.activate</span
+        >
+        <gn-ui-button
+          extraClass="w-16 h-8 font-bold"
+          type="gray"
+          (buttonClick)="validateTranslations()"
+          data-test="validateSelection"
+          >{{ 'editor.record.form.multilingual.validate' | translate }}
```
AlitaBernachot, line 35: When I click on validate with only the default language selected, it deactivates the multilingual mode.

image

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
```diff
@@ -1,14 +1,93 @@
...
+    <span class="text-3xl font-title text-black/80" translate
+      >editor.record.form.multilingual.title</span
+    >
+    <button
+      [title]="'editor.record.form.multilingual.open' | translate"
+      (click)="activateLanguageSelection()"
+      *ngIf="isMultilingual"
+      data-test="activateSelection"
+    >
+      <ng-icon class="mt-1" name="iconoirSettings"></ng-icon>
```
AlitaBernachot, line 14: I don't know if this button state should be done in this PR or not?
image

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
```diff
@@ -1,14 +1,93 @@
...
+        <ng-icon
+          *ngIf="recordLang !== formLanguage"
+          class="text-gray-800 mt-1"
+          name="iconoirCircle"
+        ></ng-icon>
+        <span [class]="getIconClass(recordLang) + 'mt-1'"></span>
+        <span [ngClass]="recordLang === formLanguage ? 'text-black' : ''">{{
+          'language.' + recordLang | translate
+        }}</span>
+      </div>
```
AlitaBernachot, line 79: Selection is not like in the mockups:

yours 
image

mockups
image

AlitaBernachot: I am wondering what type of flag is this? looks like Argentina.

image

cmoinier: Oh wow, yes it is Argentina, that's a funny mistake! Will switch it to a more fitting one :laughing:

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
```diff
@@ -1,14 +1,93 @@
...
+    <button
+      [title]="'editor.record.form.multilingual.open' | translate"
+      (click)="activateLanguageSelection()"
+      *ngIf="isMultilingual"
+      data-test="activateSelection"
+    >
+      <ng-icon class="mt-1" name="iconoirSettings"></ng-icon>
+    </button>
+  </div>
+  <div class="flex flex-col gap-2" *ngIf="editTranslations || !isMultilingual">
```
AlitaBernachot, line 17: From the mockups and the ticket description, here what i understood: when the user first opens the i18n panel and activate the multilingual mode, the language list should be displayed, and not just the default one.

translations/fr.json
```diff
@@ -238,6 +242,10 @@
   "editor.record.loadError.title": "Erreur lors du chargement",
   "editor.record.lock.format": "Les jeux de données de type 'service' et 'reuse' ne sont pas éditables",
   "editor.record.lock.reason": "Vous n'êtes pas éditeur d'un des groupes autorisés.",
+  "editor.record.multilingual.confirmation.cancelText": "Garer le(s) language(s)",
```
AlitaBernachot, line 245:
```suggestion
  "editor.record.multilingual.confirmation.cancelText": "Garder le(s) language(s)",
```

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -1,15 +1,196 @@
...
+    return languages
+      .map((lang) => ({
+        lang,
+        label: this.translateService.instant('language.' + lang),
+      }))
+      .sort((a, b) => a.label.localeCompare(b.label))
+      .map((item) => item.lang)
+  }
+
+  activateLanguageSelection() {
```
AlitaBernachot, line 89: activate and deactivate? "toggle" maybe?

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
AlitaBernachot: image

Not sure if this is the Swedish flag

libs/util/i18n/src/lib/i18n.constants.ts
```diff
@@ -19,12 +19,44 @@ marker('language.pt')
...
+  'ru',
+  'zh',
+  'sk',
+  'rm',
+  'ar',
+  'da',
+  'no',
+  'pl',
+  'sv',
+  'tr',
```
AlitaBernachot, line 57: That is cool, but why this specific list? 😄
i don't know if this list should be configurable

AlitaBernachot (approved):
Nice work, thx!

There is just a small misunderstanding about when the user first activate the multilingual mode, i think the current workflow is not exactly what is describe in the ticket. But this should be quick to fix for you!

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
cmoinier: That's a behavior I didn't try to test, thanks for noticing it! Will fix that.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
cmoinier: Good catch! Didn't see that blue dot.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
cmoinier: Not either! I guess I shouldn't have put so much trust in this lib :smile:

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
cmoinier: Oh okay! Reading it again I guess it makes more sense than just displaying a list of 1 language anyway. Thanks for pointing it out :)

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -1,15 +1,196 @@
...
+      this.selectedLanguages = this.selectedLanguages.filter(
+        (language) => language !== lang
+      )
+    } else {
+      this.selectedLanguages.push(lang)
+    }
+  }
+
+  validateTranslations() {
+    if (this.selectedLanguages.length < this.recordLanguages.length) {
```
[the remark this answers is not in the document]
cmoinier: Also a behavior I didn't try, thanks!

libs/util/i18n/src/lib/i18n.constants.ts
cmoinier: Haha, it's from https://ign-gpf-test.apps.gs-ch-prod.camptocamp.com/geonetwork/srv/eng/admin.console#/settings/languages , which I guess is the full list of GN supported languages.
I took it from the local instance though, and I see that the IGN one has a few more - wondering if I should add those languages as well? Maybe there is a way to get the list from the API, which would be safer. I will look into it.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
AlitaBernachot: Oopsy this is describe in ticket no. 252. I hope you haven't already done it. Sorry about that.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.html
cmoinier: But at the same time, a record with only 1 language is considered mono-lingual, so shouldn't we deactivate the mulitlingual mode in this case?

cmoinier:
@AlitaBernachot As per my last commits, I've : 

- [ ] Added the blue dot on the `top-toolbar` button
- [ ] Added the languages supported by IGN (https://ign-gpf-test.apps.gs-ch-prod.camptocamp.com/geonetwork/srv/eng/admin.console#/settings/languages)
- [ ] The panel now gets the languages from the GN config and not from the hard-coded values
- [ ] Changed the wrong flags to the right ones
- [ ] Fixed the panels behaviors 

However I did not fix the fact that the record "turns back" to monolingual if you validate with only the default language selected because a record with no `otherLanguages` is automatically mono-lingual, so I think this is the right behavior :slightly_smiling_face:

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
```diff
@@ -68,7 +68,16 @@
     (buttonClick)="toggleTranslatePanel()"
     type="light"
   >
-    <ng-icon name="iconoirTranslate"></ng-icon>
+    <div style="position: relative; display: inline-block">
```
AlitaBernachot, line 71: any reason why you did not use tailwind classes here? (same for the circle icon below)

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -465,6 +471,17 @@ export class Gn4Repository implements RecordsRepositoryInterface {
     )
   }
 
+  getApplicationLanguages(): Observable<string[]> {
+    return this.gn4LanguagesApi.getApplicationLanguages().pipe(
+      map(
+        (languages) =>
+          languages
+            .map((lang) => LANG_3_TO_2_MAPPER[lang.id ?? ''])
+            .filter((code): code is string => !!code) // pour exclure undefined si id est inconnu
```
AlitaBernachot, line 480: chatgpt comment?

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -47,6 +48,10 @@ import {
 import { catchError, map, tap } from 'rxjs/operators'
 import { lt } from 'semver'
 import { ElasticsearchService } from './elasticsearch'
+import {
+  LANG_2_TO_3_MAPPER,
```
AlitaBernachot, line 52: LANG_2_TO_3_MAPPER seams unused?

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -13,20 +13,27 @@ import {
...
 const extraFlagMap: { [key: string]: string } = {
+  ar: 'arab',
   en: 'gb',
   ko: 'kr',
   cs: 'cz',
   zh: 'cn',
   ca: 'es-ct',
   rm: 'ch',
   da: 'dk',
+  sv: 'se',
```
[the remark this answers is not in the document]
cmoinier: Mmmh, well so far it's not used anywhere else and I don't really see where else we shall need it (given that the language switcher in DH doesn't have flags, and the multilingual fields in the editor won't as well). So I would say that this should be done in the future if we need the flags somewhere else.

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
```diff
@@ -64,11 +64,20 @@
...
+      <ng-icon
+        size="8px"
+        name="matCircle"
+        class="text-primary"
+        style="position: absolute; top: 0; right: 0"
+        *ngIf="translatePanelOpen"
```
AlitaBernachot, line 77: The dot should display if the metadata is multilanguage, not if the panel is open

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
cmoinier: For this div, that is because when I struggle with finding proper tailwind classes I revert back to `style` to be able to quickly copy/paste classes from the dev tools - and obviously I forgot to switch back to tailwind here!

For the circle icon, I think it was because the tailwind classes were not applied but I will double check.

apps/metadata-editor/src/app/edit/components/top-toolbar/top-toolbar.component.html
cmoinier: Oh okay! Misunderstood that, thanks!

AlitaBernachot:
Thx for the changes, working good.
I think the dot should display if the metadata is multiling and not if the panel is open though.

I have one question, i did not test the behavior but, what about a metadata with a language that is not configured in the GN config, will it appear it the panel?

cmoinier:
> Thx for the changes, working good. I think the dot should display if the metadata is multiling and not if the panel is open though.
> 
> I have one question, i did not test the behavior but, what about a metadata with a language that is not configured in the GN config, will it appear it the panel?

Fixed all of the above mentionned, thanks!I will merge once the e2e tests pass.

For your question, I did have the case at first, and it will display the language as long as the language key is an ISO one (because from what I understand GN is capable of handling all ISO languages). It will just have a broken translation key but it should work : 

image

But at the same time, I can't really be added all of ISO languages keys, right?

LHBruneton-C2C:
> Do you think that on the opposite, we shouldn't be displaying a language that is not in the GN config? That's easy to do on my end but for the edition part it might be more complex.

On the functional side, I don't see any reason to display a language that's not been configured to be supported (as mentioned in GSGNEDIT-248, supported languages are part of the configuration).
A record with more languages than supported MUST NOT loose the unsupported translations.

Why would it be more complex on the edition part?

cmoinier:
> > Do you think that on the opposite, we shouldn't be displaying a language that is not in the GN config? That's easy to do on my end but for the edition part it might be more complex.
> 
> On the functional side, I don't see any reason to display a language that's not been configured to be supported (as mentioned in GSGNEDIT-248, supported languages are part of the configuration). A record with more languages than supported MUST NOT loose the unsupported translations.
> 
> Why would it be more complex on the edition part?

Okay thanks, so I will hide the languages that are not in the GN config and make sure they are not all erased when saving the new languages.

Well I just didn't read how we will be handling the multilingual fields yet, that's all :smile: It's just an assumption since on this PR it will just be a filter.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -1,15 +1,211 @@
...
+      }))
+      .sort((a, b) => a.label.localeCompare(b.label))
+      .map((item) => item.lang)
+  }
+
+  toggleLanguageSelection() {
+    this.editTranslations = !this.editTranslations
+  }
+
+  getIconClass(lang) {
```
AlitaBernachot, line 104: it would be great to add typing here

cmoinier:
With the last commits, I have made the following changes : 

- [ ] Unsupported languages (based off the gn config), are now listed in the multilingual panel and are deletable & assignable as the default language of the record. They are NOT in the list of selectable languages, since they are not supported.
- [ ] Languages can now be removed from their own three dot button in the main menu. They can also be deleted by clicking on their button in the supported languages list.

apps/metadata-editor/src/index.html
```diff
@@ -7,6 +7,10 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="assets/favicon.ico" />
     <link rel="stylesheet" href="assets/css/materials-symbols-outline.css" />
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/gh/lipis/flag-icons@7.2.3/css/flag-icons.min.css"
+    />
```
tkohr, line 13: I think we should import this in our CSS and get rid of the external dependency here. If I saw right you've already installed the npm package.

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -1162,19 +1162,22 @@ export function readOtherLanguages(rootEl: XmlElement): LanguageCode[] {
     findChildrenElement('gmd:locale', false),
     mapArray(
       pipe(
-        findChildElement('gmd:LanguageCode'),
+        findChildElement('lan:LanguageCode'),
```
tkohr, line 1165: Can't this also be `gmd:LanguageCode`? I think we definitely need unit tests for all modifications done on the read- and write-parts that will already check the changes against our existing fixtures.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -1,15 +1,316 @@
...
+  switchFormLang(lang) {
+    // TO IMPLEMENT FURTHER
+  }
+
+  switchDefaultLang(lang: string) {
+    this.formLanguage = lang
+    this.facade.updateRecordField('defaultLanguage', lang)
+    this.facade.updateRecordField(
+      'otherLanguages',
+      this.selectedLanguages.filter((lang) => lang !== this.formLanguage)
```
tkohr, line 203: Still a little confused about this point: Does this mean the `mainlanguage` should not be part of the `otherLanguage` array?

tkohr (commented):
Thanks for the huge work @cmoinier ! I've left a couple of comments.

Testing the PR, things seem to work really fine. The only thing I noticed that I was able to click on "Marquer par défaut", which broke the position of the menu. I also had the impression that it messed with the record metadata. I saw that the position has been fixed in https://github.com/geonetwork/geonetwork-ui/pull/1237 and I did not have any issues with the metadata there anymore.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
cmoinier: No, since it was its own attribute `defaultLanguage`, it has never been part of `otherLanguages` :) It's indeed in the same object in the ES response but it's split when building the record object. Since it's used in many places I don't think we should modify it.

libs/feature/editor/src/lib/components/multilingual-panel/multilingual-panel.component.ts
```diff
@@ -1,15 +1,316 @@
...
+          'editor.record.multilingual.confirmation.confirmText'
+        ),
+        cancelText: this.translateService.instant(
+          'editor.record.multilingual.confirmation.cancelText'
+        ),
+      },
+      restoreFocus: true,
+    })
+
+    dialogRef.afterClosed().subscribe((confirmed) => {
```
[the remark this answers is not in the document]
cmoinier: Will do :)

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
cmoinier: Actually this might be leftover, it should have stayed as `gmd:LanguageCode`. I will revert, thanks for catching it.

tkohr:
Thanks for the adaptions @cmoinier, looking good now!

PR #1240, opened 2025-05-14
CI: Fix backport workflow

tkohr:
### Description

Currently, the backport workflow is triggered on every merge on `main` (for both PRs with and without labels containing "backport"). This PR inverts the logic for the label check. It also converts the array to a string as the current condition does not seem to be taken account.


tkohr:
Thanks @LHBruneton-C2C ! It seems to have worked for this non-backport PR. Let's see if it also works on the next backport PR.

PR #1244, opened 2025-05-15
Adapt metadata quality widget by resource type

Guillaume-d-o:
### Description

This PR introduces [...]

### Architectural changes

### Screenshots

image

### How to test

---


Guillaume-d-o:
@LHBruneton-C2C could you just check my translation are correct and the field is the good one ( metadata.extras.catalogUuid) . Thanks :)

translations/en.json
```diff
@@ -420,6 +421,8 @@
   "record.metadata.publication": "Date of publication",
   "record.metadata.publications": "{count, plural, =0{publication} one{publication} other{publications}}",
   "record.metadata.quality": "Metadata Quality",
+  "record.metadata.quality.capabilities.failed": "Service specifications are not specified",
+  "record.metadata.quality.capabilities.success": "Service specifications are specified",
```
LHBruneton-C2C, line 425:
```suggestion
  "record.metadata.quality.capabilities.failed": "Capabilities URL is not specified",
  "record.metadata.quality.capabilities.success": "Capabilities URL is specified",
```

translations/fr.json
```diff
@@ -420,6 +421,8 @@
   "record.metadata.publication": "Date de publication",
   "record.metadata.publications": "{count, plural, =0{donnée} one{donnée} other{données}}",
   "record.metadata.quality": "Qualité des métadonnées",
+  "record.metadata.quality.capabilities.failed": "Les spécifications du service ne sont pas renseignées",
+  "record.metadata.quality.capabilities.success": "Les spécifications du service sont renseignées",
```
LHBruneton-C2C, line 425:
```suggestion
  "record.metadata.quality.capabilities.failed": "Aucune URL de Capabilities n'est renseignée",
  "record.metadata.quality.capabilities.success": "Une URL de Capabilities est renseignée",
```

translations/fr.json
```diff
@@ -431,6 +434,8 @@
   "record.metadata.quality.legalConstraints.success": "Les contraintes légales sont renseignées",
   "record.metadata.quality.organisation.failed": "Le producteur n'est pas renseigné",
   "record.metadata.quality.organisation.success": "Le producteur est renseigné",
+  "record.metadata.quality.sources.failed": "Le catalogue liée n'est pas renseignée",
+  "record.metadata.quality.sources.success": "Le catalogue liée est renseignée",
```
LHBruneton-C2C, line 438:
```suggestion
  "record.metadata.quality.sources.failed": "Aucun jeu de données source n'est renseigné",
  "record.metadata.quality.sources.success": "Un jeu de données source est renseigné",
```

libs/ui/elements/src/lib/metadata-quality/metadata-quality.component.ts
```diff
@@ -69,18 +69,38 @@ export class MetadataQualityComponent implements OnChanges {
...
+    } else if (datasetType === 'service') {
+      this.add(
+        'capabilities',
+        linkUrl.some((url) => this.hasGetCapabilities(url))
+      )
+    } else if (datasetType === 'reuse') {
+      this.add('topic', this.metadata?.topics?.length > 0)
+      this.add('organisation', !!contact?.organization)
+      this.add('source', !!this.metadata?.extras?.sourcesIdentifiers)
+    }
```
tkohr, line 103: Maybe an array or object to map this logic could improve readability?

tkohr (commented):
Thanks @Guillaume-d-o, the added logic looks good to me, but I think there is still an issue with the calculation for the service, where I always get 62% indicated (also with 5 of 6 fields present). On reuse, I also see 62% for 6 out of 8 fields present, btw.

Also I see differences between the checked fields in the results list widget and the record page widget, but I guess this is normal as we have the ES pipeline (which has not been updated in this PR) running in our local compo?

tkohr:
So I've double-checked, if I deactivate the `qualityScore` from the pipeline, the calculations are correct indeed!

However, realizing that only the score come from the pipeline, I noticed another inconsistency we already talked about before. The reuse records do not have the information displayed if they have a source dataset or not when displayed in the results list. You can actually add this easisily by adding the `recordLink` field to the response for the search records [here](https://github.com/geonetwork/geonetwork-ui/blob/adapt-metadata-quality-widget-by-resource-type/libs/feature/search/src/lib/constants.ts#L25). The rest works out of the box, as you only need the IDs and the mapping is already present. Note that it will make the response a little heavier, but I don't think we have an alternative.

libs/ui/elements/src/lib/metadata-quality/metadata-quality.component.ts
```diff
@@ -69,18 +73,60 @@ export class MetadataQualityComponent implements OnChanges {
       this.items.push({ name, value })
     }
   }
+  hasGetCapabilities(url: string): boolean {
+    const searchParams = new URLSearchParams(url)
+    return searchParams.get('REQUEST')?.toLowerCase() === 'getcapabilities'
```
tkohr, line 78: Small detail, but the pipeline uses 'capabilities' as string to be contained, as indicated in the ticket. I think, we should make sure we use exactly the same string to prevent any potential different score calculations that will be hard to find in the future.

libs/ui/elements/src/lib/metadata-quality/metadata-quality.component.ts
```diff
@@ -69,18 +73,60 @@ export class MetadataQualityComponent implements OnChanges {
...
+    if (datasetType && this.SPECIFIC_CHECKS[datasetType]) {
+      Object.entries(this.SPECIFIC_CHECKS[datasetType]).forEach(
+        ([name, check]) => {
+          this.addIfEnabled(name, check(this.metadata))
+        }
+      )
+    }
+  }
+
+  private addIfEnabled(name: string, value: boolean) {
```
tkohr, line 126: There is still the `add` method with the identical code, that is not used anymore.

libs/ui/elements/src/lib/metadata-quality/metadata-quality.component.ts
Guillaume-d-o: you are right, i make the modifications :)

tkohr (approved):
Thanks for the added e2e tests and the fix on the capabilities url @Guillaume-d-o ! It looks consistent now!

tools/pipelines/register-es-pipelines.js
```diff
@@ -93,7 +93,7 @@ if(!isService && ctx.contact != null && ctx.contact.length > 0 && ctx.contact[0]
   ok++
 }
 // this checks for multilingual Organizations (GN 4.2.3+)
-if(!isService && ctx.contact != null && ctx.contact.length > 0 && ctx.contact[0].organisationObject != null && ctx.contact[0].organisationObject.default != '') {
+if(!isService && ctx.contact != null && ctx.contact.length > 0 && ctx.contact[0].organisationObject != null && ctx.contact[0].organisationObject.default != null && ctx.contact[0].organisationObject.default != '') {
```
AlitaBernachot, line 96: Thx for the fix, working great now for `dataset/a3774ef6-809d-4dd1-984f-9254f49cbd0a`!

PR #1242, opened 2025-05-14
Datahub: Get linked internal records

tkohr:
### Description

This PR is still WIP. 

It gets internally linked records via the ES fields `sources` and `hasSources` to display these records on dataset, service and reuse pages.

### Architectural changes

The following library now depends on [...]

### Screenshots

[...]

### Quality Assurance Checklist

- [ ] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [ ] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](docs) 📚 has received the love it deserves

### How to test

Follow these instructions to confirm [...]

---

PR #1247, opened 2025-05-19
Datahub: Display linked internal records

tkohr:
### Description

This PR is a follow up on https://github.com/geonetwork/geonetwork-ui/pull/1242 to display linked internal records in the datahub.

The following relations exist and are displayed:

- datasets:
  - section for source links
    - subsection for “source” datasets
    - subsection for “associated” datasets
    - subsection for “associated” services
    - subsection for “associated” reuses
- services:
  - section with one subsection for “source” datasets
- reuses:
  - section with one subsection for “source” datasets

Note: The structure of the e2e tests has been changed a little bit with the idea to split bigger test suites in multiple ones.

Todos:

- [x] e2e tests
- [ ] remove related-records-card use from organization-page and delete component => card will be fixed in a different PR (a ticket already exists)

### Architectural changes

### Screenshots

image

### How to test

There are two new records in the dump that can be accessed for testing:
- Service: http://localhost:4200/service/ce551730-ee1a-4de3-9b49-7e432375d08a
- Reuse: http://localhost:4200/reuse/7eb795c2-d612-4b5e-b15e-d985b0f4e697

Both point to the same dataset as a source dataset: http://localhost:4200/dataset/n_tri_lill_inondable_s_059

If you want to test a bigger number of links between records you can add more "source datasets" via GN. For the e2e tests a fixture xas added to test this case without doing too much editing of the dump.

---


apps/datahub/src/app/record/record-linked-records/record-linked-records.component.html
```diff
@@ -0,0 +1,26 @@
...
+<datahub-record-internal-links
+  *ngIf="hasLinkedReuses$ | async"
+  [title]="'record.metadata.linked.reuses' | translate"
+  [records]="linkedReuses$ | async"
+></datahub-record-internal-links>
+<datahub-record-internal-links
+  *ngIf="hasLinkedServices$ | async"
+  [title]="'record.metadata.linked.services' | translate"
+  [records]="linkedServices$ | async"
+></datahub-record-internal-links>
```
cmoinier, line 26: How about a function that uses `switch` to return title & records based on the type, and in the template an ngFor loop to generate the components, rather than having 4 instances for each type?
I don't know if it would be more clear though, just a suggestion, your choice :)

apps/datahub/src/app/record/record-linked-records/record-linked-records.component.ts
```diff
@@ -0,0 +1,42 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class RecordLinkedRecordsComponent {
+  sourceDatasets$ = this.metadataViewFacade.sources$
+  linkedDatasets$ = this.metadataViewFacade.sourceOf$.pipe(
+    map((records) => records?.filter((record) => record?.kind === 'dataset'))
+  )
+  linkedReuses$ = this.metadataViewFacade.sourceOf$.pipe(
+    map((records) => records?.filter((record) => record?.kind === 'reuse'))
+  )
```
cmoinier, line 23: How about :

```suggestion
  linkedReuses$ = this.metadataViewFacade.sourceOf$.pipe(
  map((records) => records?.filter((record) => record?.kind === 'reuse') ?? [])
);
```

Instead of checking again in the getter?

cmoinier (approved):
Thank you for this big work and for the adaptations to the e2e tests - much appreciated :)

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.ts
```diff
@@ -64,12 +64,12 @@ export class NavigationBarComponent {
...
-      anchor: 'user-feedbacks',
-      label: 'record.metadata.userFeedbacks',
+      anchor: 'linked-records',
+      label: 'record.metadata.linked.records',
     },
     {
-      anchor: 'related',
-      label: 'record.metadata.related.contents',
+      anchor: 'user-feedbacks',
+      label: 'record.metadata.userFeedbacks',
```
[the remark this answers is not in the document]
tkohr: Indeed, I didn't see any in the mockups, and I just double-checked with Pierre.

apps/datahub/src/app/record/record-linked-records/record-linked-records.component.html
tkohr: Good idea, but I think I'll stick with my verbose version for now as it is very simple to read.

apps/datahub/src/app/record/record-linked-records/record-linked-records.component.ts
tkohr: Thanks for the idea! Do you mean to check for the length in the template then?

tkohr:
> 
> I have left so suggestions inline but I didn't find any issues. I just have noticed something in the mobile view, only for the dataset page (http://localhost:4200/dataset/n_tri_lill_inondable_s_059) : image It seems that the links from `datahub-record-internal-links` (not sure this is related to your PR or not) are larger than they should be. This is not the case for reuses and services - or maybe it's related to the fact that this record has 2 linked records and the other ones I tested only have 1?

Good catch! This seems to be related to the fact if the `internal-link.card` displays an organization or not. I won't touch on it in this PR as there is already https://github.com/geonetwork/geonetwork-ui/pull/1249 touching on this topic and there is another ticket addressing the card without an organization. I'll add a comment there.

apps/datahub/src/app/record/record-linked-records/record-linked-records.component.ts
cmoinier: I would find that a bit more readable than having two checks in the logic but it's a preference :)

apps/datahub/src/app/record/navigation-bar/navigation-bar.component.ts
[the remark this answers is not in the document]
cmoinier: Alright then!

apps/datahub/src/app/record/record-linked-records/record-linked-records.component.ts
tkohr: Ok I kept it as is, as I think it'd still need the length check in the template if not

PR #1249, opened 2025-05-21
Link card description instead of empty contact

Guillaume-d-o:
### Description

This PR introduce modification around the internal Link cards. When the contact is not defined, the description replace it for M and S size. If there is already a description like in L size, the maximum lines is raised from 2 to 6 lines. 

### Architectural changes

I splitted the component in two parts to separate the contact and the card

### Screenshots
For Large size : 
image
image

For M / S size ( M size in example here)  
image
image

### How to test

Find a dataset with contact
Empty the organisation name in the contacts 
you will have this display

---

### DIFF
--- a/libs/ui/elements/src/lib/internal-link-card/internal-link-card.component.ts
+++ b/libs/ui/elements/src/lib/internal-link-card/internal-link-card.component.ts
@@ -112,15 +112,17 @@ export class InternalLinkCardComponent implements OnInit {
   }
 
   getAbstractLineClamp(): string {
-    if (this.size === 'L') {
-      return this.record.ownerOrganization?.name
-        ? 'line-clamp-2'
-        : 'line-clamp-6'
+    if (this.record.ownerOrganization?.name) {
+      if (this.size === 'L') {
+        return 'line-clamp-2'
+      }
+      return ''
     }
-    if (
-      (this.size === 'M' || this.size === 'S') &&
-      !this.record.ownerOrganization?.name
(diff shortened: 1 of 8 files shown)


libs/ui/elements/src/lib/internal-link-card/internal-link-card.component.ts
```diff
@@ -125,35 +107,33 @@ export class InternalLinkCardComponent implements OnInit {
...
+    }
+
+    if (this.size === 'L') {
+      return 'line-clamp-6'
+    }
+    if (this.size === 'M' || this.size === 'S') {
+      return 'line-clamp-2'
+    }
+    return ''
   }
```
tkohr, line 129: Wondering if this could be simplified...

tkohr (approved):
Thanks @Guillaume-d-o, tested in storybook and behavior looks good!

A small detail I noticed was the indentation of the title not being the same as the description on the 'S' sized card:
image

libs/ui/elements/src/lib/internal-link-card/internal-link-card.component.ts
Guillaume-d-o: simplified. I keep the logic inside of the ts, because some of you think it's better ( but not easier) , and it is much more easier to understand unit test.

tkohr: Nice!

PR #1250, opened 2025-05-22
[Editor] : Multilingual demo feedback

cmoinier:
### Description

This PR does : 

- [ ] Change the confirmation dialog text to something less dramatic + focuses on the "delete" button
- [ ] Display unsupported languages in the dashboard tooltip for multilingual records
- [ ] Avoids having the newly-deleted languages show up in the panel in-between saving and reloading the page --> note that this was due to the isoConverter parsing the translations to `otherLanguages` all the time (loading the page, saving etc...) no matter if the list was already populated by the `gn4.field.mapper`. As far as I understand (and also with the time limit I had), this is only needed when `otherLanguages` has no languages already, so I added a check for that.

### Architectural changes

no

### Screenshots

no notable UI changes

### DIFF
--- a/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
+++ b/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
@@ -160,7 +160,7 @@ export class Gn4FieldMapper {
     },
     otherLanguage: (output, source) => {
       const langList = getAsArray(selectField<string>(source, 'otherLanguage'))
-      const languages = langList.map((lang) => LANG_3_TO_2_MAPPER[lang])
+      const languages = langList.map((lang) => LANG_3_TO_2_MAPPER[lang] ?? lang)
       const defaultLanguage = output.defaultLanguage ?? languages[0] ?? null
       const otherLanguages = languages.filter(
         (lang) => lang !== defaultLanguage
--- a/libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
+++ b/libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
@@ -183,8 +183,10 @@ export class Iso19139Converter extends BaseConverter<string> {
         }
         // add languages that have translations to the root language list
         for (const lang in fieldTranslations) {
-          if (!record.otherLanguages.includes(lang)) {
-            record.otherLanguages.push(lang)
+          if (!record.otherLanguages.length) {
+            if (!record.otherLanguages.includes(lang)) {
+              record.otherLanguages.push(lang)
+            }
           }
         }
       }
--- a/libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
+++ b/libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
@@ -7,6 +7,7 @@ import {
 import { ButtonComponent } from '@geonetwork-ui/ui/inputs'
 
 export interface ConfirmationDialogData {
+  focusCancel: string
   title: string
   message: string
   confirmText: string
@@ -22,10 +23,13 @@ export interface ConfirmationDialogData {
   imports: [MatDialogModule, ButtonComponent],
 })
 export class ConfirmationDialogComponent {
+  focusCancel = null
   constructor(
     public dialogRef: MatDialogRef<ConfirmationDialogComponent>,
     @Inject(MAT_DIALOG_DATA) public data: ConfirmationDialogData
-  ) {}
+  ) {
+    this.focusCancel = data.focusCancel ? true : null
+  }
 
   onConfirm() {
     this.dialogRef.close(true)
--- a/translations/en.json
+++ b/translations/en.json
@@ -245,7 +245,7 @@
   "editor.record.lock.reason": "You are not an editor of the allowed groups",
   "editor.record.multilingual.confirmation.cancelText": "Keep the language(s)",
   "editor.record.multilingual.confirmation.confirmText": "Remove the language(s)",
-  "editor.record.multilingual.confirmation.message": "You are about to remove one or several languages, which will delete all of the related translations and is irreversible. Do you want to proceed?",
+  "editor.record.multilingual.confirmation.message": "You are about to remove one or several languages, which will delete all of the related translations. Do you want to proceed?",
   "editor.record.multilingual.confirmation.title": "Remove languages",
   "editor.record.onlineResource.protocol.other": "Other",
   "editor.record.onlineResourceError.body": "An error happened while adding the resource:",
--- a/translations/fr.json
+++ b/translations/fr.json
@@ -245,7 +245,7 @@
   "editor.record.lock.reason": "Vous n'êtes pas éditeur d'un des groupes autorisés.",
   "editor.record.multilingual.confirmation.cancelText": "Garder le(s) language(s)",
   "editor.record.multilingual.confirmation.confirmText": "Supprimer le(s) language(s)",
-  "editor.record.multilingual.confirmation.message": "Vous allez supprimer un ou plusieurs languages, ce qui supprimera toutes les traductions qui y sont liées et est irréversible. Voulez-vous procéder ?",
+  "editor.record.multilingual.confirmation.message": "Vous allez supprimer un ou plusieurs languages, ce qui supprimera toutes les traductions qui y sont liées. Voulez-vous procéder ?",
   "editor.record.multilingual.confirmation.title": "Suppression de language(s)",
   "editor.record.onlineResource.protocol.other": "Autre",
   "editor.record.onlineResourceError.body": "Une erreur est survenue lors de l'ajout de la ressource :",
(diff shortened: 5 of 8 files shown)


libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -160,7 +160,7 @@ export class Gn4FieldMapper {
     },
     otherLanguage: (output, source) => {
       const langList = getAsArray(selectField<string>(source, 'otherLanguage'))
-      const languages = langList.map((lang) => LANG_3_TO_2_MAPPER[lang])
+      const languages = langList.map((lang) => LANG_3_TO_2_MAPPER[lang] ?? lang)
```
tkohr, line 163: Maybe adding a test/fixture with an example of a code that can not be found in the mapper would be worth it.

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
```diff
@@ -183,8 +183,10 @@ export class Iso19139Converter extends BaseConverter<string> {
         }
         // add languages that have translations to the root language list
         for (const lang in fieldTranslations) {
-          if (!record.otherLanguages.includes(lang)) {
-            record.otherLanguages.push(lang)
+          if (!record.otherLanguages.length) {
```
tkohr, line 186: Hum, I'm not quite sure if this is the right way to go, I see there is a UT breaking in the converter. Maybe this can give an indication?

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
```diff
@@ -183,8 +183,10 @@ export class Iso19139Converter extends BaseConverter<string> {
...
+            if (!record.otherLanguages.includes(lang)) {
+              record.otherLanguages.push(lang)
+            }
```
tkohr, line 189: Btw, I noticed that the language list in the tooltip in the dashboard also does not get updated as long as the change has not been synchronized with the backend. At the same time, this is pretty logic as we get the info from ES and I think it shouldn't bother the user too much as (s)he can see that changes are still pending in the draft system.

libs/ui/elements/src/lib/confirmation-dialog/confirmation-dialog.component.ts
```diff
@@ -22,10 +23,13 @@ export interface ConfirmationDialogData {
   imports: [MatDialogModule, ButtonComponent],
 })
 export class ConfirmationDialogComponent {
+  focusCancel = null
   constructor(
     public dialogRef: MatDialogRef<ConfirmationDialogComponent>,
     @Inject(MAT_DIALOG_DATA) public data: ConfirmationDialogData
-  ) {}
+  ) {
+    this.focusCancel = data.focusCancel ? true : null
```
tkohr, line 31: Good idea to make it configurable!

translations/en.json
```diff
@@ -245,7 +245,7 @@
   "editor.record.lock.reason": "You are not an editor of the allowed groups",
   "editor.record.multilingual.confirmation.cancelText": "Keep the language(s)",
   "editor.record.multilingual.confirmation.confirmText": "Remove the language(s)",
-  "editor.record.multilingual.confirmation.message": "You are about to remove one or several languages, which will delete all of the related translations and is irreversible. Do you want to proceed?",
+  "editor.record.multilingual.confirmation.message": "You are about to remove one or several languages, which will delete all of the related translations. Do you want to proceed?",
```
tkohr, line 248: Shouldn't "languages" be "langues" in French? Note, there are various translations where this term is used.

tkohr (commented):
Thanks, @cmoinier, the first two commits look good to me, but I also have a doubt about the last one. It works well when testing this case manually but there is a UT failing I think.

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
AlitaBernachot: This mapping is also done in `resourceLanguage` (legacy) and `mainLaguage` (legacy), why fix the mapping for `otherLanguage` only?

cmoinier: Because `resourceLanguage` and `mainLanguage` are always supported languages, so the ISO mapping always works for them :) 
Technically we could import a record with an unsupported language. It's very unlikely, but I will also change their mapping then.

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
cmoinier: Yes, will work on that with Olivia asap.

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
cmoinier: Oh good catch. It should change since when we change a title but don't save, the new title is still shown in the dashboard. I will fix that if it doesn't take too much time.

translations/en.json
cmoinier: I think both work in French? But "langues" is more correct I think. I will edit the French file :)

libs/common/fixtures/src/lib/records.fixtures.ts
```diff
@@ -972,3 +972,55 @@ export const SAMPLE_RECORD = {
...
+        description: { fr: 'Description FR', de: 'Schlusselwort DE' },
+      },
+    },
+  ],
+  onlineResources: [], // todo online resources
+  licenses: [],
+  legalConstraints: [], // TODO constraints
+  securityConstraints: [],
+  otherConstraints: [],
+  contacts: [], // TODO contacts
```
tkohr, line 1004: Are these still todos, or can we remove the comments?

translations/fr.json
```diff
@@ -244,10 +244,12 @@
   "editor.record.lock.harvested": "Les jeux de données moissonnés ne sont pas éditables",
   "editor.record.lock.owner": "Vous n'êtes pas éditeur d'un des groupes autorisés",
   "editor.record.lock.resourceType": "Les fiches d'un type autre que jeu de données ne sont pas éditables",
-  "editor.record.multilingual.confirmation.cancelText": "Garder le(s) language(s)",
-  "editor.record.multilingual.confirmation.confirmText": "Supprimer le(s) language(s)",
-  "editor.record.multilingual.confirmation.message": "Vous allez supprimer un ou plusieurs languages, ce qui supprimera toutes les traductions qui y sont liées et est irréversible. Voulez-vous procéder ?",
-  "editor.record.multilingual.confirmation.title": "Suppression de language(s)",
+  "editor.record.lock.format": "Les jeux de données de type 'service' et 'reuse' ne sont pas éditables",
+  "editor.record.lock.reason": "Vous n'êtes pas éditeur d'un des groupes autorisés.",
```
tkohr, line 248: Did these two maybe slip in here accidentally?

tkohr (commented):
Thanks for the big continued work on this @cmoinier ! As far as I see the changes, LGTM. 

As discussed, I just noticed that a deleted language sometimes pops up again when the app is reloaded before saving. And in this case one has to reload the app again after saving to make the language disappear. However, this behavior is not systematic and sometimes everything works fine.

translations/fr.json
cmoinier: That must be a rebase error, good catch, thanks :)

libs/common/fixtures/src/lib/records.fixtures.ts
cmoinier: Must be from a copy/paste, I will remove them

libs/api/metadata-converter/src/lib/iso19139/iso19139.converter.ts
```diff
@@ -398,10 +398,11 @@ export class Iso19139Converter extends BaseConverter<string> {
         this.writers['spatialExtents'](record, rootEl)
       fieldChanged('lineage') && this.writers['lineage'](record, rootEl)
     }
-
     fieldChanged('otherLanguages') &&
       this.writers['otherLanguages'](record, rootEl)
 
+    fieldChanged('translations') && this.writers['translations'](record, rootEl)
```
AlitaBernachot, line 404: see l. 163 https://github.com/geonetwork/geonetwork-ui/pull/1250/files#diff-fb106e4a7ec88ef749c3d7f2b3a5c0dc559b7ddb9792f134b62cb0be82d4a3f3R163 the dedicated write does nothing.

AlitaBernachot:
@cmoinier thanks a lot for all these fixes! this is a huge topic, i must admit i had some difficulties to dive into it. I hope i have identified and fixed the case you were talking about.

tkohr (approved):
Thanks for the additional fixes @AlitaBernachot on the converter ! Not sure if I got everything, but deleting one or all languages and reloading the app before saving works flawlessly now :tada:

Also, I noticed that the field values are changed on the fly when switching the default language now, which was not the case before I believe. Nice! (Just as a note: This is not the case for the title, maybe due to the type of input box it uses!? One has to reopen or reload the app to see the title field updated.)

PR #1269, opened 2025-06-19
Datahub: Make navigation bar appear on scroll

tkohr:
### Description

This PR reworks the integration of the "quick access" navigation bar into the `header-record.component`, following the mockups:
- Initially it displays the following three buttons within the header: navigation back, favorite star, language (optional)
- When scrolling down, the navigation bar appears and displays the same three buttons in a different style along with the section titles for the quick access

### Screenshots

Desktop:
nav-bar-desktop

Mobile:
nav-bar-mobile

### How to test

Visit dataset, service and reuse pages on desktop and mobile and check if the header and navigation bar display correctly.

Activate and deactivate languages in the config and check if both cases display well.

`languages = ['en', 'fr', 'de']`

---

### DIFF
--- a/apps/datahub/src/app/organization/organization-header/organization-header.component.html
+++ b/apps/datahub/src/app/organization/organization-header/organization-header.component.html
@@ -18,7 +18,7 @@
             size="20px"
             class="align-middle"
           ></ng-icon>
-          <span class="mx-2 mt-0.5 text-[16px] content-center" translate
+          <span class="mx-2 mt-0.5 text-base content-center" translate
             >datahub.search.back</span
           >
         </gn-ui-button>
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -26,7 +26,7 @@
               size="20px"
               class="align-middle"
             ></ng-icon>
-            <span class="mx-2 mt-0.5 text-[16px] content-center" translate
+            <span class="mx-2 mt-0.5 text-base content-center" translate
               >datahub.search.back</span
             >
           </gn-ui-button>
--- a/apps/datahub/src/app/record/header-record/header-record.component.ts
+++ b/apps/datahub/src/app/record/header-record/header-record.component.ts
@@ -67,7 +67,6 @@ export const HEADER_HEIGHT_MOBILE_THUMBNAIL = 554
 })
 export class HeaderRecordComponent {
   @Input() metadata: DatasetRecord | ServiceRecord | ReuseRecord
-  @Input() expandRatio: number
   backgroundCss =
     getThemeConfig().HEADER_BACKGROUND ||
     `center /cover url('assets/img/header_bg.webp')`


apps/datahub/src/app/organization/organization-header/organization-header.component.html
```diff
@@ -11,23 +11,21 @@
...
         >
           <ng-icon
             name="matArrowBack"
             size="20px"
             class="align-middle"
           ></ng-icon>
-          <span
-            class="mx-2 mt-0.5 text-[16px] tracking-widest content-center opacity-75 uppercase"
-            translate
+          <span class="mx-2 mt-0.5 text-[16px] content-center" translate
```
cmoinier, line 21:
```suggestion
          <span class="mx-2 mt-0.5 text-base content-center" translate
```

apps/datahub/src/app/record/header-record/header-record.component.html
```diff
@@ -1,72 +1,136 @@
...
-    >
-    </gn-ui-image-overlay-preview>
-  </div>
-</header>
+      <datahub-navigation-bar
+        *ngIf="(facade.isMetadataLoading$ | async) === false"
+        [metadata]="metadata"
+        [style.opacity]="-1 + (1 - expandRatio) * 2"
+        [style.pointer-events]="
+          -1 + (1 - expandRatio) * 2 <= 0 ? 'none' : 'auto'
```
cmoinier, line 131: Should we have that kind of calculations in the template? I feel like it belongs more in the logic (unless `expandRatio` is only available here)

apps/datahub/src/app/record/header-record/header-record.component.ts
```diff
@@ -50,26 +67,44 @@ import {
...
+  showLanguageSwitcher = getGlobalConfig().LANGUAGES?.length > 0
 
   showOverlay = true
 
+  isMobile$ = fromEvent(window, 'resize').pipe(
+    startWith(window.innerWidth),
+    map(() => window.innerWidth < MOBILE_MAX_WIDTH)
+  )
+
+  thumbnailUrl$ = this.facade.metadata$.pipe(
```
cmoinier, line 84: I don't know if it's related to the fact that `thumbnailUrl` is now an Observable, but there is a glitch on http://localhost:4200/dataset/ee965118-2416-4d48-b07e-bbc696f002c2 where the thumbnail appears and disappears when loading

cmoinier (approved):
Thank you @tkohr for the rework :) It works really well IMO, just a few suggestions and a small glitch I noticed on the thumbnail overview !

apps/datahub-e2e/src/e2e/service/serviceDetailPage.cy.ts
```diff
@@ -25,12 +25,26 @@ describe('service pages', () => {
...
-        .should('be.visible')
-      cy.get('datahub-navigation-bar')
-        .find('[data-cy="data-preview"]')
-        .should('not.be.visible')
+      cy.get('datahub-record-page')
+        .find('gn-ui-service-capabilities')
+        .should('exist')
+      cy.scrollTo(0, 1000, { ensureScrollable: false })
+      cy.window().then((win) => {
+        if (win.scrollY > 0) {
```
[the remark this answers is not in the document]
tkohr: Hum, it stays a kind of workaround.

apps/datahub/src/app/record/header-record/header-record.component.html
tkohr: Good point, I completely agree with you, and you just made me realize that I don't use the `expandRatio` input anymore since I moved the sticky header within this component. The `let-expandratio` directive exposes the value only to the template and it would actually mean more changes (like an additional output in the sticky header component) to move this logic out of the template. So I suggest to keep it as is.

apps/datahub/src/app/record/header-record/header-record.component.ts
tkohr: Good catch, I double-checked and we have the same behavior on main. Note that this occurs only, if the record has a thumbnail url, but the url is erroneous.

tkohr:
Thanks for the review @cmoinier ! I've added a commit to address your comments.

apps/datahub/src/app/record/header-record/header-record.component.html
cmoinier: Okay, it makes sense then :)

PR #1273, opened 2025-06-24
[Datahub]: Close autocomplete suggestions on search

cmoinier:
### Description

This PR makes sure that the autocomplete suggestion panel closes when a search is done, either through the enter touch or the search button.
The ticket reported that the panel was already closing when clicking on the search button, but it was actually only closing on the second click. This is fixed too.
I have also removed duplicate e2e tests about the datasets sorting.

### Screenshots

No UI changes.

### How to test

Do a search on the main header.

### DIFF
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
@@ -296,11 +296,11 @@ export class AutocompleteComponent
     this.inputRef.nativeElement.focus()
   }
 
-  handleSearch(any?: string) {
+  handleSearch() {
     if (!this.cancelEnter && this.allowSubmit) {
       this.isSearchActive.emit(true)
       this.searchActive = true
-      this.inputSubmitted.emit(any ?? this.inputRef.nativeElement.value)
+      this.inputSubmitted.emit(this.inputRef.nativeElement.value)
     }
     this.triggerRef?.closePanel()
   }
--- a/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.html
+++ b/libs/ui/inputs/src/lib/autocomplete/autocomplete.component.html
@@ -14,7 +14,7 @@
       [placeholder]="placeholder"
       [formControl]="control"
       [matAutocomplete]="auto"
-      (keyup.enter)="handleSearch(searchInput.value)"
+      (keyup.enter)="handleSearch()"
       [ngClass]="{
         'text-primary': searchActive && enterButton,
         'text-gray-900': !searchActive,


apps/datahub-e2e/src/e2e/datasets.cy.ts
```diff
@@ -501,50 +501,6 @@ describe('datasets', () => {
     })
   })
 
-  describe('sorting results', () => {
```
tkohr, line 504: Where are the duplicates of these?

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -296,16 +296,13 @@ export class AutocompleteComponent
     this.inputRef.nativeElement.focus()
   }
 
-  handleEnter(any: string) {
+  handleSearch(any?: string) {
     if (!this.cancelEnter && this.allowSubmit) {
       this.isSearchActive.emit(true)
       this.searchActive = true
-      this.inputSubmitted.emit(any)
+      this.inputSubmitted.emit(any ?? this.inputRef.nativeElement.value)
```
tkohr, line 303: Just wondering if we still actually need the `any` input here, or if we could always use `this.inputRef.nativeElement.value`?

tkohr (approved):
Thanks for the quick fix @cmoinier ! I've tested the autocomplete in the datahub and the editor and both work fine :smiley:

apps/datahub-e2e/src/e2e/datasets.cy.ts
cmoinier: They are not exact duplicates (the commands differ), but in both cases they are testing that the sorting of the datasets work by date and by popularity : https://github.com/geonetwork/geonetwork-ui/blob/9134515579b7cd3f5e62c0be96c16bd24cc47374/apps/datahub-e2e/src/e2e/search-header/filters.cy.ts#L15 (and below)

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
cmoinier: Good question, I will try it out!

libs/ui/inputs/src/lib/autocomplete/autocomplete.component.ts
```diff
@@ -183,7 +183,7 @@ export class AutocompleteComponent
       switchMap((value) => this.action(value)),
       tap((suggestions) => {
         // forcing the panel to open if there are suggestions
-        if (suggestions.length > 0) {
+        if (suggestions.length > 0 && !this.isSearchActive) {
```
jahow, line 186: shouldn't this be `this.searchActive` instead? `this.isSearchActive` is the output emitter.

cmoinier: Ah damn, you're right. Weirdly it works too, but I will slip that fix in my other PR (which will be merged today)

PR #1275, opened 2025-06-26
Datahub: Fix quick access navigation bar for mobile

tkohr:
### Description

This PR fixes the nav bar that currently does not expand on mobile, because it is contained in the sticky header that shrinks to a height of 60px when scrolling down the page.

Follow-up PR of https://github.com/geonetwork/geonetwork-ui/pull/1269

### Screenshots

image

### How to test

Open a record on mobile screen size and try to use the nav bar. Check if opening and closing is intuitive to the user.

---

### DIFF
--- a/apps/datahub/src/app/record/header-record/header-record.component.html
+++ b/apps/datahub/src/app/record/header-record/header-record.component.html
@@ -126,7 +126,7 @@
       <datahub-navigation-bar
         *ngIf="(facade.isMetadataLoading$ | async) === false"
         [metadata]="metadata"
-        [style.opacity]="-1 + (1 - expandRatio) * 2"
+        [style.display]="-1 + (1 - expandRatio) * 2 < 1 ? 'none' : 'block'"
         [style.pointer-events]="
           -1 + (1 - expandRatio) * 2 < 1 ? 'none' : 'auto'
         "


apps/datahub-e2e/src/e2e/dataset/datasetHeader.cy.ts
```diff
@@ -126,4 +126,29 @@ describe('dataset: header', () => {
...
+      .should('not.be.visible')
+    cy.get('#data-preview').scrollIntoView()
+
+    //it should display only the (active) data-preview section anchor
+    cy.get('datahub-navigation-bar')
+      .find('[data-cy="data-preview-mobile-title"]')
+      .should('be.visible')
+    cy.get('datahub-navigation-bar')
+      .find('[data-cy="resources-mobile-menu"]')
+      .should('not.be.visible')
```
cmoinier, line 152: Could we also check that the menu expands on mobile?

cmoinier (approved):
Thanks for the fixes :) Just one suggestion for the e2e tests, but since it's also tested in UT I'll let you chose if you wanna do it.

Also, I don't know if it's an easy fix, but on mobile when you are on a lower section and click on "Description" in the menu, this is the view when the animation finishes : 
image
I think it would be best to hide it in this case, WDYT?

apps/datahub-e2e/src/e2e/dataset/datasetHeader.cy.ts
tkohr: This is actually what I wanted to do here, but somehow I was not able to come up with a test that passes systematically (often needing two clicks to be executed).

PR #1277, opened 2025-07-02
Datahub: Make record kind quick filter optional

tkohr:
### Description

This PR adds a `record_kind_quick_filter` setting to the config which is read in the datahub and conditions `<gn-ui-results-hits-search-kind>` to display or not.

Also slips in a commit changing the application-banner to `light` (see last screenshot).

### Architectural changes

### Screenshots

`record_kind_quick_filter = true` (default)
image

`record_kind_quick_filter = false`
image

Application banner color change:

image

### How to test

- Check if quick filter is visible by default on `/search`
- Set `record_kind_quick_filter = false` in `default.toml` and check if it displays well without
- Check if doc is clear

---

### DIFF
--- a/apps/datahub/src/app/home/home-header/home-header.component.html
+++ b/apps/datahub/src/app/home/home-header/home-header.component.html
@@ -6,7 +6,7 @@
     <gn-ui-application-banner
       [message]="bannerMessage"
       [closeEnabled]="true"
-      type="secondary"
+      type="light"
     ></gn-ui-application-banner>
   </div>
   <div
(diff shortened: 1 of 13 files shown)


cmoinier:
> Application banner color change:
> 
> image

Noticing that the close icon is not aligned with the text anymore. It's not related to this PR though so you can chose to fix it or not :)

tkohr:
> > Application banner color change:
> > image
> 
> Noticing that the close icon is not aligned with the text anymore. It's not related to this PR though so you can chose to fix it or not :)

I think we actually placed it intentionally there, with the same padding to the top and right borders, which especially makes sense when there are multiple lines. When there is just one you can argue about it indeed, I'll keep as is for now.

PR #1280, opened 2025-07-04
Fix storybook build / CI for main

tkohr:
### Description

This PR should fix the CI on main that is currently red due to the failing storybook build.


cmoinier:
I think the changes you made to the fixture had an impact on the `ISO19139 converter` UT (they are red, there is an extra tag returning), but it should be easily fixable hopefully.

cmoinier (approved):

PR #1288, opened 2025-07-15
[Tools] Fix quality score pipeline

cmoinier:
### Description

This PR does a fix to the quality score pipeline, which returned a lower score than the calculated score in the frontend.
For context, the "contacts" fields was being checked twice : 
--> once looking for "organisation" (v 4.2.2)
--> another time looking for "organisationObject" (v 4.2.3+)

But it was normal, since the calculation was expecting 8 OKs (on the total of 9 checks, because the two checks on "contacts" were never going to both return OK, being on two different GN-UI versions), so it made sense. So that's a bit of a mystery on my end.

After merging the two "contacts" checks together, the scores returned by the pipeline started to be over-evaluated for records that had an "organisation" --> that was because the "name" field inside of it was not accounted for.

In conclusion : The proposed fix merges the two checks together & looks out for a "name" in the "organisation", to be better aligned with the "organisationObject", where a "default" field is being looked out for.

### Architectural changes

none

### Screenshots

no UI changes

### How to test

Remember to activate the quality widget in the config :)

### DIFF
--- a/tools/pipelines/register-es-pipelines.js
+++ b/tools/pipelines/register-es-pipelines.js
@@ -50,7 +50,7 @@ program
 
 program.parse(process.argv)
 
-const VERSION = 102 // increment on changes
+const VERSION = 103 // increment on changes
 
 const GEONETWORK_UI_PIPELINE = {
   description: 'GeoNetwork-UI pipeline',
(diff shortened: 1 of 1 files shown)


tkohr (commented):
Thanks for the PR @cmoinier ! As discussed, I'm not sure if the change is really needed, as I get the correct score when updating the pipeline manually on `main` as well. I guess we rather have an issue with the updating mechanism. Will try to investigate this.

PR #1296, opened 2025-07-31
fix special caractère et queryparam

Guillaume-d-o:
### Description

This PR introduces [...]

### Architectural changes

### Screenshots

[...]

### How to test

---

### DIFF
--- a/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
+++ b/libs/api/repository/src/lib/gn4/platform/gn4-platform.service.ts
@@ -400,10 +400,21 @@ export class Gn4PlatformService implements PlatformServiceInterface {
           return parsed
         }
 
-        const decoded = atob(parsed)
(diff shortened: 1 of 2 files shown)


LHBruneton-C2C (approved):
Looking good and local test OK, thanks!

jahow:
Oh ok, thank you for the clarification! the GN4 platform implementation has the `apiVersion$` observable so when a logic has to vary depending on the GN version the best is to rely on this instead of try/catch. This would also make the code easier to read and understand IMO.

cmoinier:
> Oh ok, thank you for the clarification! the GN4 platform implementation has the `apiVersion$` observable so when a logic has to vary depending on the GN version the best is to rely on this instead of try/catch. This would also make the code easier to read and understand IMO.

I didn't know that! Thanks, that's valuable info :) I will make a PR about that soon-ish.

PR #1344, opened 2025-09-10
[Datahub]: custom scroll handling for query param change only

LHBruneton-C2C:
### Description

This PR introduces the most relied upon workaround to prevent scroll on query param change:
https://gist.github.com/iffa/9c820072135d25a6372d58075fe264dd 

### Architectural changes

The scroll logic on navigation is no longer managed by Angular 'scrollPositionRestoration'.

### How to test

Freely navigate between the news, search, organization and records pages and see that the scroll comes back to the top when actually changing page. The quick access menu is still working correctly too.
Then on the search page, scroll a bit and change some search parameters (filter, order) and see that the scroll position is kept.
From a scrolled search result list, navigate to a record, then navigate backward (browser navigation), and see that the scroll position on the result list is kept.


LHBruneton-C2C:
Well, the scroll stays were it should when disabling `scrollPositionRestoration` so I guess it's not just a matter of page size.

I don't see how the scroll handling of Angular could know when the data is finished loading, could the order of events have an impact too?

jahow:
I don't think the scroll handling is the problem, instead I think the fact that the page becomes almost empty for a little while is the problem. But I might be wrong! I'll try this PR

PR #1353, opened 2025-09-16
[Datahub] Allow disabling auth-related functions in the Datahub

ronitjadhav:
### Description

This PR introduces the ability to disable authentication related functions in the Datahub application, enabling it to operate as a read only catalog browser. When auth is disabled via configuration, all authentication UI elements (favorite stars, user feedback forms, login buttons) are completely hidden, providing a clean interface for public data discovery.

### Architectural changes

A new `AuthUtilsService` has been added to `libs/feature/auth` to provide centralized authentication state management. This creates a dependency relationship where components across multiple libraries and apps can now check authentication disable status without tight coupling to configuration utilities.

### Screenshots

BeforeAfter2

### How to test

Follow these instructions to confirm the auth disable functionality works:

1. **Test default behavior (auth enabled):**
   - Verify all favorite stars, user feedback forms, and auth buttons are visible
   - Confirm existing functionality works unchanged

2. **Test auth disabled mode:**
   ```toml
   # In conf/default.toml, uncomment:
   disable_auth = true

3. Verify favorite stars disappear from all search results and record cards
4. Confirm user feedback forms are hidden from record detail pages
5. Check that header favorites button is hidden
6. Ensure clean layout with proper space reclamation


libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
cmoinier, line 1: I don't think you should be checking for Auth thing inside of this component at all. I would rather add a condition in the `ngIf` of all its implementations, because keeping the `ngIf` at root here would only create an empty container, and that might be why you had extra spaces left in the cards.

cmoinier (changes requested):
Thanks for this big work ! :slightly_smiling_face: This is a very good first version.
I have left remarks along the way, and will add more once I've tested the PR.
Don't hesitate to reach out if there's something unclear!

Also, I think you should add e2e tests to check that all items that should disappear when the Auth is disabled are indeed not here.

apps/datahub/src/app/record/record-internal-links/record-internal-links.component.spec.ts
```diff
@@ -2,19 +2,65 @@ import { ComponentFixture, TestBed } from '@angular/core/testing'
 import { RecordInternalLinksComponent } from './record-internal-links.component'
 import { MockBuilder } from 'ng-mocks'
 
+jest.mock('@geonetwork-ui/util/app-config', () => {
+  let _disableAuth = false
+  return {
+    getMetadataQualityConfig: () => ({ ENABLED: false }),
+    getOptionalSearchConfig: () => ({
+      LIMIT: 20,
+    }),
```
cmoinier, line 11: I'm not sure that those two elements are needed here

apps/datahub/src/app/record/record-metadata/record-metadata.component.spec.ts
```diff
@@ -24,6 +24,22 @@ import { provideI18n } from '@geonetwork-ui/util/i18n'
 import { REUSE_FORM_URL } from '../record-data-preview/record-data-preview.component'
 import { PlatformServiceInterface } from '@geonetwork-ui/common/domain/platform.service.interface'
 
+jest.mock('@geonetwork-ui/util/app-config', () => {
+  let _disableAuth = false
+  return {
+    getMetadataQualityConfig: () => ({ ENABLED: false }),
+    getOptionalSearchConfig: () => ({
+      LIMIT: 20,
+    }),
```
cmoinier, line 33: Same remark here. Since you didn't change anything related to the metadataquality or searchconfig, that that those mocks weren't here before, I don't think you need to add them.

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.component.html
```diff
@@ -19,63 +19,72 @@ <h2 class="gn-ui-section-title" translate>record.metadata.userFeedbacks</h2>
...
-      <a
-        [href]="loginUrl"
-        class="gn-ui-btn-primary uppercase flex flex-row items-center gap-2 ml-auto"
-        translate
-      >
-        <ng-icon class="grow-0" name="matAccountBoxOutline"></ng-icon>
-        button.login
-      </a>
+      <ng-icon class="grow-0" name="matEditOffOutline"></ng-icon>
+      <span translate>record.metadata.userFeedbacks.authDisabled</span>
```
cmoinier, line 85: That's better, but still not as simple as I thought. I was thinking basically zero changes to the template, expect here https://github.com/geonetwork/geonetwork-ui/blob/3c7a48ceb6ecb489c25a495fc86d4ba8c1762cc7/apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.component.html#L66C18-L66C32 where you would condition the translation and the icon depending on `showAuthUI` (because if there is a user authenticated, then the `showAuthUI` is obviously going to be false. So the condition is really needed when the user is anonymous only. That would remove one `ng-container` and keep both options in the same one :)

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.component.spec.ts
```diff
@@ -22,6 +22,21 @@ import { Gn4PlatformMapper } from '@geonetwork-ui/api/repository'
 import { MockBuilder, MockProvider } from 'ng-mocks'
 import { provideI18n } from '@geonetwork-ui/util/i18n'
 
+jest.mock('@geonetwork-ui/util/app-config', () => {
+  let _disableAuth = false
+  return {
+    getOptionalSearchConfig: () => ({
+      LIMIT: 20,
+    }),
```
cmoinier, line 30: Same here, this option is not necessary

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.component.ts
```diff
@@ -116,7 +129,9 @@ export class RecordUserFeedbacksComponent implements OnInit, OnDestroy {
     private readonly mapper: Gn4PlatformMapper,
     private readonly platformServiceInterface: PlatformServiceInterface
   ) {
-    this.activeUser$ = this.platformServiceInterface.getMe()
+    this.activeUser$ = getGlobalConfig().DISABLE_AUTH
```
cmoinier, line 132: Again I think it's not needed to check for the disabled Auth here :)

libs/feature/search/src/lib/favorites/favorite-star/favorite-star.component.ts
cmoinier, line 1: Are the changes here still needed?

cmoinier (changes requested):
Thanks for the rework! A lot of things are already looking much better. 
There are still some things to be addressed in the feedback section, and some cleanup to be done.
Also, the e2e tests are still needed IMO :)

apps/datahub/src/app/record/record-metadata/record-metadata.component.spec.ts
cmoinier: Ah right, my bad, those are not optional so indeed it's needed for the mock. Please ignore my other comments :)

apps/datahub/src/app/record/record-user-feedbacks/record-user-feedbacks.component.ts
ronitjadhav: Yes, agree.
Reverted it back to the original.

jahow:
Also **the libs should not depend on the app-config lib** (only apps should depend on it); I think currently there's no linting rule against this but this is technical debt, as an app config is not always available in every context.

libs/ui/search/src/lib/record-preview-feed/record-preview-feed.component.html
```diff
@@ -49,7 +49,7 @@
       </div>
     </div>
     <div class="pt-5 pb-5 px-10 relative">
-      <div class="absolute top-[0.85em] right-[0.85em]">
+      <div class="absolute top-[0.85em] right-[0.85em]" *ngIf="showFavorites">
```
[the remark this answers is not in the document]
ronitjadhav: image

translations/en.json
```diff
@@ -475,6 +475,7 @@
   "record.metadata.otherConstraints": "Additional Conditions",
   "record.metadata.owner": "Catalog of origin",
   "record.metadata.preview": "Preview",
+  "record.metadata.preview.config.": "",
```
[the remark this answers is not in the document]
jahow: yes, figure out where it comes from and fix it there :)

PR #1376, opened 2025-10-10
Use `prepareExternalUrl` in `AuthService` to handle application base href

f-necas:
### Description

This PR fixes a change introduced while upgrading Angular version:
https://github.com/geonetwork/geonetwork-ui/commit/9b561bdd865c00e5e4884a529861bf917737971d#diff-a8cc3af3857477c6843f5f9a8853413d96800709fe5418693a2097bbcd8bef12R36

The URL constructed from the Angular location path is not working when the application is deployed with a base href.

Angular provides the `prepareExternalUrl` function for such cases.

### How to test

Tests should be run locally both without a base href (http://localhost:4200) and with a base href (http://localhost:8080/application).

On the Datahub, check:
- login from favorite (search result or record page)
- login from Q/A section on a record page
- use of the proxy for the calls to outside services in the dataviz

On the Metadata Editor, check:
- logout
- login from a deep route

For example directly accessing http://localhost:8080/metadata-editor/edit/uuid when not already logged in should redirect to http://localhost:8080/metadata-editor/edit/uuid after login.

Follow the instructions in "docs/guide/run.md#with-docker" to check with an application deployed on "http://localhost:8080/application/".

To simulate the deployment of the Datahub on a path such as http://localhost/datahub you can try injecting the APP_BASE_HREF token, see: https://angular.dev/api/common/APP_BASE_HREF

(this token is set automatically when the app is built with --base=/datahub/)

### DIFF
--- a/libs/api/repository/src/lib/gn4/auth/auth.service.spec.ts
+++ b/libs/api/repository/src/lib/gn4/auth/auth.service.spec.ts
@@ -10,12 +10,14 @@ const translateServiceMock = {
   currentLang: 'fr',
 }
 
-let windowLocation
+let mockAppBaseHref
+let mockLocationPath
 
 describe('AuthService', () => {
   let service: AuthService
   beforeEach(() => {
-    windowLocation = 'http://localhost'
+    mockAppBaseHref = ''
+    mockLocationPath = '/'
 
     TestBed.configureTestingModule({
       providers: [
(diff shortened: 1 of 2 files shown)


LHBruneton-C2C:
> Then maybe we can use `this.location.prepareExternalUrl(this.location.path())` ([doc](https://angular.dev/api/common/Location#prepareExternalUrl))?

@jahow Thanks for the tip, that did the trick!

I updated the PR description, and kept only the diff that were needed, as the path with the base is used only for the auth service when generating the redirect URL.

jahow (approved):
Thank you!

PR #1386, opened 2025-10-21
[Datahub]: Handle STAC protocol

AlexianMasson:
### Description

This PR introduces **STAC (SpatioTemporal Asset Catalog) protocol support** in geonetwork-ui, enabling the application to properly recognize, classify, and display STAC API links as part of dataset metadata.

The implementation adds STAC as a new service protocol type that gets classified as an API usage, allowing users to access STAC endpoints directly from dataset pages.

### Architectural changes

The following libraries have been modified to support STAC protocol:

- **`libs/common/domain`**: Added `stac` as a new `ServiceProtocol` type in the metadata model
- **`libs/util/shared`**: Enhanced `LinkClassifierService` to classify STAC links as API usage
- **`libs/api/metadata-converter`**: Updated distribution mapper to handle "STAC Items" protocol strings
- **`libs/common/fixtures`**: Added STAC link fixtures for testing

### Screenshots

image

### How to test

Launch geonetwork and datahub locally then go to this page: http://localhost:4200/dataset/zzz_nl_test_wfs_syth_la_ciotat

You should see the STAC link inside one of the API links.

### Notes

This work is sponsored by [INRAE](https://www.inrae.fr/).

### DIFF
--- a/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
+++ b/libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
@@ -513,7 +513,7 @@ export class Gn4FieldMapper {
       /TMS/i.test(protocol) ||
       /ogc\W*api\W*features/i.test(protocol) ||
       /^DB:POSTGIS/i.test(protocol) ||
-      /^STAC/.test(protocol) ||
+      /STAC/i.test(protocol) ||
       (/^WWW:DOWNLOAD-/.test(protocol) && /data.geopf.fr/.test(url)) // TO DO : change with the good protocol when decided
     ) {
       return 'service'


libs/api/metadata-converter/src/lib/common/distribution.mapper.ts
```diff
@@ -10,6 +10,7 @@ export function matchProtocol(protocol: string): ServiceProtocol {
   if (/esri/i.test(protocol)) return 'esriRest'
   if (/DOWNLOAD-1/i.test(protocol)) return 'GPFDL'
   if (/DB:POSTGIS/i.test(protocol)) return 'postgis'
+  if (/stac\W*items/i.test(protocol)) return 'stac'
```
jahow: see my other comment, the match should only be on `/stac/i` here

libs/api/metadata-converter/src/lib/iso19139/write-parts.ts
```diff
@@ -255,6 +255,8 @@ export function getServiceDistributionProtocol(
       return 'OGC:WMS'
     case 'wps':
       return 'OGC:WPS'
+    case 'stac':
+      return 'STAC Items'
```
jahow: See my other comments
```suggestion
      return 'STAC'
```

AlexianMasson: Out of scope, I'm gonna remove this change.

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -513,6 +513,7 @@ export class Gn4FieldMapper {
       /TMS/i.test(protocol) ||
       /ogc\W*api\W*features/i.test(protocol) ||
       /^DB:POSTGIS/i.test(protocol) ||
+      /^STAC/.test(protocol) ||
```
tkohr: As we are being as generic/permissive as possible for now, I think it would even make sense to ignore the case and start of the character chain here and just test `/STAC/i` like we do for `TMS` and `OGC API FEATURES`.

AlexianMasson: Done

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -513,6 +513,7 @@ export class Gn4FieldMapper {
       /TMS/i.test(protocol) ||
       /ogc\W*api\W*features/i.test(protocol) ||
       /^DB:POSTGIS/i.test(protocol) ||
+      /^STAC Items/.test(protocol) ||
```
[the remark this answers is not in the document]
LHBruneton-C2C: The harvester is currently producing links to the source collection entrypoint, with a protocol of "STAC Collection". This protocol will be changed to a simple "WWW:LINK-1.0-http--link" as the link is only kept for further reference.

The Datahub should not extract the items links from the STAC Collection, but instead the harvester will extract a new link with the protocol "STAC Items".

We are being specific on the protocol here, as both entrypoints deliver different features.

And we want to keep extracting both collection and items links. That's why I still think identifying "STAC Items" in the same manner as "OGC API Features" would be best.

PR #1394, opened 2025-10-27
Enhance nginx configuration for caching and gzip

f-necas:
Updated gzip settings and Cache-Control headers for better performance.

Index.html is not cached as it controls versioning

### Description

This PR introduces Gzip and cache control for docker images as default. 

Time has been incresed to 1 year for js resources as thez are invalidated if index.html change.

### DIFF
--- a/tools/docker/nginx.apps.conf
+++ b/tools/docker/nginx.apps.conf
@@ -10,17 +10,16 @@ server {
 
         gzip on;
         gzip_types text/plain text/css application/javascript application/json image/svg+xml;
-        gzip_min_length 256;
 
 
         location ~ /index.html|.*\.toml|.*\.json$ {
           expires -1;
-          add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0' always;
+          add_header Cache-Control 'no-store, no-cache, must-revalidate, proxy-revalidate, max-age=0';
         }
 
         location ~ .*\.css$|.*\.js$ {
           # cached for 1 year because hash ensures cache invalidation when content changes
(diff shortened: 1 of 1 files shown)


tools/docker/nginx.apps.conf
```diff
@@ -4,16 +4,23 @@ server {
         server_name  localhost;
 
         root   /usr/share/nginx/html;
+        index  index.html;
 
         server_tokens off;
 
+        gzip on;
+        gzip_types text/plain text/css application/javascript application/json image/svg+xml;
+        gzip_min_length 256;
```
jahow: not sure if we need this

tools/docker/nginx.apps.conf
```diff
@@ -22,8 +29,7 @@ server {
 
         location / {
           try_files $uri $uri/ /APP_NAME/index.html;
-
-          add_header Cache-Control 'max-age=86400'; # 24h
+          add_header Cache-Control "no-cache" always;
```
jahow: I don't think this is used at all since `try_files` does a redirect on the other locations. We might as well get rid of it I guess?

tools/docker/nginx.apps.conf
jahow: (the gzip_min_length I mean)

LHBruneton-C2C:
> Doing a longer cache on JS and CSS files would make sense though, as we are 100% sure that they're immutable.

The PR you mention describes:

> This avoids to have deprecated resources in cache for too long.

As we are now planning a release every ~3 month, we could set the cache duration to 4 month for all immutable resources.

PR #1412, opened 2025-11-14
chore(i18n): clean data services error translation keys

LHBruneton-C2C:
This PR introduces specific translation keys for tms and stac endpoints errors.

It also renames dataset error translation keys that were at some point converted to wfs errors.


tkohr (approved):
Thanks for the clean-up @LHBruneton-C2C ! Watch out, one UT needs to be adapted to pass.

jahow (commented):
I'm not sure I get it; why have duplicated keys for each service type? Isn't that going to add exponentially more clutter as we add other protocols?

LHBruneton-C2C:
> I'm not sure I get it; why have duplicated keys for each service type? Isn't that going to add exponentially more clutter as we add other protocols?

I agree, this is the reason why I never merged even though it's been approved.

The other solution would be to rename existing keys and translations to be a bit more generic, and not use `ogc` as key for STAC and TMS.

PR #1428, opened 2025-12-03
implement background filtering in search input (fix #1427)

### LINKED ISSUE #1427: Add background filter support for search input web components
### Description

As discussed in the [OSGeo Discourse thread](https://discourse.osgeo.org/t/embedding-geonetwork-on-another-site/148198/8), there is a need to allow web components to have "built-in filters" that restrict search results to specific subsets of metadata when embedding gnui components on external sites.

Currently, when using the `gn-search-input` web component, there's no way to apply background filters that would automatically restrict both search results and autocomplete suggestions to a specific organization, group, or other criteria. This limitation makes it challenging to embed focused search interfaces on partner sites.

Use Case
A typical scenario would be:

An organization wants to embed a GeoNetwork search input on their website
They want search results and autocomplete suggestions to be automatically filtered to show only their organization's metadata records such as `{"groupOwner":{"xx":true}}` or even more complex queries.

The filtering should be transparent to end users (applied in the background)

maxcollombin:
fix #1427

### DIFF
--- a/apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
+++ b/apps/webcomponents/src/app/components/gn-search-input/gn-search-input.component.ts
@@ -31,8 +31,7 @@ import { of } from 'rxjs'
 })
 export class GnSearchInputComponent
   extends BaseComponent
-  implements AfterViewInit
-{
+  implements AfterViewInit {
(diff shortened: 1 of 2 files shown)


maxcollombin:
I am closing this PR since a workaround could finally be found (#1427).

PR #1433, opened 2025-12-11
Fix search and results wc records catalog redirection

### LINKED ISSUE #1407: Webcomponents demo for search and results - can't navigate to a suggestion
### Describe the bug

On the search and results webcomponents demo page, clicking on an autocomplete suggestion doesn't navigate to this record.

### Expected behaviour

Clicking on an autocomplete suggestion should redirect to this record in the catalog based on the search webcomponent attributes.

### Steps To Reproduce

- Visit the demo page for the search and results webcomponents:
  https://geonetwork.github.io/geonetwork-ui/main/demo/webcomponents/gn-search-input-and-results.sample.html
- Type a search, for example "eau"
- Check that autocomplete suggestions are displayed
- Click on a suggestion
- See that the redirection to the record does not happen

maxcollombin:
### Description
 Fix #1407


LHBruneton-C2C (approved):
Thanks for the clean git history and the fix!

PR #1465, opened 2026-01-28
[DataHub]: Support search by reuse without cl_presentationForm

LHBruneton-C2C:
### Description

This PR is a follow up on https://github.com/geonetwork/geonetwork-ui/pull/1415.

The search filter on `recordKind` was mixing rules on both `recordType` and `cl_presentationForm`, in a way that made `cl_presentationForm` mandatory when searching for `reuse` only.

The new rule now supports both `dataset` and `document` `resourceType` with a `mapDigital` or `mapHardCopy` `cl_presentationForm`, and other `reuse` types with or without `cl_presentationForm`.

### Architectural changes

There is now a direct logical dependency between `feature-search` and `api-repository`, on the `gn-ui-crossFieldFilter`.
This should be improved, but needs heavy changes on the filter model, to allow for multiple fields with an `OR` logic on one filter criteria.

### How to test

Create a map or application record, without a `cl_presentationForm`.
In the quick `recordKind` search, select `reuse`.
Make sure the newly created record is present in the results.
Make sure the previous `dataset` or `document` record reclassified as `reuse` due to their `cl_presentationForm` are still present too. 

---

**This work is sponsored by [GeoRhena](https://geoportal.georhena.eu/)**.


tkohr (commented):
Thanks @LHBruneton-C2C! As you already mentioned this is not pretty but it works. I don't see a simple cleaner solution neither. 

You may need to make a small adjustment in the [pipeline](https://github.com/geonetwork/geonetwork-ui/blob/dh-reuse-search-presentation-form-optional/tools/pipelines/register-es-pipelines.js#L74-L84) as well to keep the quality score calculation for reuse inline with these changes.

PR #1525, opened 2026-03-19
Attempt at reducing the bundle sizes & removing dependencies

jahow:
### Description

This PR is a general attempt to reduce the bundles sizes (Datahub, Webcomponents...).

What's been done:
* Remove some dependencies: axios (used fetch instead), basiclightbox (used cdk overlay instead; a storybook entry for the image preview component was added for that purpose)
* ~Moved to esbuild for the build steps to have better tree-shaking~
* Avoid including all the default material indigo CSS; instead we use an extract of this theme that's enough for gn-ui to work
* Removed an unused material outlined font everywhere; ng-icons relies on SVG so it's not needed anymore

> Note: moving to esbuild was postponed because it breaks the webcomponents bundle; because Esbuild only produces ESM, it's impossible to merge all the ESM modules once the `webcomponents` app is built to obtain a single bundle... this will need more investigation

Also:
* Updated many dependencies in the project, including OpenLayers, ngx-translate (this required a migration)
* Removed the unused data-platform app which brought many dependencies with it (nestjs etc.)
* Removed other unused libs in the project, like moment, pg, ng-table-virtual-scroll, etc.; this has no impact on the bundles, just makes for a cleaner project overall

### Architectural changes

Many changes to the root package.json, see above.

Made it so that the data-fetcher lib cannot be built anymore, hopefully this will speed up build times a little.

### Screenshots

none

- [x] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test

All these commands should work:

```bash
npm install
npx nx build datahub
npm run build:webcomponents
node package/generate-package.js
```


tailwind.base.css
```diff
@@ -1,3 +1,5 @@
+@import './material-styles.css';
```
ronitjadhav, line 1: This import breaks the npm package for consumers: `generate-package.js` copies `tailwind.base.css` into the package dist but not `material-styles.css`, so the import can't resolve. 

Tested with a scratch Angular app (npm pack the dist, install, import geonetwork-ui/style.css):

package.json
```diff
@@ -124,23 +107,21 @@
...
-    "@nestjs/schematics": "11.0.9",
-    "@nestjs/testing": "11.1.19",
+    "@fetch-mock/jest": "~0.2.20",
     "@ngrx/component": "20.1.0",
     "@ngrx/effects": "20.1.0",
     "@ngrx/router-store": "20.1.0",
     "@ngrx/schematics": "20.1.0",
     "@ngrx/store": "20.1.0",
     "@ngrx/store-devtools": "20.1.0",
+    "@nx/angular": "~22.6.0",
```
ronitjadhav, line 114: Other @nx/* packages are still at 22.0.4, and `npx nx report` flags it:
The following packages should match the installed version of nx
   - @nx/angular@22.6.5
   - @nx/module-federation@22.6.5
   - @nx/web@22.6.5
   - @nx/webpack@22.6.5

 To fix this, I guess we can run `nx migrate nx@22.6.5`

ronitjadhav (commented):
Went through the PR and tested a few things.

 - `typeorm` and `express/@types/express` are still in the root `package.json`, but with `data-platform` gone I couldn't find any remaining imports of either. Can they be dropped too, or is something else using them that I'm missing?

 - `libs/util/data-fetcher/package.json` (`@geonetwork-ui/data-fetcher@0.0.1`)  with the lib's build target removed, this file looks orphaned. Should it go as well?

tailwind.base.css
jahow: good catch thanks! I'll fix this

package.json
jahow: You're right! I think I'm going to postpone the NX migration for now, this should definitely be done though. but it's not trivial

jahow:
Thanks for the first review @ronitjadhav, i've addressed all comments

PR #1563, opened 2026-05-04
Editor lib: Add DISABLE_DRAFT token to allow disabling the draft system

tkohr:
### Description

This PR introduces a DISABLE_DRAFT injection token at the repository level so external apps embedding the editor library can opt out of the localStorage draft system.

Note: The new setting primarily targets external apps using the lib and does not affect the UI of the gn-ui metadata editor app.

### How to test

- Check that modifications of a metadata record are not saved in the draft system anymore when setting `disableDraft: true` where in `provideGn4`.
- Behavior can be tested it in the gn-ui metadata-editor app or an external editor like the datafeeder.

---


jahow (commented):
I would suggest putting this option at the Repository level, such as here: https://github.com/geonetwork/geonetwork-ui/blob/main/libs/api/repository/src/lib/gn4/gn4.provider.ts

The GN4 repository already has a couple of options when it's being provided, and it doesn't necessitate as much changes downstream

tkohr:
> I would suggest putting this option at the Repository level, such as here: https://github.com/geonetwork/geonetwork-ui/blob/main/libs/api/repository/src/lib/gn4/gn4.provider.ts
> 
> The GN4 repository already has a couple of options when it's being provided, and it doesn't necessitate as much changes downstream

Oh good point, thanks for the architectural alignment! It has been adapted.

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -350,9 +355,28 @@ export class Gn4Repository implements RecordsRepositoryInterface {
...
+                  [record, recordAsXml, true] as [
+                    CatalogRecord,
+                    string,
+                    boolean,
+                  ]
+              )
+          )
+        )
+      )
+    }
```
jahow: This is duplicated code; you can probably use something like

```typescript
const draft$ = this.disableDraft ? of(null) : of(this.getRecordFromLocalStorage(uniqueIdentifier))
```

tkohr: Oups, good catch! I've adapted this as well as the test which as the only one was actually returning false positives.

jahow (approved):
thanks!

PR #1564, opened 2026-05-06
Upgrade ogc-client

tkohr:
### Description

This PR upgrades ogc-client to make the `setCacheExpiryDuration` and `getCacheExpiryDuration` functions available.

### How to test

Check if the functions can now be imported properly.


jahow (commented):
I think you also need to update it in the package/package.json file

tkohr:
> I think you also need to update it in the package/package.json file

Oh good catch, thanks ! I've updated the commit.

jahow (approved):

PR #1570, opened 2026-05-18
[Metadata-Editor] Allow deleting last contact and show empty-state warning immediately

ronitjadhav:
### Description

This PR fixes issues in the Metadata Editor contact fields:

1. In the **Metadata Contacts** field, the delete button was hidden when only one contact remained. Users were unable to delete the last contact.
2. In the **Data Responsible Party** field, deleting the last contact did not immediately show the "no contact" warning, a page refresh was required.
3. In the **Metadata Contacts** autocomplete, users with no name set were displayed as `"undefined undefined"` in the search results after adding the contact.

### Screenshots

BeforeAfterLandscape_contacts_1

BeforeAfterLandscape_contacts_2

### How to test

1. Open the Metadata Editor and load a record that has exactly **one** metadata point of contact (Contacts section).
   - Confirm a delete button is now visible on the single contact.
   - Click delete and confirm the "Please provide at least one point of contact" warning appears **immediately** without a page refresh.

2. Load a record that has Data managers contact.
   - Delete it delete button.
   - Confirm the "Please provide at least one point of contact responsible for the data" warning appears **immediately**.
   - Confirm the role picker buttons are fully restored (all roles available to add again).

3. In the Metadata Contacts autocomplete, search for a user that has no first/last name set (eg: admin or Barbara Roberts)
   - Confirm the result no longer shows `"undefined undefined"` after adding the contact.


libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts/form-field-contacts.component.ts
```diff
@@ -117,9 +117,11 @@ export class FormFieldContactsComponent implements OnDestroy, OnChanges {
...
    */
   displayWithFn: (user: UserModel) => string = (user) =>
-    `${user.name} ${user.surname} ${
-      user.organisation ? `(${user.organisation})` : ''
-    }`
+    user.name
+      ? `${user.name} ${user.surname} ${
+          user.organisation ? `(${user.organisation})` : ''
+        }`
+      : ``
```
jahow: this could be a separate util and used in other places where we want to display a contact using a single line text :)

libs/feature/editor/src/lib/components/record-form/form-field/form-field-contacts-for-resource/form-field-contacts-for-resource.component.ts
```diff
@@ -67,25 +68,29 @@ export class FormFieldContactsForResourceComponent
...
+  rolesToPick: Role[] = this.getPickableRoles()
 
   roleSectionsToDisplay: Role[] = []
 
   allOrganizations: Map<string, Organization> = new Map()
 
   ngOnChanges() {
+    if (this.value.length === 0) {
+      this.roleSectionsToDisplay = []
+      this.rolesToPick = this.getPickableRoles()
```
jahow: I'm not sure about this behavior; if we delete the last contact then all existing sections will be suddenly deleted. Can't we show the warning without having this potentially surprising behavior?

libs/util/shared/src/lib/utils/user-display.ts
```diff
@@ -0,0 +1,7 @@
+import { UserModel } from '@geonetwork-ui/common/domain/model/user'
+
+export function getUserDisplayName(user: UserModel): string {
+  if (!user.name) return ''
```
jahow: fallback to somethig else would be good here, instead of nothing. And even if there's no name there can be a surname, or an organization, or maybe something else

jahow (approved):
looking good, thanks! one good thing would be to have a better fallback when no name is present, if you have any idea for this? I'm approving so you can decide what's best and merge, thanks

libs/util/shared/src/lib/utils/user-display.ts
ronitjadhav: Updated! If no name/surname, falls back to organisation, then username, then email.

PR #1571, opened 2026-05-18
[Metadata-Editor] Navigate to field from quality panel

tkohr:
### Description

  When a user clicks an invalid criterion in the metadata quality panel, the editor now navigates to the page containing that field and scrolls it into view.

  ### Changes

  - **`EditorFacade`** — adds `navigateToQualityField(page, model)` and a `pendingScrollToField$` signal to coordinate navigation and scroll across components
  - **`MetadataQualityPanelComponent`** — invalid criteria are now clickable; `pageIndex` is tracked per criterion so the right page can be targeted; fields without a form field (`organisation`) get `pageIndex: -1` and remain non-navigable
  - **`record-form.component.html`** — adds `[id]` to each `gn-ui-form-field` so fields can be located in the DOM after page navigation
  - **`EditPageComponent`** — subscribes to `pendingScrollToField$`, uses `afterNextRender` to scroll the container to top then to the target field once the new page has rendered
  - Fixes `topics` incorrectly treated as a future field — it already has a form field in the editor

  ### Tests

  Unit tests for facade, quality panel, and edit-page; e2e tests covering navigation to an invalid criterion, no navigation on valid criteria, and no navigation on future fields.

### How to test

Click on fields in the editor's metadata quality panel and check if the click navigates to the correct field.

---

### DIFF
--- a/apps/metadata-editor/src/app/edit/edit-page.component.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.ts
@@ -38,7 +38,8 @@ import { TopToolbarComponent } from './components/top-toolbar/top-toolbar.compon
 import { SpinningLoaderComponent } from '@geonetwork-ui/ui/widgets'
 import { SearchHeaderComponent } from '../dashboard/search-header/search-header.component'
 import { PageErrorComponent } from './components/page-error/page-error.component'
-import { DateService, ValidatorMapperKeys } from '@geonetwork-ui/util/shared'
+import { DateService } from '@geonetwork-ui/util/shared'
+import { CatalogRecordKeys } from '@geonetwork-ui/common/domain/model/record'
 
 marker('editor.record.form.bottomButtons.comeBackLater')
 marker('editor.record.form.bottomButtons.previous')
@@ -181,9 +182,9 @@ export class EditPageComponent implements OnInit, OnDestroy {
         .pipe(
           filter((field) => !!field),
           switchMap(async (field) => ({
-            field: field as ValidatorMapperKeys,
+            field: field as CatalogRecordKeys,
             pageIndex: await this.getPageIndexForField(
-              field as ValidatorMapperKeys
+              field as CatalogRecordKeys
             ),
           }))
         )
@@ -245,13 +246,11 @@ export class EditPageComponent implements OnInit, OnDestroy {
     }
   }
 
-  onCriterionClicked(model: ValidatorMapperKeys) {
+  onCriterionClicked(model: CatalogRecordKeys) {
     this.facade.setFocusedField(model)
   }
 
-  async getPageIndexForField(
-    model: ValidatorMapperKeys
-  ): Promise<number | null> {
+  async getPageIndexForField(model: CatalogRecordKeys): Promise<number | null> {
     const config = await firstValueFrom(this.facade.editorConfig$)
     const pageIndex = config.pages.findIndex((page) =>
       page.sections.some((section) =>
--- a/apps/metadata-editor/src/app/edit/edit-page.component.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.ts
@@ -179,10 +179,9 @@ export class EditPageComponent implements OnInit, OnDestroy {
       this.facade.pendingScrollToField$
         .pipe(filter((field) => !!field))
         .subscribe((field) => {
-          afterNextRender(
-            () => this.scrollToQualityField(field),
-            { injector: this.injector }
-          )
+          afterNextRender(() => this.scrollToQualityField(field), {
+            injector: this.injector,
+          })
         })
     )
 
@@ -238,7 +237,7 @@ export class EditPageComponent implements OnInit, OnDestroy {
     this.scrollContainer?.nativeElement.scroll({ top: 0, behavior: 'instant' })
     document
       .getElementById(field)
-      ?.scrollIntoView({ behavior: 'smooth', block: 'start' })
+      ?.scrollIntoView({ behavior: 'instant', block: 'start' })
     this.facade.clearPendingScrollField()
   }
 
--- a/apps/metadata-editor/src/app/edit/edit-page.component.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.ts
@@ -18,6 +18,7 @@ import {
   MetadataQualityPanelComponent,
   MultilingualPanelComponent,
   RecordFormComponent,
+  ANCHOR_ID_PREFIX,
 } from '@geonetwork-ui/feature/editor'
 import {
   NotificationsContainerComponent,
@@ -30,14 +31,14 @@ import {
   TranslateService,
 } from '@ngx-translate/core'
 import { combineLatest, filter, firstValueFrom, Subscription, take } from 'rxjs'
-import { map, skip } from 'rxjs/operators'
+import { map, skip, switchMap } from 'rxjs/operators'
 import { SidebarComponent } from '../dashboard/sidebar/sidebar.component'
 import { PageSelectorComponent } from './components/page-selector/page-selector.component'
 import { TopToolbarComponent } from './components/top-toolbar/top-toolbar.component'
 import { SpinningLoaderComponent } from '@geonetwork-ui/ui/widgets'
 import { SearchHeaderComponent } from '../dashboard/search-header/search-header.component'
 import { PageErrorComponent } from './components/page-error/page-error.component'
-import { DateService } from '@geonetwork-ui/util/shared'
+import { DateService, ValidatorMapperKeys } from '@geonetwork-ui/util/shared'
 
 marker('editor.record.form.bottomButtons.comeBackLater')
 marker('editor.record.form.bottomButtons.previous')
@@ -176,9 +177,20 @@ export class EditPageComponent implements OnInit, OnDestroy {
     )
 
     this.subscription.add(
-      this.facade.pendingScrollToField$
-        .pipe(filter((field) => !!field))
-        .subscribe((field) => {
+      this.facade.focusedField$
+        .pipe(
+          filter((field) => !!field),
+          switchMap(async (field) => ({
+            field: field as ValidatorMapperKeys,
+            pageIndex: await this.getPageIndexForField(
+              field as ValidatorMapperKeys
+            ),
+          }))
+        )
+        .subscribe(({ field, pageIndex }) => {
+          if (pageIndex !== null) {
+            this.facade.setCurrentPage(pageIndex)
+          }
           afterNextRender(() => this.scrollToQualityField(field), {
             injector: this.injector,
           })
@@ -233,12 +245,26 @@ export class EditPageComponent implements OnInit, OnDestroy {
     }
   }
 
+  onCriterionClicked(model: ValidatorMapperKeys) {
+    this.facade.setFocusedField(model)
+  }
+
+  async getPageIndexForField(
+    model: ValidatorMapperKeys
+  ): Promise<number | null> {
+    const config = await firstValueFrom(this.facade.editorConfig$)
+    const pageIndex = config.pages.findIndex((page) =>
+      page.sections.some((section) =>
+        section.fields.some((field) => field.model === model)
+      )
+    )
+    return pageIndex >= 0 ? pageIndex : null
+  }
+
   scrollToQualityField(field: string) {
-    this.scrollContainer?.nativeElement.scroll({ top: 0, behavior: 'instant' })
     document
-      .getElementById(field)
+      .getElementById(ANCHOR_ID_PREFIX + field)
       ?.scrollIntoView({ behavior: 'instant', block: 'start' })
-    this.facade.clearPendingScrollField()
   }
 
   private scrollToTop() {
--- a/apps/metadata-editor/src/app/edit/edit-page.component.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.ts
@@ -2,9 +2,11 @@ import { CommonModule } from '@angular/common'
 import {
   Component,
   ElementRef,
+  Injector,
   OnDestroy,
   OnInit,
   ViewChild,
+  afterNextRender,
   inject,
 } from '@angular/core'
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner'
@@ -71,6 +73,7 @@ export class EditPageComponent implements OnInit, OnDestroy {
   private translateService = inject(TranslateService)
   private router = inject(Router)
   private dateService = inject(DateService)
+  private injector = inject(Injector)
 
   subscription = new Subscription()
 
@@ -172,6 +175,17 @@ export class EditPageComponent implements OnInit, OnDestroy {
       })
     )
 
+    this.subscription.add(
+      this.facade.pendingScrollToField$
+        .pipe(filter((field) => !!field))
+        .subscribe((field) => {
+          afterNextRender(
+            () => this.scrollToQualityField(field),
+            { injector: this.injector }
+          )
+        })
+    )
+
     // if we're on the /duplicate route, go to /edit/{uuid} to update the uuid
     if (this.route.snapshot.routeConfig?.path.includes('duplicate')) {
       this.router.navigate(['edit', currentRecord.uniqueIdentifier], {
@@ -218,6 +232,14 @@ export class EditPageComponent implements OnInit, OnDestroy {
       this.facade.setCurrentPage(currentPage + 1)
       this.scrollToTop()
     }
+  }
+
+  scrollToQualityField(field: string) {
+    this.scrollContainer?.nativeElement.scroll({ top: 0, behavior: 'instant' })
+    document
+      .getElementById(field)
+      ?.scrollIntoView({ behavior: 'smooth', block: 'start' })
+    this.facade.clearPendingScrollField()
   }
 
   private scrollToTop() {
--- a/apps/metadata-editor/src/app/edit/edit-page.component.ts
+++ b/apps/metadata-editor/src/app/edit/edit-page.component.ts
@@ -2,11 +2,9 @@ import { CommonModule } from '@angular/common'
 import {
   Component,
   ElementRef,
-  Injector,
   OnDestroy,
   OnInit,
   ViewChild,
-  afterNextRender,
   inject,
 } from '@angular/core'
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner'
@@ -18,7 +16,6 @@ import {
   MetadataQualityPanelComponent,
   MultilingualPanelComponent,
   RecordFormComponent,
-  ANCHOR_ID_PREFIX,
 } from '@geonetwork-ui/feature/editor'
 import {
   NotificationsContainerComponent,
@@ -31,7 +28,7 @@ import {
   TranslateService,
 } from '@ngx-translate/core'
 import { combineLatest, filter, firstValueFrom, Subscription, take } from 'rxjs'
-import { map, skip, switchMap } from 'rxjs/operators'
+import { map, skip } from 'rxjs/operators'
 import { SidebarComponent } from '../dashboard/sidebar/sidebar.component'
 import { PageSelectorComponent } from './components/page-selector/page-selector.component'
 import { TopToolbarComponent } from './components/top-toolbar/top-toolbar.component'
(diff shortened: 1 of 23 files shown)


libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,46 +44,68 @@ marker('editor.record.form.field.organisation')
...
+    pageIndex: number
+  }) {
+    if (!property.value && property.pageIndex >= 0) {
+      this.facade.navigateToQualityField(property.pageIndex, property.model)
+    }
+  }
+
+  getExtraClass(checked: boolean, pageIndex: number): string {
+    const isClickable = !checked && pageIndex >= 0
+    const baseClasses = `flex flex-row justify-between rounded mb-1 h-[34px] w-full focus:ring-0 hover:border-none border-none hover:text-black text-black ${isClickable ? 'cursor-pointer' : 'cursor-default'}`
```
jahow: why add a cursor-pointer on a button? I don't think it's needed

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,46 +44,68 @@ marker('editor.record.form.field.organisation')
   styleUrl: './metadata-quality-panel.component.css',
 })
 export class MetadataQualityPanelComponent implements OnChanges {
+  private readonly facade = inject(EditorFacade)
```
jahow: This doesn't feel right; the component takes a record and config as input, but it's now also injecting the facade. It's either one or the other

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,46 +44,68 @@ marker('editor.record.form.field.organisation')
...
+
   propsToValidate: ValidatorMapperKeys[] = getAllKeysValidator()
-  propertiesByPage: { label: string; value: boolean }[][] = []
+  propertiesByPage: {
+    label: string
+    value: boolean
+    model: ValidatorMapperKeys
+    pageIndex: number
```
jahow: page index on an array of objects by page looks really strange

libs/feature/editor/src/lib/+state/editor.facade.ts
```diff
@@ -77,6 +82,15 @@ export class EditorFacade {
...
   }
 
+  navigateToQualityField(page: number, model: string) {
+    this.setCurrentPage(page)
+    this._pendingScrollToField$.next(model)
+  }
+
+  clearPendingScrollField() {
+    this._pendingScrollToField$.next(null)
+  }
```
jahow: this is strange; the facade could simply offer an event called focusedField or something, no need to clear it

jahow: and why specify the page? the facade has the config already

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -172,6 +175,16 @@ export class EditPageComponent implements OnInit, OnDestroy {
...
     )
 
+    this.subscription.add(
+      this.facade.pendingScrollToField$
+        .pipe(filter((field) => !!field))
+        .subscribe((field) => {
+          afterNextRender(() => this.scrollToQualityField(field), {
+            injector: this.injector,
+          })
+        })
```
jahow: this looks hacky; why not use ViewportScroller?

libs/feature/editor/src/lib/components/record-form/record-form.component.html
```diff
@@ -24,6 +24,7 @@
           ) {
             @if (!field.config.hidden) {
               <gn-ui-form-field
+                [id]="field.config.model"
```
jahow: this is way too generic; having an element with `id="name"` is the guarantee to have conflicts in the future

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,46 +44,68 @@ marker('editor.record.form.field.organisation')
...
-            value: validator(),
-          }))
+        .map(({ pageIndex, models }) =>
+          getQualityValidators(this.record, models).map(
+            ({ name, validator }) => ({
+              label: `editor.record.form.field.${name}`, // use same translations as in fields.config.ts
+              value: validator(),
+              model: name as ValidatorMapperKeys,
+              // FIXME: see above
+              pageIndex: name === 'organisation' ? -1 : pageIndex,
```
jahow: this is technical debt; do we need it?

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,46 +44,68 @@ marker('editor.record.form.field.organisation')
...
-      )
-      // FIXME: temporarily add topics and organisation to the first and third page
-      // as long as they are not handled by the editor
-      if (fieldsByPage.length > 0) {
-        fieldsByPage[0].includes('topics') || fieldsByPage[0].push('topics')
-        fieldsByPage[2].includes('organisation') ||
-          fieldsByPage[2].push('organisation')
+        ),
+      }))
+      // FIXME: organisation is not yet a form field in the editor; show it on page 2 as non-navigable
```
jahow: the organisation field is deduced from the main contact of the record; it should point to contacts then I guess

jahow (changes requested):
Thanks, please see comments!

libs/feature/editor/src/lib/+state/editor.facade.ts
tkohr: Indeed I did a 
```
   const config = await firstValueFrom(this.editorConfig$)
   const pageIndex = config.pages.findIndex((page) =>
     page.sections.some((section) =>
       section.fields.some((field) => field.model === model)
     )
   )
```
in the facade at some point. Is this what you have in mind?

I had the feeling that it blew up the facade, but it may be the cleaner approach in the end.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
tkohr: I think there were some parts missing, but I'll double-check.

libs/feature/editor/src/lib/components/record-form/record-form.component.html
tkohr: Good point! Do you think adding a prefix is sufficient or do you think we should not rely at all on the id for scrolling? At some point I followed an approach to handle the srcoll via `ViewChild` but this finally seemed more complicated than a simple `document.getElementById`.

jahow: using an id is fine as long as it has a prefix; look out for other prefixes used elsewhere, maybe `gn-ui--name` for instance?

libs/feature/editor/src/lib/+state/editor.facade.ts
jahow: I think this is too much logic for the facade indeed; it should not have any logic. Any component down the line can use the config to know on which page is the focused component.

apps/metadata-editor/src/app/edit/edit-page.component.ts
tkohr: I gave it a try but `ViewportScroller` seems to be meant for `window` scrolls and not sub containers.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
tkohr: This line is not there anymore now, but I still need to check if we still need the organisation workaround above and do some testing on my side.

libs/feature/editor/src/lib/+state/editor.facade.ts
```diff
@@ -77,6 +79,10 @@ export class EditorFacade {
     this.store.dispatch(EditorActions.setCurrentPage({ page }))
   }
 
+  setFocusedField(model: ValidatorMapperKeys | null) {
```
jahow: the model should be a `CatalogRecordKeys`

libs/feature/editor/src/lib/+state/editor.reducer.ts
```diff
@@ -24,6 +27,7 @@ export interface EditorState {
   changedSinceSave: boolean
   editorConfig: EditorConfig
   currentPage: number
+  focusedField: ValidatorMapperKeys | null
```
jahow: this should not be kept in the state? it is merely an event

jahow: there can still be an action attached, but there's no need to store it in the state I think.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -44,9 +50,14 @@ marker('editor.record.form.field.organisation')
 })
 export class MetadataQualityPanelComponent implements OnChanges {
   propsToValidate: ValidatorMapperKeys[] = getAllKeysValidator()
-  propertiesByPage: { label: string; value: boolean }[][] = []
+  propertiesByPage: {
+    label: string
+    value: boolean
+    model: ValidatorMapperKeys
```
jahow: This should be CatalogRecordKeys

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -20,6 +20,7 @@ import { CatalogRecordKeys } from '@geonetwork-ui/common/domain/model/record'
   imports: [CommonModule, FormFieldComponent, TranslateDirective],
 })
 export class RecordFormComponent {
```
jahow: this can probably be the right place to listen to the focusedField event/action? and then focus the right component. I think it can be "built-in" the record form and not be the responsibility of the app (otherwise every app relying on an editor form will have to reimplement it).

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -220,6 +245,28 @@ export class EditPageComponent implements OnInit, OnDestroy {
     }
   }
 
+  onCriterionClicked(model: ValidatorMapperKeys) {
+    this.facade.setFocusedField(model)
+  }
```
jahow: same here, I think this isn't the responsability of the app to do this. Why not connect the metadata quality component to the facade and let it be smart and dispatch that action itself?

apps/metadata-editor/src/app/edit/edit-page.component.ts
```diff
@@ -172,6 +176,27 @@ export class EditPageComponent implements OnInit, OnDestroy {
...
+        )
+        .subscribe(({ field, pageIndex }) => {
+          if (pageIndex !== null) {
+            this.facade.setCurrentPage(pageIndex)
+          }
+          afterNextRender(() => this.scrollToQualityField(field), {
+            injector: this.injector,
+          })
+        })
+    )
```
jahow: this might be too much responsibility for the app; see my other comment on doing this in the RecordFormComponent

libs/feature/editor/src/lib/+state/editor.actions.ts
```diff
@@ -73,3 +74,8 @@ export const canEditRecord = createAction(
   '[Editor] User can edit record',
   props<{ canEditRecord: boolean }>()
 )
+
+export const setFocusedField = createAction(
+  '[Editor] Set focused field',
+  props<{ model: ValidatorMapperKeys | null }>()
```
jahow: Should be CatalogRecordKeys

jahow: Also, no need to allow focusing a "null" field; the focus happens instantly and does not need to be cleared

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -72,17 +80,29 @@ export class MetadataQualityPanelComponent implements OnChanges {
...
+    const isClickable = !checked && pageIndex >= 0
+    const baseClasses = `flex flex-row justify-between rounded mb-1 h-[34px] w-full focus:ring-0 hover:border-none border-none hover:text-black text-black`
+    if (checked) {
+      return `${baseClasses} bg-neutral-100 hover:bg-neutral-100 cursor-default`
+    }
+    // FIXME: for fields that are not yet handled by the editor (pageIndex -1)
+    if (isClickable) {
+      return `${baseClasses} bg-transparent hover:bg-gray-100`
+    }
+    return `${baseClasses} bg-transparent hover:bg-transparent cursor-default`
```
jahow: this logic is honestly quite obscure. Why so many conditions? besides, if a button has a cursor-default, it is still clickable right? I haven't tested recently but it looks like the approach isn't quite right. I'll let you judge on it though :)

libs/feature/editor/src/lib/models/editor-config.model.ts
```diff
@@ -1,5 +1,6 @@
 import { CatalogRecordKeys } from '@geonetwork-ui/common/domain/model/record'
 
+export const ANCHOR_ID_PREFIX = 'gn-ui--field-'
```
jahow: this isn't really a model per se, more like a constant. Which should IMO be better scoped inside the RecordformComponent instead of being published outside of this lib (see my other comments)

jahow (changes requested):
thank you, some more remarks!

tkohr:
Thanks for all the feedback @jahow ! I think I've addressed most of it. I'll just need to check on the expected behavior for the `organisation` field with @LHBruneton-C2C. The thing is that an "organisation" field does not (yet) exist in the editor (field config), but was asked for in the quality component (if I remember right).

Another thing worth noting is that place keywords are not visible in the editor form field, but have the effect that the field is checked in the quality component.

Also a small scroll offset might make sense for a better user experience.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -43,8 +47,13 @@ marker('editor.record.form.field.organisation')
   styleUrl: './metadata-quality-panel.component.css',
 })
 export class MetadataQualityPanelComponent implements OnChanges {
+  facade = inject(EditorFacade)
```
jahow, line 49: we're back to having both the facade injected and the record & config as inputs :) this is contradictory

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -40,4 +80,24 @@ export class RecordFormComponent {
...
+      )
+    )
+    return pageIndex >= 0 ? pageIndex : null
+  }
+
+  scrollToQualityField(field: string) {
+    this.el.nativeElement.scrollIntoView({
+      behavior: 'instant',
+      block: 'start',
+    })
```
jahow: why do we need this?

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
jahow: again I don't understand why we can't just point the organisation criteria to the metadata point of contact in the form.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
```diff
@@ -57,32 +66,35 @@ export class MetadataQualityPanelComponent implements OnChanges {
...
+
   getExtraClass(checked: boolean): string {
-    const baseClasses =
-      'flex flex-row justify-between rounded mb-1 h-[34px] w-full focus:ring-0 hover:border-none border-none hover:text-black text-black cursor-default'
-    return checked
-      ? `${baseClasses} bg-neutral-100 hover:bg-neutral-100`
-      : `${baseClasses} bg-transparent hover:bg-transparent`
+    const baseClasses = `flex flex-row justify-between rounded mb-1 h-[34px] w-full focus:ring-0 hover:border-none border-none hover:text-black text-black`
+    if (checked) {
+      return `${baseClasses} bg-neutral-100 hover:bg-neutral-100 cursor-default`
```
jahow: again, this is going to be clickable even if with a cursor-default; isn't that weird?

jahow (changes requested):

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
tkohr: Right, I got rid of the inputs now.

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
tkohr: Otherwise the scroll after changing pages does not works well. I've separated the calls differently so that this is only called now when changing the page.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
tkohr: True I got rig of this now using `disabled`.

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.ts
tkohr: I've added ea1e6241d4be3104f7f637a2f5f5c1fed73b240f to use the validator object (with an optional alias) to be the one to be filtered and get rid of the FIXMEs

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.html
```diff
@@ -7,11 +7,16 @@
...
     >
   </div>
+  @let propertiesByPage = (propertiesByPage$ | async) ?? [];
   @for (properties of propertiesByPage; track properties; let i = $index) {
     <div class="flex flex-col gap-2">
       @for (property of properties; track property) {
         <gn-ui-button
-          [extraClass]="getExtraClass(property.value)"
+          [extraClass]="
+            'justify-between rounded mb-1 h-[34px] w-full focus:ring-0 hover:border-none border-none hover:text-black text-black bg-transparent hover:bg-gray-100 disabled:opacity-100'
```
jahow: same thing as in #1572, the extra classes here are partly redundant with what the gn-ui-button offer. I think this is technical debt, because any change in gn-ui-button might break this as well in ways hard to anticipate. But I'll let you judge on that

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -19,13 +26,49 @@ import { CatalogRecordKeys } from '@geonetwork-ui/common/domain/model/record'
...
+            ),
+          }))
+        )
+        .subscribe(({ field, pageIndex }) => {
+          if (pageIndex !== null) {
+            this.facade.setCurrentPage(pageIndex)
+            this.el.nativeElement.scrollIntoView({
+              behavior: 'instant',
+              block: 'start',
+            })
```
jahow, line 58: are you sure this scroll is necessary? in my tests it causes a flash if we're already on the right page

libs/feature/editor/src/lib/components/metadata-quality-panel/metadata-quality-panel.component.html
tkohr: Indeed I was hesitating about this point. So I guess you'd get rid of `extraClass` entirely (in general) and add the missing CSS vars to the gn-ui-button? Or does `extraClass` still make sense (for very specific things such as `justify-between` for example) ? I must admit I still find the CSS var customization system pretty cumbersome.

jahow: justify-between is justified yes, the size (w and h) why not, but then why remove the focus ring and the border? why force the text color?

For the UI to be consistent we need all buttons to look kind of similar. this button currently does not look like any other.

tkohr: Ok I've cleaned up the styles removing some customization and getting rid of the `extraClass` completely. Regarding the ring/border/text color I think it was due to following the initial mockup, but we can get back to this when implementing the new mockups if necessary.

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
tkohr: Yes it's necessary when changing the page. However the code was actually missing a check if the page was being changed before applying this scroll, which I've added now.

jahow (approved):

PR #1572, opened 2026-05-19
[Datahub] Add "Contacts for the resource" expandable panel

ronitjadhav:
### Description

This PR introduces a new "Contacts for the resource" expandable panel to the metadata info, positioned between the "License and Conditions" and "About the data" panels. Contacts are grouped by role and displayed as pills in a responsive 2-column grid. Each pill shows the contact's display name (organization name, or first/last name, or email as fallback) 

### Architectural changes

Add `ContactPillComponent` (libs/ui/elements) which helps to render a single contact as a button.

### Screenshots

image

On Hover:
image

Mobile View:
image

### How to test

1. Open any dataset detail page in Datahub that has contacts
2. Verify a new "Contacts for the resource" expandable panel appears between the usage and details panels
3. Expand the panel: contacts should be grouped under their role label, displayed as pills in a 2-column grid
4. Hover over a pill to see the highlight effect and the tooltip on truncated names
5. Alternatively you can also test it in storybook.


libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -105,9 +105,47 @@
...
+  >
+    <div class="flex flex-col gap-1 pt-3 pb-4">
+      @for (group of contactGroups; track group.role) {
+        <div class="flex flex-col gap-1 rounded bg-gray-50 py-4 px-2">
+          <p class="text-xs font-normal text-black">
+            {{ getRoleTranslationKey(group.role) | translate }}
+          </p>
+          <div class="grid gap-0.5 grid-cols-1 md:grid-cols-2">
+            @for (contact of group.contacts; track contact.email) {
+              <div
```
jahow: this should probably be a separate container, for instance in libs/ui/elements, so that we can test it separately

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -132,12 +141,21 @@ export class MetadataInfoComponent {
     return getTemporalRangeUnion(temporalExtents, this.dateService)
   }
 
-  get shownOrganization() {
-    return this.metadata.ownerOrganization
+  get contactGroups(): ContactGroup[] {
+    return groupContactsByRole(this.metadata?.contactsForResource ?? [])
```
jahow: I don't think we need a separate util here, it's just a few lines of logic which are not used anywhere else.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -132,12 +141,21 @@ export class MetadataInfoComponent {
...
+  }
+
+  getRoleTranslationKey(role: string): string {
+    return RoleLabels.get(role) ?? 'domain.contact.role.other'
```
jahow: no need for this fallback I think?

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
```diff
@@ -132,12 +141,21 @@ export class MetadataInfoComponent {
...
 
-  get resourceContact() {
-    return this.metadata.contactsForResource?.[0]
+  getContactDisplayName(contact: Individual): string {
+    if (contact.organization?.name) return contact.organization.name
+    const first = contact.firstName?.trim()
+    const last = contact.lastName?.trim()
+    if (first && last) return `${first} ${last}`
+    if (first || last) return first || last
+    return contact.email || ''
```
jahow: this will be reworked later on once we show more information about the contact right?

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -105,9 +105,47 @@
...
+                class="group flex items-center justify-between min-h-12 w-full rounded bg-primary-white py-2 pr-4 pl-5 gap-3 cursor-pointer hover:bg-primary transition-colors duration-100"
+                data-test="contact-pill"
+              >
+                <span
+                  class="font-title font-medium text-base leading-none truncate group-hover:text-white"
+                  [title]="getContactDisplayName(contact)"
+                  >{{ getContactDisplayName(contact) }}</span
+                >
+                <button
```
jahow: use gn-ui-button here

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -105,9 +105,47 @@
...
+    <div class="flex flex-col gap-1 pt-3 pb-4">
+      @for (group of contactGroups; track group.role) {
+        <div class="flex flex-col gap-1 rounded bg-gray-50 py-4 px-2">
+          <p class="text-xs font-normal text-black">
+            {{ getRoleTranslationKey(group.role) | translate }}
+          </p>
+          <div class="grid gap-0.5 grid-cols-1 md:grid-cols-2">
+            @for (contact of group.contacts; track contact.email) {
+              <div
+                class="group flex items-center justify-between min-h-12 w-full rounded bg-primary-white py-2 pr-4 pl-5 gap-3 cursor-pointer hover:bg-primary transition-colors duration-100"
```
jahow: cursor-pointer without any interaction is bad for accessibility

jahow (changes requested):

libs/ui/elements/src/lib/metadata-info/metadata-info.component.ts
ronitjadhav: Yes, this will be reworked, but I guess we might still need it to ensure the pill never renders empty.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
```diff
@@ -260,12 +260,11 @@ describe('MetadataInfoComponent', () => {
...
-      it('should display the resourceContact section', () => {
-        expandPanel(fixture, 'details-panel')
-        const displayedElement = fixture.debugElement.query(
-          By.css('[data-test="details-panel-resource-contact"]')
+      it('should render the contacts panel', () => {
+        const contactsPanel = fixture.debugElement.query(
+          By.css('[data-test="contacts-panel"]')
         )
-        expect(displayedElement).toBeTruthy()
+        expect(contactsPanel).toBeTruthy()
```
jahow: this test doesn't really make sense here anymore as it's a separate panel

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
```diff
@@ -0,0 +1,34 @@
+@if (contactGroups.length) {
+  <div class="flex flex-col gap-1 pt-3 pb-4">
+    @for (group of contactGroups; track group.role) {
+      <div class="flex flex-col gap-1 rounded bg-gray-50 py-4 px-2">
```
jahow: my suggestion was to make a single contact a presentation component, but why not...

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
```diff
@@ -0,0 +1,34 @@
...
+              data-test="contact-pill"
+            >
+              <span
+                class="font-title font-medium text-base leading-tight truncate group-hover:text-white"
+                [title]="getContactDisplayName(contact)"
+                >{{ getContactDisplayName(contact) }}</span
+              >
+              <gn-ui-button
+                type="outline"
+                extraClass="w-10 h-8 rounded-lg shrink-0 !bg-primary-white hover:!bg-primary-white group-hover:!bg-primary group-hover:!border-white group-hover:!text-white"
```
jahow: that's a lot of !important, are you sure this is a good idea? this looks super brittle. gn-ui-button components should do most of what you need outside the box. See the storybook entry for the component to learn how to use it

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
```diff
@@ -0,0 +1,34 @@
...
+  <div class="flex flex-col gap-1 pt-3 pb-4">
+    @for (group of contactGroups; track group.role) {
+      <div class="flex flex-col gap-1 rounded bg-gray-50 py-4 px-2">
+        <p class="text-xs font-normal text-black">
+          {{ getRoleTranslationKey(group.role) | translate }}
+        </p>
+        <div class="grid gap-0.5 grid-cols-1 md:grid-cols-2">
+          @for (contact of group.contacts; track contact.email) {
+            <div
+              class="group flex items-center justify-between min-h-12 w-full rounded bg-primary-white py-2 pr-4 pl-5 gap-3 hover:bg-primary transition-colors duration-100"
```
jahow: why do a hover effect if it's not interactive? this is very confusing and not consistent with the rest of the app

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
jahow: also this has another separate hover? really strange

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
```diff
@@ -0,0 +1,34 @@
+@if (contactGroups.length) {
```
jahow: this check already exists before; this is why handling the whole collection in a presentation component is awkward IMO

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
jahow: the thing is the way this collection is presented is very specific to the app; what's interesting is to make smaller components part of a lib so that they can be polished and tested individually, and then apps arrange these components into whatever layout they need

jahow: but doing this is OK I guess, as long as it is a pure presentation component and can be easily tested :)

jahow (commented):
More comments, thanks!

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
ronitjadhav: The issue is that `gn-ui-button` manages its own background/border, which can't be conditionally overridden on parent hover without !important. According to the design, the button needs to visually merge with the pill on hover (matching background, white border/text). 

As per my understanding, a plain <button> with Tailwind utilities would be cleaner here if we want to stick to the Figma design.

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
ronitjadhav: The hover effect is intentional, it's specified in the design. The pill highlights on hover.

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
jahow: I do find this confusing to be honest: is the button the whole pill or just the small icon?

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
ronitjadhav: I misunderstood, yes a single contact as a presentation component will make more sense.

I have Replaced `MetadataResourceContactsComponent` with a focused single contact `ContactPillComponent`

libs/ui/elements/src/lib/metadata-resource-contacts/metadata-resource-contacts.component.html
ronitjadhav: The entire pill is a button now and the small icon is part of the button.

libs/ui/elements/src/lib/metadata-info/metadata-info.component.spec.ts
```diff
@@ -446,6 +422,22 @@ describe('MetadataInfoComponent', () => {
...
+      } as DatasetRecord
+      fixture.detectChanges()
+    })
+    it('should render the contacts panel', () => {
+      const contactsPanel = fixture.debugElement.query(
+        By.css('[data-test="contacts-panel"]')
+      )
+      expect(contactsPanel).toBeTruthy()
+    })
+  })
```
jahow, line 467: there's no test for the logic that distributes contacts in the different roles

libs/ui/elements/src/lib/contact-pill/contact-pill.component.ts
```diff
@@ -0,0 +1,36 @@
...
+export class ContactPillComponent {
+  @Input() contact: Individual
+  @Output() contactClick = new EventEmitter<Individual>()
+
+  get displayName(): string {
+    if (this.contact?.organization?.name) return this.contact.organization.name
+    const first = this.contact?.firstName?.trim()
+    const last = this.contact?.lastName?.trim()
+    if (first && last) return `${first} ${last}`
+    return first || last || ''
```
jahow: didn't we create a function for rendering a contact in a single text string in another PR? I think it's been merged, you might as well reuse it maybe

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
```diff
@@ -0,0 +1,17 @@
+<gn-ui-button
+  type="outline"
+  extraClass="group w-full min-h-12 gap-3 justify-between bg-primary-white hover:bg-primary hover:border-primary hover:text-white py-2 pl-5 pr-4 rounded"
```
jahow: that's a lot of extra classes here, I think this is still hacky. Buttons are reusable components and should be actually reused, not just hacked for each different use. If the existing button does not match what we need then either 1/ change the button to make it more versatile or 2/ adjust the mockup to match the other buttons.

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
```diff
@@ -0,0 +1,17 @@
...
+  data-test="contact-pill"
+  (buttonClick)="contactClick.emit(contact)"
+>
+  <span
+    class="font-title font-medium text-base leading-tight truncate group-hover:text-white"
```
jahow, line 7: this should not be needed in a reusable button

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
```diff
@@ -0,0 +1,17 @@
...
+    [title]="displayName"
+    >{{ displayName }}</span
+  >
+  <div
+    class="flex items-center justify-center w-10 h-8 rounded-lg border border-gray-300 shrink-0 group-hover:border-white group-hover:text-white"
```
jahow: see other comments, this is a lot of custom classes whereas reusable buttons should just work without needing to do all this

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
jahow: especially it looks like the "hover" state of the button does not match the "hover" state of other buttons in the app

jahow (changes requested):

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
```diff
@@ -0,0 +1,17 @@
+<gn-ui-button
+  type="outline"
```
jahow: oh and also in the mockup there's no outline to the button

libs/ui/elements/src/lib/contact-pill/contact-pill.component.ts
ronitjadhav: Yes, we created `getUserDisplayName` but it expects a `UserModel` as input but whereas here we are using type `Individual`. I guess they aren't the same, we will need to convert before using it here.

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
ronitjadhav: This is needed to wrapping the info icon inside a box shape container according to the mockup. I can remove the custom box and just keep the info icon. This is how it will look:

image

image

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
ronitjadhav: If we have to stick to the current mockup, I can update the button component to introduce a new `primary-white` button type that encapsulates these theme styles (color, border, hover, active, and focus ring) natively instead of overriding them here. What do you think?"

```
.gn-ui-btn-primary-white {
    @apply gn-ui-btn
      bg-primary-white hover:bg-primary focus:bg-primary hover:text-white focus:text-white active:bg-primary-darker
      border-primary-white hover:border-primary focus:border-primary focus:ring-4 focus:ring-primary-lightest;
  }
```

libs/ui/elements/src/lib/contact-pill/contact-pill.component.ts
jahow: probably a `toIndividual(user: UserModel)` util would make sense here

libs/ui/elements/src/lib/contact-pill/contact-pill.component.html
jahow: the issue is that these "contact pills" were made to match the style of the keyword badges a little bit above; so these two parts should use the same code

ronitjadhav:
@jahow following things are also updated now:

- Added primary-light type to gn-ui-button
- Using gn-ui-button for keywords and categories instead of gn-ui-badge
- Update getUserDisplayName and reuse it in contact-pill and editor form fields

image

image

libs/util/shared/src/lib/utils/user-display.ts
```diff
@@ -1,8 +1,23 @@
 import { UserModel } from '@geonetwork-ui/common/domain/model/user'
+import { Individual } from '@geonetwork-ui/common/domain/model/record'
 
-export function getUserDisplayName(user: UserModel): string {
-  const nameParts = [user.name, user.surname].filter(Boolean).join(' ')
-  const orgPart = user.organisation ? ` (${user.organisation})` : ''
+export function getUserDisplayName(individual: Individual): string {
```
jahow: should be called getPersonDisplayName or getIndividualDisplayName maybe

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
```diff
@@ -346,10 +307,11 @@
...
             @for (topic of metadata.topics; track topic) {
-              <gn-ui-badge
-                [clickable]="false"
-                class="inline-block mr-2 mb-2 lowercase"
-                >{{ topic }}</gn-ui-badge
+              <gn-ui-button
+                type="primary-light"
+                class="inline-block opacity-70 pointer-events-none mr-2 mb-2"
+                extraClass="lowercase text-sm py-1 px-2"
+                >{{ topic }}</gn-ui-button
```
jahow: why make this a button? it's not interactive

jahow (commented):
almost there :)

libs/ui/elements/src/lib/metadata-info/metadata-info.component.html
ronitjadhav: I have reverted it back to use `gn-ui-badge`

libs/util/shared/src/lib/utils/user-display.ts
ronitjadhav: It is renamed to `getIndividualDisplayName`

tailwind.base.css
```diff
@@ -134,6 +134,12 @@
       bg-white border-white focus:ring-4 focus:ring-gray-300;
   }
 
+  .gn-ui-btn-primary-light {
+    @apply gn-ui-btn text-primary-darkest
+      bg-primary-white hover:bg-primary focus:bg-primary hover:text-white focus:text-white
```
jahow: it's missing an active state

jahow (approved):

tailwind.base.css
ronitjadhav: Done! Added 👍🏼

PR #1578, opened 2026-05-29
Datahub Web Component / generate correct links to internal records

jahow:
### Description

This PR fixes an issue where internal record links (e.g. in search results or related records) in the Datahub Web Component were given an invalid `href` attribute. The Web Component relies on hash-based navigation but these links were generated with the assumption that the navigation was url-based.

### Architectural changes

none

### Screenshots

image

### How to test

1. start support services
2. run `npm run docs:dev`
3. run `npm run dev:webcomponents`
4. Go to http://localhost:5173/webcomponents/gn-datahub.html
5. Try right clicking a record and clicking on "open in new tab"
6. The new tab should correctly open on the selected record inside the web component

---

**This work is sponsored by DataGrandEst**

### DIFF
--- a/apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
+++ b/apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
@@ -12,7 +12,8 @@ it('gn-datahub', () => {
   cy.title().as('initialPageTitle')
 
   // should display the search bar and header buttons
-  cy.get('gn-ui-fuzzy-search', { timeout: 10000 }).should('be.visible')
+  cy.get('gn-ui-fuzzy-search', { timeout: 6000 }) // we use a higher timeout because the web component takes a while to boot in dev mode
+    .should('be.visible')
   cy.get('datahub-navigation-menu').find('button').should('have.length', 3)
 
   // should display news page
--- a/apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
+++ b/apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
@@ -12,7 +12,7 @@ it('gn-datahub', () => {
   cy.title().as('initialPageTitle')
 
   // should display the search bar and header buttons
-  cy.get('gn-ui-fuzzy-search').should('be.visible')
+  cy.get('gn-ui-fuzzy-search', { timeout: 10000 }).should('be.visible')
   cy.get('datahub-navigation-menu').find('button').should('have.length', 3)
 
   // should display news page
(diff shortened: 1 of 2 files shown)


apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
```diff
@@ -12,7 +12,7 @@ it('gn-datahub', () => {
   cy.title().as('initialPageTitle')
 
   // should display the search bar and header buttons
-  cy.get('gn-ui-fuzzy-search').should('be.visible')
+  cy.get('gn-ui-fuzzy-search', { timeout: 10000 }).should('be.visible')
```
tkohr: Is this hardcoded timeout necessary?

apps/webcomponents/src/app/components/gn-datahub/gn-datahub.component.ts
```diff
@@ -37,6 +54,28 @@ import { AppComponent } from '@geonetwork-ui/apps/datahub/app.component.ts'
     },
     provideGn4({ disableAuth: true }),
     DATAHUB_CONFIG_PROVIDERS,
+    // we're overriding the URL providers so that they work with our hash-based navigation
+    // TODO: rely on the router for generating these links!
+    {
```
tkohr, line 59: Could it make sense to wrap these objects into a function with an optional prefix to prevent the code duplication?

tkohr (approved):
Thanks @jahow, code LGTM and works as expected! Watch out the CI is still red.

I don't think it's related to this PR, but I often see errors in the console when navigating in the WC:

Image

apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
jahow: yes, when running the e2e tests in dev mode it takes a while for the web component to show.

apps/webcomponents/src/app/components/gn-datahub/gn-datahub.component.ts
jahow: ~I don't think this matters much to be honest, it's just a couple of providers~
I'd rather not add another indirection (a function that generates these providers with a prefix) because this needs to be refactored IMO; the url generation should be the responsibility of the router, and should not be decided by an application providers

jahow:
Thanks for the review!

apps/webcomponents/cypress/e2e/gn-datahub.cy.ts
jahow: I'm going to lower it because it might cause failing tests to take a long time

jahow:
> I don't think it's related to this PR, but I often see errors in the console when navigating in the WC:
> 
> Image

This is because of the copyable code blocks in vite press, there's a conflict with the datahub web component and the copy button it contains I think. To be fixed later on I guess...

PR #1579, opened 2026-05-29
[Metadata-Editor] Make default language for new records configurable

ronitjadhav:
### Description

This PR introduces a configurable default language for newly created metadata records in the metadata-editor app.
Previously, the default language was hardcoded to `'en'` in `new-record.resolver.ts`. Now the default is `current`.

#### Changes

- `default.toml` : added a new `[editor]` section with the `new_record_default_language` key
- `configure.md` : document the new `[editor]` section
- `app-config` : added `EditorConfig` model, `[editor]` section parser, `checkNewRecordDefaultLanguage()` and `getOptionalEditorConfig()` getter
- `metadata-editor` : resolver reads from config, falls back to `'current'`
- `e2e-test` : Added e2e test and a fixture to change the language in config

### How to test

1. In `default.toml`, set `new_record_default_language = "fr"` under the `[editor]` section.
2. Start the metadata-editor app.
3. Create a new record, its default language should be French, not English.
4. Open the newly created record's XML view and confirm the language element shows `codeListValue="fre"`. If the config is absent or invalid, it should fall back to `codeListValue="eng"`.
5. Remove or comment out the key, confirm the default falls back to English.

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -151,6 +151,14 @@ background_color = "#fdfbff"
 # enabled = true
 # If u want to use metadata quality widget this configuration is required
 
+### EDITOR SETTINGS
+
+# This section contains settings specific to the Metadata Editor application.
+[editor]
+# Optional; defines the default language assigned to newly created records (e.g. 'en', 'fr', 'de').
+# If not set, defaults to 'en'.
+# new_record_default_language = "fr"
+
 ### MAP SETTINGS
 
 # The map section allows to customize how maps are configured.
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -156,7 +156,8 @@ background_color = "#fdfbff"
 # This section contains settings specific to the Metadata Editor application.
 [editor]
 # Optional; defines the default language assigned to newly created records (e.g. 'en', 'fr', 'de').
-# If not set, defaults to 'en'.
+# Use "current" to assign the current UI language at the time of record creation.
+# If not set, defaults to the current UI language.
 # new_record_default_language = "fr"
 
 ### MAP SETTINGS
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -151,12 +151,11 @@ background_color = "#fdfbff"
 # enabled = true
 # If u want to use metadata quality widget this configuration is required
 
-### EDITOR SETTINGS
+### EDITING SETTINGS
 
-# This section contains settings specific to the Metadata Editor application.
-[editor]
+# This section contains settings for metadata editing behavior in the Metadata Editor application.
+[editing]
 # Optional; defines the default language assigned to newly created records (e.g. 'en', 'fr', 'de').
-# Use "current" to assign the current UI language at the time of record creation.
 # If not set, defaults to the current UI language.
 # new_record_default_language = "fr"
 
--- a/libs/util/app-config/src/lib/app-config.spec.ts
+++ b/libs/util/app-config/src/lib/app-config.spec.ts
@@ -257,7 +257,7 @@ describe('app config utils', () => {
           () =>
             minimalAppConfigFixture() +
             `
-[editor]
+[editing]
 new_record_default_language = "fr"
 `
         )
@@ -276,7 +276,7 @@ new_record_default_language = "fr"
           () =>
             minimalAppConfigFixture() +
             `
-[editor]
+[editing]
 new_record_default_language = "fre"
 `
         )
@@ -295,7 +295,7 @@ new_record_default_language = "fre"
           () =>
             minimalAppConfigFixture() +
             `
-[editor]
+[editing]
 new_record_default_language = "xyz"
 `
         )
--- a/libs/util/app-config/src/lib/app-config.spec.ts
+++ b/libs/util/app-config/src/lib/app-config.spec.ts
@@ -3,6 +3,7 @@ import {
   _reset,
   getCustomTranslations,
   getGlobalConfig,
+  getOptionalEditorConfig,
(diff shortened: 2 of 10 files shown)


conf/default.toml
```diff
@@ -151,6 +151,15 @@ background_color = "#fdfbff"
 # enabled = true
 # If u want to use metadata quality widget this configuration is required
 
+### EDITOR SETTINGS
+
+# This section contains settings specific to the Metadata Editor application.
+[editor]
```
tkohr: "This section contains settings specific to the Metadata Editor application." already sounds good to me, but I'm wondering if it's enough for admins to understand quickly that theming and url configs for the editor are outside this section. Could it perhaps make sense to name the section `[editing]` for example?

conf/default.toml
```diff
@@ -151,6 +151,15 @@ background_color = "#fdfbff"
...
+# Optional; defines the default language assigned to newly created records (e.g. 'en', 'fr', 'de').
+# Use "current" to assign the current UI language at the time of record creation.
+# If not set, defaults to the current UI language.
```
tkohr: Do we actually need the 'current' option/value (as the default behavior when not set is already the same)?

libs/util/app-config/src/lib/app-config.spec.ts
```diff
@@ -248,6 +249,73 @@ describe('app config utils', () => {
     })
   })
 
+  describe('when the configuration file contains new_record_default_language', () => {
```
tkohr, line 252: Maybe add a test for "current" as well if we really need it?

libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -137,3 +137,24 @@ export function checkMetadataLanguage(
   }
   return parsedConfigSection
 }
+
+export function checkNewRecordDefaultLanguage(
```
tkohr, line 141: Claude comment which makes sense to me: "The function both mutates parsedConfigSection in place AND returns it. The mutation means callers don't strictly need the return value, but the caller in app-config.ts reassigns it anyway. Pick one pattern — either mutate in place (no return value) or return a new object (no mutation). Currently it's inconsistently both."

libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -137,3 +137,24 @@ export function checkMetadataLanguage(
...
+  parsedConfigSection: any,
```
tkohr, line 142: Misses typing

libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -137,3 +137,24 @@ export function checkMetadataLanguage(
...
+  outWarnings: string[]
+) {
+  if (parsedConfigSection.new_record_default_language === 'current') {
+    return parsedConfigSection
```
tkohr: Returning `null` here could simplify the logic in the resolver, but as mentioned above I don't think we actually need the 'current' logic, do we ?

libs/util/app-config/src/lib/parse-utils.ts
```diff
@@ -137,3 +137,24 @@ export function checkMetadataLanguage(
...
+  }
+  const lang2 = toLang2(
+    parsedConfigSection.new_record_default_language.toLowerCase()
```
tkohr, line 146: Could be useful to apply the `toLowerCase` to the 'current' case above already to make the config case insensitive if 'current' is needed.

tkohr (commented):
Thanks @ronitjadhav , did not test it, but the code looks mostly good to me!

conf/default.toml
ronitjadhav: I guess not. I included `current` because I took inspiration from the existing config for `metadata_language`:

```
# metadata_language = "current"
```
But in our case we need a default, which can already be the current UI language when the setting is omitted, so we might not need this option explicitly.

libs/util/app-config/src/lib/parse-utils.ts
ronitjadhav: `checkNewRecordDefaultLanguage` now returns a normalized copy instead of mutating `parsedConfigSection` in place.

libs/util/app-config/src/lib/parse-utils.ts
ronitjadhav: I kept the typing as is for now because `parseConfigSection()` currently returns a generic `Record<string, unknown> | null` for all config sections, and similar function such as `checkMetadataLanguage()` also use the same pattern.

Typing this properly would need either a section-specific type guard/assertion at the call site, or a broader refactor where `parseConfigSection()` can return typed section shapes.

libs/util/app-config/src/lib/parse-utils.ts
ronitjadhav: This is not required now as we removed option `current`

libs/util/app-config/src/lib/parse-utils.ts
ronitjadhav: This is not required now as we removed option `current`

libs/util/app-config/src/lib/app-config.spec.ts
ronitjadhav: This is not required now as we removed option `current`

tkohr (approved):
Thanks for the adaptions, which LGTM! Did not test it.

ronitjadhav:
Thanks for the review, I tested it again myself just to be sure.

PR #1580, opened 2026-05-29
Fix spatial extent display on load

tonio:
### Description

This PR fixes spatial extent display in zoneless app & error occuring with nullish map passed to `applyContextDiffToMap`.

With the help of @claude as zone's thing is not my favourite part of Angular 😁.

  ### Bug 1 — Spatial extent not displayed on load

  Root cause: The app uses provideZonelessChangeDetection(). In SpatialExtentComponent, mapContext$ used switchMap(async ...) which returns a Promise resolving outside Angular's change
  detection cycle. In zoneless + OnPush mode this means MapContainerComponent.ngOnChanges is never called with the real context → map stays at world view.

  ### Bug 2 — TypeError: Cannot read properties of undefined (reading 'getLayers')

  Root cause: ngAfterViewInit is async but Angular doesn't await it. ngOnChanges (non-first change) could fire before createMapFromContext resolved, leaving this.olMap undefined when
  passed to applyContextDiffToMap.

**Watchout**: as the key factor of the first bug was the use of components in zoneless app, I asked Claude if there was others places in the lib in the code where the same kind of bug could occur, looks like there are 3:

image

This PR is only related to spatial extent for now, should I take the opportunity to proactively fix those potential issues here?

### DIFF
--- a/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
+++ b/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from '@angular/core'
+import { ChangeDetectorRef, Component, inject, Input } from '@angular/core'
 import { CommonModule } from '@angular/common'
 import { Geometry } from 'geojson'
 import { GeoJSONFeatureCollection } from 'ol/format/GeoJSON.js'
@@ -10,8 +10,8 @@ import {
   MapContextLayer,
 } from '@geospatial-sdk/core'
 import { MapContainerComponent } from '../map-container/map-container.component'
-import { BehaviorSubject, Observable } from 'rxjs'
-import { switchMap } from 'rxjs/operators'
+import { BehaviorSubject, from, Observable, of } from 'rxjs'
+import { map, switchMap, tap } from 'rxjs/operators'
 import { DatasetSpatialExtent } from '@geonetwork-ui/common/domain/model/record'
 
 @Component({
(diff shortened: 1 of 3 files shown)


jahow:
I'm not sure I'm following the deal about zoneless apps here. I don't know of any GN-UI app that is zoneless. Besides, for the spatial extent component isn't it just enough to call `cdr.markForCheck()`?

tonio:
> I don't know of any GN-UI app that is zoneless

Datafeeder is. (-:

jahow:
Ok, good to know thanks :)

I suggest we only fix the necessary issues in this PR, and tackle support for zoneless change detection in another PR according to https://angular.dev/guide/zoneless

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
```diff
@@ -61,11 +63,10 @@ export class SpatialExtentComponent {
...
         },
       }
-      const view = await createViewFromLayer(layer)
-      return {
-        view,
-        layers: [layer],
-      }
+      return from(createViewFromLayer(layer)).pipe(
+        map((view) => ({ view, layers: [layer] }) as MapContext),
+        tap(() => this._cdr.markForCheck())
```
jahow, line 68: I'm still surprised that we need this, the AsyncPipe directive should automatically do it

https://angular.dev/api/common/AsyncPipe#description

jahow (approved):
The change generally looks good; before merging, could you please give it a last glance to make sure that it doesn't work with the `async` function and the `markForCheck` call? Using promises inside RxJS pipes is a super useful feature and it would be good to know if we should avoid  it in the future.

thanks :)

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
tonio: `markForCheck()` is redundant if template is using | async directly on mapContext$ in a app with zone.js, but in a zoneless app, AsyncPipe is not enough because there's no global scheduler to trigger detection cycle —  markForCheck() is the explicit way to do it.

PR #1582, opened 2026-05-29
feat: create feature/notify-reuse

tonio:
### Description

Add new Nx Angular library for editing reuse records, with a NotifyReuseFormComponent shell (typed against ReuseRecord), path alias, module boundary config, unit tests, and a Storybook story. Form field configuration out of scope.

### Screenshots

Empty shell for now


libs/feature/notify-reuse/project.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "notify-reuse",
```
jahow: the project name should be `feature-notify-reuse` for consistency with others

libs/feature/notify-reuse/jest.config.ts
```diff
@@ -0,0 +1,21 @@
+export default {
+  displayName: 'notify-reuse',
+  preset: '../../../jest.preset.js',
+  setupFilesAfterEnv: ['<rootDir>/src/test-setup.ts'],
```
jahow: not need, it's part of the preset :)

libs/feature/notify-reuse/jest.config.ts
```diff
@@ -0,0 +1,21 @@
...
+        tsconfig: '<rootDir>/tsconfig.spec.json',
+        stringifyContentPathRegex: '\\.(html|svg)$',
+      },
+    ],
+  },
+  transformIgnorePatterns: ['node_modules/(?!.*\\.mjs$)'],
+  snapshotSerializers: [
+    'jest-preset-angular/build/serializers/no-ng-attributes',
+    'jest-preset-angular/build/serializers/ng-snapshot',
+    'jest-preset-angular/build/serializers/html-comment',
```
jahow: also part of the preset, not needed

package.json
```diff
@@ -174,6 +177,7 @@
     "jest-canvas-mock": "2.5.2",
     "jest-environment-jsdom": "29.7.0",
     "jest-preset-angular": "14.6.2",
+    "jest-util": "^29.7.0",
```
jahow: not needed I think?

package.json
```diff
@@ -146,6 +146,9 @@
     "@schematics/angular": "20.3.12",
     "@storybook/angular": "9.1.9",
     "@stylistic/eslint-plugin-ts": "^2.12.1",
+    "@swc-node/register": "~1.9.1",
+    "@swc/core": "~1.5.7",
+    "@swc/helpers": "~0.5.11",
```
jahow: do we need that? I'm not sure

jahow (commented):
Thanks! a couple of comments but apart from that the new lib looks consistent with the rest :)

package.json
tonio: automatically added by nx when generating new library

package.json
tonio: same here «These are pulled in by the Nx Angular library generator as part of the Jest+SWC test runner configuration. They're legitimate dev dependencies — the library's jest.config.ts uses SWC as the TypeScript transformer for faster test compilation.»

jahow:
I think we can skip the new dependencies; the project runs fine without them and I'd really like to avoid bloat in the package.json if possible. I know these are probably present somewhere in the dependencies tree already, but if they're not required explicitly from the project we should keep them out of the explicit dependencies list. What do you think?

tonio:
> What do you think?

Given the choice of using nx, I would have kept generated changes. But no strong opinion on this, I can remove if you prefer.

jahow (approved):

PR #1587, opened 2026-06-04
Read/Write lineage sources for ISO 19139 and ISO 19115-3

ronitjadhav:
### Description

This PR implements reading and writing of the `source` part of a lineage for both ISO 19139 and ISO 19115-3 converters.

Only the `uuidref` attribute is the authoritative identifier. However, `xlink:title` and `xlink:href` are read alongside it when present in existing XML. So that to keep all editions non-destructive of other information, so we should keep all attributes of other sources which are not updated.

When writing, if the uuidref (uuid) changes, the entire gmd:source / mrl:source element is replaced which erases `title` and `href` attributes

### Architectural changes

A new optional field `lineageSources` is added to metadata model.

### How to test

This PR only covers converter read/write support, the other implementations are yet to be completed for proper testing.  

For now run the converter unit tests and check the new `readLineageSources` / `writeLineageSources` cases for both ISO 19139 and ISO 19115-3.


libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -254,10 +254,20 @@ export interface DatasetTemporalExtent {
   end?: Date
 }
 
+/**
+ * Represents a source dataset referenced from a lineage entry.
+ */
+export interface LineageSource {
+  uuid: string
```
jahow: I don't think this should be mandatory; a source could simply be an href link.

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -894,6 +895,33 @@ export function readLineage(
...
+    findNestedElements(
+      'gmd:dataQualityInfo',
+      'gmd:DQ_DataQuality',
+      'gmd:lineage',
+      'gmd:LI_Lineage',
+      'gmd:source'
+    ),
+    mapArray((el) => {
+      const uuid = readAttribute('uuidref')(el)
+      if (!uuid) return null
```
jahow: see other comment, this should not be mandatory

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
```diff
@@ -291,6 +292,27 @@ export function readLineage(
   )(rootEl)
 }
 
+export function readLineageSources(
```
jahow: you can probably factorize this code by using the same as the iso19139 one from the `mdb:resourceLineage` element; otherwise this looks a lot like duplicated code

jahow (commented):
I thik the "LineageSource" name isn't the best, it isn't very descriptive. I think "SourceRecord" would describe it better. What do you think?

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
ronitjadhav: Done 👍🏼

libs/common/domain/src/lib/model/record/metadata.model.ts
ronitjadhav: Done 👍🏼

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.ts
ronitjadhav: Yes, I should have already done that, thanks for pointing it out.

libs/api/metadata-converter/src/lib/iso19115-3/write-parts.ts
```diff
@@ -564,3 +565,22 @@ export function writeOtherLanguages(record: DatasetRecord, rootEl: XmlElement) {
...
+      'mdb:resourceLineage',
+      'mrl:LI_Lineage'
+    )(rootEl)
+    if (lineageEl) removeChildrenByName('mrl:source')(lineageEl)
+    return
+  }
+
+  pipe(
+    findNestedChildOrCreate('mdb:resourceLineage', 'mrl:LI_Lineage'),
+    appendSourceRecords('mrl:source', sources)
```
jahow: so you probably don't need to specify an element name here. If you look here: https://github.com/geonetwork/geonetwork-ui/blob/d7fca85607f83610c57b443fd515526fdeacf3ed/libs/api/metadata-converter/src/lib/iso19115-3/iso19115-3.converter.ts#L87, the Iso19115-3 converted will rename any elements coming from the Iso19139 standard to its correct name.

Same for deletion, `removeChildrenByName` ignores namespace so you can also reuse that part of the code in the Iso19139 converter.

libs/api/metadata-converter/src/lib/iso19115-3/write-parts.ts
```diff
@@ -564,3 +565,22 @@ export function writeOtherLanguages(record: DatasetRecord, rootEl: XmlElement) {
...
+  if (sources === undefined) return
+
+  if (sources.length === 0) {
+    const lineageEl = findNestedElement(
+      'mdb:resourceLineage',
+      'mrl:LI_Lineage'
+    )(rootEl)
+    if (lineageEl) removeChildrenByName('mrl:source')(lineageEl)
+    return
+  }
```
jahow: why are you only deleting the children if the sources array is empty? I think it's risky to reuse existing ones without looking at what's inside them.

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -254,10 +254,20 @@ export interface DatasetTemporalExtent {
...
+  uuid?: string
+  title?: string
+  href?: string
+}
+
 export interface DatasetRecord extends BaseRecord {
   kind: 'dataset'
   status: RecordStatus
   lineage: string // Explanation of the origin of this record (e.g: how, why)"
+  sourceRecords?: Array<SourceRecord>
```
jahow: if you look at the other models, if a property holds an array then it is never optional; it can simply be an empty array. Basically we should not have to check for the presence _and_ the length of an array; either it's empty or it has elements, but it should not be undefined

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -894,6 +895,43 @@ export function readLineage(
   )(rootEl)
 }
 
+export function extractSourceRecords(
+  liLineageEl: XmlElement
+): SourceRecord[] | undefined {
```
jahow: see my other comment on the model; this should not return undefined.

libs/common/domain/src/lib/model/record/metadata.model.ts
jahow: (this is actually not true for the resouceIdentifiers field, and that was an oversight I think)

ronitjadhav: Done 👍🏼

libs/api/metadata-converter/src/lib/iso19115-3/write-parts.ts
ronitjadhav: I have removed it.

libs/api/metadata-converter/src/lib/iso19115-3/write-parts.ts
ronitjadhav: Thanks, Understood I am using the renameElements now.

jahow (approved):
Thank you, looking good!

PR #1590, opened 2026-06-05
[Metadata-Editor] Make new record standard configurable

ronitjadhav:
### Description

This PR introduces a configurable metadata standard for newly created records in the metadata-editor app.

Previously, new records without a reference source were serialized as ISO19139 by default. Platforms can now configure `[editing].new_record_standard` with either `iso19139` or `iso19115-3`.

#### Changes

- `default.toml`: added a commented `new_record_standard` example under `[editing]`
- `configure.md`: documented the new `[editing].new_record_standard` setting
- `app-config`: added `NEW_RECORD_STANDARD`, parser support, validation, and unit tests
- `metadata-editor`: new record resolver reads the configured standard and pre-serializes ISO19115-3 records when requested
- `feature-editor`: accepts nullable record sources at the facade/service boundary and passes `undefined` to the repository when no source exists
- `e2e-test`: added coverage and fixture for creating records with ISO19115-3 configured

### How to test

1. In `default.toml`, set `new_record_standard = "iso19115-3"` under the `[editor]` section.
2. Start the metadata-editor app.
3. Create a new record.
4. Publish the record
5. Confirm the saved XML contains `<mdb:MD_Metadata`.
6. Remove or comment out `new_record_standard`.
7. Create another new record and confirm the saved XML uses ISO19139, for example `<gmd:MD_Metadata`.

### DIFF
--- a/apps/metadata-editor/src/app/new-record.resolver.ts
+++ b/apps/metadata-editor/src/app/new-record.resolver.ts
@@ -4,13 +4,14 @@ import {
   Individual,
   Organization,
 } from '@geonetwork-ui/common/domain/model/record'
-import { Observable, of } from 'rxjs'
+import { from, Observable, of } from 'rxjs'
 import { map, switchMap } from 'rxjs/operators'
 import { PlatformServiceInterface } from '@geonetwork-ui/common/domain/platform.service.interface'
 import { OrganizationsServiceInterface } from '@geonetwork-ui/common/domain/organizations.service.interface'
 import { TranslateService } from '@ngx-translate/core'
 import { NOT_KNOWN_CONSTRAINT } from '@geonetwork-ui/feature/editor'
 import { getOptionalEditorConfig } from '@geonetwork-ui/util/app-config'
+import { Iso191153Converter } from '@geonetwork-ui/api/metadata-converter'
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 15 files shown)


apps/metadata-editor/src/app/new-record.resolver.ts
```diff
@@ -53,7 +56,22 @@ export class NewRecordResolver {
...
+                [catalogRecord, recordSource, false] as [
+                  CatalogRecord,
+                  string,
+                  false,
+                ]
+            )
+          )
+        }
+        return of([catalogRecord, null, false] as [CatalogRecord, null, false])
       })
```
tkohr: This could probably be simplified into something like:
```
      switchMap(async (catalogRecord) => {
        const recordSource =
          editorConfig?.NEW_RECORD_STANDARD === 'iso19115-3'
            ? await this.iso191153Converter.writeRecord(catalogRecord)
            : null
        return [catalogRecord, recordSource, false] as [
          CatalogRecord,
          string | null,
          boolean,
        ]
      })
```

libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -42,7 +42,7 @@ export class EditorService {
     }
 
     return this.recordsRepository
-      .saveRecord(savedRecord, recordSource, publishToAll)
+      .saveRecord(savedRecord, recordSource ?? undefined, publishToAll)
```
tkohr: Do we really need this mapping?

tkohr (approved):
Nice feature @ronitjadhav, which works as expected!

apps/metadata-editor/src/app/new-record.resolver.ts
```diff
@@ -53,7 +56,22 @@ export class NewRecordResolver {
...
+              (recordSource) =>
+                [catalogRecord, recordSource, false] as [
+                  CatalogRecord,
+                  string,
+                  false,
+                ]
+            )
+          )
+        }
+        return of([catalogRecord, null, false] as [CatalogRecord, null, false])
```
jahow: I would suggest making this always return an XML document, either iso19139 or 19115-3, instead of relying on the fallback that's way deeper when not asking for 19115-3.

jahow (commented):
FYI the fallback to iso19139 is here: https://github.com/geonetwork/geonetwork-ui/blob/7e58935b21c3b5a8d58070e793399a75a2e53bd8/libs/api/repository/src/lib/gn4/gn4-repository.ts#L604-L613

jahow:
Actually we might offer an injection token in the GN4 repository service to let us decide what is the default converter; this would let the app not have to worry about generating XML at creation

ronitjadhav:
Done. I have updated the approach as suggested.
Added a DEFAULT_RECORD_CONVERTER injection token in gn4repository with Iso19139Converter as default.

jahow (approved):
thanks!

PR #1606, opened 2026-06-16
[Metadata-Editor] Duplicate metadata records into the user's group

ronitjadhav:
### Description

This PR updates metadata duplication so duplicated records are created in the current user’s group instead of always using group `2`.

The selected group follows the expected priority:
- first group where the user has reviewer permissions
- otherwise first group where the user has editor permissions
- throw error when no writable group found for duplication

### Architectural changes

Adds `GroupModel` in the common domain user models.

Adds `getUserPermissionsByGroup()` to the platform service interface and implements it for GN4.

`Gn4Repository.openRecordForDuplication()` now asks the platform service for group permissions before calling the GeoNetwork duplicate endpoint.

### How to test

1. Log in as `barbie`
2. Confirm `/geonetwork/srv/api/me` returns editor permissions for group `102`:

```
{
  "id": "101",
  "username": "barbie",
  "name": "Barbara",
  "surname": "Roberts",
  "organisation": "Barbie Inc.",
  "admin": false,
  "groupsWithRegisteredUser": [],
  "groupsWithEditor": [102],
  "groupsWithReviewer": [],
  "groupsWithUserAdmin": [103]
}
```
3. Duplicate a metadata record from the metadata editor.
4. In the browser Network tab, check the request to `/geonetwork/srv/api/records/duplicate` it should include: `group=102`
5. Copy the new UUID returned by the duplicate request.
6. Confirm the stored owner group: `http://localhost:8080/geonetwork/srv/api/records/<new-uuid>/sharing`
7. Expected response contains:
```
{
  "groupOwner": "102"
}
```
8. To test other conditions, log in as `admin`, open the GeoNetwork admin console, update users’ group, then repeat the duplication flow. For example, give a user a Reviewer group and an Editor group to confirm the Reviewer group is selected first.

### DIFF
--- a/libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
+++ b/libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
@@ -109,6 +109,26 @@ let _supportsAuthentication = true
 class PlatformServiceInterfaceMock {
   getApiVersion = jest.fn(() => of('4.2.5'))
   supportsAuthentication = jest.fn(() => _supportsAuthentication)
+  getUserPermissionsByGroup = jest.fn(() =>
+    of([
+      {
+        groupId: 105,
+        groupName: 'Groupe Reviewers',
+        isMember: true,
+        canEdit: true,
+        canApprove: true,
+        canAdministrate: false,
+      },
+      {
+        groupId: 103,
+        groupName: 'Groupe Editors',
+        isMember: true,
+        canEdit: true,
+        canApprove: false,
+        canAdministrate: false,
+      },
+    ])
+  )
 }
 
 let allowEditHarvested = false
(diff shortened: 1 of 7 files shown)


libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
```diff
@@ -744,6 +764,53 @@ describe('Gn4Repository', () => {
...
+          expect.anything(),
+          expect.anything(),
+          undefined,
+          expect.anything(),
+          expect.anything(),
+          undefined,
+          expect.anything(),
+          expect.anything(),
+          expect.anything()
+        )
```
jahow: I'm not sure about this fallback, why are we targeting this group again? maybe if we cannot find any group that fits we can simply throw an error?

jahow (commented):
Looking good, thanks! I'm just not sure about the fallback to the group "2", can't remember why that was there to be honest

libs/api/repository/src/lib/gn4/gn4-repository.spec.ts
ronitjadhav: It was mentioned in requirement ticket: `If the user doesn't have any groups, then the previous behavior is kept (duplicating into group 2).`

However, if we think throwing an error is a better approach now, I'm happy to change it. Let me know if you'd like me to update that!

jahow: It depends why we did put this group in the first place; I'd be in favor of an error indeed, since duplicating should always be allowed for a user that has access to the metadata editor app

ronitjadhav: I have removed the '2' fallback. It now throws an Error instead

jahow:
following @LHBruneton-C2C's question, what happens for an admin account? which group is selected?

ronitjadhav:
For an admin, `getUserPermissionsByGroup()` returns all groups with all flags set to true, so the first group from the GeoNetwork groups list is selected, in this test geonetwork case group 2, which is the first entry returned by `/api/groups` is used.

jahow (approved):

PR #1607, opened 2026-06-16
Datahub: Add notify reuse form

tkohr:
### Description

This PR adds a "Declare a reuse" form, opened from a button on dataset pages (logged-in users only). The form collects a title, URL, and contact email (pre-filled with the dataset's organization email), then saves a reuse record in GeoNetwork linked back to the source dataset.

Also fixes a converter bug where lineage and sourceRecords were only serialized for dataset records — reuse records were saved without their source link (and showed up as datasets).

**Changes**

- New NotifyReuseFormComponent (overlay form + validation + save)
- text-input: only show the invalid state after a field is touched
- spinning-loader: configurable size/color via layoutClasses
- Converter: write lineage/sourceRecords for reuse records too
- Unit + e2e test coverage

**To-dos**

- [x] check url/~~email~~ formats with user feedback
- [x] display potential error after request => via `NotificationService`
- [x] send to editor page => page does not display yet since `canEdit` in [gn4-repository](https://github.com/geonetwork/geonetwork-ui/blob/main/libs/api/repository/src/lib/gn4/gn4-repository.ts#L340) does not include kind === `reuse` yet. => to be decided how to allow access without implementing full reuse support for editor (may be part of follow up PR)
- [x] add a `redirect_on_leave` query param to let the user come back to the DH reuse page from the ME light edit

Note that the "Champs incorrects" button from the mockups has not been implemented yet, as it would either need the submit button to become enabled when fields are invalid or already display before clicking submit (which seems quite intrusive). To be discussed if this button is really necessary.

### Screenshots

image

### How to test

1. Start datahub
2. Start metadata-editor
3. Point `reuse_form_url` to local metadata-editor
4. Login with any user with write access to local GN
5. Go to a NON dataset page and verify the "Declare a reuse" button does not appear.
6. Go to a dataset page and verify the "Declare a reuse" button DOES appear.
7. Declare a reuse.
8. Verify the reuse was created in GN and the metadata-editor opens. Note that there will still be an error message displayed in the metadata-editor, due to the last point mentioned in the To-dos.
9. Verify that errors are handled.

---


fvanderbiest:
Would it be possible to have the email pre-set when used in a geOrchestra context ? (email is provided as a http header when logged-in)

tkohr:
> Would it be possible to have the email pre-set when used in a geOrchestra context ? (email is provided as a http header when logged-in)

Good thing you point this out, I might have mixed something up here! The email is actually already being prefilled with the [owner organisation's email of the record](https://github.com/geonetwork/geonetwork-ui/blob/notify-reuse-form/libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts#L89). I guess it should rather be the email of the user who's logged in?

fvanderbiest:
> I guess it should rather be the email of the user who's logged in?

Good point. I ... don't know. 
Sometimes it will make more sense to have the user's email, and sometimes not.
I think we can try to read the user's email and it not available, default to the owner's organisation email.

jahow:
the email of the user should definitely work out

jahow:
I'm picking up work on this

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
```diff
@@ -199,7 +200,17 @@ export class NotifyReuseFormComponent implements OnDestroy {
...
+        const baseHrefClean = this.locationStrategy
+          .getBaseHref()
+          .replace(/^\/+|\/+$/g, '')
+        const prefix = baseHrefClean ? `/${baseHrefClean}` : ''
+        const rawRedirectUrl = `${window.location.origin}${prefix}/reuse/${uniqueIdentifier}`
+        const redirect_on_leave = encodeURIComponent(rawRedirectUrl)
+        window.open(
+          `${baseUrl}/edit/${uniqueIdentifier}?redirect_on_leave=${redirect_on_leave}`,
+          '_self'
+        )
```
jahow, line 213: this should be simply handled by the reuseFormUrl, like so:

```
reuse_form_url = https://my.site/metadata-editor/light-edit/${uniqueIdentifier}?redirect_on_leave=${currentUrl}
```

translations/de.json
```diff
@@ -563,40 +575,26 @@
...
-  "record.metadata.title": "Titel",
-  "record.metadata.topics": "Kategorien",
-  "record.metadata.uniqueId": "Eindeutige Kennung",
-  "record.metadata.update": "Aktualisierungsdatum",
-  "record.metadata.updateFrequency": "Aktualisierungsfrequenz der Daten",
-  "record.metadata.updatedOn": "Informationen geändert",
-  "record.metadata.usage": "Nutzung und Einschränkungen",
-  "record.metadata.userFeedbacks": "Fragen / Antworten",
-  "record.metadata.userFeedbacks.anonymousUser": "Um einen Kommentar zu schreiben, melden Sie sich bitte an.",
-  "record.metadata.userFeedbacks.authDisabled": "Kommentare sind deaktiviert.",
```
jahow, line 577: these translations are gone; there were lost translations in other languages too

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
LHBruneton-C2C: We don't want to redirect to the current dataset url, but to the reuse url in the DataHub.

So, the workflow is:
- notify from a dataset in the DH
- redirect to the light edit page of the reuse in the ME
- redirect to the reuse page in the DH

jahow: Ok then maybe the light reuse form could simply handle the redirection itself, without needing a url query param for it.

What I mean is if we hardcode `?redirect_on_leave` in the code here, it doesn't even make sense to offer the `reuse_form_url` config since everything is wired to only work with the metadata editor

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
```diff
@@ -130,7 +130,7 @@ export class NotifyReuseFormComponent implements OnDestroy {
 
     this.overlayRef = this.overlay.create({
       positionStrategy,
-      hasBackdrop: true,
```
jahow: why? the overlay should close if there are other interactions on the page right?

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
```diff
@@ -142,6 +142,7 @@ export class NotifyReuseFormComponent implements OnDestroy {
   }
 
   closeOverlay() {
+    this.clearInputs()
```
jahow, line 146: do we really want this? closing the overlay would mean losing information that was written by the user

libs/api/metadata-converter/src/lib/gn4/gn4.field.mapper.ts
```diff
@@ -84,7 +84,7 @@ export class Gn4FieldMapper {
           'resourceAbstractObject',
           this.getLocalizedIndexKey
         ),
-        'no title'
+        this.translateService.instant('record.metadata.abstract.null')
```
jahow: maybe `record.metadata.abstract.empty` instead? it would be clearer

jahow: maybe `record.metadata.abstract.empty` instead? it would be clearer

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
LHBruneton-C2C: `redirect_on_leave` is computed by the Datahub to return to the DataHub. I don't see where this is wired to work only with the metadata editor. Any external app can decide to use or not this query param. It's even more flexible that way, as the DataHub can change the behavior wanted without the app changing anything.

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
LHBruneton-C2C: This doesn't work with the error notification. That's why it's been changed to only explicit closing on the X button or on the declaration button.

libs/feature/notify-reuse/src/lib/notify-reuse-form/notify-reuse-form.component.ts
LHBruneton-C2C: Closing the overlay is now limited to the X button and declaration button.

It felt weird actually, scrolling around, doing things, and having this form re-open pre-filled with potentially wrong inputs. The "normal" workflow is filling and submitting (leaves the page). On error, the form remains open, to let the user fix the inputs. Closing starts again from scratch.

LHBruneton-C2C (approved):
Looking good and tested OK. Great team effort!

PR #1614, opened 2026-06-24
feat(record): display metadata spatial extent on the data preview map

fgravin:
All code has been generated by Claude.

One remark concerning the addition of the GeoJSON layer for the extent: do you think it would have been better to rely on the map state, or is it OK to add it on top of the layers like this?

No dedicated overlay tests yet.
No extent drawn if there is no map.

Screenshot from 2026-06-24 13-29-29

## Context

In the datahub record **data preview** (map tab), the metadata often declares a spatial extent (a bbox, sometimes a polygon). Until now this extent was only used to compute the map's _initial zoom_ as a fallback — it was never **drawn** on the map.

This PR overlays the metadata's declared spatial extent on top of the previewed data, so users can see the dataset's announced footprint alongside the actual data.

## What changed

- **Shared geometry helpers** (`libs/util/shared`, `utils/geojson.ts`): the `bbox → polygon` and `spatialExtents → FeatureCollection` logic that previously lived (privately) inside `SpatialExtentComponent` is extracted into pure, reusable helpers — `bboxToPolygon()`, `spatialExtentToGeometry()` and `spatialExtentsToFeatureCollection()`. They are placed in their canonical home next to the existing `getGeometryBoundingBox()` / `BoundingBox`, following the repo guideline that cross-cutting helpers belong in `util/shared`, not tucked inside a `ui/*` lib.
- **`MapUtilsService.getRecordExtentLayer()`** (`libs/feature/map`): a new method, sitting right beside the existing `getRecordExtent()`, that builds the overlay layer from a record's `spatialExtents` (each extent uses its `geometry` when present, otherwise its `bbox` converted to a polygon). The layer is `clickable: false` so it never interferes with feature selection, and the method returns `null` when no usable extent exists.
- **`SpatialExtentComponent`** (`libs/ui/map`): now consumes `spatialExtentsToFeatureCollection()`, dropping its duplicated `bboxCoordsToGeometry` method and its direct OpenLayers dependency. Its visual rendering is unchanged.
- **`MapViewComponent.mapContext$`** (`libs/feature/record`): appends `mapUtils.getRecordExtentLayer(metadata)` **last** in the `layers` array, so the extent is drawn on top of the data.

## Why it's non-intrusive

- The extent layer is appended **after** the data layer, so `createViewFromLayer(layers[0])` still computes the initial zoom from the data — the `view` logic is untouched (it keeps falling back to the record extent only when the data provides no view of its own). The display of the extent and the computation of the initial view are kept as two distinct concerns.
- The legend (`gn-ui-map-legend`) only reads `layers[0]`, so no spurious legend entry is added.
- No cross-layer leak: pure helpers live in `util/shared`, the map↔record transform lives in `feature/map`, and the smart component in `feature/record` just orchestrates.
- The overlay is styled to stay readable without hiding the data underneath (dashed black outline over a very light fill).

## Tests

- Existing `spatial-extent` and `map-view` suites pass. The direct unit test on the removed `bboxCoordsToGeometry` method is dropped (the `bbox → polygon` path is still covered end-to-end by the existing "layer with bbox" test); the `map-view` `MapUtilsService` mock gains a `getRecordExtentLayer` stub returning `null` (the fixture metadata has no `spatialExtents`), so the existing `layers` assertions still hold.
- Dedicated unit tests for the new helpers and for the extent actually being overlaid in the preview are intentionally left as a follow-up.

## Manual check

Open a record that has a map and a declared extent, e.g. `/dataset/implantation-des-arceaux-velos-a-roubaix`, and look at the map tab.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
+++ b/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
@@ -1,9 +1,5 @@
 import { ChangeDetectorRef, Component, inject, Input } from '@angular/core'
 import { CommonModule } from '@angular/common'
-import { Geometry } from 'geojson'
-import { GeoJSONFeatureCollection } from 'ol/format/GeoJSON.js'
-import GeoJSON from 'ol/format/GeoJSON.js'
-import { Polygon } from 'ol/geom.js'
 import {
   createViewFromLayer,
   MapContext,
@@ -13,6 +9,7 @@ import { MapContainerComponent } from '../map-container/map-container.component'
 import { BehaviorSubject, from, Observable, of } from 'rxjs'
 import { map, switchMap, tap } from 'rxjs/operators'
 import { DatasetSpatialExtent } from '@geonetwork-ui/common/domain/model/record'
+import { spatialExtentsToFeatureCollection } from '@geonetwork-ui/util/shared'
 
 @Component({
   selector: 'gn-ui-spatial-extent',
@@ -33,29 +30,9 @@ export class SpatialExtentComponent {
       if (extents.length === 0) {
         return of(null)
       }
-      const featureCollection: GeoJSONFeatureCollection = {
-        type: 'FeatureCollection',
-        features: [],
-      }
-      extents.forEach((extent) => {
-        if (extent.geometry) {
-          featureCollection.features.push({
-            type: 'Feature',
-            properties: {},
-            geometry: extent.geometry,
-          })
-        } else if (extent.bbox?.length >= 0) {
-          featureCollection.features.push({
-            type: 'Feature',
-            properties: {},
-            geometry: this.bboxCoordsToGeometry(extent.bbox),
-          })
-        }
-      })
-
       const layer: MapContextLayer = {
         type: 'geojson',
-        data: featureCollection,
+        data: spatialExtentsToFeatureCollection(extents),
         label: 'Spatial extents',
         style: {
           'stroke-color': 'black',
@@ -71,18 +48,4 @@ export class SpatialExtentComponent {
   )
 
   error = ''
-
-  bboxCoordsToGeometry(bbox: [number, number, number, number]): Geometry {
-    const geometry = new Polygon([
-      [
-        [bbox[0], bbox[1]],
-        [bbox[0], bbox[3]],
-        [bbox[2], bbox[3]],
-        [bbox[2], bbox[1]],
-        [bbox[0], bbox[1]],
-      ],
-    ])
-
-    return new GeoJSON().writeGeometryObject(geometry)
-  }
 }
--- a/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
+++ b/libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
@@ -1,15 +1,11 @@
 import { ChangeDetectorRef, Component, inject, Input } from '@angular/core'
 import { CommonModule } from '@angular/common'
-import {
-  createViewFromLayer,
-  MapContext,
-  MapContextLayer,
-} from '@geospatial-sdk/core'
+import { createViewFromLayer, MapContext } from '@geospatial-sdk/core'
 import { MapContainerComponent } from '../map-container/map-container.component'
 import { BehaviorSubject, from, Observable, of } from 'rxjs'
 import { map, switchMap, tap } from 'rxjs/operators'
 import { DatasetSpatialExtent } from '@geonetwork-ui/common/domain/model/record'
-import { spatialExtentsToFeatureCollection } from '@geonetwork-ui/util/shared'
+import { createSpatialExtentLayer } from '../../map-utils'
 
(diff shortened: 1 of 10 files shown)


jahow (commented):
I've feel like i've seen this utils already two or three times. Also it doesn't use the existing spatial extent thing from the service records at all I think

fgravin:
> I've feel like i've seen this utils already two or three times. Also it doesn't use the existing spatial extent thing from the service records at all I think

You are right.
It seems that what you want to say is that the logic of building the extent geojson in the map context was done in the `spatial-context.component` ? 

Claude said that is saw it, but judged not to use it for 2 reasons
1. I asked in my prompt to not be intrusive (which was a mistake because it prevents refactoring).
2. In one `AGENTS.md`, I insisted on reusing existing component, Claude judged I could not reuse the component so it bypassed the logic inside the component.

So I've cleared the context, modify my `AGENTS.md` to also insist on reusing logic, and remove my _not intrusive_ statement.

I think that the second round is correct.
The utils should not go in this folder though, do you confirm ?
So I will add a rule about that.

WDYT @jahow ?

jahow:
I don't know, it's just weird to put geojson-related utilities in the ui/map library I think? how is anyone supposed to figure out that this exists later on?

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ * a translucent grey fill, so the extent stays visible above any underlying
+ * data without hiding it.
+ */
+export const SPATIAL_EXTENT_LAYER_STYLE: VectorStyle = {
+  'stroke-color': 'black',
+  'stroke-width': 2,
+  'fill-color': 'rgba(153, 153, 153, 0.3)',
+}
+
+function boundingBoxToPolygon([minX, minY, maxX, maxY]: BoundingBox): Polygon {
```
jahow: this already exists in the shared/utils/geojson file

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.utils.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Style used to render a record's spatial extents: a solid black outline over
+ * a translucent grey fill, so the extent stays visible above any underlying
+ * data without hiding it.
+ */
+export const SPATIAL_EXTENT_LAYER_STYLE: VectorStyle = {
+  'stroke-color': 'black',
+  'stroke-width': 2,
+  'fill-color': 'rgba(153, 153, 153, 0.3)',
```
jahow: this is probably way too visible for the data preview map; a dashed line was perfect IMO

jahow (commented):
Also now we don't just have `map-utils.ts` and `map-utils.service.ts` and `utils/geojson.ts`... we also have `spatial-extent.utils.ts`! 🎉

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.utils.ts
fgravin: yes, during the refactoring attempt, after your comments, it put it in `geojson.utils.ts`
But after restarting from scratch, it created a new util ...

I will try a new attempt with adding new rules in the `AGENTS.md`

fgravin:
The convention that helped putting the code in the right helper/module :

- **Place new helpers in their canonical home — don't spawn new util files**: when you genuinely need a new pure helper, add it to the *existing* module that owns its domain (dates, strings/URLs, links, geometry, …), choosing the location by *what the helper operates on*, not by where it is first called. Cross-cutting helpers belong in `util/shared` (or the matching sibling lib), never tucked inside a `ui/*` or `feature/*` lib where nobody will find them again. Creating a new `*.utils.ts` next to the consumer is a smell — extend an existing util module instead; if truly nothing fits, surface that gap rather than scattering one more parallel helper. Adding to an already-crowded area (e.g. several `*-utils` / `*.utils.ts` files for the same concern) is itself the signal to consolidate, not to add another file.

At some point, we may need to list the utils and their goal, to help AI and human understand where things are.
Humans also often badly organize helpers...

libs/ui/map/src/lib/components/spatial-extent/spatial-extent.component.ts
```diff
@@ -33,29 +30,9 @@ export class SpatialExtentComponent {
...
-        } else if (extent.bbox?.length >= 0) {
-          featureCollection.features.push({
-            type: 'Feature',
-            properties: {},
-            geometry: this.bboxCoordsToGeometry(extent.bbox),
-          })
-        }
-      })
-
       const layer: MapContextLayer = {
```
fgravin: Ideally, this should call `mapUtils.getRecordExtentLayer()`
Meaning `getRecordExtentLayer` should take the `spatialsExtents` instead of the `record`

jahow (approved):

PR #1619, opened 2026-06-26
Editor / allow editing reuse and service records, use expressions to hide certain parts of the form

jahow:
### Description

This PR introduces an expanded expression system, which is able to evaluate expressions such as `${record.kind != 'dataset'}`. This is used in the editor form to hide or show parts depending on the record being edited.

As such I have lifted the restriction on editing only datasets, and added a "record kind" facet in the editor search for better useability.

Note that this is still bare bones as many fields from the service records cannot be edited yet. Also many labels would benefit from being adjusted for this change, I think this can be done later on.

### Architectural changes

none

### Screenshots

search interface:

image

no temporal extent or dataset links on a service:
image
image

### How to test

- Open the Metadata Editor
- Filter records by type = reuse
- Open a reuse record => only a single online resource is available on page 2 (annexes)
- Filter records by type = service
- No temporal extent on page 1

---


jahow:
oops, I just realized I'm missing E2E tests

libs/feature/editor/src/lib/components/record-form/record-form.component.html
```diff
@@ -3,7 +3,7 @@
     section of facade.currentSections$ | async;
     track sectionTracker($index, section)
   ) {
-    @if (!section.hidden) {
+    @if ((evaluateExpression(section.hidden) | async) !== true) {
```
tkohr: I couldn't confirm this (see my other comment), but Claude gave me some feedback regarding this section:

> Notification spam + recompilation on every change-detection cycle record-form.component.html calls evaluateExpression(section.hidden) / evaluateExpression(field.config.hidden) directly in the template (inside @if (… | async)). Each change-detection pass:
>   - returns a new Observable, forcing the async pipe to unsubscribe/resubscribe every cycle;
>   - re-runs evaluate(), recompiling the expression from scratch each time (record-form.component.ts:113);
>   - and critically, in the error branch calls this.notifications.showNotification(...) (record-form.component.ts:118).
> 
> So a single malformed expression in the config produces a warning toast on every CD cycle, indefinitely — not a one-time warning. Even with valid config, you're recompiling N expressions per CD for every section/field.
> 
> Suggest compiling once (the evaluate() result is pure and config is static) — e.g. memoize the {evaluator, errors} per config node, surface the compile error once, and expose a pre-built observable rather than a method call in the template. This kills both the spam and the per-CD recompile.

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
```diff
@@ -29,6 +39,7 @@ import { switchMap } from 'rxjs/operators'
 })
 export class RecordFormComponent implements OnInit, OnDestroy {
   facade = inject(EditorFacade)
+  notifications = inject(NotificationsService)
```
tkohr, line 43: I don't see error notifications showing up when putting a typo in the field config hidden expressions. I might be missing something.

libs/feature/editor/src/lib/expressions.ts
```diff
@@ -1,16 +1,270 @@
...
+export type ExpressionEvaluator = (
+  context: EvaluationContext
+) => EditorFieldValue
+
+const TOKEN_FUNCTION_CALL_START = '('
+const TOKEN_FUNCTION_CALL_END = ')'
+const TOKEN_TERNARY_FIRST = '?'
+const TOKEN_TERNARY_SECOND = ':'
+const TOKEN_PROPERTY_ACCESS = '.'
+const TOKEN_OPERATOR_LOGICAL = ['==', '==', '!=', '!=']
```
tkohr: This looks like a typo. Did you mean '===' and '!==' as alternative syntax?

libs/feature/editor/src/lib/expressions.ts
```diff
@@ -1,16 +1,270 @@
...
+        if (!/[a-zA-Z]/.test(readNext(expression, pos))) {
+          break
+        }
+        propertyName += readNext(expression, pos)
+        pos++
+      }
+      const variableName = currentSymbol
+      currentEvaluator = (context) => {
+        if (!(variableName in context.globals)) {
+          throw new Error(`Unknown global variable '${variableName}'`)
```
tkohr: Another comment from claude:

> Property access throws at evaluation time (expressions.ts:136,139) for unknown global/property. Inside the template async pipe this errors the binding. Safe today (only record.kind, always present), but referencing a field absent on some record kinds would break form rendering rather than degrade gracefully. Consider returning undefined + a compile-time/eval warning instead of throwing, given this feeds the UI.

libs/feature/editor/src/lib/expressions.ts
```diff
@@ -1,16 +1,270 @@
...
+      currentEvaluator = (context) => {
+        if (conditionEvaluator(context)) {
+          return ifTrueEvaluator(context)
+        }
+        return ifFalseEvaluator(context)
+      }
+      continue
+    }
+
+    // LOGICAL OPERATORS
```
tkohr, line 263: Some more feedback I got from claude that seems to make sense resolving:

> Operator precedence: ==/!= and ternary greedily consume the rest of the expression
> readNextExpression just returns input.substring(start) (the whole remainder), so the RHS of a comparison swallows any following ternary. record.kind == 'dataset' ? 'a' : 'b' parses as record.kind == ('dataset' ? 'a' : 'b'), i.e. record.kind == 'a' — wrong. Chained comparisons (a == b == c) misparse similarly.
> 
> Not triggered by current config (only simple ${record.kind == "service"} / != "dataset" are used), so this is latent — but it's a trap for the next config author, and there's no test combining a comparison with a ternary. Worth either fixing precedence or documenting the "one operator per expression" limitation explicitly.

libs/feature/editor/src/lib/fields.config.ts
```diff
@@ -1,318 +1,288 @@
...
+            'editor.record.form.section.associatedResources.description'
+          ),
+          fields: [
+            {
+              model: 'onlineResources',
+              modelSpecifier: 'onlineResourceType:!link',
+              formFieldConfig: {},
+            },
+          ],
+        },
```
tkohr, line 181: I'm still wondering how we'll be able to display two different components for the same field/model and modelSpecifier, like `FormFieldOnlineSingleLinkResourceComponent` for reuse and `FormFieldOnlineLinkResourcesComponent` for datasets for example?

tkohr (commented):
Wow, very flexible and complex new system! I think it would be great to add some doc about it, on how to use it! Running `npm run i18nextract` will remove an unused key I think, to be checked.

libs/feature/editor/src/lib/fields.config.ts
LHBruneton-C2C: Answering here too, to help find the answer in the future: look for `componentName` at the same level as `model`.

I didn't check if this has been kept with the new system.

libs/feature/editor/src/lib/components/record-form/record-form.component.html
jahow: indeed, will try to improve this

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
jahow: could you please give more details? what expression did you try out?

libs/feature/editor/src/lib/expressions.ts
jahow: No I meant to have only '==' and '!=' as operators

libs/feature/editor/src/lib/components/record-form/record-form.component.ts
jahow: I think typos will not trigger a notification; only malformed expression will

LHBruneton-C2C:
For history, the reuse will be unlocked, but the quality score cannot reach 100% without the source field, and this one is not ready yet for edition.

jahow:
I addressed all comments, this is ready for final review

LHBruneton-C2C:
> I addressed all comments, this is ready for final review

Could you please fill the How to test section? Thanks!

LHBruneton-C2C (changes requested):
I mostly did some testing, and a quick review of the code, to look for potential regression, but not a full code review.

Here is some feedback:
- the multi-select search on type is broken when selecting both `service` and `reuse`.
  We had to face a similar issue when selecting from the Datahub, both types are somewhat overlapping, with the new `presentationForm` field that needs to be taken into account.
- I think `reuse` record should allow the edition of resources, both as file (PDF) and as service (WMS)
- I think `service` record should not allow the edition of file resources, only services and annexes, but that might be a bit too complex to implement
- I think `service` record should not allow the edition of topics and update frequency, as those fields are clearly meant for the data, maybe even the update date is not appropriate
- All the labels should be reworked when mentioning dataset

Overall, I think this PR is a great improvement, but not ready to be merged as it. It would be best to switch off the unlocking of service and reuse, merge the PR, and then take some time to define the fields that can be edited, and a mechanism for the labels (or switch back again to generic labels, but we already did the reverse on purpose).

jahow:
@LHBruneton-C2C I've made the labels to adapt to the record kind; I'm going to fix the reuse+service facets issue in another PR

LHBruneton-C2C (changes requested):
Thanks @jahow for the improvements!

There are still some issues though.

Some icons are broken (blocking):

Image

Image

And some labels are still misleading (not blocking):

Image

Image

Image

Image

Which makes me wonder if services and reuses should not have their own dashboards. This isn't blocking, but should be clearly mentioned in the release notes, if released in this current state.

And I still think services should have less fields available, but then again, this should be discussed with members of the community using the Metadata-Editor, and managing services on their platforms.

jahow:
Icons should be fixed with #1639 

I left out labels that could not be easily changed (i.e. "Datasets" when showing all records). This will need a bit more UX study.

As for fields in services, I left all the fields as they are allowed in the metadata model in gn-ui. Once the form can be configured by the platform admins it will be easy to tailor it to what people really need.

LHBruneton-C2C (approved):
Thanks for this huge work!

PR #1620, opened 2026-06-29
fix(editor): align ISO topic keys with ISO codelist

### LINKED ISSUE #1615: Metadata Editor / Mismatch of INSPIRE topic keys
### Describe the bug

The hardcoded category keys do not correspond to the INSPIRE keys in geonetwork.

Consequence: some categories from GeoNetwork are not displayed correctly in GN-UI:

Image
The middle category is also "Society" but with the correct key, not recognized by GN-UI

The other way round, when a mismatched category is defined in GN-UI, it cannot be saved in GeoNetwork.

The problematic keys are:
GN UI: geoscientific information, Location, Oceans, Society, Structure, Transportation
whereas InsPIRE (and GeoNetwork) uses lower case camelCase keys:
geoscientificInformation, location, oceans, society, structure, transportation

- GeoNetwork-UI version used: [`2.10.0-dev.0d752aeb4`]

- Application: [datafeeder]

### Expected behaviour

The application should use the same keys as INSPIRE

please fix here:
https://github.com/geonetwork/geonetwork-ui/blob/main/libs/feature/editor/src/lib/fields.config.ts#L366

geoscientific information, Location, Oceans, Society, Structure, Transportation

### Screenshots

Image
The middle category is also "Society" but with the correct key, not recognized by GN-UI

fgravin:
## Problem

Fixes #1615

`ISO_TOPICS` (formerly `INSPIRE_TOPICS`) in `libs/feature/editor/src/lib/fields.config.ts` hardcodes the ISO topic categories. 6 of the 19 `value`s used wrong casing/spacing and did **not** match the canonical ISO `gmd:MD_TopicCategoryCode` codes the GeoNetwork backend uses:

| before | after (ISO code) |
|---|---|
| `geoscientific information` | `geoscientificInformation` |
| `Location` | `location` |
| `Oceans` | `oceans` |
| `Society` | `society` |
| `Structure` | `structure` |
| `Transportation` | `transportation` |

### Why it broke

The editor reads/writes records through the XML ISO converter:

- **Display** — `readIsoTopics` (`iso19139/read-parts.ts`) reads the raw `MD_TopicCategoryCode` text (e.g. `society`); `FormFieldTopicsComponent.getTranslatedTopic()` matches `value === topic` against `ISO_TOPICS`. No match → blank badge.
- **Save** — `writeTopics` (`iso19139/write-parts.ts`) writes the value verbatim into the XML. Selecting e.g. "Society" produced `<gmd:MD_TopicCategoryCode>Society</gmd:MD_TopicCategoryCode>`, which is outside the codelist and not recognized by GeoNetwork.

(The `iso19115-3` converter reuses the same readers/writers, so it had the same bug.)

## Fix

- Correct the 6 `value`s to the canonical ISO codes.
- Add `fields.config.spec.ts`: a coherence test asserting values are valid ISO `MD_TopicCategoryCode` codes and that each label key stays in sync with its value (catches future casing/typo regressions).

## Naming cleanup

These topics are ISO categories, not INSPIRE. Renamed accordingly:

- `INSPIRE_TOPICS` → `ISO_TOPICS`
- `INSPIRE_topic` type → `ISOTopic`, moved from the shared domain model into `fields.config.ts` (only used there)
- translation keys `editor.record.form.topics.inspire.*` → `editor.record.form.topics.iso.*` (9 locale files)

## Out of scope

Records already saved by GN-UI with the old invalid codes remain broken on the backend — that's a pre-existing data cleanup, not addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


libs/feature/editor/src/lib/fields.config.ts
jahow: I would include the whole theme in the label key, to be consistent
```suggestion
    label: 'editor.record.form.topics.inspire.climatologyMeteorologyAtmosphere',
```

jahow: same for all topics

jahow: these are actually not INSPIRE topics but ISO topics; not a big deal though

libs/feature/editor/src/lib/fields.config.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+  'utilitiesCommunication',
+]
+
+describe('INSPIRE_TOPICS', () => {
+  it('exposes the same set of values as the ISO MD_TopicCategoryCode codelist', () => {
+    expect(INSPIRE_TOPICS.map((topic) => topic.value).sort()).toEqual(
+      [...ISO_TOPIC_CATEGORY_CODES].sort()
+    )
+  })
+})
```
jahow: Still don't understand how this test brings anything...

libs/feature/editor/src/lib/fields.config.ts
jahow: even better would be to use the `marker` function inline to clean things up:


```suggestion
label: marker('editor.record.form.topics.inspire.climatologyMeteorologyAtmosphere'),
```

jahow (commented):
I have double checked the new values and they are correct.

I'm wondering if we did not create invalid ISO records because of this list in the past though... Should we mention this change in the release notes and tell people how to fix invalid topics in their record if needed?

jahow:
@copilot remove the "‎libs/feature/editor/src/lib/fields.config.spec.ts‎" file

PR #1629, opened 2026-07-08
[Datahub] Delete a reuse from its record page

ronitjadhav:
### Description

This PR introduces a "delete" button on **reuse** record pages in the Datahub. It is shown only when:
- the user has edit rights on the record (owner / editor / admin), and
- `reuse_form_url` is configured on the platform (i.e. the reuse-notification feature is enabled).

Behavior:
- On success: navigate to the search page.
- On error: stay on the reuse page and show an error notification.

### Relation to #1607

This builds on the "notify reuse" work in #1607 (still open). There is intentional overlap.

### Screenshots

image

### How to test

1. Set `reuse_form_url` in the config. 
2. Log in as a user with write access to the reuse's group.
4. Open the reuse record page: the delete button appears in the sidebar.
5. Delete: verify redirect to the search page.
6. Force the DELETE request to fail (e.g. DevTools : Offline) : verify an error notification appears and you stay on the page.


apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -288,4 +320,30 @@ export class RecordMetadataComponent {
       }
     })
   }
+
+  deleteReuse() {
+    this.metadataUuid$.pipe(take(1)).subscribe((uuid) =>
+      this.recordsRepository.deleteRecord(uuid).subscribe({
+        next: () => {
+          this.recordsRepository.clearRecordDraft(uuid)
```
jahow: why?

apps/datahub-e2e/src/e2e/reuse/reuseDelete.cy.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    // on success: navigate to the search page
+    cy.intercept('DELETE', '**/api/records/*', { statusCode: 204 }).as(
+      'deleteOk'
+    )
+    cy.get('[data-cy="deleteReuseButton"]').click()
+    cy.wait('@deleteOk')
+    cy.url().should('include', '/search')
+  })
+})
```
jahow: please use an existing e2e test for this to save on execution time

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -233,6 +255,16 @@ export class RecordMetadataComponent {
...
+  showDeleteReuseButton$: Observable<boolean> =
+    this.metadataViewFacade.metadata$.pipe(
+      switchMap((record) =>
+        record?.kind === 'reuse' && this.reuseFormUrl
+          ? this.recordsRepository.canDelete(record as CatalogRecord)
+          : of(false)
+      ),
+      startWith(false)
+    )
+
```
jahow, line 270: I don't think we need a `startWith(false)` here?

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -288,4 +320,30 @@ export class RecordMetadataComponent {
       }
     })
   }
+
+  deleteReuse() {
+    this.metadataUuid$.pipe(take(1)).subscribe((uuid) =>
```
jahow, line 347: do a confirm dialog first

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -288,4 +320,30 @@ export class RecordMetadataComponent {
...
+            {
+              type: 'error',
+              title: this.translateService.instant(
+                'record.reuse.deleteError.title'
+              ),
+              text: this.translateService.instant(
+                'record.reuse.deleteError.body'
+              ),
+            },
+            7000,
```
jahow: we don't usually put a timeout for error notifications I think

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
```diff
@@ -288,4 +320,30 @@ export class RecordMetadataComponent {
...
+        error: (error) => {
+          this.notificationsService.showNotification(
+            {
+              type: 'error',
+              title: this.translateService.instant(
+                'record.reuse.deleteError.title'
+              ),
+              text: this.translateService.instant(
+                'record.reuse.deleteError.body'
+              ),
```
jahow: we need to show an error to the user here

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -83,6 +83,19 @@
                   ></ng-icon>
                 </gn-ui-button>
               }
+              @if (showDeleteReuseButton$ | async) {
+                <gn-ui-button
+                  class="self-end"
+                  type="outline"
+                  extraClass="w-fit bg-white"
```
jahow: I don't understand why we need these classes

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
```diff
@@ -83,6 +83,19 @@
...
+                  [style.--gn-ui-button-rounded]="'10px'"
+                  (buttonClick)="deleteReuse()"
+                  data-cy="deleteReuseButton"
+                >
+                  <ng-icon class="text-[1em]" name="matDeleteOutline"></ng-icon>
+                  <span class="sr-only" translate>record.action.delete</span>
```
jahow: we don't do sr-only I think? what would be nice would be to have this translated label as a title of the button

jahow (commented):
Thanks, we need a confirm dialog before the reuse is deleted. Also, please remember to use existing e2e tests as much as possible instead of introducing new ones, the E2E test runs are already super long for the Datahub

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
ronitjadhav: Done 👍🏼

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
ronitjadhav: Simplified to just bg-white.

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
ronitjadhav: Done 👍🏼

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
ronitjadhav: Removed 👍🏼

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
ronitjadhav: Done, added a `ConfirmationDialogComponent` before delete.

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
ronitjadhav: Removed 👍🏼

apps/datahub-e2e/src/e2e/reuse/reuseDelete.cy.ts
ronitjadhav: Done 👍🏼 folded into `reuseDetailPage.cy.ts`

apps/datahub/src/app/record/record-metadata/record-metadata.component.ts
ronitjadhav: Its not needed here, it has no drafts to clear.  I Removed it.

apps/datahub/src/app/record/record-metadata/record-metadata.component.html
jahow: but why do we need bg-white? :)

LHBruneton-C2C: Please, be consistent with "ask a question" and "declare a reuse".

ronitjadhav: Yes, the "ask a question" and "declare a reuse" buttons right above both use bg-white.

jahow: I don't get it; they're "outline" buttons, they should have the same style? why specify another class?

ronitjadhav: `type: outline` doesn't default to white. It inherits the theme background `--color-background` (`#fdfbff`, off-white); `.gn-ui-btn-outline` only sets text/border. That's why all three buttons add bg-white.
 
Dropping it just here would leave delete off-white next to the other two. 
We could instead drop bg-white from all three and let them use the default off-white, if that's preferred (but it wont look good) or any other solution in mind?

jahow: ok, then let's leave these bg-white classes. The outline button is used in inconsistent ways in other places anyway

LHBruneton-C2C: I see the class was finally dropped, resolving here.

LHBruneton-C2C (approved):
LGTM, tested OK, thanks!

PR #1630, opened 2026-07-08
Light edition page for reuses

ronitjadhav:
### Description

This PR introduces a **light edition page** in the Metadata Editor (`/light-edit/:uuid`) for editing **reuse** records, reached from the DataHub after declaring a reuse.

Compared to the full editor, this page:

- has **no dashboard sidebar, page selector, or previous/next navigation**;
- shows a **reduced form** : title, description, reuse URL, thumbnail, and the point of contact (edited via `ContactDetailsFormComponent`);
- has a slim top bar with a **leave** and a **save** button (save disables both buttons and shows a spinner while in progress), **no draft status, no publish button, no metadata-quality indicator and toggle, no multilingual toggle**;
- renders the DataHub-style themed **header band at 15% opacity**;
- supports a documented **`redirect_on_leave`** query parameter so the calling app can send the user back where they came from.

On save/leave errors the user stays on the page and a notification is shown via the notification service.

> **Depends on:**
> - **#1619** (allow editing reuse/service records) — provides the `canEdit` relaxation this page needs to render for reuse records. Not included here; see _How to test_.
> - **#1607** (notify reuse form) — provides the DataHub entry point that navigates to `/light-edit/:uuid?redirect_on_leave=…`.

### TODO

- [ ] DataHub "edit this reuse" button (follow-up, waiting for #1629 to be merged)
- [x] save without publishing (follow-up, waiting for #1644 to be merged)
- [ ] deactivate draft  (follow-up, waiting for ??? to be merged)

### Architectural changes

- New route `light-edit/:uuid` in the metadata-editor app with `LightEditPageComponent`, reusing the existing `EditRecordResolver`.
- New `REUSE_LIGHT_CONFIGURATION` editor config and a new `onlineResourceType:singleLink` field specifier that wires the existing-but-unused `FormFieldOnlineSingleLinkResourceComponent`.
- `ContactDetailsFormComponent` is now exported from `feature/editor` and consumed by the metadata-editor app (to edit the reuse's point of contact).
- `record-form` section cards now carry an explicit `bg-white` (they were transparent and relied on a white page behind them).

### Screenshots

localhost_4200_light-edit_7eb795c2-d612-4b5e-b15e-d985b0f4e697

### How to test

> To reach the page (instead of the permission error), you must
> **locally** apply the withheld authorization change : in
> `libs/api/repository/src/lib/gn4/gn4-repository.ts`, method `canEdit()`:
>
> ```diff
> -      record.kind === 'dataset' &&
> +      (record.kind === 'dataset' || record.kind === 'reuse') &&
> ```

1. `npx nx serve metadata-editor` (with the GeoNetwork backend running).
2. Log in as a user with edit rights, and open  `/light-edit/<reuse-uuid>. 
3. Confirm: no sidebar / page selector / publish button / draft status; the form shows title, description, reuse URL, thumbnail and point of contact; the themed header band shows at 15% opacity behind the form.


apps/metadata-editor-e2e/src/e2e/edit/light-edit.cy.ts
```diff
@@ -0,0 +1,69 @@
+// Skipped: every test here requires opening /light-edit/<reuse-uuid> to
+// actually show the record form, which depends on Gn4Repository.canEdit()
+// allowing kind === 'reuse'. That relaxation is a separate, undecided
+// authorization change (see PR #1607's own to-do) and is not shipped with
+// this PR. Un-skip once that change lands (either here or in its own PR).
+describe.skip('light edit page', () => {
```
LHBruneton-C2C: TODO: remove skip

apps/metadata-editor-e2e/src/e2e/edit/light-edit.cy.ts
```diff
@@ -0,0 +1,69 @@
...
+      cy.get('md-editor-publish-button').should('not.exist')
+      cy.get('[data-cy="undo-button"]').should('not.exist')
+      cy.get('[data-cy="save-status"]').should('not.exist')
+      cy.get('[data-test="light-edit-background"]').should('exist')
+      cy.get('[data-cy="leave-button"]').should('be.visible')
+      cy.get('[data-cy="save-button"]').should('be.visible')
+    })
+  })
+
+  describe('save', () => {
```
LHBruneton-C2C: It's a bit ugly, but I think we should apply the same strategy as in the DataHub e2e tests here, and merge all the tests of one use case into a single describe/it. You can keep your lines but commented out, to read it like a standard test file.

apps/metadata-editor/src/app/light-edit/light-edit-page.component.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        fixture.nativeElement.querySelector('gn-ui-record-form')
+      ).toBeFalsy()
+    })
+  })
+
+  describe('notifications', () => {
+    beforeEach(() => {
+      fixture.detectChanges()
+    })
+    it('shows an error notification on publish version error', () => {
```
LHBruneton-C2C: I think this can be discarded as the light edit page will always save without publishing.

apps/metadata-editor/src/app/light-edit/light-edit-page.component.ts
```diff
@@ -0,0 +1,174 @@
...
+  subscription = new Subscription()
+
+  isLoading = true
+
+  backgroundCss =
+    getThemeConfig().HEADER_BACKGROUND ||
+    `center /cover url('assets/img/header_bg.webp')`
+
+  private contacts: Individual[] = []
+  // copy of the first contact: ContactDetailsFormComponent mutates its input
```
LHBruneton-C2C:
```suggestion
  // copy of the initial contact: ContactDetailsFormComponent mutates its input
```

Didn't review that part yet, but when I read "ContactDetailsFormComponent mutates its input" my first reaction is that it shouldn't. It should either emit changes to its parent, or send changes to a store.

apps/metadata-editor/src/app/light-edit/light-edit-page.component.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    this.subscription.add(
+      this.facade.record$.subscribe((record) => {
+        this.contacts = record.contacts ?? []
+      })
+    )
+
+    this.subscription.add(
+      this.facade.saveError$.subscribe((error) => {
+        if (error instanceof PublicationVersionError) {
```
LHBruneton-C2C: Same remark as in the tests, I think this can be discarded.

apps/metadata-editor/src/app/light-edit/light-edit-page.component.ts
```diff
@@ -0,0 +1,174 @@
...
+    ])
+  }
+
+  leave() {
+    const url = this.route.snapshot.queryParamMap.get('redirect_on_leave')
+    // only allow http(s) redirect targets (e.g. no javascript: urls)
+    if (url && /^https?:\/\//.test(url)) {
+      window.open(url, '_self')
+    } else {
+      this.router.navigate(['catalog', 'search'])
```
LHBruneton-C2C: The whole point of the light edit page is to not take the user to the full Metadata-Editor.

Thus, I think the leave button should not be displayed if the `redirect_on_leave` query param is missing or unusable.

docs/apps/editor.md
```diff
@@ -62,6 +63,16 @@ The REVIEWER rights are also needed for the publication step.
...
 
+## Light edition page
+
+Opening `/light-edit/{uuid}` shows a simplified edition page for the given record, without the dashboard sidebar or page navigation. It is used by the Datahub to edit reuse records, but any application can link to it.
+
+The top bar offers a "leave" button and a "save" button; there is no draft handling and no separate publication step on this page.
+
+### URL parameters
+
+- `redirect_on_leave` (optional): a URL-encoded absolute `http(s)` URL the user is sent back to when clicking the "leave" button. The calling application is responsible for building the complete URL, including the record's UUID if it wants the user to come back to the record's page (e.g. `/light-edit/{uuid}?redirect_on_leave=https%3A%2F%2Fexample.com%2Fdatahub%2Freuse%2F{uuid}`). When the parameter is absent or invalid, the user is redirected to the editor dashboard instead.
```
LHBruneton-C2C:
```suggestion
- `redirect_on_leave` (optional): a URL-encoded absolute `http(s)` URL the user is sent back to when clicking the "leave" button. The calling application is responsible for building the complete URL, including the record's UUID if it wants the user to come back to the record's page (e.g. `/light-edit/{uuid}?redirect_on_leave=https%3A%2F%2Fexample.com%2Fdatahub%2Freuse%2F{uuid}`). When the parameter is absent or invalid, the "leave" button is not displayed.
```

LHBruneton-C2C (changes requested):
Already tested and working great!

Just a few minor corrections needed here, mostly removing the leave button when the `redirect_on_leave` param is off.

translations/en.json
```diff
@@ -322,6 +324,12 @@
   "editor.record.resourceError.body": "There was an issue with the {recordKind, select, dataset{dataset} service{service} reuse{reuse} other{record}} attachments:",
   "editor.record.resourceError.closeMessage": "Understood",
   "editor.record.resourceError.title": "Error with the {recordKind, select, dataset{dataset} service{service} reuse{reuse} other{record}} attachments",
+  "editor.record.reuse.leave": "Leave without saving",
+  "editor.record.reuse.save": "Save the reuse",
+  "editor.record.reuse.saveError.body": "The reuse could not be saved:",
+  "editor.record.reuse.saveError.title": "Saving failed",
+  "editor.record.reuse.saveSuccess.body": "The reuse was successfully saved.",
+  "editor.record.reuse.saveSuccess.title": "Reuse saved",
```
jahow: there are already translation keys which take the record type as input; they should be used instead

translations/en.json
```diff
@@ -261,6 +262,7 @@
   "editor.record.form.section.geographicalCoverage.label": "Geographical coverage",
   "editor.record.form.section.metadataPointOfContact.description": "This information concerns the metadata",
   "editor.record.form.section.metadataPointOfContact.label": "Metadata point of contact",
+  "editor.record.form.section.reusePointOfContact.label": "Point of contact",
```
jahow: the `editor.record.form.section.metadataPointOfContact.label` should take the record type as input instead of defining a new one for reuses

libs/api/repository/src/lib/gn4/gn4-repository.ts
```diff
@@ -361,8 +361,10 @@ export class Gn4Repository implements RecordsRepositoryInterface {
   }
 
   openRecordForEdition(
-    uniqueIdentifier: string
+    uniqueIdentifier: string,
+    disableDraft?: boolean
```
jahow, line 367: this should be done at the repository level

docs/developers/editor-draft.md
```diff
@@ -0,0 +1,31 @@
...
+  ```
+  {
+    path: 'light-edit/:uuid',
+    component: LightEditPageComponent,
+    resolve: { record: EditRecordResolver },
+    data: {
+      disableDraft: true,
+    },
+  }
+  ```
```
jahow, line 31: what is that file?? there's a developers doc on the website

apps/metadata-editor/src/app/light-edit/light-edit-page.component.html
```diff
@@ -0,0 +1,54 @@
...
+                </div>
+              </div>
+              @if (pointOfContact$ | async; as pointOfContact) {
+                <gn-ui-contact-details-form
+                  [contact]="pointOfContact"
+                  (contactChange)="handleContactChange($event)"
+                ></gn-ui-contact-details-form>
+              }
+            </div>
+          </div>
```
jahow, line 40: this needs to be part of the editor form right?

apps/metadata-editor/src/app/light-edit/components/light-top-toolbar/light-top-toolbar.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    }),
+  ],
+  templateUrl: './light-top-toolbar.component.html',
+  styleUrls: ['./light-top-toolbar.component.css'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class LightTopToolbarComponent {
+  private editorFacade = inject(EditorFacade)
+
+  @Output() leave = new EventEmitter<void>()
```
jahow: this can be a smart component and handle the redirection itself

translations/en.json
LHBruneton-C2C: I'm not so sure we should use the same translations. The workflow is different here, we will never publish for example. I'd keep the distinction, but on `light` instead of `reuse`.

translations/en.json
LHBruneton-C2C: Same remark as above. The labels on the light edit page don't need to be as precise/complex as the full edit page. Here, the translation key was added to simplify the label. On the full page edit we need to make a distinction between metadata point of contact and data point of contact. On the light edit page, the user only needs to know that this is the contact information (I even hesitated to change "Point of contact" into "Contact information", as the first one is unknown outside of the metadata world).

libs/api/repository/src/lib/gn4/gn4-repository.ts
LHBruneton-C2C: I tried and tried and tried again, I couldn't overload the injection dependency through the facade, effect, service and repository... Could you give it a try?

docs/developers/editor-draft.md
LHBruneton-C2C: Disabling the draft system was never documented, not even the full app token, but might be bug prone. Where would you document this?

apps/metadata-editor/src/app/light-edit/components/light-top-toolbar/light-top-toolbar.component.ts
LHBruneton-C2C: I agree

apps/metadata-editor/src/app/light-edit/light-edit-page.component.html
LHBruneton-C2C: Yes. It felt really to special at first, but now that we are discussing creating/editing contact from the ME, this looks like a first step.

This will need some refactoring, to be able to switch to the details component, for the same model key `contacts`.

docs/developers/editor-draft.md
jahow: yes sorry, I thought this file wasn't part of the docs website

apps/metadata-editor/src/app/light-edit/components/light-top-toolbar/light-top-toolbar.component.html
```diff
@@ -0,0 +1,38 @@
+<div
+  class="flex flex-row items-center w-full h-14 shrink-0 border-b-[1px] px-5 gap-5"
+>
+  <gn-ui-button
+    type="outline"
+    extraClass="border-primary-lighter"
```
jahow: it looks better without this I think

LHBruneton-C2C: These comments should be discussed on figma, as the mockup explicitly uses `primary-lighter` for the border.

apps/metadata-editor/src/app/light-edit/light-edit-page.component.html
ronitjadhav: Left it untouched for now.

apps/metadata-editor/src/app/light-edit/components/light-top-toolbar/light-top-toolbar.component.ts
ronitjadhav: Done 👍🏼

translations/en.json
ronitjadhav: Kept them distinct, just renamed reuse.* → light.*.

libs/api/repository/src/lib/gn4/gn4-repository.ts
jahow: I've tried too and i think to achieve this we need to significantly rework the way the app providers are organized, by having two distinct paths in the app: one with the full GN4 repository with draft, and the other without (for the light edit). This requires quite a bit of work unfortunately, and would be easier if we migrated to a more recent version of angular and ngRX, so I suggest we do this later on.

This being said, the solution used here is going to cause side effects if the user navigates from edit/ to light-edit/, which can technically happen.

jahow: also I think we should mark this with a big FIXME

LHBruneton-C2C (approved):
Thanks for this nice feature!

PR #1632, opened 2026-07-13
chore: remove Coveralls

jahow:
Test coverage is still collected but not surfaced any more. I could not find any easy enough tool to use, and I don't think it's worth it to invest more time in a metrics that is not that essential anyway.

Also worth noting that coveralls 1. is currently down and 2. used to give false metrics because of the way tests were run in the mono repo.

Note that the code coverage of the project can still be inspected locally when running tests for instance.


jahow:
let's keep it for now...

PR #1631, opened 2026-07-10
[Datahub] Display keyword thesaurus / hierarchy path on hover

ronitjadhav:
### Description

This PR introduces a hover tooltip on keyword pills of the datahub record page, showing where each keyword comes from .

- Hierarchical thesaurus → full path: `Aménagement du territoire > Risques et contraintes > **keyword**`
- Flat thesaurus → its title: `GEMET - INSPIRE themes, version 1.0 > **keyword**`
- Free keyword → translated type label: `Theme > **éolienne**`, `Keywords > **données ouvertes**`

The hierarchy comes from the `th_<key>_tree` fields already in the Elasticsearch record, no backend change. 
Reuses `gn-ui-popover` (tippy) with a new `dark` theme matching the Figma.

Thesaurus names are read from the group's `multilingualTitle`, falling back to the flat `title` .

### Decisions to validate
- **`ThesaurusModel.id` is now optional** to represent cited-but-unregistered thesauri. Note the required `id` was already violated at runtime by the ISO reader (`readThesaurus` yields `id: undefined` when the XML citation has no `gmd:identifier`) and by `gn4-platform.mapper` (`id: matchedThesaurus?.key`)
- **Poly-hierarchy display**: when a keyword sits at several places in a tree, the **shortest** path is shown (first on tie). 

### Architectural changes

- New read-only model field `Keyword.hierarchyPath?: string[]`, set only by the GN4/Elasticsearch converter.
- `gn-ui-popover` gains a `dark` theme; new i18n keys `domain.record.keywordType.{theme,place,temporal,other}`.

### Screenshots
image

image

image

image

### Quality Assurance Checklist

- [ ] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [ ] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test
1. Open a dataset record
2. Hover keyword pills
3. Verify the tooltip

PR #1638, opened 2026-07-17
[Metadata-editor] Display the original thesaurus of keywords on hover

ronitjadhav:
### Description

Follow-up to #1631: the same keyword-origin tooltip is now shown in the metadata-editor, on the keyword badges of the record form. 

To avoid duplicating that logic, the tooltip content was extracted from `MetadataInfoComponent` into a new shared dumb component `gn-ui-keyword-tooltip` 

The `.tippy-box[data-theme~='dark']` styling moved from the datahub's
`styles.css` to the shared `tailwind.base.css` so both apps share a single copy.

### Architectural changes

- New shared presentation component `gn-ui-keyword-tooltip` in `ui/elements`
- `node_modules/tippy.js/dist/tippy.css` added to the metadata-editor build
  styles

### Screenshots

image

### How to test

1. Start metadata-editor
2. open a record for edition and hover the keyword badges
3. It should show a tooltip


jahow (commented):
can you factor everything into a KeywordBadge component that would go in ui/element? thanks

tailwind.base.css
```diff
@@ -242,3 +242,12 @@
...
 }
+
+/* dark theme for gn-ui-popover tooltips (tippy.js) */
+.tippy-box[data-theme~='dark'] {
+  background-color: var(--color-gray-950);
+}
+
+.tippy-box[data-theme~='dark'] .tippy-arrow {
+  color: var(--color-gray-950);
+}
```
jahow: why do we need that theme again? tooltips were always dark

libs/ui/elements/src/lib/keyword-badge/keyword-badge.component.html
```diff
@@ -0,0 +1,32 @@
...
+      }
+    </gn-ui-badge>
+  } @else {
+    <gn-ui-button
+      type="primary-light"
+      class="inline-block opacity-70"
+      extraClass="lowercase text-sm py-1 px-2"
+      (buttonClick)="keywordClick.emit(keyword)"
+      >{{ keyword.label }}</gn-ui-button
+    >
```
jahow: this doesn't look good; this component is doing two very different things

ronitjadhav: I have updated it so `KeywordBadge` only handles the removable badges used in the editor, while clickable keywords in Datahub remain buttons. 
I also restored the shared `KeywordTooltip` component so both can reuse the same tooltip content.

tailwind.base.css
ronitjadhav: The new colour was actually from the design mockup. 

For now I've reverted to tippy's default dark styling, and I'll confirm with Pierre whether the mockup color should be reinstated tomorrow.

jahow: you can maybe apply this tippy theme to all tooltips, not just this new one

ronitjadhav:
I’ve reworked based on the comments:

- The editor and Datahub now both use `KeywordBadgeComponent`. It always renders `gn-ui-badge`, with `editMode` deciding whether it is removable or clickable.
- I moved the tooltip content into this component and removed the separate `KeywordTooltipComponent`
- Refactored `BadgeComponent`. A clickable badge now renders through `gn-ui-button`, giving it  button behaviour such as keyboard activation and making it accessible. 
- Added an `unstyled` button type for this case. It provides the button behaviour without applying a button theme, so the appearance is still entirely controlled by `.gn-ui-badge`.

jahow (approved):

PR #1644, opened 2026-07-21
[Editor]: new facade method to save without publish

LHBruneton-C2C:
### Description

This PR introduces a new method in the editor facade, to allow saving with or without publishing distinctly. 

It also fixes the publish state when publishing for the first time on a record. The publication status was only checked on open. Now it's always set to true when saving with publishing in success.

Note: no changes in e2e tests as of now, this will need to be tested when used in the reuse light edit page.

Breaking Change: the `saveRecord` of the editor facade will only save as of now, use the `saveAndPublishRecord` to get the same behavior as previous `saveRecord`

### Architectural changes

The editor service is only responsible for saving with or without publishing as it's told.
The publish boolean set to false for new records comes from higher up in the application (editor-page).

### How to test

No regression:
Open the Metadata-Editor and create a new record. 
Make sure that the record is correctly saved but not published in GeoNetwork.
Edit the record and publish.
Check that the top tooltap displays the translation for `editor.record.saveStatus.recordUpToDate`.
Make sure the changes are correctly saved and the record is published in GeoNetwork.

New feature:
In the publish button in the Metadata-Editor, replace `this.facade.saveAndPublishRecord()` with `this.facade.saveRecord()`.
Open the Metadata-Editor and create a new record. 
Make sure that the record is correctly saved but not published in GeoNetwork.
Edit the record and save.
Check that the top tooltap displays the translation for `editor.record.saveStatus.recordNotPublished`.
Make sure the changes are correctly saved and the record is not published in GeoNetwork.

---

**This work is sponsored by [DataGrandEst](https://www.datagrandest.fr)**.

### DIFF
--- a/libs/feature/editor/src/lib/services/editor.service.ts
+++ b/libs/feature/editor/src/lib/services/editor.service.ts
@@ -16,7 +16,8 @@ export class EditorService {
   saveRecord(
     record: CatalogRecord,
     recordSource: string,
-    fieldsConfig: EditorConfig
+    fieldsConfig: EditorConfig,
+    publish: boolean
   ): Observable<[CatalogRecord, string]> {
     const savedRecord = { ...record }
 
@@ -34,15 +35,9 @@ export class EditorService {
         })
       }
     }
-    let publishToAll = true
-    // if the record is new, generate a new unique identifier and pass publishToAll as false
-    if (!record.uniqueIdentifier) {
-      savedRecord.uniqueIdentifier = null
-      publishToAll = false
-    }
 
     return this.recordsRepository
-      .saveRecord(savedRecord, recordSource, publishToAll)
+      .saveRecord(savedRecord, recordSource, publish)
       .pipe(
         switchMap((uniqueIdentifier) =>
           this.recordsRepository.openRecordForEdition(uniqueIdentifier)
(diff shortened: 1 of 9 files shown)


libs/feature/editor/src/lib/services/editor.service.ts
```diff
@@ -34,15 +35,9 @@ export class EditorService {
         })
       }
     }
-    let publishToAll = true
-    // if the record is new, generate a new unique identifier and pass publishToAll as false
-    if (!record.uniqueIdentifier) {
-      savedRecord.uniqueIdentifier = null
-      publishToAll = false
-    }
```
jahow, line 42: Where did this logic go?

jahow (commented):
Thanks! not tested but it looks good. Only thing I'm not sure of is where the logic pointed out in my comment went?

libs/feature/editor/src/lib/services/editor.service.ts
LHBruneton-C2C: This is what's described in "Architectural changes". The service was responsible for detecting if the record is new or not, which feels out of it's scope. The special case of saving a newly created record is handled in the edit-page `apps/metadata-editor/src/app/edit/edit-page.component.ts`.

This part of the code wasn't modified, as it already called `this.facade.saveRecord()` which now saves without publishing.

jahow (approved):

PR #1655, opened 2026-07-23
Restore the full Material CSS

jahow:
Fix following #1525 where a stripped-down version of the Material Theme CSS was used to reduce the bundle size. This had too many side effects so we're restoring the full Material CSS.

Stripped down version was ~18kB, full CSS is ~100kB


LHBruneton-C2C (approved):
Working much better, thanks!

PR #1661, opened 2026-07-28
Read/Write associated records for ISO 19139 and ISO 19115-3

ronitjadhav:
### Description

This PR implements reading and writing of record associations (`aggregationInfo` / `associatedResource`) for both the ISO 19139 and ISO 19115-3 converters.

An association is made of the UUID of the linked record and an association type (the raw `DS_AssociationTypeCode` `codeListValue`). All associations present in a record are read, in document order; associations missing a UUID or a type are ignored.

The two standards store the reference differently, so reading shares one extractor (`extractAssociatedRecords`) with a thin wrapper per standard, while writing is done per standard:
- ISO 19139: `gmd:aggregationInfo/gmd:MD_AggregateInformation`, UUID in `gmd:aggregateDataSetIdentifier/gmd:MD_Identifier/gmd:code`
- ISO 19115-3: `mri:associatedResource/mri:MD_AssociatedResource`, UUID in `mri:metadataReference/@uuidref`

This PR only covers converter read/write support; the editor field (toggle + UUID input + type dropdown, editing a single association) comes in a follow-up.

Known limitations:

- Only the UUID + type are modeled, so editing associations rebuilds the elements and drops un-modeled siblings such as `initiativeType` or the aggregate dataset name citation. Records whose associations are left untouched are unaffected.
- Associations identified only by a citation/name (no UUID) are not read, and are removed if the associations of a record are edited.

### Architectural changes

A new `AssociatedRecord` model is added and `associatedRecords` is added to `BaseRecord`, consistently with the other identification-level fields (`keywords`, `topics`, `overviews`, `resourceIdentifiers`), so that it also applies to service records. Being a mandatory array, it is added to all record fixtures; the DCAT-AP and GN4 converters return an empty array as they cannot represent it.

### How to test

This PR only covers converter read/write support, the editor field is yet to be implemented.

For now run the converter unit tests and check the new `readAssociatedRecords` / `writeAssociatedRecords` cases for both ISO 19139 and ISO 19115-3


libs/api/metadata-converter/src/lib/fixtures/generic-dataset+metawal.iso19115-3.xml
```diff
@@ -227,71 +227,6 @@ Cette section contient des *caractères internationaux* (ainsi que des "caractè
                     </mmi:userDefinedMaintenanceFrequency>
                 </mmi:MD_MaintenanceInformation>
             </mri:resourceMaintenance>
-            <mri:associatedResource>
```
LHBruneton-C2C, line 230: Why remove this part of the fixture?

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -144,6 +144,7 @@ export interface BaseRecord {
   resourceCreated?: Date
   resourcePublished?: Date
   resourceUpdated?: Date
+  associatedRecords: Array<AssociatedRecord>
```
LHBruneton-C2C: There is a subtility here, as `gmd:identificationInfo` can contain either `gmd:MD_DataIdentification` or `gmd:MD_ServiceIdentification`.

For now, I think it would be best if association where only available for Datasets and Reuses, and not Services.

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.spec.ts
```diff
@@ -573,4 +574,133 @@ describe('read parts', () => {
...
+        </mri:MD_AssociatedResource>
+      </mri:associatedResource>
+    </mri:MD_DataIdentification>
+  </mdb:identificationInfo>
+</mdb:MD_Metadata>`)
+        )
+        expect(readAssociatedRecords(root)).toEqual([])
+      })
+    })
+    describe('unknown, non-standard association type', () => {
```
LHBruneton-C2C: Although this test is truthy, there is no specific code to handle unknown association types, so I don't think we should explicitly test for it.

The same applies to other read/write tests and for other standards.

LHBruneton-C2C (commented):
Thanks for this first implementation!

libs/api/metadata-converter/src/lib/fixtures/generic-dataset+metawal.iso19115-3.xml
ronitjadhav: That fixture is writeRecord(GENERIC_DATASET_RECORD, METAWAL_DATASET), writing the generic record (which has associatedRecords: []) onto the metawal XML as reference. Since associatedRecords is now managed, the writer sees [] vs the reference's 5 existing associations and removes them. This is the same behavior sourceRecords already has for mrl:source in this fixture. So 0 associations is the correct output, and I updated the fixture to match.

If you'd rather keep them in this fixture, I can add matching associations to GENERIC_DATASET_RECORD instead. Let me know.

LHBruneton-C2C: Yes, I'd rather we add parts to the fixture and be more and more complete, than remove those. Thanks!

ronitjadhav: Done 👍🏼

libs/common/domain/src/lib/model/record/metadata.model.ts
ronitjadhav: Done 👍🏼

libs/api/metadata-converter/src/lib/iso19115-3/read-parts.spec.ts
ronitjadhav: Done 👍🏼

LHBruneton-C2C (approved):
Thanks for the review fixes! The fixtures and tests now cover all that's needed.

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -260,11 +260,20 @@ export interface SourceRecord {
   href?: string
 }
 
+/**
+ * Represents another record associated with this one (e.g. a parent/sibling dataset).
+ */
+export interface AssociatedRecord {
+  uuid: string
```
jahow, line 267: **it's NOT necessarily a UUID**; a unique identifier could be anything. The field is called **uniqueIdentifier** for a reason

libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -260,11 +260,20 @@ export interface SourceRecord {
...
+  associationType: string
```
jahow, line 268: this is terrible practice, leaving a field like this inside the core model without any kind of typing; this should absolutely have been typed as a string union

libs/api/metadata-converter/src/lib/iso19115-3/write-parts.spec.ts
```diff
@@ -824,6 +825,166 @@ describe('write parts', () => {
...
+        const sample = parseXmlString(`
+<root>
+    <gmd:identificationInfo>
+        <gmd:MD_DataIdentification>
+            <mri:associatedResource>
+                <mri:MD_AssociatedResource>
+                    <mri:associationType>
+                        <mri:DS_AssociationTypeCode codeListValue="crossReference"/>
+                    </mri:associationType>
+                    <mri:metadataReference uuidref="old-uuid"/>
```
jahow, line 845: I don't think this passes XML validation? meaning that the XSD requires other fields to be there

LHBruneton-C2C: The fixtures produced match exactly what's created when running the same steps purely from GeoNetwork though.

ronitjadhav: Checked both relevant schemas.
In MD_AssociatedResource_Type (https://schemas.isotc211.org/19115/-3/mri/1.0/identification.xsd) for ISO 19115-3 and gmd:MD_AggregateInformation_Type (https://schemas.isotc211.org/19139/-/gmd/1.0/identification.xsd) for ISO 19139, associationType is the only required child element. The other child elements have minOccurs="0", so they are optional.

libs/common/domain/src/lib/model/record/metadata.model.ts
ronitjadhav: I renamed it to `uniqueIdentifier` in #1677.

libs/common/domain/src/lib/model/record/metadata.model.ts
ronitjadhav: I have added an `AssociationType` union in #1677

PR #1677, opened 2026-08-04
Type association types, rename identifier field

ronitjadhav:
### Description

Follow-up to #1661, addressing review feedback:

- `AssociatedRecord.associationType` was a bare `string`; it's now an `AssociationType` union derived from `associationTypeValues`
- `AssociatedRecord.uuid` is renamed to `uniqueIdentifier`

### How to test

No behavior change to verify manually.

### DIFF
--- a/libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
+++ b/libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
@@ -939,7 +939,7 @@ export function extractAssociatedRecords(
     mapArray((el) => {
       const aggregateEl = findChildElement(aggregateName, false)(el)
       const uniqueIdentifier = readIdentifier(aggregateEl)
-      // kept as read, so that a value outside the codelist survives a round trip
+      // kept as read, so that a value outside the shared set survives a round trip
       const associationType = pipe(
         findNestedElement('gmd:associationType', 'gmd:DS_AssociationTypeCode'),
         readAttribute('codeListValue')
--- a/libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
+++ b/libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
@@ -1,6 +1,5 @@
 import {
   AssociatedRecord,
-  AssociationType,
   Constraint,
   ConstraintTranslations,
   DatasetOnlineResource,
@@ -52,6 +51,7 @@ import {
   readText,
   XmlElement,
 } from '../xml-utils'
+import { getAssociationTypeFromCode } from './utils/association-type.mapper'
 import { readGeometry } from './utils/geometry'
 import { fullNameToParts } from './utils/individual-name'
 import { getKeywordTypeFromKeywordTypeCode } from './utils/keyword.mapper'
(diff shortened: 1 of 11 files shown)


libs/common/domain/src/lib/model/record/metadata.model.ts
```diff
@@ -260,12 +260,29 @@ export interface SourceRecord {
...
+  'largerWorkCitation',
+  'partOfSeamlessDatabase',
+  'stereoMate',
+  'source',
+  'isComposedOf',
+  'collectiveTitle',
+  'series',
+  'dependency',
+  'revisionOf',
+] as const
```
tkohr, line 269: I'm unsure about including all values here as the ticket states "keep only the shared types for now". I guess the goal is to preserve existing values in records. However, I think this can produce xml in one standard with a value of the other, if we don't have any check on it in the write parts.

ronitjadhav: I've narrowed it to the four shared values.

libs/api/metadata-converter/src/lib/iso19139/read-parts.ts
```diff
@@ -931,19 +932,20 @@ export function readSourceRecords(rootEl: XmlElement): SourceRecord[] {
...
-  readUuid: ChainableFunction<XmlElement, string>
+  readIdentifier: ChainableFunction<XmlElement, string>
 ): ChainableFunction<XmlElement, AssociatedRecord[]> {
   return pipe(
     findChildrenElement(containerName, false),
     mapArray((el) => {
       const aggregateEl = findChildElement(aggregateName, false)(el)
-      const uuid = readUuid(aggregateEl)
+      const uniqueIdentifier = readIdentifier(aggregateEl)
+      // kept as read, so that a value outside the shared set survives a round trip
```
tkohr, line 942: Is this behavior wanted and this comment needed? Concerns the same topic as below.

libs/common/domain/src/lib/model/record/metadata.model.ts
tkohr: Ok thanks, still not sure what's the best approach here as this mostly concerns changes from #1661. Do you have any opinion on this @jahow @LHBruneton-C2C ? Should the type only include the shared values? Are there more explicit checks in the write parts necessary?

LHBruneton-C2C: I think we should offer (for input) only the 4 shared values, but keep any unrecognized type.

tkohr (approved):
Thanks for the adaptions @ronitjadhav !

libs/common/domain/src/lib/model/record/metadata.model.ts
jahow: this is not how we handle other codelists from iso schemas FYI; we usually have a mapper that only keeps certain values considered as the most common ones, and maps the rest to "other". I have personally never encountered values such as "collectiveTitle" and "partOfSeamlessDatabase" in the real world. Another useful inspiration is to look at how these kind of associations are handled in DCAT-AP and GeoDCAT-AP, especially the SEMIC-EU converter, to see what other people decided for these.

Mapping to "other" means that our type system is actually reliable, and that records at runtime will not end up having different values than the one that we expect.

LHBruneton-C2C: If this can be augmented easily, we need to keep 'crossReference' and 'largerWorkCitation' at the very least.

Does that mean that all other types will be kept "as is" in the record, but displayed as "other"? If so, this will be good enough for now.

jahow: no, the other types would be cast to "other" in the record. This is what we do for other codelists

LHBruneton-C2C: I think we are talking about different things. You talk about casting in the record, but what I have in mind is what remains in the XML on save after edit. I don't have any objection with the record object in the application knowing only "other" and a few more types.

ronitjadhav: I have updated the approach, added a `getAssociationTypeFromCode` mapper in `iso19139/utils/`, similar to `role.mapper.ts` and `update-frequency.mapper.ts`. It keeps the 4 values shared by both standards and maps everything else to `other` (let me know if you would also want to drop `partOfSeamlessDatabase`). 
 
 Hope this approach is good.

LHBruneton-C2C:
As discussed outside of GitHub, we should keep only the ISO19115-3 list, and map any unrecognized type to cross reference.

ronitjadhav:
Done, now keeping the ISO19115-3 codelist (9 values) and mapping any unrecognized type to crossReference.

tkohr (approved):
Thanks @ronitjadhav , the latest commit LGTM!

PR #1616, opened 2026-06-25
[WIP] [Datahub] Replace alasql with duckdb wasm in data-fetcher

ronitjadhav:
### Description

This PR is a WIP migration of `libs/util/data-fetcher` from `AlaSQL` to `DuckDB WASM` as the query backend. DuckDB brings real SQL semantics, better type support, and active development.

What's done:
- New `duckdb-utils.ts`: lazy singleton DB init, WASM assets served locally via `assets/duckdb/`, `queryJsonItems()` replacing AlaSQL calls in readers
- SQL dialect updated: double-quoted identifiers, `FROM data, SUM(...)::DOUBLE` cast (Arrow/HUGEINT serialisation fix)
- AlaSQL removed from all `package.json` files

### TODO

- Wire up remaining apps so the feature works everywhere (webcomponents, data-platform)
- Fix broken unit tests
- Verify existing end-to-end tests still pass
- Review and assess the impact on app loading size
- Check compatibility with stricter server security configurations

### Architectural changes

`@duckdb/duckdb-wasm` added as a dependency (replaces `alasql`). WASM + worker assets must be copied to assets/duckdb/ via project.json asset globs, currently wired for datahub and metadata-editor only.

### Screenshots

[...]

### Quality Assurance Checklist

- [ ] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [ ] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test

1. Start the Datahub
2. Open a dataset with a CSV, JSON or GeoJSON source, table view should load and display data
3. Sort a column and paginate, results should update correctly
4. Switch to chart view, select a field, aggregation type (sum/avg/count etc.), chart should render

PR #1679, opened 2026-08-04
Migrate data-fetcher lib to DuckDB

### LINKED ISSUE #1635: Legends not displayed in the preview for WMS-R
### Describe the bug

In the preview of a record, WMS-R layers do not display a legend, even though it is correctly specified in the GetCapabilities document : 

Image

Image

It affects only WMS-R layers, not the other webservices like WMS-V, WMTS or TMS

- GeoNetwork-UI version used: main

### Expected behaviour

Image

jahow:
### Description

Continuation of #1616 

The data-fetcher lib now does most if its work using the [DuckDB WASM build](https://duckdb.org/docs/lts/clients/wasm/overview) client.

An `Engine` singleton is lazily created and instantiates a DuckDB worker on creation. This worker contains an in-memory database where we load whichever datasets are needed during the Datahub navigation. This will allow us to have a very simple and consistent access to the data across the various data types.

Exceptions are:
* WFS is still handled differently depending on whether we can use the service to do pagination/sorting/filtering  
   Note: I've tried to use DuckDB and GDAL to handle WFS as well but could not get it to work unfortunately
* Excel XLS files are converted to JSON using the `xlsx` package and then processed by DuckDB; this is because [DuckDB only supports XLSX format](https://duckdb.org/docs/lts/core_extensions/excel)

There might be slight changes in the way the data is handled, but overall the behavior should stay the same. Performance for filtering, aggregating etc. should be much better as it is done by compiled C code.

Limitations:
* when reloading the page the DuckDB worker is recreated; this means that we don't keep cache across page reloads!  
   This could be addressed by wrapping the DuckDB worker in a [SharedWorker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker), but this is outside of this PR's scope

Remaining to do:
- [x] check that a dataset isn't already loaded first before creating a new table
- [x] reproduce the logic of enabling/disabling cache by clearing existing data in DuckDB
- [x] correctly handle errors when ingesting data into DuckDB
- [x] Avoid crashing when loading very large datasets; an error should happen instead

### Architectural changes

- removed `papaparse` and `alasql` dependencies from the project
- added `@duckdb/duckdb-wasm` and `apache-arrow`
- updated `ogc-client` to latest dev version

Note: the DuckDB worker is [loaded through DelivrJS](https://duckdb.org/docs/lts/clients/wasm/instantiation#cdnjsdelivr) and not using a local asset. This was done because the worker is ~35MB in size which would make the docker image much bigger, and also would make deploying the Datahub more complicated because it would require more care when setting cache headers on the assets.

### Screenshots

no visual changes

### How to test

- start support-services docker composition
- start datahub with `nx serve datahub`
- open the following URLS and make sure that both the map, table and diagram previews work well when selecting vector layers:
http://localhost:4200/dataset/ed34db28-5dd4-480f-bf29-dc08f0086131
http://localhost:4200/dataset/fr-120066022-jdd-f20f8125-877e-46dc-8cf8-2a8a372045eb 
http://localhost:4200/dataset/ee965118-2416-4d48-b07e-bbc696f002c2
http://localhost:4200/dataset/6d0bfdf4-4e94-48c6-9740-3f9facfd453c
http://localhost:4200/dataset/04bcec79-5b25-4b16-b635-73115f7456e4

---


libs/util/data-fetcher/src/lib/readers/wfs.ts
```diff
@@ -1,14 +1,53 @@
...
+export function parseGml(
+  text: string,
+  featureType: string,
+  version: string
+): DataItem[] {
+  const parts = featureType.split(':')
+  const regex = new RegExp(`xmlns:${parts[0]}=["']([^'"]*)["']`)
+  const match = regex.exec(text)
+  if (match && match.length >= 2) {
+    const wf = new WFS({
```
tkohr:
```suggestion
    const wfs = new WFS({
```

tkohr (commented):
Pfiou, I gave it a human review, but didn't spot anything. Claude did find some things which mostly seem to make sense and I permit myself to paste a summary of them here:

> The design is sound; the problems are in the seams. The blockers: 
> - sql-utils.ts:129 concatenates ORDER BY before WHERE, so any filter+sort combination emits invalid SQL (and the spec enshrines it); 
> - the two new CSV "Caching" tests are identical placeholders that cannot pass; 
> - every DuckDB load failure is reported as 'parse', so 404/403/CORS now surface as "malformed file"; 
> - a failed makeInit() is memoised forever, killing all previews for the session; 
> - the WFS sort guard weakened from sort.length > 0 to if (this.sort), which makes a bare orderBy() build ['A', undefined] and throw inside ogc-client. 
> 
> The rest are narrower: leaked connections, force_not_null = NULL on all-numeric CSVs, parameterised DuckDB types (DECIMAL(18,3), DOUBLE[]) never matching the type map, the new 'boolean' type having no consumer, raw URL interpolation into SQL, and the package/package.json dependency pins drifting from the root.

When testing I discovered the following behaviours, maybe you can double check:
- http://localhost:4200/dataset/ed34db28-5dd4-480f-bf29-dc08f0086131 => first click on table: data stays empty, switching to chart and back loads data, no data visible in chart. once data is cached table also displays on reload of app
- http://localhost:4200/dataset/fr-120066022-jdd-f20f8125-877e-46dc-8cf8-2a8a372045eb => WFS table and chart not working, but map does => `PROPERTYNAME=` seems to create a Whitelabel Error Page, when checking requests manually 
- http://localhost:4200/dataset/ee965118-2416-4d48-b07e-bbc696f002c2 => chart not working with following error in error popup: `Binder Error: No function matches the given name and argument types 'sum(DATE)'. You might need to add explicit type casts. Candidate functions: sum(DECIMAL) -> DECIMAL sum(BOOLEAN) -> HUGEINT sum(SMALLINT) -> HUGEINT sum(INTEGER) -> HUGEINT sum(BIGINT) -> HUGEINT sum(HUGEINT) -> HUGEINT sum(DOUBLE) -> DOUBLE sum(BIGNUM) -> BIGNUM LINE 1: SELECT SUM("datappro") as "sum(datappro)", "id" as "distinct(id... ^`
- http://localhost:4200/dataset/6d0bfdf4-4e94-48c6-9740-3f9facfd453c => CORS error => when using CORS plugin in browser following error displays in error popup: `Les données ont été chargées mais leur décodage a échoué: "DuckDB encountered an error when loading the data: Invalid Input Error: read_xlsx: Failed to parse cell 'J49': Could not convert string 'Autre' to DOUBLE"`
- http://localhost:4200/dataset/04bcec79-5b25-4b16-b635-73115f7456e4 => some issue on first table load similar as first record above but not systematic, issue with pagination on table: `Le chargement des données a échoué en raison d'une erreur HTTP: "Error 400 {"timestamp":"2026-08-19T15:34:08.240+00:00","path":"/geoserver/insee/wfs","status":400,"error":"Bad Request","requestId":"24b256af-3079403","georchestraStylesheet":"/public/georchestra.css","logoUrl":"/public/logo-g2f.png"}"`

Regarding the cdn approach I found [this PR](https://github.com/geonetwork/geonetwork-ui/pull/857) from the past, but I don't know if there was a particular motivation/constraint to get rid of the remote resources that could be an issue here.

jahow:
Thanks for the review!

> Regarding the cdn approach I found https://github.com/geonetwork/geonetwork-ui/pull/857 from the past, but I don't know if there was a particular motivation/constraint to get rid of the remote resources that could be an issue here.

I think it was mostly GDPR preventing us from using Google Fonts unless we specifically asked the user if they were OK with sharing their user data.

tkohr (approved):
Thanks for the improvements @jahow ! 

There are still some issues with some charts and tables in the records from the "how to test" section, but they are also present on main, except for one: The chart of the first record (http://localhost:4200/dataset/ed34db28-5dd4-480f-bf29-dc08f0086131) still does not display, but maybe this just occurs on my side.

tkohr:
The chart of the first record (http://localhost:4200/dataset/ed34db28-5dd4-480f-bf29-dc08f0086131) now displays fine on my side following 8f479866424cdef4ee9ec193067afeaa6f8616b7, thanks!

jahow:
Thanks for the reviews!

Issue #1457, opened 2026-01-19
[Datahub]: Broken search when no record in the catalog has a revision date


LHBruneton-C2C:
Since https://github.com/geonetwork/geonetwork-ui/pull/1373, when a catalog doesn't contain any record with a revision date, the Datahub search by date is broken.

PR #1684, opened 2026-08-12
Sort by `resourceDate` with explicit syntax

### LINKED ISSUE #1669: [Datahub]: Sort by date incorrect for records without all three revision, publication and creation dates
### Describe the bug

When opening the Datahub, navigate to the search page, and sort by date, with or without a search term, or other filters. The records are ordered first by records with a revision date (correct order inside that group), then records with a publication date (correct order inside that group), then records with a creation date (correct order inside that group).

- GeoNetwork-UI version used: 4.2.5, 4.4.10

- Application: Datahub 

In relation with https://github.com/geonetwork/geonetwork-ui/issues/1457

### Expected behaviour

The records should be ordered altogether based on their most recent date between revision, publication and creation, and a missing date should fallback to the other dates.

### Steps To Reproduce

- Run the local docker composition
- Open GeoNetwork
- Create a record, set “azerty - revision” as title, set 08/02/2026 as revision date
- Create a record, set “azerty - publication” as title, set 08/03/2026 as revision date
- Open the Datahub
- Search for “azerty”, order by date
- See that the record “azerty - revision” appears before the record “azerty - publication” even though the record “azerty - publication” is more “recent” than the record “azerty - revision”.

### Screenshots

Image

Image

Image

LHBruneton-C2C:
### Description

This PR fixes https://github.com/geonetwork/geonetwork-ui/issues/1669.

It replaces the sort on revision date, publication date then creation date by a sort on all resource dates.

The sort on `resourceDate` is only working with the explicit syntax
```
"sort": [
  {
    "resourceDate.date": {
      "order": "desc",
      "mode": "max",
      "missing": "_last",
      "nested": {
        "path": "resourceDate"
      }
    }
  }
]
```
even though the short form 
```
"sort": [
  { "resourceDate.date": "desc" }
]
```
should be working. It's mostly the `nested` part that's needed.

TODO: 
- [X] e2e test specific to the sort by resource date
- [ ] check if this fixes https://github.com/geonetwork/geonetwork-ui/issues/1457
       => couldn't reproduce the issue

### How to test

- Run the local docker composition
- Open GeoNetwork
- Create a record, set “azerty - revision” as title, set 08/02/2026 as revision date
- Create a record, set “azerty - publication” as title, set 08/03/2026 as revision date
- Open the Datahub
- Search for “azerty”, order by date
- See that the record “azerty - revision” now appears after the record “azerty - publication”

### DIFF
--- a/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
+++ b/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
@@ -11,6 +11,7 @@ import {
   AggregationsParams,
   FieldFilter,
   FieldFilters,
+  FieldSort,
   FilterQuery,
   FiltersAggregationParams,
   SortByField,
@@ -200,25 +201,26 @@ export class ElasticsearchService {
 
   private buildPayloadSort(sortBy: SortByField): SortParams {
     if (sortBy === null) return undefined
-    // Sort by nested array of dates only works with the explicit syntax
-    if (
-      typeof sortBy[1] === 'string' &&
-      (sortBy[1] as string).endsWith('.date')
-    ) {
-      const nestedPath = sortBy[1].slice(0, sortBy[1].lastIndexOf('.date'))
-      return {
-        [sortBy[1]]: {
-          order: sortBy[0] as 'desc' | 'asc',
-          mode: sortBy[0] === 'desc' ? 'max' : 'min',
-          missing: '_last',
-          nested: {
-            path: nestedPath,
+    const fields = Array.isArray(sortBy[0])
+      ? (sortBy as FieldSort[])
+      : [sortBy as FieldSort]
+    return fields.map((field) => {
+      // Sort by nested array of dates only works with the explicit syntax
+      if (field[1].endsWith('.date')) {
+        const nestedPath = field[1].slice(0, field[1].lastIndexOf('.date'))
+        return {
+          [field[1]]: {
+            order: field[0] as 'desc' | 'asc',
+            mode: field[0] === 'desc' ? 'max' : 'min',
+            missing: '_last',
+            nested: {
+              path: nestedPath,
+            },
           },
-        },
+        }
       }
-    }
-    const fields = Array.isArray(sortBy[0]) ? sortBy : [sortBy]
-    return fields.map((field) => ({ [field[1]]: field[0] }))
+      return { [field[1]]: field[0] }
+    })
   }
 
   private injectLangInQueryStringFields(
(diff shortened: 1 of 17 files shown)


libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -200,6 +200,19 @@ export class ElasticsearchService {
 
   private buildPayloadSort(sortBy: SortByField): SortParams {
     if (sortBy === null) return undefined
+    // Sort by resourceDate only works with the explicit syntax
+    if (sortBy[1] === 'resourceDate') {
```
LHBruneton-C2C, line 204: Also to be noted, we might soon need the same logic but with `recordDate` for MEL (new indexed dates coming from georchestra/maelstro). Should we try to make it more generic?

jahow (commented):
What's weird is that it _should_ be possible to give multiple sort criteria one after the other. The issue that appeared with the resource dates can very well appear with other fields, and this fix seems very focused on the resource dates only.

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
jahow: this `if` looks like a code smell to be honest

LHBruneton-C2C:
@jahow I completely agree! But what could be a viable alternative? Clearly we won't fix GN/ES behavior.

I was thinking maybe an array of field names that require the explicit syntax, with a proper mapping to the nested value and other options.

LHBruneton-C2C:
Discussed outside of GitHub: if the sorting field contains `.date`, we switch to the explicit syntax.

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -200,6 +200,23 @@ export class ElasticsearchService {
...
+        [sortBy[1]]: {
+          order: sortBy[0] as 'desc' | 'asc',
+          mode: sortBy[0] === 'desc' ? 'max' : 'min',
+          missing: '_last',
+          nested: {
+            path: nestedPath,
+          },
+        },
+      }
+    }
```
jahow: I don't think this is covered by a unit test?

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
```diff
@@ -200,6 +200,23 @@ export class ElasticsearchService {
...
     const fields = Array.isArray(sortBy[0]) ? sortBy : [sortBy]
     return fields.map((field) => ({ [field[1]]: field[0] }))
```
jahow: Also I would put the change here, inside the `map`; because a sortBy value can be an array of fields

jahow (commented):
Not tested

libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
LHBruneton-C2C: This revealed an unpredicted bug on empty sort arrays. Now handled properly too.

jahow (approved):
Thanks

PR #1686, opened 2026-08-13
feat(elasticsearch): add hard spatial extent filter

fredj:
### Description

This PR adds support for a hard spatial-extent filter in `ElasticsearchService.filtersToQuery`. Any filter value shaped like a bounding box (`[minX, minY, maxX, maxY]`) is detected via a new `isBoundingBox` type guard, excluded from the generic Lucene query-string building (same pattern already used for `gn-ui-crossFieldFilter` and date-range filters), and turned into a `geo_shape` `envelope` filter clause (relation `intersects`) added to the ES query's `bool.filter` array.

This is intentionally distinct from the existing `geometry` parameter, which only adds `should` (relevance-boost) clauses — the new spatial filter hard-excludes non-intersecting records.

This is a technical prerequisite for spatial search by GeoJSON file upload / geolocation service. Wiring this filter into the Datahub's search-field configuration system (`SpatialExtentSearchField`, URL sync, enable/disable config) is tracked separately.

### Architectural changes

- `libs/common/domain/src/lib/model/search/filter.model.ts`: added a `QueryGeoShape` type to the `FilterQuery` union, replacing a previously untyped cast.
- `libs/util/shared/src/lib/utils/geojson.ts`: added `isBoundingBox`, a type guard reused by the ES service.

### Screenshots

N/A — backend query-building change only, no UI changes.

- [ ] The documentation website 📚 has received the love it deserves

### How to test

1. Run the unit tests: `npx nx test api-repository --test-file=libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.spec.ts` and `npx nx test util-shared --test-file=libs/util/shared/src/lib/utils/geojson.spec.ts`
2. Inspect the `handles spatial filter special case` test in `elasticsearch.service.spec.ts` to see the expected `geo_shape` clause produced from a bounding-box filter value.

### DIFF
--- a/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
+++ b/libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
@@ -36,6 +36,7 @@ import {
 import { TranslateService } from '@ngx-translate/core'
 import {
   bboxToPolygon,
+  BoundingBox,
   getGeometryBoundingBox,
   isBoundingBox,
 } from '@geonetwork-ui/util/shared'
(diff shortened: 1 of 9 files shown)


libs/api/repository/src/lib/gn4/elasticsearch/elasticsearch.service.ts
tkohr, line 446: Of what I understood from last week's discussion, the idea was to use this for both the existing boost and the new filter, either by pushing it to should with the input geometry or pushing it to filter with the converted filter geometry, but I might be wrong (cc @jahow ).

jahow (approved):

PR #1709, opened 2026-09-04
refactor(search,map)!: move geocoding to feature/search, add LocationSearchComponent

fredj:
### Description

This PR moves geocoding out of `feature/map` and introduces a reusable, map-agnostic search-and-select component (`LocationSearchComponent`), laying the groundwork for a spatial-extent search filter.

- Adds an optional `itemTemplate` input to `AutocompleteComponent` for custom per-suggestion markup (falls back to plain-text rendering when unset).
- Moves `GeocodingService` from `feature/map` to `feature/search` — it has no map-specific logic. `GEOCODING_PROVIDER` now has a root-level factory default so it works standalone.
- Removes the old `gn-ui-geocoding` component and its usage in `map-viewer`. That app is unmaintained with no funded roadmap, so breaking it here is acceptable; a `TODO` marks where to restore it.
- Adds `gn-ui-location-search` (`LocationSearchComponent`), wrapping `AutocompleteComponent` + `GeocodingService`. It just emits the selected `GeocodingResult` via `(resultSelected)`, leaving what to do with it (recenter a map, set a search filter...) to the consumer.

### Architectural changes

`GeocodingService`, `GEOCODING_PROVIDER` and `GeocodingProvider` move from `libs/feature/map` to `libs/feature/search`. `feature/map` no longer depends on the geocoding SDK.

### Screenshots

Not applicable — no consumer of `LocationSearchComponent` exists yet (the map-viewer search box was removed rather than migrated; a follow-up will build the spatial-extent search filter on top of this component).

### Quality Assurance Checklist

- [x] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [x] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [x] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test

1. `npx nx test ui-inputs feature-map feature-search`
2. `npx nx build map-viewer --configuration=development` and `npx nx build datahub --configuration=development` — both should build cleanly
[...]

PR #1711, opened 2026-09-04
feat(search): make the geocoding provider configurable via default.toml

fredj:
Continuation of #1709 (location-search-component), stacked on top of it.

Adds `geocoding_provider`/`geocoding_provider_options` to the `[search]`
config section, backed by the existing `GEOCODING_PROVIDER` injection
token (defaults to `geonames` when unset).

Also renames the `data-geopf-fr` provider to `geoplateforme`, matching
the provider rename in camptocamp/geospatial-sdk#75 (`base-adresse-nationale-fr`
was rejected as misleading: the poi/administratif config used here returns
communes/departments/regions, not BAN's address data).

Bumps `@geospatial-sdk/{core,geocoding,legend,openlayers}` to
`0.0.5-dev.83.sha-b8ae1e5`, which is required for the `GeoplateformeOptions`
and `queryGeoplateforme` exports used above (`dev.78` predates the rename).


jahow (commented):
Looking good, missing a few more tests on the app config side! thanks!

libs/feature/search/src/lib/geocoding/geocoding.service.ts
```diff
@@ -5,25 +5,40 @@ import {
...
-    factory: () => ['geonames', { maxRows: 5 }] as GeocodingProvider,
+    factory: (): GeocodingProvider => {
+      const config = getOptionalSearchConfig()
+      if (!config?.GEOCODING_PROVIDER) {
+        return DEFAULT_GEOCODING_PROVIDER
+      }
+      return [
+        config.GEOCODING_PROVIDER,
+        config.GEOCODING_PROVIDER_OPTIONS ?? {},
+      ] as GeocodingProvider
```
jahow: no the libs should _not_ read from the configuration; only the applications should!

jahow (commented):
See comment; also this needs to be documented thoroughly (possible values for the provider, links to the respective documentations, etc.)

jahow (approved):
Thanks!

docs/guide/configure.md
```diff
@@ -213,6 +213,30 @@ For a detailed explanation on the classification system, see [this documentation
 
   If set to `true`, the two default pre-configured search badges ("The latest" and "The most popular") will not be shown under the main search bar. Defaults to `false`. Note that the "My favorites" badge cannot be disabled by configuration.
 
+- `geocoding_provider` (optional)
+
+  Selects the geocoding service used for location search (searching for a place name to move the map to). Defaults to `'geonames'`. One of:
+
+  - `'geoadmin'`: the [Swiss Federal Geoportal search service](https://api3.geo.admin.ch/services/sdiservices.html#search)
```
jahow, line 220: @fredj this link does not work

PR #1712, opened 2026-09-04
feat(search): add SpatialExtentGeocodingSearchComponent

### LINKED ISSUE #1710: AutocompleteComponent: search icon and clear button not vertically centered
### Description

In `AutocompleteComponent` (`libs/ui/inputs/src/lib/autocomplete`), both the search icon on the left and the clear/close button on the right are not vertically centered in the input field. This is visible in the component's own Storybook stories, e.g. "No Submit".

Reproducible on the published `main` Storybook: https://geonetwork.github.io/geonetwork-ui/main/storybook/demo/?path=/story/inputs-autocompletecomponent--no-submit

### Likely cause

**Search icon** (left): `autocomplete.component.css` positions it with a hardcoded magic-number offset instead of real vertical centering:

```css
ng-icon.search {
  margin-top: 0.6rem;
  margin-left: 0.6rem;
}
```

The wrapper div is already absolutely positioned and vertically stretched (`absolute inset-y-[--icon-padding] left-[--icon-padding] w-[--icon-width]`), so a cleaner fix would be to make that div `flex items-center` (and `justify-center`) and drop the `margin-top` hack.

**Clear/close button** (right): `getExtraClass()` in `autocomplete.component.ts` combines `absolute`, a fixed `h-8`, and `inset-y-2` on the same element. When `top`, `bottom`, and `height` are all specified on an absolutely positioned element, CSS resolves this by honoring `top` + `height` and ignoring `bottom` — so it isn't actually centered, it just happens to look right when the input's height matches the assumed value exactly.

### Steps to reproduce

1. Open the `AutocompleteComponent` "No Submit" story
2. Observe the search icon is slightly above vertical center
3. Type some text to reveal the clear (X) button on the right; observe it's also off-center

### Scope

Confirmed pre-existing on `main` (not introduced by any in-flight branch) — checked via `git diff` against `main` and against the published `main` Storybook.

fredj:
Stacked on top of #1711 (geocoding_provider config), which is itself a
continuation of #1709 (location-search-component).

Adds a `[search.spatial_extent_service]` config section
(`main/secondary/tertiary_label_json_pointer`) and
`parseSpatialExtentResult()`, which derives a result's secondary/tertiary
label from these JSON Pointers (RFC 6901), resolved against the whole
`GeocodingResult`. `main_label_json_pointer` defaults to `/label`; the
other two have no default.

The label formatting and bounding-box emission (from the selected
result's `geom`, which every provider already normalizes) live directly
on `LocationSearchComponent` as its default item template/`bboxSelected`
output -- no separate wrapper component.

fixes #1710

### DIFF
--- a/conf/default.toml
+++ b/conf/default.toml
@@ -116,6 +116,15 @@ background_color = "#fdfbff"
 # will not be shown under the main search bar. Defaults to false.
 # do_not_use_default_search_preset = false
 
+# Optional; JSONPaths (resolved against the geocoding result) used to derive the label
+# and geometry shown when searching by spatial extent. main_label_jsonpath and
+# geometry_string_jsonpath default to '$.label' and '$.geom'.
+# [search.spatial_extent_service]
+# main_label_jsonpath = "$.label"
+# secondary_label_jsonpath = "$.properties.citycode[0]"
+# tertiary_label_jsonpath = "$.properties.category[1]"
+# geometry_string_jsonpath = "$.geom"
+
 # One or several search presets visible in form of badges can be defined here; every search preset is composed of:
 # - a name (which can be a translation key)
 # - a sort criteria (prepend the field name with - to do a descending sort):
--- a/libs/feature/search/src/lib/geocoding/spatial-extent-result.parser.ts
+++ b/libs/feature/search/src/lib/geocoding/spatial-extent-result.parser.ts
@@ -1,0 +1,49 @@
+import { JSONPath } from 'jsonpath-plus'
+import { GeocodingResult } from '@geospatial-sdk/geocoding'
+import { Geometry } from 'geojson'
+
+export interface SpatialExtentJsonPaths {
+  mainLabel?: string
+  secondaryLabel?: string
+  tertiaryLabel?: string
+  geometry?: string
+}
+
+export interface SpatialExtentResult extends GeocodingResult {
+  secondaryLabel?: string
+  tertiaryLabel?: string
+}
+
+const DEFAULT_MAIN_LABEL_PATH = '$.label'
+const DEFAULT_GEOMETRY_PATH = '$.geom'
+
+function resolveJsonPath(json: object, path: string): unknown {
+  return JSONPath({ path, json, wrap: false, eval: false })
+}
+
+function resolveJsonPathAsString(
+  json: object,
+  path?: string
+): string | undefined {
+  if (!path) return undefined
+  const value = resolveJsonPath(json, path)
+  return typeof value === 'string' && value ? value : undefined
+}
+
+export function parseSpatialExtentResult(
+  result: GeocodingResult,
+  paths: SpatialExtentJsonPaths
+): SpatialExtentResult {
+  const label = resolveJsonPath(
+    result,
+    paths.mainLabel ?? DEFAULT_MAIN_LABEL_PATH
+  )
+  const geom = resolveJsonPath(result, paths.geometry ?? DEFAULT_GEOMETRY_PATH)
+  return {
+    ...result,
+    label: typeof label === 'string' && label ? label : result.label,
+    geom: (geom as Geometry) ?? result.geom,
+    secondaryLabel: resolveJsonPathAsString(result, paths.secondaryLabel),
+    tertiaryLabel: resolveJsonPathAsString(result, paths.tertiaryLabel),
+  }
+}
--- a/libs/feature/search/src/lib/spatial-extent-geocoding-search/spatial-extent-geocoding-search.component.ts
+++ b/libs/feature/search/src/lib/spatial-extent-geocoding-search/spatial-extent-geocoding-search.component.ts
@@ -1,0 +1,53 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  EventEmitter,
+  Output,
+} from '@angular/core'
+import { GeocodingResult } from '@geospatial-sdk/geocoding'
+import { BoundingBox, getGeometryBoundingBox } from '@geonetwork-ui/util/shared'
+import { getOptionalSearchConfig } from '@geonetwork-ui/util/app-config'
+import { LocationSearchComponent } from '../location-search/location-search.component'
+import {
+  parseSpatialExtentResult,
+  SpatialExtentJsonPaths,
+} from '../geocoding/spatial-extent-result.parser'
+
+@Component({
+  selector: 'gn-ui-spatial-extent-geocoding-search',
+  templateUrl: './spatial-extent-geocoding-search.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  imports: [LocationSearchComponent],
+})
+export class SpatialExtentGeocodingSearchComponent {
+  @Output() bboxSelected = new EventEmitter<BoundingBox>()
+
+  private get jsonPaths(): SpatialExtentJsonPaths {
+    const config = getOptionalSearchConfig()?.SPATIAL_EXTENT_SERVICE
+    return {
+      mainLabel: config?.MAIN_LABEL_JSONPATH,
+      secondaryLabel: config?.SECONDARY_LABEL_JSONPATH,
+      tertiaryLabel: config?.TERTIARY_LABEL_JSONPATH,
+      geometry: config?.GEOMETRY_STRING_JSONPATH,
+    }
+  }
+
+  getSecondaryLabel(result: GeocodingResult): string | undefined {
+    return parseSpatialExtentResult(result, this.jsonPaths).secondaryLabel
+  }
+
+  getMainLabel(result: GeocodingResult): string {
+    const { label, tertiaryLabel } = parseSpatialExtentResult(
+      result,
+      this.jsonPaths
+    )
+    return tertiaryLabel ? `${label}, ${tertiaryLabel}` : label
+  }
+
+  handleResultSelected(result: GeocodingResult) {
+    const { geom } = parseSpatialExtentResult(result, this.jsonPaths)
+    if (!geom) return
+    this.bboxSelected.emit(getGeometryBoundingBox(geom))
+  }
+}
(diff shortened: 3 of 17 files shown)


libs/util/app-config/src/lib/fixtures.ts
```diff
@@ -43,6 +43,12 @@ filter_geometry_url = 'https://my.domain.org/geom.json'
 do_not_use_default_search_preset = false
 advanced_filters = ['publicationYear', 'documentStandard', 'inspireKeyword', 'topic', 'license']
 
+[search.spatial_extent_service]
+main_label_jsonpath = '$.properties.name[0]'
+secondary_label_jsonpath = '$.properties.citycode[0]'
+tertiary_label_jsonpath = '$.properties.category[1]'
+geometry_string_jsonpath = '$.properties.truegeometry'
```
jahow: these should be geocoder options

libs/feature/search/src/lib/spatial-extent-geocoding-search/spatial-extent-geocoding-search.component.ts
```diff
@@ -0,0 +1,53 @@
...
+    )
+    return tertiaryLabel ? `${label}, ${tertiaryLabel}` : label
+  }
+
+  handleResultSelected(result: GeocodingResult) {
+    const { geom } = parseSpatialExtentResult(result, this.jsonPaths)
+    if (!geom) return
+    this.bboxSelected.emit(getGeometryBoundingBox(geom))
+  }
+}
```
jahow: I really don't get why this is not just done by the Location Search component; to me it seems to be an unnecessary layer of complexity

conf/default.toml
```diff
@@ -116,6 +116,15 @@ background_color = "#fdfbff"
...
 # do_not_use_default_search_preset = false
 
+# Optional; JSONPaths (resolved against the geocoding result) used to derive the label
+# and geometry shown when searching by spatial extent. main_label_jsonpath and
+# geometry_string_jsonpath default to '$.label' and '$.geom'.
+# [search.spatial_extent_service]
+# main_label_jsonpath = "$.label"
+# secondary_label_jsonpath = "$.properties.citycode[0]"
+# tertiary_label_jsonpath = "$.properties.category[1]"
+# geometry_string_jsonpath = "$.geom"
```
jahow: I think we should seriously discuss using JSON pointers here instead of the JSONPath spec which is going to be very costly to support

jahow (changes requested):

libs/feature/search/src/lib/spatial-extent-geocoding-search/spatial-extent-geocoding-search.component.ts
```diff
@@ -0,0 +1,53 @@
...
+export class SpatialExtentGeocodingSearchComponent {
+  @Output() bboxSelected = new EventEmitter<BoundingBox>()
+
+  private get jsonPaths(): SpatialExtentJsonPaths {
+    const config = getOptionalSearchConfig()?.SPATIAL_EXTENT_SERVICE
+    return {
+      mainLabel: config?.MAIN_LABEL_JSONPATH,
+      secondaryLabel: config?.SECONDARY_LABEL_JSONPATH,
+      tertiaryLabel: config?.TERTIARY_LABEL_JSONPATH,
+      geometry: config?.GEOMETRY_STRING_JSONPATH,
```
jahow: oh and why are we reading the geometry again? besides, each provider will put the geometry in a specific place so no need to make this configurable

conf/default.toml
fredj: done

libs/feature/search/src/lib/spatial-extent-geocoding-search/spatial-extent-geocoding-search.component.ts
fredj: done

libs/feature/search/src/lib/location-search/location-search.component.ts
```diff
@@ -13,7 +13,13 @@ import {
   AutocompleteComponent,
   AutocompleteItem,
 } from '@geonetwork-ui/ui/inputs'
+import { BoundingBox, getGeometryBoundingBox } from '@geonetwork-ui/util/shared'
+import { getOptionalSearchConfig } from '@geonetwork-ui/util/app-config'
```
jahow, line 17: there should be no dependency to app-config in the libs, only in the apps; use injectable tokens instead

docs/guide/configure.md
```diff
@@ -237,6 +237,17 @@ For a detailed explanation on the classification system, see [this documentation
...
 
+- `[search.geocoding_result_labels]` (optional)
+
+  JSON Pointers (resolved against a geocoding result) used to derive the label shown when searching by spatial extent. `main_label_json_pointer` defaults to `/label`.
+
+  ```toml
+  [search.geocoding_result_labels]
+  main_label_json_pointer = "/label"
+  secondary_label_json_pointer = "/properties/citycode/0"
+  tertiary_label_json_pointer = "/properties/category/1"
```
jahow, line 248: I think the concept of these three labels need to be clarified a bit. Maybe with a small snapshot of what the dropdown looks like with the results in it.

libs/feature/search/src/lib/geocoding/spatial-extent-result.parser.ts
jahow, line 1: I'm struggling to see the added value of this file to be honest :)

libs/feature/search/src/lib/location-search/location-search.component.ts
```diff
@@ -28,13 +34,39 @@ export class LocationSearchComponent {
   @Input() placeholder = ''
   @Input() displayWithTemplate?: TemplateRef<{ $implicit: GeocodingResult }>
```
jahow, line 35: Is that used anywhere? If not let's remove it, this component used to be part of an unmaintained app so we won't break anything in custom apps

libs/feature/search/src/lib/location-search/location-search.component.stories.ts
```diff
@@ -4,14 +4,56 @@ import { importProvidersFrom } from '@angular/core'
...
 import { GeocodingResult } from '@geospatial-sdk/geocoding'
 import { provideI18n } from '@geonetwork-ui/util/i18n'
+import { loadAppConfig } from '@geonetwork-ui/util/app-config'
 import { LocationSearchComponent } from './location-search.component'
 import { GeocodingService } from '../geocoding/geocoding.service'
 
+// LocationSearchComponent reads its JSON Pointer config from
+// getOptionalSearchConfig(), which is only populated by loadAppConfig(); mock
+// fetch just for that call so the story actually exercises the secondary/
+// tertiary label rendering, instead of always falling back to the plain label.
```
jahow, line 14: maybe the LLMs should realize that the app-config utils should not be called in the libs (I think it's written somewhere)

libs/util/app-config/src/lib/fixtures.ts
```diff
@@ -46,6 +46,11 @@ geocoding_provider_options.category = "administratif"
 geocoding_provider_options.limit = 5
 advanced_filters = ['publicationYear', 'documentStandard', 'inspireKeyword', 'topic', 'license']
 
+[search.geocoding_result_labels]
+main_label_json_pointer = '/properties/name/0'
+secondary_label_json_pointer = '/properties/citycode/0'
+tertiary_label_json_pointer = '/properties/category/1'
```
jahow, line 52: So just to be clear: the original specification mentioned these configuration keys

```toml
geocoding_provider = "geoplateforme"
geocoding_provider_options.index = "poi"
geocoding_provider_options.category = "administratif"
geocoding_provider_options.limit = 5
geocoding_provider_options.main_label_jsonpath = "$.properties.name[0]"
geocoding_provider_options.secondary_label_jsonpath = "$.properties.citycode[0]"
geocoding_provider_options.tertiary_label_jsonpath = "$.properties.category[1]"
```

Since we're not using JSON Path anymore, and to disambiguate from the options that are passed to the provider, how about this?

```toml
geocoding_provider = "geoplateforme"
geocoding_provider_options.index = "poi"
geocoding_provider_options.category = "administratif"
geocoding_provider_options.limit = 5
geocoding_provider_labels.main = "/properties/name/0"
geocoding_provider_labels.secondary = "/properties/citycode/0"
geocoding_provider_labels.tertiary = "/properties/category/1"
```

(note that we don't suffix these with "json_pointer" anymore because 1/ it makes it less verbose and 2/ we don't do it for other keys, we just specify what types they are in the comments)

Also keep in mind that this needs to have sensible defaults for each provider; these options are **optional**

libs/feature/search/src/lib/location-search/location-search.component.ts
fredj: yes, it's unused, i will remove it

libs/feature/search/src/lib/location-search/location-search.component.stories.ts
fredj: The eslint check was not working, see #1746

libs/util/app-config/src/lib/fixtures.ts
fredj: for the tertiary label ("/properties/category/1"), it also depends on category, for example:
category=administratif: index 0 is always "administratif" and index 1 is the useful type.
```
Beaufort              ["administratif","commune"]
Paris 16e Arrdt       ["administratif","arrondissement municipal"]
``` 

without category:
```
Mont Blanc            ["sommet","élément topographique ou forestier","détail orographique"]
Mont Blanc Plage      ["camping","zone d'activité ou d'intérêt"]
```

libs/feature/search/src/lib/geocoding/spatial-extent-result.parser.ts
fredj: inlined

libs/util/app-config/src/lib/fixtures.ts
jahow: That's right but here we're always asking for POIs of type "administratif"

this being said, maybe we can support pointing to an array instead of a specific value, and then output the elements of the array joined with commas?

PR #1718, opened 2026-09-07
Connect spatial extent dropdown for files to search filter

tkohr:
### Description

This PR connects the spatial extent dropdown to the search filters.

- The bbox computed from an imported GeoJSON file is now applied as a `spatialExtent` search filter (and persisted in the URL) instead of just being logged.
- The dropdown gained an `initialBbox` input, so a bbox coming from the URL/filters is restored in the UI on page load.
- The selection button labels now distinguish a bbox imported from a file (`bbox from {fileName}`) from a pre-selected one (`Pre-selected bbox`), replacing the previous prefix + filename concatenation.

Claude was used for label translation handling, test generation and test adaptions.

### Architectural changes

None — no new libraries and no dependency changes.

A reusable `cy.uploadFile()` command was added to the shared `tools/e2e/commands.ts` (selects a file into a component's hidden `<input type="file">`), and `datasets.cy.ts` was restructured to the flat, comment-separated style (49 nested tests → 5 flat `it` blocks, spec runtime ~4 min → ~35 s).

### Screenshots

image

### Quality Assurance Checklist

- [x] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [x] If **new logic** ⚙️ is introduced: unit tests were added
- [x] If **new user stories** 🤏 are introduced: E2E tests were added
- [x] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created

### How to test

Follow these instructions to confirm the spatial extent filter is applied:

1. Enable the filter in `conf/default.toml`: `advanced_filters = ['spatialExtent', ...]` (optionally `spatial_extent_max_file_size = 1`), then `npm start`.
2. On `/search`, open the "Spatial extent" dropdown and import a GeoJSON file — the results are filtered on its bbox and `spatialExtent=<minX>,<minY>,<maxX>,<maxY>` appears in the URL. The selection reads `bbox from <filename>`.
3. Reload that URL (or open it in a new tab) — the filter is still applied and the selection now reads `Pre-selected bbox`.
4. Remove the selection — the filter and the URL parameter are cleared.
5. Import an invalid file (wrong extension, malformed JSON, no geometry, or over the configured size) — an error notification is shown and no filter is applied.

PR #1727, opened 2026-09-10
[DataHub] Add search by resource date

vladyslav-tk:
### Description

This PR introduces a new advanced filter, `resourceCreationRevisionDate`, letting users filter records by a resource's creation and/or revision date. A record matches if either its creation date (`creationDateForResource`) or its revision date (`revisionDateForResource`) falls within the selected range. Since this isn't a physically indexed ES field, it's implemented as an ES runtime field: a Painless script emits every value from both source date fields into one virtual multi-valued `date` field, queried with a standard inclusive `range` filter.

### Architectural changes

- `ElasticsearchService.registerRuntimeField` now accepts an optional `type: 'keyword' | 'date'` (previously always `'keyword'`), and `processRuntimeFields` now also detects runtime fields referenced inside `range` filter queries.
- `ElasticsearchService.filtersToQuery` now supports multiple simultaneous date-range filters (e.g. `changeDate` + `resourceCreationRevisionDate` active at once) instead of only honoring the first one found.

### Screenshots

Reuses the existing date-range dropdown UI.

- [x] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test

1. Add `resourceCreationRevisionDate` to `advanced_filters` in your `default.toml` 
2. Run the DataHub app
3. Open the search filters panel, open the "Resource date" filter, and pick a start and/or end date.
4. Verify results update after each change, and that the URL reflects the selected range (`?resourceCreationRevisionDate=...`).

---


apps/datahub-e2e/src/e2e/search-header/filters.cy.ts
```diff
@@ -141,3 +141,35 @@ describe('filters and sorts', () => {
     })
   })
 })
+
+describe('Filter on resource creation/revision date', () => {
```
tkohr: The naming of the test files may currently be a bit misleading. I'd rather put the tests in `datasets.cy.ts` around `describe('filter by geometry', () => {`. However this will create conflicts with #1718 so we may also move it later.

tkohr (commented):
Thanks @vladyslav-tk ! I wasn't able to check the details on the runtime field yet, but the PR reveals that the drop-down UI does not yet display quite correctly in the datahub. The width and height of the `date-range-dropdown` should not stretch as far as they do.

Edit: Here's a screenshot of what I meant. Maybe it's ok like this, just noticed that it displayed larger than in the editor. Isn't caused by this PR in any case though.
image

jahow (approved):
Thanks, working and looking good!

I haven't reproduced the issue mentioned by @tkohr so maybe make sure that the issue is gone before merging, thanks

apps/datahub-e2e/src/e2e/search-header/filters.cy.ts
vladyslav-tk: I rebased on the latest main and moved the test

vladyslav-tk:
> Thanks @vladyslav-tk ! I wasn't able to check the details on the runtime field yet, but the PR reveals that the drop-down UI does not yet display quite correctly in the datahub. The width and height of the `date-range-dropdown` should not stretch as far as they do.
> 
> Edit: Here's a screenshot of what I meant. Maybe it's ok like this, just noticed that it displayed larger than in the editor. Isn't caused by this PR in any case though. image

@tkohr I checked figma, and the dropdown size should be fixed, but it's now a bit bigger than in figma. Should I change it?


image

tkohr:
Thanks for checking! This looks good on your screenshot. I think you'll have to keep it at `w-[296px]` for the content to fit inside correctly. Somehow this width class doesn't seem to be applied on my side, I'll double check.

tkohr (approved):
Ok confirmed, this was some tailwind issue present in the cache on my side. Clearing the `.angular/cache` resolved it. Sorry for the noise!

PR #1527, opened 2026-03-23
Don't reject on first error when getting feature types

LHBruneton-C2C:
### Description

This PR fixes a wrong behavior when getting WFS layers:
- for display in the service capabilities section of a record (DH)
- for the dropdown in the online resource record form (ME)

If only one WFS layer was failing, none were shown. Now, all WFS layers correctly described are shown. 

### Quality Assurance Checklist

- [X] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [X] If **new logic** ⚙️ is introduced: unit tests were added
- [ ] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](https://geonetwork.github.io/geonetwork-ui/main/docs/) 📚 has received the love it deserves

### How to test

Find a server with WFS capabilities listing a WFS layer which will fail on DescribeFeatureType request.
Or trust the completed unit test.

PR #1208, opened 2025-04-07
[Datahub]: Service page

cmoinier:
### Description

This PR introduces a new page to display ServiceRecords in detail. This page is based off the current DatasetRecord page with some differences (see in ticket's figma).

Key Changes : 

- [x] Did NOT create a full new page because it was just too much duplicated code. Instead re-used the `record-metadata` page & children and made it work for both types without much more code (and for `reuse` types, it will also be simpler - will just need to add some more display conditions)
- [x] New UI component : `service-capabilities` displays the service's layers, and allows to click on them to display their properties --> note that the list of displayed properties is inspired from the list in figma (which based itself on the WFS format) and has a few free additions from my end for other formats. It is possible to search through the layer's list. Autocomplete was not used for this case because we didn't want the dropdown & a simple input was simpler.
- [x] New UI component : `metadata-spatial-extent`, which displays the spatial extent for service records in the collapsable menus, top of the page.
- [x] New UI button "Ask a question" which scrolls down to the user feedbacks section
- [x] Routing to `/service/uuid` for service records, and an error displayed if a service is directly accessed from the `/dataset/uuid` path (the same goes for dataset records accessed from a service path).
- [x] "Related datasets" moved under the "User feedbacks section" --> aware that it looks ugly right now because of the cards & backgrounds having the same color, but once https://github.com/geonetwork/geonetwork-ui/pull/1202 is merged and implemented it should look pretty again!
- [x] Not visible in service page : API section, Preview section, Catalog name
- [x] In the `ServiceEndpoint` model, the keys "protocolUrl" and "protocol" have been harmonized with other link types and renamed "url" and "accessServiceProtocol" everywhere, to facilitate the reading of service record's links.
- [x] Small UI changes common to dataset/service pages

Out of scope for this ticket : 

- [x] Associated records section
- [x] Reading OGC links for the service capabilities component

### Architectural changes

none

### Screenshots

image

image

### Quality Assurance Checklist

- [x] Commit history is devoid of any _merge commits_ and readable to facilitate reviews
- [x] If **new logic** ⚙️ is introduced: unit tests were added
- [x] If **new user stories** 🤏 are introduced: E2E tests were added
- [ ] If **new UI components** 🕹️ are introduced: corresponding stories in Storybook were created
- [ ] If **breaking changes** 🪚 are introduced: add the `breaking change` label
- [ ] If **bugs** 🐞 are fixed: add the `backport <release branch>` label
- [ ] The [documentation website](docs) 📚 has received the love it deserves

[...]

PR #1732, opened 2026-09-11
Make deep details optional on getLayers

LHBruneton-C2C:
### Description

This PR is a follow up on https://github.com/geonetwork/geonetwork-ui/pull/1527.

At the time, only fixing the behavior when encountering an error on fetching one of the layers details was fixed.

Looking at the repo history, we can see that:
- `getLayers` (as the standalone link-utils function) is only used twice:
  - https://github.com/geonetwork/geonetwork-ui/blob/main/libs/feature/editor/src/lib/components/online-service-resource-input/online-service-resource-input.component.ts#L153
  - https://github.com/geonetwork/geonetwork-ui/blob/main/libs/ui/elements/src/lib/service-capabilities/service-capabilities.component.ts#L129

The second one has been added in https://github.com/geonetwork/geonetwork-ui/pull/1208.
In the same PR, we can see the changes that fetch the WFS and WMS details on each layer.

This current PR simply makes the details fetching optional, and activates the option only for the service capabilities component in the Datahub. This way, the original `getLayers` as used in the Metadata-Editor to add an online resource remains details free and efficient.

### How to test

Datahub
- create a `service` record with a WMS online resource
- make sure the capabilities section is still displayed correctly 
- create a `service` record with a WFS online resource
- make sure the capabilities section is still displayed correctly 

Metadata-Editor
- create a new record or navigate to an existing one
- on the resource page, link to a service, load a WFS GetCapabilities URL
  preferably pointing to a WFS service with many layers such as
  https://mel.integration.apps.gs-fr-prod.camptocamp.com/geoserver/ows?service=WFS&acceptversions=2.0.0&request=GetCapabilities
- see that the loading is faster than before (clean the ogc-client cache before)
- see that the displaying is faster than before


jahow (approved):
Looks OK, I still think the full layers should be lazily fetched instead of loaded eagerly but that would be a bigger change :)

LHBruneton-C2C:
@jahow Thanks for the review! I'll fix the tests and merge then.

PR #1733, opened 2026-09-15
[DataHub] Fix search by resource date interval

vladyslav-tk:
### Description

This PR is a follow up on https://github.com/geonetwork/geonetwork-ui/pull/1727. It fixes the matching semantics of the `resourceCreationRevisionDate` DataHub filter (resource date range filter, based on a record's creation/revision dates).

Previously, a record matched only if **at least one** of its dates (creation or revision) individually fell inside the selected filter range. 

The fix treats a record's dates as an interval and matches it against the filter range.

### Architectural changes

- `ResourceCreationRevisionDateSearchField` now registers two ES runtime fields, `resourceCreationRevisionDateMin` and `resourceCreationRevisionDateMax`, instead of a single field emitting every raw date value.
- `ElasticsearchService.buildDateRangeQuery` detects when a date-range filter field has matching `<field>Min`/`<field>Max` runtime fields registered, and in that case builds an interval-intersection query (`bool.filter` with two `range` clauses) instead of a plain single-field `range` query. Fields without such siblings (e.g. `changeDate`) keep the previous plain `range` behavior.

### Screenshots

No UI changes.

### How to test

1. Enable `resourceCreationRevisionDate` in `advanced_filters`
2. In DataHub, open the "Resource date" filter and pick a range from `10/21/2023` to `04/25/2024` (fits existing data)


LHBruneton-C2C:
Tested OK on both existing filter in the Metadata-Editor and new filter in the Datahub.

fredj (approved):

jahow (commented):
For the record, it might have been possible to use a date_range field here, see https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/range

vladyslav-tk:
> For the record, it might have been possible to use a date_range field here, see https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/range

date_range is not a valid type for runtime mappings
https://www.elastic.co/docs/manage-data/data-store/mapping/map-runtime-field

PR #1737, opened 2026-09-17
chore: ignore Nx-coupled packages in Dependabot config

fredj:
## Summary
- Add `.github/dependabot.yml` to stop Dependabot from opening individual PRs for Angular, Nx, ngrx and Storybook packages, since these must be upgraded together via `nx migrate` rather than as one-off version bumps.
- The ignore list was derived from `npx nx report` (installed `@nx/*` plugins and community plugins `@ngrx/*`, `@storybook/angular`) cross-referenced with `package.json`.


jahow (approved):
Sweet, thanks!

PR #1741, opened 2026-09-21
Config: Support custom filters

tkohr:
### Description

This PR introduces new `[[custom_filter]]`s in the config, containing a `name`, `base_filter` and `label_key`. Multiple custom filters can be added in the config and referenced withing the `advanced_filters` by their name.

The PR currently maps the custom filters to the advanced filters within the `app-config.ts`.

### Architectural changes

### Screenshots

[...]

### How to test

---


libs/util/app-config/src/lib/app-config.ts
```diff
@@ -280,7 +351,12 @@ export function loadAppConfig(configUrl = 'assets/configuration/default.toml') {
                 name: param.name,
                 filters: param.filters,
               })),
-              ADVANCED_FILTERS: parsedSearchSection.advanced_filters,
+              ADVANCED_FILTERS: resolveAdvancedFilters(
+                parsedSearchSection.advanced_filters as string[],
+                customFilters,
+                warnings
+              ),
+              CUSTOM_FILTERS: customFilters,
```
tkohr: My idea is to handle the "mapping" of custom filters into advanced filters in the `app-config.ts`, so I think we may not even need to expose the `CUSTOM_FILTERS` from the config (as everything will be part of the ADVANCED_FILTERS). Wondering if this is the right approach or if I'm missing something, especially regarding the `include/exclude_values` properties that will follow 🤔 .

jahow: I wouldn't include too much logic into the app-config lib. I think it's fine that the config object exposes a list of filter names, some of which might be built-in while others are custom, and also expose a list of custom filter definitions.

Then the actual processing of these custom fields can be done in the FieldService for instance.

I would argue that the role of the app-config library is to offer configuration values in an easy-to-use way, organized into dicts/arrays/strings, and leave all the actual logic to the lib/feature modules down the line.

Also keep in mind that whatever the app-config lib provides, someone should be able to also provide "manually" using Angular injectable token providers, so it has to make sense for a human too

docs/guide/configure.md
```diff
@@ -204,11 +204,49 @@ The filters should be provided as an array, for instance:
...
+  It is only displayed if it is referenced by its name in the `advanced_filters` setting.
+
+  Every custom filter is composed of:
+
+  - `name` (mandatory): the name of the filter, which must hold a prefix separated by a colon; this is the name
+    used to reference it in `advanced_filters`
+  - `base_filter` (mandatory): the search field the filter is based on; only `'keyword'` is supported for now,
+    any other value will make the custom filter be ignored
+  - `label_key` (optional): a translation key used as the label of the filter; it can be defined in the
+    `[translations]` sections. Defaults to the label of the base filter.
```
jahow, line 228: Might be a matter of opinion, but I think this description is a bit hard to understand. How about:

```md
  Custom filters allow customizing default search filters by changing some parameters to them.
  They can be referenced by their name in the `advanced_filters` setting of the `[search]` section.

  Every custom filter is composed of:

  - `name` (mandatory): name of the filter as it should appear in the `advanced_filters` setting; custom filters must have a prefix separated by a colon in their name, e.g.: "myOrg:myOrgKeywords"; case sensitive
  - `base_filter` (mandatory): the search field the filter is based on; only `'keyword'` is supported for now
  - `label_key` (optional): a translation key used as the label of the filter; it can be defined in the
    `[translations]` sections. Defaults to the label of the base filter.
```

jahow (commented):
Thank you, I put some feedback on your current approach. Also why not add all the fields we'll need right away? They will be wired later on anyway

libs/util/app-config/src/lib/app-config.ts
tkohr: Ok thanks for your input, I've moved the resolving logic from the app-config lib to the `FieldsService`. This makes the app-config, really just provide the config. I haven't added the `include/excludeValues` in this PR yet.

docs/guide/configure.md
tkohr: Thanks for the reformulation, I've updated it.

libs/feature/search/src/lib/utils/service/fields.service.ts
```diff
@@ -109,6 +115,66 @@ export class FieldsService {
...
+        )}). This filter will be ignored.`
+      )
+      return null
+    }
+    return {
+      fieldName: baseFilter,
+      labelKey: labelKey ?? `search.filters.${baseFilter}`,
+    }
+  }
+
```
jahow: I'm not sure about this change; sounds like the topics of "advanced filters" and "configuration" are leaking into this service where they don't really belong. The fields service has one purpose: to define which fields are used for filtering and searching.

TBH I think we can just keep the config part in this PR and leave out the rest to another PR, what do you think?

tkohr: I agree, this does not really feel to be the right place to resolve the advanced filters and check the config. As the config checking cannot be done without resolving, the app-config lib does not seem to be the right place either (like we discussed before). Since this is not the responsibility of the `SearchFiltersComponent` neither (IMO), what do you think about introducing a new service or utils for this (in this PR or the follow-up)? I'm thinking about a `FiltersConfigService` in `feature/search/utils` for example.

tkohr: Something like: 8c9ddd3f59002914959639a5dc25a76c76bee6b6

tkohr: As discussed I've just kept the app-config part in this PR and I'll re-tackle the rest in the follow-up with a different approach.

jahow (approved):
Looking good!

Two things that might be improved:
- maybe don't use `ign:` in all the docs, examples and fixtures; this ends up looking weirdly specific. A more neutral prefix like `myOrg:` or maybe `org:` would be less eye catching
- it might also be worth clarifying that include_values and exclude_values will only have an effect when showing possible values for a filter (e.g. in the filter dropdown); they will essentially have no effect when the custom field is used e.g. in the url

